PR #230, opened 2015-06-04
Fixed: Fixed tuplets in the Schubert Lindenbaum example

dinamix1:
Previously the tuplets in the measures 4 and 8 were not properly encoded. On measure 4, the final g was missing and on measure 8, it was not included in the tuplet.


lpugin:
There are also some new 2012 examples, aren't there? Do we want this? Shouldn't we have them in 2013 version directly?

dinamix1:
We only changed the mistakes in the 2013 version. Would you like them to be changed in the 2012 version as well?

lpugin:
I am seeing new files (samples/MEI2012/Handcodings/Schubert_Erlkönig.mei) in the 5 files changed.

lpugin:
Sorry, I just understood, one commit is about the change in the filename. Did you want to change this as well?

ahankinson:
The ö in Erlkönig seemed to be causing some problems with git, and the file was showing up as marked for deletion. I think re-committing it fixed that issue.

PR #231, opened 2015-06-05
Fixed: removed sameas attribute in Grieg opus 43 example

minamouse:
Previously, @sameas was used within note elements in some places in the Grieg opus 43 example, specifically in measure 7 where the notes were not the same and were actually supposed to have different durations. This fixes this problem and has the correct note durations.


lpugin:
Looks good, but what was actually wrong with the sameas attributes?

ahankinson:
The `@sameas` on a note pointed to another note, but the durations between them were actually different (therefore the notes were not the same...).

Issue #232, opened 2015-06-08
Ornaments visual attribute classes should be members of att.altsym


raffazizzi:
Currently it's not possible to specify an alternative symbol for an ornament, even though the guidelines say (from the desc of `@form` on mordent): "Another attribute in the visual domain would be necessary in order to be completely explicit about which visual symbol is actually to be rendered."

The following attribute classes should be members of att.altsym:
- att.mordent.vis
- att.turn.vis
- att.trill.vis

raffazizzi:
I would also suggest that this is a bug, not an enhancement: the element's description clearly states that there should be an attribute in the visual domain that is currently missing.

PR #233, opened 2015-06-08
Made ornament visual classes members of att.atlsym. 

raffazizzi:
Fixes issue #232


ahankinson:
Changing to `Needs Review` so that it's clear that someone should review the patch and comment. `Needs Discussion` should probably be reserved for issues... 

#growingpains

ahankinson:
Since this is our first go-round with this type of development, I'm wondering if we should also ask for a small example file that will fail validation without this patch, but pass with this patch. These test cases could then form the basis of an automated unit testing system...

raffazizzi:
Good idea :+1:

I added a test file, let me know if you think it should be in a different location.

I also regenerated the schema... this is another thing we should request when submitting a fix: regenerate at least mei-all.rng to show that there are no errors introduced in the schema generation.

ahankinson:
Hm. I wonder if this should in fact go in a top-level `tests` directory, and not in the sample encodings. 

It should also probably have the pull request number in it somewhere, just so we can find the reason for the file later if we need to.

ahankinson:
BTW, if you branch locally for a specific fix and don't issue your pull request from your own `master` branch but from your local issue-specific branch, you won't need to do a merge if things change upstream...

ahankinson:
Other than that, looks great!

raffazizzi:
> Hm. I wonder if this should in fact go in a top-level tests directory, and not in the sample encodings.

I had the same thought, but then aren't the samples tests themselves when it comes to check for schema validity / coherence? That's why I thought bundling them together might make sense. Maybe.

ahankinson:
RE: Schema re-generation, we've got the Jenkins CI server set up to do this for us:

http://dev.simssa.ca:8080/jenkins/

I've not updated it to pull from GitHub yet, but it will re-compile and re-validate all the samples from the latest commit of the MEI source and customizations.

http://dev.simssa.ca:8080/jenkins/job/music-encoding-initiative/48/console

raffazizzi:
RE: Jenkins, cool, but it would only kick off after the PR is merged right? Or is there a way to test the PR on Jenkins before merging?

ahankinson:
I think Jenkins can be configured to test PRs and report back on their status. Failing that, there's also TravisCI

PR #237, opened 2015-06-26
New customization testing layout

ahankinson:
This pull request updates the build script to perform testing on provided sample customization files. A new `tests` directory was added with sub-folders to hold test files for each customization. When `build.sh` is called, the build script will validate the corresponding test set in the `tests` directory against the built MEI customization schema.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/tests/.DS_Store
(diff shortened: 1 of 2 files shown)


lpugin:
This looks good to me.

PR #239, opened 2015-06-26
Fixing ODD validity

raffazizzi:
This PR introduces changes to the ODD to make it valid against the current TEI ODDs schema. Some of the changes are substantial, but the resulting schema will have only one real change (which itself is a fix, see below).
## Changes with _no consequences_ for the schema:
- associated latest ODD schema served directly from TEI servers instead of local copy. The URL always points to latest stable release of the schema, so that we know right away if we need to make changes for validity.
- `attDef` elements defining attributes with semiclosed lists were not valid against a schematron check that imposes either a datatype or a closed list. This was fixed by adding a generic "string" datatype.
- `attDef` elements  defining attributes without any datatype are not valid, so I added a generic "string" datatype. NB a lot of these need to be improved! Either with a token datatype (eg like `data.word` in TEI), or a more specific one, but this is for another bug fix / enhancement.
- datatype macros for shared attribute values (such as `data.ARTICULATION`) where defined with  `rng:choice`, which is fine, but the RNG documentation element (in a different namespace) is not allowed by the TEI schema. I replaced this with `tei:valList`, which is an element used specifically for describing attribute values. This is a more semantic element than `rng:choice` and allows using the same `tei:desc` element that is used for documentation elsewhere. This is a major change in the ODD, but it has no consequences for the schema.
## Change _with consequences_ for the schema (a FIX):

"One-or-more-value" datatypes are currently defined as separate datatypes. For example `data.SLURS` is defined with an `rng:content` of `rng:oneOrMore` occurrences of `data.SLUR`. This is not processed by Roma and currently these datatypes are broken in the schema generation. The change proposed here follows common practice in the TEI schema:
- only define single datatypes
- when specifying the datatype of an attribute, use `tei:datatype[@maxOccurs='unbounded']`.
  This means that "one-or-more-value" datatypes are not necessary and are therefore _removed_ in this PR. 

Now Roma processes the datatypes correctly and multiple values are allowed by the schema as originally intended.

NB Removing "one-or-more-value" datatypes (such as `data.SLURS`, `data.TIES`, `data.URLS` and many more) means that they will not be listed in the datatypes documentation of the guidelines, and will result in a change to the HTML and PDF guidelines. In my opinion this should not be too alarming unless known MEI customizations have been calling these datatypes directly, in which case they should switch to "single" datatypes with `tei:datatype[@maxOccurs='unbounded']`. A notification to MEI-L should be sufficient in my opinion.


raffazizzi:
Noted the failing build, looking into the problem

ahankinson:
Classifying as a bug, since as I understand it, this fixes ODD validation

raffazizzi:
@ahankinson we were validating against a local copy of an old TEI ODD schema. Our ODD is invalid against the latest TEI ODD schema. Here I'm also proposing to always validate from the latest published schema instead. If that seems like a good idea, we should probably remove out local copy (the PR doesn't do that for now).

lpugin:
How often and what type of changes should we expect to see in the ODD schema? If we validate against the latest one, we need to make sure we do suddenly have our schema not validating anymore because of some changes there.

raffazizzi:
If we keep the latest schema, we will only have to deal with small issues from time to time as soon as they arise. In my experience these are typically introduced by stricter schematron rules. If we stick to one version and don't update regularly, we would have to introduce larger changes at less regular times, such as the ones proposed in this PR.

We depend on the TEI Stylesheets and they evolve together with the ODD schema. It is in our interest to keep up with these changes to make sure that we always can generate schemata without errors or missing bits (as it was happening for "one-or-more-values" datatypes).

ahankinson:
I would go with @raffazizzi on this one, but we should make sure to test it first. Would an effective test case be one that hits `data.SLURS` or `data.TIES`?

ahankinson:
I would also like to hear opinions about this from others who are more experienced with TEI. @bwbohl? @kepper? @pe-ro ?

kepper:
This is beyond my scope in TEI – I'm happy to use, but don't oversee the implications at this level. One thing @pe-ro and I wondered about in NY is what would happen to semi-open lists. These lists allow you to recommend some values, but ultimately leave it to the user to write whatever he wants. However, oxygen hints to the suggested values, making it less error-prone than a fully open list. Are they unaffected, or will they go away?

raffazizzi:
@kepper the semiclosed list will work in the same way as before. The only difference is that now there is a "string" datatype defined for the custom values. This is sufficient for validity and does not change the current behaviour.

In general, unless I've done something wrong, there will be no change to the schema, except for values with repeated datatypes, which will now work (they where broken before as you and @pe-ro pointed out to me in NY).

pe-ro:
I think "NMTOKEN" is a better choice than "string" for the datatype of items in a valList.  This will work seamlessly in most cases.  Unless there's a way to qualify some values, however, this won't work for MEI datatypes like data.TEXTRENDITION where values and datatypes are intermingled rather than defined once for all the values in the list.

raffazizzi:
> I think "NMTOKEN" is a better choice than "string" for the datatype of items in a valList.

I agree. Though we should take care of that in another issue or pull request. Adopting "string" does not change the status quo, as that's what we were already adopting silently. Switching to NMTOKEN instead affects the schema and current usage, so it needs its own discussion, IMO.

raffazizzi:
> this won't work for MEI datatypes like data.TEXTRENDITION

`data.TEXTRENDITION` is a closed list, so there is no need to provide an alternative datatype: the list is the datatype (see [here](https://github.com/raffazizzi/music-encoding/blob/96167b2e9d397bf121f770cac9a064fe6339bcf8/source/specs/mei-source.xml#L2141)). 

The only reason there must be a datatype specified for semiclosed lists is to cover and regulate the "open" part of the list.

pe-ro:
If you look closely, you'll see that data.TEXTRENDITION is not really a closed list, at least not in the conventional sense.  Down near the end of the definition rng:data is used to "parameterize" certain values --

```
<rng:data type="string">
  <rng:param name="pattern"
  >(underline|overline|line-through|strike|letter-spacing|line-height)\(\d+\)</rng:param>
</rng:data>
<rng:data type="string">
  <rng:param name="pattern"
 >(letter-spacing|line-height)\((\+|-)?\d+(\.\d+)?\)</rng:param>
</rng:data>
```

The new TEI approach seems to disallow this kind of thing by insisting on the same datatype for every value in the <valList> and by not providing this kind of parameterization.

pe-ro:
My attempt at including a chunk of the ODD didn't work, so everyone should look at the mei-source file.

raffazizzi:
Oh I see. My bad! I missed the datatypes at the bottom. Let me look for a ODD-friendly solution.

pe-ro:
Thanks.  If the rng: namespace is no longer supported, I think the only solution might be created (at least) 2 different MEI datatypes -- one for the plain values and one for the parameterized ones -- then allow attributes to utilize values from these 2 dataypes.  But I'm also not sure of how to accomplish that either.

pe-ro:
I don't agree that using NMTOKEN is a change to the schema.  We've not been silently using "string", we've been silently using "NMTOKEN".  Making that explicit won't change anything.

pe-ro:
Is this the correct syntax for semi-open lists now? At least for those with values that don’t require parameterization?  --

<attDef ident="unit" usage="opt">
  <desc>indicates the unit used for a measurement of size or duration.</desc>
  <datatype>
    <rng:data type="NMTOKEN"/>
  </datatype>
  <valList type="semi">
    <valItem ident="byte">
      <desc>byte.</desc>
    </valItem>
    
  </valList>
</attDef>

raffazizzi:
I worry about attributes like `expan` that [had no datatype](https://github.com/music-encoding/music-encoding/blob/develop/source/specs/mei-source.xml#L9011). Those may as well be treated as strings and not NMTOKEN which, if I understand correctly, does not allow spaces. This changes the schema, I think.

So while I agree that NMTOKEN is better suited than string for a lot of the attributes we're updating, I think we should look at each case individually. I would be inclined to resolving this issue with string, which is safe enough, and then start working on which attributes should get NMTOKEN right away, but in a different place.

If not, perhaps a middle ground would be introducing NMTOKEN to datatypes only, but leave the others to review another time?

pe-ro:
Yes, those attributes without a datatype assume string values.  Value lists in MEI datatypes, on the other hand, assume NMTOKEN values.  There are only a few outlying cases that use other datatypes, such as nonNegativeInteger, token, decimal, positiveInteger, etc.  But in all of those cases, the datatype is already specified and shouldn’t be a problem.

raffazizzi:
Ok :) I'll add nmtoken to the datatypes and leave string elsewhere, then.

pe-ro:
I’m working through all the datatypes now.  I’ll send you the file ASAP.  I think we only need to change those places where we use <a:documentation>.  In all other cases, <rng:data type=”whatever”> works fine.

pe-ro:
One question, though: since <valList> doesn’t require @type, is @type assumed to be “string” when it’s missing?

raffazizzi:
`<valList>` has a default value of `open` for `@type`. So if it's not specified it's an open list.

pe-ro:
Thanks.

raffazizzi:
For the record, 3576a61 merges changes by @pe-ro. A list of changes is in the commit message.

c99501c introduces a change to get rid of `rng:list` in `data.DURATION.gestural`. This is not necessary for validity, but strengthens consistency with all the other changes from rng to ODD notation. 
To do this I had to introduce two partial datatypes (similary to what we've been doing so far):
- data.DURATION.gestural.prop for proportional duration
- data.DURATION.gestural.pat for the ppq, beats, or seconds pattern
  Happy to change the names if you have better suggestions.

Now I'll figure out what's wrong with the build.

raffazizzi:
It looks like the master branch of the TEI Stylesheet is ZAPping too many elements and classes from our ODD. The same problem exists with the one on master now. Possibly Travis CI should use the latest release tag instead? I think @sebastianrahtz uses the master branch for development and can become unstable.

raffazizzi:
Ok this is going through now. Should we merge? :)

pe-ro:
If the ODD file parses and the customization service works as it should, then I say we're ready to proceed.  But, I'd like to wait for @lpugin to approve it as well.

pe-ro:
Seeing Travis in action, so to speak, I'm wondering what should happen at the next release for changes that break backward compatibility.  We should have a stylesheet that migrates the example files to the new schema, but where does it "plug into" the build process?

ahankinson:
Since 'tags' are a snapshot of the repo as it was at a given point in time, then any of the samples or tests that are included in that tag should build against that version in perpetuity. So if we're releasing MEI 2015 we (on the develop branch):

1) Roll the customization into mei-source.
2) Validate the examples and update where needed. This can be manually or with an XSLT.
3) Once everything is built and tested we merge `develop` to `master` and tag it.

I'm hoping we can gradually build up a test set in 'tests' that will validate the various customizations too.

ahankinson:
I will take over merging it, but I'll wait for Laurent to say OK too.

sebastianrahtz:
I  am a bit distressed I seem to have broken MEI builds;I am in the process of implementing Pure ODD datatypes. Is there anything still wrong?

raffazizzi:
All is good! Thanks for the quick fix @sebastianrahtz

lpugin:
Looks good to me, but considering the topic, I would rather ask for @pe-ro 's OK.

ahankinson:
Ha! He already asked for yours. ;)

https://github.com/music-encoding/music-encoding/pull/239#issuecomment-118115399

pe-ro:
Who wants to hit the button?  The sooner it's done, the sooner we can get to work on the next release.

ahankinson:
I'll do it.

pe-ro:
Unfortunately, it appears that the customization service doesn't work.  It produces an RNG, but all new or replaced datatype declarations (data.BETYPE and data.ACCIDENTAL.EXPLICIT, for example) result in empty definitions.

ahankinson:
Yes, the customeization service does not work just yet. I'm working on an update... Sorry for the delay!

(PS: Anyone out there who wants to help, get in touch...)

pe-ro:
Just checking on its status, didn’t intend to rush you.

PR #240, opened 2015-06-27
Fixed: Updated paths in musicxml2mei script

ahankinson:
Previously the musicxml2mei XSLT script pointed to the 2.1.0 release on Google Code.

This commit fixes it to point to 2.1.1. I have also changed it to the GitHub version, but I think this should probably be temporary, since there should be a more accessible version of the .rng officially published for each release.


lpugin:
This looks good to me

PR #244, opened 2015-06-28
New Contributing Guidelines and other documentation

ahankinson:
This Pull Request adds a couple of new files in the root of the MEI source directory: A README, intended to provide information on how to build and customize MEI; a CONTRIBUTING, which attempts to set out the structure and nature of contributions; and a LICENSE, which basically copies the MEI ECL from the source file. The LICENSE also updates the date, and assigns the copyright to the new MEI Board since the Council no longer exists.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,47 @@
+# Music Encoding Initiative Contribution Guidelines
+
+This document sets out the methods and practices for contributing to the Music Encoding Initiative.
+
+# Contributors
+
+The Music Encoding Initiative (MEI) schema is developed and maintained by the MEI Technical Team. Membership in this team is open to anyone and everyone who expresses an interest, and has a demonstrated ability to contribute. There are two co-leaders of the Technical Team, who act as administrative representatives to the board and manage the technical infrastructure of the project (e.g., GitHub administrative access).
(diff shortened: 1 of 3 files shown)


ahankinson:
FYI, you can browse the repo as it would appear after this pull request by clicking on the most recent commit and choosing "Browse files" in the header. This should take you to this page:

https://github.com/music-encoding/music-encoding/tree/de74e80124358992762a77955f41b78bca23df45

lpugin:
Looks good to me as a first version.

lpugin:
Don't we want this in the master already? This does not follow the workflow we agreed on, but I think we need people to see something when coming to GitHub. Otherwise we need a minimal README in master that links to it

ahankinson:
Currently `develop` is set to be the primary branch, so it's the one that people will see when they first come to GitHub. I did this to help the transition, since most of the activity will be on `develop` until we get 2015 out the door.

I will still need to fix the typos that Raffaele pointed out. In general, I think we should wait to have an agreement to merge before merging it in -- I would have preferred to fix these on the pull request before merging it.

lpugin:
OK, if develop is the default branch, then this solves the issue.

Sorry, I realized the merge was a bit quick.

Issue #253, opened 2015-07-26
Create a generic ornament element


raffazizzi:
As suggested in a [MEI-L discussion](https://lists.uni-paderborn.de/pipermail/mei-l/2015/001612.html), I tried using `<dir>` for encoding ornaments that are not mordents, turns, or trils, but it lacks some attributes that would be quite helpful when encoding ornaments, such as `@accidupper` and `@accidlower`. Would it make sense to consider introducing a generic `<ornament>` element for ornaments that are not mordents, turns, or trills?

I would imagine it would have attributes that all three current supported ornaments share, plus duration attributes to support ornaments that extend like a trill. The `@form` attribute instead could probably be skipped because it wouldn't be possible to pre-define forms of a generic ornament. The `@altsym` attribute could be used to refer to a specific glyph and an expansion should probably be provided in the stream of events like @pe-ro recommended.

For example a Doppelt-Cadence now can be encoded as:

``` xml
<dir tstamp="1">
   <symbol ref="#my_doppel-cadence_symbol"/>
</dir>
```

With a generic ornament it would be encoded very similarly:

``` xml
<ornament altsym="#my_doppel-cadence_symbol" tstamp="1" />
```

But, unlike `<dir>` it would also allow me to add an accidental:

``` xml
<ornament altsym="#my_doppel-cadence_symbol" tstamp="1"  accidupper="s" />
```

Thoughts? Here is a possible ODD definition:

``` xml
<elementSpec ident="ornament" module="MEI.cmnOrnaments">
      <desc>An element indicating an ornament that is not a mordent, turn, or trill.
      </desc>
      <classes>
        <memberOf key="att.common"/>
        <memberOf key="att.facsimile"/>
        <memberOf key="att.ornament.log"/>
        <memberOf key="att.ornament.vis"/>
        <memberOf key="att.ornament.ges"/>
        <memberOf key="att.ornament.anl"/>
        <memberOf key="model.ornamentLike.cmn"/>
      </classes>
      <content>
        <rng:empty/>
      </content>
      <constraintSpec ident="ornament_start-type_attributes_required" scheme="isoschematron">
        <constraint>
          <sch:rule context="mei:ornament">
            <sch:assert test="@startid or @tstamp or @tstamp.ges or @tstamp.real">Must have one of
              the attributes: startid, tstamp, tstamp.ges or tstamp.real</sch:assert>
          </sch:rule>
        </constraint>
      </constraintSpec>
      <remarks>
[...]

PR #254, opened 2015-08-10
New generic ornament element and documentation

raffazizzi:
This PR introduces a new `<ornament>` elements as proposed and discussed in issue #253. It also contributes related documentation to the Guidelines.

In more detail:
- Added `elementSpec` for `ornament` to `MEI.shared` module
- Added related attribute classes following the typical structure for element attribute classes addressing four domains: logical, visual, gestural, and analyitical
- The `elementSpec` and the `classSpec` elements have been added in alphabetical order of `@ident`, following current (inferred) convention
- Added a new subsection to Chap 1 of the Guidelines (shared element) documenting `<ornament>`
- Added a new section to Chap 8 of the Guidelines (CMN ornaments) documenting use of `<ornament>` in a CMN context. This includes an example from Bach's table in the Klavierbüchlein für Wilhelm Friedemann Bach. The image is from Wikipedia and is in the public domain.


ahankinson:
Can you write a couple tests just to double-check that this works and validates?

A refresher: You can write MEI fragments in `tests/mei-all_anyStart/<your testname here>.mei` which will automatically get tested against the mei-all_anyStart.rng. If you want to test it against another customization, you can put it in the appropriate folder in the tests/ directory and it will be automatically run and tested.

PR #255, opened 2015-08-24
add missing note

bwbohl:
Measure 4 on the first staff missed the third note of the tuplet. This
commit adds the note, nevertheless it still misses an xml:id.


ahankinson:
Thanks!

Issue #259, opened 2015-09-21
Issue with grace + tuplet conversion


rigaux:
This score: 
http://neuma.huma-num.fr/opus/timbres:cdc:0037
MusicXML: http://neuma.huma-num.fr/rest/timbres/cdc/0037/score.xml
MEI: http://neuma.huma-num.fr/rest/timbres/cdc/0037/mei.xml

Measures 5 and 9: each time, one note from the tuplet is missing, as well as the grace note

ahankinson:
I believe this one is fixed now.

pe-ro:
Partially fixed, I believe.  The stylesheet raises an error --

"Description: The attribute axis starting at an attribute node will never select anything
Start location: 10431:0"

and the problems reported in #258 aren't completely resolved; that is, the beamed grace notes in m. 4 and m. 5 aren't beamed properly.

kepper:
will look into this asap. Paid no attention to #258 (yet).

Am 28.09.2015 um 19:42 schrieb pe-ro notifications@github.com:

pe-ro:
Sorry, didn’t mean to get ahead of you.

## 

p.

kepper:
All good. I hadn't noticed #258 as separate issue, so it's good you pointed me at it :-)

Am 28.09.2015 um 19:49 schrieb pe-ro notifications@github.com:

ahankinson:
This issue was moved to music-encoding/encoding-tools#5

PR #260, opened 2015-09-26
improved handling of tuples and graces

kepper:
this should address #259 by @rigaux


ahankinson:
This looks good and I'll merge it into develop. Just a few admin notes (because this process is new):
- It's easier on your end if you create a special branch just for this fix. So, instead of putting the fix on `develop`, put it on `fix-tuplet-handling` on your end.
- "Fixes" is the magic word to automatically close an issue from the pull request: "Fixes #259".

PR #261, opened 2015-09-28
Fixed: Allow @visible on <accid>

ahankinson:
This commit adds the @visible parameter to `<accid>`, providing a way to record an accidental that is present, but hidden. Note that this is not to be confused with the function of cautionary accidentals (i.e., accidentals that would normally be shown but for the convention that they were already shown in the bar.

Also included in this commit is a test case to check that `<accid>` takes @visible.

Note that this pull request is based on the latest from the `develop-perry` branch and, if accepted, should be merged into this branch before being merged into `develop.`


kepper:
Just to understand: Is this the element solution matching the purpose of @accid.ges? If so, could we name it more consistently, on either side? Or at least amke it more explicit?

pe-ro:
Exactly what is the need for `@visible` on `<accid>`?  When you say this provides "a way to record an accidental that is present, but hidden", under what circumstances would/does this occur?  Playing dumb for the sake of argument, if the accidental is invisible, how do you know that it's "present"?

zolaemil:
I'd like to second the above questions. We use @accid.ges on every element to record the pitch, and use @accid to record the "visual presence", if you like.

ahankinson:
The issue I'm trying to solve here can probably be easily summed up with a picture:

sharp-non-vis

In this case, I have:
- A preceding sharp before the bar line
- A supplied sharp (above the staff)
- An "invisible" sharp after the barline, added because it was needed (in Sibelius) to make it a sounding F#, but explicitly hidden because it does not follow the convention of showing cautionary accidentals after each bar line.

This is also using the `<accid>` element, and not `@accid`. As it is, I've just discovered `@accid.ges` is not currently allowed on `<accid>`, so maybe that's the best way forward, rather than `@visibility`?

pe-ro:
Is the purpose of adding the sharp to be able to go back to Sibelius for some reason, such as the creation of a playable version/MIDI file?  Are you anticipating the loss of the supplied accidental (which I presume you're encoding using <supplied>) when translating the MEI back into Sibelius' internal format?  

I can't tell from the Sibelius documentation of the "Add Ficta Above Note" plug-in if ficta accidentals are represented differently than "normal" ones.  If they're different somehow, then neither `@visible` nor `@accid.ges` are necessary on <accid> since <supplied><artic/></supplied> can serve as the "trigger" for ficta in Sibelius.  If they're not, then round-tripping Sibelius->MEI->Sibelius without some kind of Sib-specific kludge is much more difficult.  I hesitate to add `@accid.ges` to <accid> just to work around Sibelius' inferior representation.  

It might be worth considering revising the values of `accid/@func` to differentiate/include ficta. But this is simply adding a shortcut version of <supplied><accid/></supplied> and means breaking the current 'editorial' function into multiple values that target sub-functions of the editorial process, like error correction, normalization between sources, etc.  Adding 'ficta' to this list immediately causes confusion between accidentals coded as ficta in this manner vs. those coded within <supplied>.

I'd like to better understand/see what the ficta plug-in does before going further.

ahankinson:
I will be among the last to defend Sibelius' internal representations of music, but in this case I think there's an important distinction.

Supplied accidentals are not directly tied to any Sibelius objects -- they're placed visually, but with no bearing on the actual note that is below it. Placed ficta are just symbols, so they don't even have a bearing on the actual pitch unless (as this person has done) you place the ficta visually, then hide the accidental on the note so you don't get both. So the supplied accidental is not what I'm worried about right now.

There is actually no way to tell whether an accidental has a visible or invisible accidental, using the `<accid>` child element. If I put `<accid accid="s">` then, in my reading of the spec, this means that it's an explicitly written and visible accidental. If I put `<accid accid.ges="s">` then that means that it's an accidental that is performed, but not written.

In other words, without `@visible` _or_ `@accid.ges` then it's not possible to tell whether an accidental was actually written in the original.

raffazizzi:
> Supplied accidentals are not directly tied to any Sibelius objects -- they're placed visually, but with no bearing on the actual note that is below it.

I can see how this can cause problems when converting from Sibelius to MEI, but in itself is not sufficient to justify encoding two accidentals in MEI. In the example that you posted, there is only one accidental for what MEI is concerned and it's the ficta one. I think your final output should end up with one accidental, something like:

``` xml
<note><accid accid="s" place="above" func="edit" /></note>
```

but it might take more than one step to get there, e.g. two `<accid>`s with enough info to merge them into one at a second pass.

Having said that, I can see how `@accid.ges` can be useful on `<accid>`: the performed accidental may be different from the written one (for some unforeseeable reason). But I wouldn't use it to encode an invisible accidental; the accidental is very visible in this case.

pe-ro:
I think it depends on what you think of as the _original_.  Maybe I'm just not getting it, but it seems to me that what you're trying to do is reconstruct the Sibelius file _before the plug-in was invoked_ -- the **original** original as it were.  But, after applying the "Add Ficta Above Note" plug-in, isn't there, in fact, _no accidental actually on the note_ in the Sibelius file?  If you want to record an accidental on the note, can't the supplied accidental be interpreted as an explicit one?  This would presumably restore the _original_ original.  The plug-in could be re-run with the same MEI result as before.

I suspect this is not totally satisfying in your case because the MEI doesn't directly represent the _original_ original without the MEI -> Sibelius conversion step described above.  But, adding features to the MEI representation to compensate for deficiencies in Sibelius would not set a good precedent.

ahankinson:
It's really very simple. Like I said, forget about the ficta.

sharps

In this example, what would the last quarter note have as a value if I wanted to use the `<accid>` element? By my reading:

```
<note pname="f">
   <accid accid.ges="s" />
</note>
```

which is currently not allowed. Either that, or:

```
<note pname="f">
   <accid accid="s" visible="false" />
</note>
```

Which is also currently not allowed. I would take this to mean that, according to the conventions of CMN there _should_ be an accidental on this note, but it has been hidden.

raffazizzi:
:+1: Ok, yes - that works :)

pe-ro:
What about --

```
<note pname="f">
  <accid func="gestural" accid="s"/>
</note>
```

ahankinson:
But that would a) add a third allowed value to `@func` that may not make sense on all the other elements it belongs to, and b) be different than an established pattern for recording the same thing on `<note />`.

ahankinson:
I was just looking at the source -- I think that the `att.accidental.performed` and `att.accid.ges` groups may be collapsed into `att.accid.ges` -- they essentially record the same thing, and the `att.*.ges` is a consistent pattern used in the rest of the source.

ahankinson:
I just updated the source and tests to allow `@accid.ges` on `<accid />`

pe-ro:
The classes att.accid.ges and att.accidental.performed are _not_ the same thing.  The att.accid.ges class are those attributes that apply to the accid element in the gestural domain.  The att.accidental.performed class defines attributes that encode gestural accidentals.  The first group (along with all the other att.[element].ges classes) exist so that attributes can be turned off en-masse for a particular element.  The second group actually defines the attributes.  As you can see, <accid> is a member of att.accid.ges, but att.accid.ges is empty, meaning that <accid> can't have an accid.ges attribute.  Collapsing these two classes would disallow handling `@accid.ges` independently on different elements.

If you want to allow <accid> to have an `@accid.ges` attribute, then make att.accid.ges a member of att.accidental.performed.

ahankinson:
OK, I've made the change. How does that look?

zolaemil:
I can see the point this, i.e. to handle the accid and accid.ges attributes on the same element but just for the sake of argument, one could do:

```
<note accid.ges="s">
 <accid accid="s"/>
</note>
<note accid.ges="n">
 <accid accid="n"/>
</note>
<note accid.ges="s">
 <accid accid="s"/>
</note>
<note accid.ges="s">
</note>
```

In fact, I prefer to keep @accid.ges on the same level as @pname and @oct and treat it effectively mandatory for in order to encode the pitch, regardless of the written accidental. This way the encoding is explicit about the otherwise implied convention of not repeating accidentals within measures, and also makes it trivial to deal with cautionary accidentals.

lpugin:
I don't think it is correct to have both accid (attribute or element) and accid.ges when they are the same since the definition of accid.ges is "records the performed pitch inflection when it differs from the written".

I am lost with the attribute class definitions, but I can certainly see the use of having accid.ges in /accid.

lpugin:
To follow-up with Andrew's example, let's imagine we have a second source that does not have the sharp on the third note. You would end up for the last two notes with:

```
<note pname="f">
    <app>
        <rdg id="src1">
            <accid accid="s" />  
        </rdg>
        <rdg id="src2"/> 
    </app>
</note>
<note pname="f">
    <app>
        <rdg id="src1">
            <accid accid.ges="s" />  
        </rdg>
        <rdg id="src2"/>  
    </app>
</note>
```

zolaemil:
Forgive me for not escaping the code in my message correctly (my phone's input isn't ideal for this). For the record here it is:

```
<note accid.ges="s">
  <accid accid="s/>
</note>
<note accid.ges="n"/>
  <accid accid="n"/>
</note>
<note accid.ges="s"/>
  <accid accid="s"/>
</note>
<note accid.ges="s"/>
</note>
```

Laurent,

> I don't think it is correct to have both accid (attribute or element) and accid.ges when they are the same > since the definition of accid.ges is "records the performed pitch inflection when it differs from the written".

You're right...

In this case, provided we agree on the benefits I described above, would it be possible to allow @accid.ges to describe the performed pitch, regardless whether it differs from the written one?

ahankinson:
I see them as mutually exclusive: `@accid.ges` is used **only if** `@accid` is not used, and vice versa. Encoding both of them sets up a situation where one will need to supersede another; i.e., if both `@accid.ges` and `@accid` are present, then `.ges` takes precedence over controlling the visibility of the accidental.

I don't think we should get into a practice where a child accidental element should control the behaviour of an attribute on the parent. I see them as alternate, but separate, ways to encode the information and you should choose one or the other. I chose the child element because it allows me to add things like positioning, color, etc. should it be needed.

pe-ro:
I think Zoltan’s suggestion is a good one -- `@accid.ges` describes the performed pitch. Its value can often be inferred from `@accid` and/or the prevailing key signature.  The addition of the language “when it differs from the written” was intended to convey the reason why `@accid.ges` is optional and when it would be a good idea to use it.  I’d be happy to drop this phrase or replace it with something better, but it still needs to be succinct as this description is used as a “tool tip” in Oxygen.

[Ed.: AH removed the e-mail reply at the bottom of this message]

ahankinson:
I strongly disagree.

`@accid.ges` is currently the only way of telling whether an accidental is visible or not. Having both `@accid` and `@accid.ges` used sets it up so that we have to infer a negative -- that is, if `@accid` is alone, it's visible, but if both `@accid` and `@accid.ges` are there then we hide it? And isn't it redundant to have both have values of `s`?

Inferring whether an accidental is shown from the key signature will not work in all cases, since you would _also_ need to look at all previous notes in the bar to infer whether the accidental is shown or not, according to convention. And this convention is not always consistent.

It is much, much simpler to say that `@accid` is used to record a written accidental, and `@accid.ges` is used to record a performed, but not written, accidental, rather than set up a situation where two attributes (on possibly different elements) need to be evaluated to determine whether or not the accidental is shown. As far as I can tell this is true in both cases of software writing and hand encoding.

lpugin:
It is too easy if you change the rules during the game ;-) If this should be the rule, then we also need to fix our examples. I might be wrong, but I think that most of them never have accid and accid.ges at the same time. This is quite of a change, actually much more than adding accid.ges in /accid.

Also, I am not sure I can see an easy way to encode the example I gave with an /app and /rdg. What would be your recommendation for such a case?

kepper:
I fully support Andrew, but I'm not sure if we're all discussing the same thing. I believe the problems come with something like

<note accid="s" accid.ges="f"/>

But in this case, I would still follow the rules Andrew just stated, that is: we see a sharp, but hear a flat. I have no clue why this could happen, but at least I know of same cases where sharps are used to 'naturalize' flats and vice versa, so something like

<note accid="s" accid.ges="n"/>

would actually make a lot of sense in these cases.

Am 29.09.2015 um 17:17 schrieb Andrew Hankinson notifications@github.com:

pe-ro:
I agree -- `@accid` is for a written accidental, `@accid.ges` is for a performed accidental.  But, visibility is not a factor.  There is no implication re: visibility when both are present.

ahankinson:
OK, so we're back to adding `@visible` on `<accid />`? And a new attribute `@accid.visible` on `<note />`?

pe-ro:
`@visible` on `<accid>` ain't so bad, though I would argue that it's slavishly explicit.  `@att.visible` on `<note>` is abhorent.

ahankinson:
Without _some_ indication of visibility, how would we encode whether an accidental was visible or not? Especially if it doesn't correspond to the conventions of CMN. I don't see that as slavishly explicit, any more than recording how many lines on a staff there are (for example).

There was a mistake in my first version of the post -- `@accid.visible`, not `@att.visible`.

kepper:
but why do we need visibility? If something is written, it is visible. If something is (only) audible, it is not visible. If it's neither written nor audible, it's just not there. Why visible? I believe the combination of @accid and @accid.ges fully supports all cases.

However, Laurent is right that the lack of @accid.ges doesn't necessarily say that it's not audible. MEI has many places to hide this information (keysig being one of the more easy to access). In essence, sometimes this information about the sound of a note is only implicitly contained, sometimes (when @accid.ges is used consistently throughout the file), it is explicitly contained. Of course it's convenient to keep that information implicit when your project has no interest in that kind of information, but I see that this can make life much harder for others. I'm not sure if we can enforce an explicit encoding, but at least we need to be more clear about this in the Guidelines, and we should come up with some best practice recommendations.

I still don't see @att.visible on <note/>, but I'm more than happy to agree on @accid.ges on <accid/>.

Am 29.09.2015 um 17:26 schrieb Andrew Hankinson notifications@github.com:

raffazizzi:
Shouldn't we be encoding _the reason_ why an accidental is visible or not, rather whether it is visible? `@accid.ges` is used to encode performed accidental (I don't have an issue with it confirming or contradicting `@accid` because they're on two different domains) and that may sometimes correspond to presence/absence of the accidental on the page. So @ahankinson's example is better encoded with `@accid.ges` than `@visibility` because one is clearer about the reason for the encoding. And @lpugin's example needs it too, as he showed.

So yes, we need `@accid.ges` on `<accid>`.

Are there other cases where we'd need to be explicit about visibility? If yes, let's find the reason and encoded that, not the visibility.

pe-ro:
+1 re: Jo's comments.  I too have no problem w/ `@accid.ges` on `<accid>`. I will reluctantly accede to `@visible` on `<accid>`, but `@accid.visible` is ugly and unnecessary.

ahankinson:
BTW, I would prefer just `@accid.ges` on `<accid />` too. It's nice, tidy, symmetrical, the only change is allowing an existing attribute on an existing element, and it's semantically meaningful ('written' vs 'performed' rather than a strict on/off for a renderer).

The second issue of `@accid` and `@accid.ges` on the same note is separate. I would argue that it's bad practice since it's both redundant and confusing (no offence, @zolaemil! :)

ahankinson:
I've changed my mind about `@visible` -- I don't think it's necessary IF we use `@accid` and `@accid.ges` to record the difference between written and performed accidentals.

kepper:
Am 29.09.2015 um 17:40 schrieb Andrew Hankinson notifications@github.com:

> The second issue of @accid and @accid.ges on the same note is separate. I would argue that it's bad practice since it's both redundant and confusing (no offence, @zolaemil! :)

I think this is true if they take the same value: @accid.ges should – if present – default to the value of @accid. However, there are cases when they contradict, so it must be possible to specify both independently – as a last resort, and only by clearly separating the different domains.

ahankinson:
Agreed. We wouldn't restrict it by rule since you _may_ want to do it, but this would be a corner case and not standard practice.

zolaemil:
1. I may not have made myself clear enough on the original proposition: I think allowing `@accid.ges` on `<accid>` is a good idea.
2. re allowing both `@accid` and `@accid.ges`  in the same time: redundant? Maybe. Confusing? Maybe my description, but certainly not what I meant to suggest. `@accid` is visual `@accid.ges` is performed.

The performed performed pitch of `<note pname="4" oct="4"/>` can, and should be calculated from the context (key signature [+ preceding notes]), the same way as the performed pitch of `<note loc="0"/>` can and should be calculated from the context (clef + key signature [+ preceding notes]). Note that dependency on preceding notes are subject to convention.

Adding redundancy to the encoding means context-independence: using my suggested definition the performed pitch of `<note pname="4" oct="4" accid.ges="n"/>` is unambiguous regardless of its context. 

Regarding the visual domain:
`<note pname="4" oct="4" accid="n"/>` have a natural sign. 
`<note pname="4" oct="4"/>` does not have any accidental. 
Note that in these examples there is no _explicit_ statement about the performed pitch - implicit information is encoded in the visual domain.

Whether the visual and performance domain attributes can contradict each-other: I am sure it's not hard to assert this in schematron if need be, but I am not sure it's even necessary, as others pointed out above.

I _like_ this kind of redundancy. (If it even counts for one - depending whether we allow visual and performed attributes contradict each-other).

For the record, I _do not_ like confusion. Credit to every criticism regarding that. I also appreciate the argument for backward compatibility, should this be the reason for dismissal.

pe-ro:
I'm in total agreement w/ @zolaemil.

ahankinson:
We should wait for agreement to merge, since I needed to update some of the tests because they were not passing. I'll commit these changes directly to `develop-perry`.

ahankinson:
I'm trying to make sense of @zolaemil's comment. 

I really, really do not think we should go down the road of relying on software to calculate context, since context depends on rules, and the first rule of music encoding is that if there's a rule, it _will_ be broken. I can think of scores that I have sung where the copyist chose to insert an accidental that was not needed, or in the case of my first example (ignoring the ficta!) where the engraver chose to hide an accidental that would have normally been displayed. Most music may follow convention, but it's exactly the exceptions that we need to preserve, and it's the exceptions that software will almost always get wrong.

Using both `@accid` and `@accid.ges` at the same time, unless it's for the exceptions that Johannes gave, gives us little (since they will have the exact same value 99.9% of the time) but removes a lot (since we can no longer tell the difference between a performed AND visible accidental or a performed but NOT visible accidental).

th-we:
@ahankinson concerning distinction of performed + visible / performed + invisible: My understanding was that if `@accid` is present, we have a visible accidental, if it's absent, there's no visible accidental, regardless of whether `@accid.ges` is present or not.  That's at least how I understood it on `<note>`. Is this understanding wrong or is there any reason why this couldn't be made the rule?

zolaemil:
I am also trying to make sense of your comment, @ahankinson. Seems we are on two completely different pages. I'll give it another go:

Using both `@accid` and `@accid.ges` at the same time does not remove any clarity:
 performed AND visible accidental:
`<note pname="4" oct="4" accid="s" accid.ges="s"/>`

performed but NOT visible accidental:
`<note pname="4" oct="4" accid.ges="s"/>`

ahankinson:
OK, but you will (99.9% of the time) have the same value for both `@accid` and `@accid.ges` in your first example, right?

zolaemil:
Yes. In those 99.9% the mere presence of `@accid` could be interpreted "`@accid.visible="true"`". In 00.1% it even contains information.

ahankinson:
Right, but as Johannes pointed out, these are highly exceptional cases where you might want to record the difference between the written and performed accidental. It's very often that a written accidental will differ from a performed accidental (e.g., the instances where convention holds), but it's very rare where the performed accidental and the visual accidental will be different accidentals.

So we save the presence of _both_ for the rare cases, and use a general 'either/or' approach for almost all others.

ahankinson:
@th-we My interpretation is that if `@accid` is present, it's a visible _and_ performed accidental. If `@accid.ges` is present it's an unwritten but performed accidental.

lpugin:
It seems that the schema discussion has reached a consensus with the addition of accid.ges in accid, but not the discussion on the guidelines. I personally don't think it would be a good idea to change from recommending to have accid.ges only when it differs from accid to a solution where accid.ges is provided regardless of accid. This would indeed generate a significant amount of redundancy in the data, plus create some backward incompatibility (even though the word is a bit strong) with existing data. I do not see a good reason to change this.

pe-ro:
Leaving aside the idea of visibility for a moment (because it's a red herring; that is, a diversion from the main issue), I think we're all saying pretty much the same thing --

`<note pname="f" oct="4" accid="s"/>` indicates a written accidental, while 
`<note pname="f" oct="4" accid.ges="s"/>` indicates a performed accidental

When both attributes are used, (`<note pname="f" oct="4" accid="s" accid.ges="s"/>`) there is a very great probability that both values will be the same, but, while it increases the size of the XML file, I don't believe the redundancy has a deleterious effect because the attributes are concerned with different domains -- `@accid` in the visual and `@accid.ges` in the gestural domain.  It would be incorrect, however, to use both attributes where there is no written accidental.

An either/or approach is valid as well, since the missing one must be handled as though it has some "default" value.  For example, `<note pname="f" oct="4" accid="s"/>` implies `<note pname="f" oct="4" accid="s" accid.ges="s"/>`.  However, `<note pname="f" oct="4" accid.ges="s"/>` expressly indicates the absence of a written accidental.

When both attributes are present but the values differ, it indicates a conflict between the visual and the performance domains.  Other than its low frequency, there's nothing special about `<note pname="f" oct="4" accid="s" accid.ges="n"/>`.

lpugin:
The size of the XML file is not the issue with redundancy but rather the fact the redundancy often creates maintenance issues. They are other cases in MEI where this should be addressed, I think, but it is not the main problem here.

Of course both approaches are valid. However, by not clarifying which one we use and by leaving it open we will generates some significant ambiguity. Having both accid and accid.ges encoded independently (encoding accid.ges regardless of the presence or absence of accid) makes it easy to read the sounding content of a file. You just need to look at accid.ges, (and since we are all lazy we will all like it).

Now if I read a file in C-major, and see this: `<note pname="f" accid="s">`. If I assume the prima practica (accid.ges only if it differs) then I would read it as f-sharp. Now if I assume the seconda practica, then I would read a f(-natural) since accid.ges is encoded independently and I have no reason to look at accid. Correct? We can probably fix it by expecting accid.ges on each note if accid.ges is encoded independently, including accid.ges="n" on every note, but otherwise I don't think it would work unambiguously.

zolaemil:
I think Perry's right, we can salvage both practices.

> <note pname="f" accid="s">
> ...
> Now if I assume the seconda practica, then I would read a f(-natural)

This is only if you assume @accid.ges defaults to "n". If it defaults to
"undefined", then there's no ambiguity: @accid.ges simply does not help you
in determining the pitch. Prima practica does however, and you can conclude
that the performed pitch is the same as the written one.

In other words, if the encoding does not provide @accid.ges at all times,
one should assert for the absence of both attributes when determining the
performed pitch.

2015-09-30 6:37 GMT+01:00 Laurent Pugin notifications@github.com:

kepper:
I think what Laurent is complaining about relates to the multitude of factors influencing the performed pitch. Consider

<note pname="b"/>

In seconda practica, it's a b natural. In prima practica, it's b natural as well. Looking at the key signature, though, it may end up as b flat. So you're not only enforced to look after another attribute (@accid), but also for a key signature at the begin of the file, a key change that might occur somewhere along the way to this note, preceding notes in the same measure, or, in case of ties, in one or more preceding measures. In older repertoires, it's not even required to tie those notes. Looking up all these places is tedious, and so I understand that from a software developer perspective, the proposed model is inconvenient.

However, I still believe it's the best model from an MEI perspective, and I wouldn't like to enforce the use of @accid.ges, as many projects just don't care about the audio domain in what they do. In the Freischütz project, I've written an XSLT that will automatically fill up an MEI file with all required @accid.ges, based on the rules sketched above. It's available from https://github.com/Freischuetz-Digital/Tools/blob/develop/14%20reCore/addAccid.ges.xsl. This stylesheet doesn't do the job for everyone, as it's tied to the Freischütz' custom model, but it should be fairly easy to remove these dependencies and use it in other contexts. I'm happy to clean everything up (including all the other XSLTs of this project) and move them to MEI, but not until December.

Also, assuming there's a convenient way to fill up a file with all these @accid.ges attributes, we still lack a mechanism to specify how @accid.ges is used in a given MEI instance. The choices here are 1) fully ignored, 2) used when necessary, 3) used consistently throughout the file. Looking at the data, it's not easy to identify the method in use without re-doing all the calculations. In essence, what we may need is some kind of indicator (preferably in <encodingDesc>) that informs about the practice in use.

Ideas?

Am 30.09.2015 um 09:11 schrieb Zoltan Komives notifications@github.com:

zolaemil:
> `<note pname="b"/>` 
> In seconda practica, it's a b natural. In prima practica, it's b natural as well.

Let me argue with this. _Assuming the encoding intends to record performed pitch_, and the current key signature is F-major, and there's no preceding b-natural, `<note pname="b"/>` is an incorrect encoding according to both prima practica and seconda practica:
1. prima practica [a] requires `@accid.ges` since the performed pitch is different from the written one, 
2. seconda practica [b] requires `@accid.ges` simply on the basis of recording performed pitch.

If the performed pitch is unimportant, both practices allow the omission of `@accid.ges` all together.

Your question is relevant, however: how to distinguish between prima practica encodings that care about performed pitch from ones that do not? (Assuming seconda pratica, the distinction is trivial, and one could restrict the schema to require the gestural attribute if need be). When the use of prima practica is practical (pun not intended), `<encodingDesc>` sounds like a plausible place to clarify this.

Terminology reference, just be be sure we talk about the same things
[a]: prima practica: current definition of `@accid.ges`: records the performed pitch inflection when it differs from the written accidental.
[b]: seconda practica - proposed definition of `@accid.ges`: records the performed pitch inflection

lpugin:
@zolaemil : yes, for the terminology with prima and seconda practica is what I meant. One thing I do not understand: if you mean that no accid.ges with the seconda practica means "undefined", what would you expect then? To look at accid? Then they are not encoded independently. Or would you expect all notes to have an accid.ges="n" (even in C-major with no accidental)?

zolaemil:
@lpugin I see the source of confusion now. Yes, I'd expect, according to seconda pratica to have @accid.ges="n" on every notes (given that the encoding is supposed to record the pitch).

But I see where the ambiguity lies if one tried to interpret the following:
`<note pname="c" accid.ges="n"/>`
According to seconda practica, this would simply be a performed c natural, according to prima practica though it is confusing because it implies that the written pitch is not natural which contradicts the absence of @accid.

Hmm, it may not be that easy to salvage both practices...

raffazizzi:
Why can't we have an undefined `accid.ges` take the value of `@accid`? Sure, `@accid` encodes the visual domain, but when dealing with the aural domain in MEI we generally rely on the the visual domain unless we need to clarify a difference. So while prima pratica is perhaps more in line with MEI's general assumptions between the graphemic and aural domain, I don't see a problem with using `@accid.ges` to restate, confirm, or update for the aural domain what `@accid` (or the key signature) already states.

lpugin:
If in the end you expect readers to use the value of accid when accid.ges is not provided, then this is indeed the prima pratica and what is currently recommended by MEI. Now if we start to have some MEI data sets where we have accid.ges on every note (including when duplicating accid) it probably means that we are going to see projects and tools relying on accid.ges only for generating sound. And these tools will inadequately play files that from the prima pratica (if not I do not see the goal of the seconda practica). And I don't think this would be a good situation. So I am not against the seconda pratica, but we should either have a way to indicate if accid.ges is provided at all, and if yes how it should be interpreted (prima or seconda pratica). Alternatively, we can switch to the seconda practica, but this needs to be a clear change with an update of the guidelines and of all our encoding examples.

raffazizzi:
> And these tools will inadequately play files that from the prima pratica

I'm not sure this is something we really need to worry about. There are a lot of cases where there is more than one way of doing the same thing in MEI; why start policing now? Example: using both `<tie>` and `@tie` - I wouldn't necessarily do it, but the guidelines don't restrict their concurrent use. It's the application's responsibility to either support both and decide which one to prioritize in case of conflict, or pick one and notify the user (I would do that by creating an "application profile" ODD to validate files before input).

I'd agree with you if we were designing a tighter version of MEI for a smaller set of uses (e.g. interchange).

lpugin:
`@tie` together with `<tie>` is once case of redundancy which I thought too requires clarification, but let's keep it for another thread.

It is not a question of policing, it is a question of _basic_ interoperability since we are talking about music notation fundamentals. If people just don't care, then I can see somebody five years down the road making some harmonic analysis on a large set of MEI files gathered from different places. Quite likely the person will not read all the guidelines (how is doing it?), an might very well assume that accid.ges gives the alteration for the sounding pitch since some of the files do it. With mixed practices, the results will be wrong if this is not noticed (all written accidental missed, this is pretty bad), or the person will complain that MEI is not consistent and have to write a sanitizer. We know that this type of situations arise, and we have to prevent them as much as possible, from now on.

For accid.ges, the guidelines are pretty clear (for once) and currently producing `<note accid="s" accid.ges="s">` is just wrong. It is a case where we I think we should add a schematron rule to check it. Unless we decide to change how it works.

lpugin:
OK, it actually seems that our examples have accid.ges also when identical to accid. Sorry for my last messages. I will fix the guidelines for accid.ges and we can definitely forget the prima practica.

kepper:
Don't be too fast. I'm still in favor of what has been identified as prima practica. For me, it's not a problem if that means we're mixing visual and gestural domains. It's actually more like connecting them, and I believe the separation of domains isn't easy in real life either. Instead of requiring @accid.ges independently of @accid to capture audio, I would like to have a standardized mechanism somewhere in <encodingDesc> that describes how specific parts of MEI are used. This could inform how @accid / @accid.ges is used, but also @tie vs. <tie> etc.

Of course this will result in even more documentation. I don't buy the argument that people don't read documentation anyway. It might be true that we're lacking short introductions into certain topics, but please keep in mind that MEI is a framework, not a strictly defined, ready-to-use format. I know this contradicts to the usual software development perspective, but from my experience in various projects, it's definitely a benefit. To model a problem is a scholarly task in its own right, and it requires the full flexibility MEI offers, and it requires a proper documentation to avoid tag abuse. How to generate, process and use this model and the conforming data with software is then just the second step – no less important, but we shouldn't short-circuit the first step for everyone just because of comfort. It'll make life easier for some, but definitely harder for others. I think the way to get there is to come up with different profiles of MEI, like TEI did with TEI Lite etc.

Just my 2 cent,
jo

Am 01.10.2015 um 22:22 schrieb Laurent Pugin notifications@github.com:

pe-ro:
At the risk of beating the proverbial "dead horse", I want to revisit this topic with the hope that we can finally put it to rest.

I think _Behind Bars_, in the section “Accidentals relevant to more than one part on a stave” provides instructive examples.  In the example at the bottom of p. 79, the notes F and C following the first are to be sounded as F-sharp and C-sharp even though they don’t carry a written accidental.  In this case

```
<note pname=”f” accid.ges=”s”/>
```

is necessary to indicate the sounded pitch of the note without having to backtrack to see if there was an accidental on a prior note in the measure. By not providing `@accid`, it also records the fact that there’s no written accidental.  If we were to insist that `@accid.ges` can only appear concurrently with `@accid`, then there must be a “none” value for `@accid`, e.g.,

```
<note pname=”f” accid=”none” accid.ges=”s”/>
```

but this is too redundant too much, isn’t it? :-)

In the example on p. 80, all of the accidentals are explicit, so the F-sharp in the 2nd chord should be encoded 

```
<note pname=”f” accid=”s”/>
```

From these examples I conclude that
- when `@accid` appears _without_ `@accid.ges`, it applies to **both the visual and gestural domains**; that is, a written accidental implies a performed one of the same value,
- when `@accid.ges` is used, it is redundant unless a contradictory value is needed; that is, when the source has a written accidental but the written accidental has a different meaning in this particular repertoire and should be performed differently, and 
- `@accid.ges` may appear alone to manage situations like the one presented in Gould, p. 79.

So, I propose the following steps
- review the Guidelines and in-schema documentation, making sure that it's clear that when `@accid` appears alone it applies to **_both**_ the visual and gestural domains,
- update the examples to remove `@accid.ges` when its value duplicates that of `@accid`,
- add Schematron to the att.accidental.performed class

```
<classSpec ident="att.accidental.performed" module="MEI.shared" type="atts" mode="replace">
  <desc>Attributes for capturing momentary pitch inflection in the gestural domain.</desc>
  <attList>
    <attDef ident="accid.ges" usage="opt">
      <desc>Records the performed pitch inflection.</desc>
      <datatype>
        <rng:ref name="data.ACCIDENTAL.IMPLICIT"/>
      </datatype>
      <constraintSpec ident="no_duplication_of_accidentals" scheme="isoschematron">
        <constraint>
          <sch:rule context="@accid.ges">
            <sch:assert test="not(@accid.ges eq @accid)">The value of @accid.ges cannot
              duplicate the value of @accid.</sch:assert>
          </sch:rule>
        </constraint>
      </constraintSpec>
    </attDef>
  </attList>
</classSpec>
```

to disallow any future duplication of `@accid` by `@accid.ges`.

N.B. I believe this should also clear up any confusion about the visibility of accidentals.

Of course, there are other ways this could be accomplished, but I think this solution is relatively easy to understand.  It also facilitates unambiguous implementation since the performed pitch of any note can readily by determined by looking at `@accid.ges` (if it exists) or `@accid` (if `@accid.ges` isn't present).

Sound good?

pe-ro:
Also, the MusicXML to MEI stylesheet should be revised so that it doesn't generate `@accid.ges` attributes that duplicate `@accid`.

ahankinson:
Can you open a separate issue for the stylesheet so we can track that task separately?

pe-ro:
Will do.

zolaemil:
How does this relate to https://github.com/music-encoding/music-encoding/issues/275? I assume we'll need to drop this then.

pe-ro:
Yes, this also resolves #275, which I will close momentarily.  It raises another question, however.  Can this change wait until v. 3.1.1 (the next, next release) or should it go into the current planned release?

zolaemil:
Thanks Perry! 

Just for future reference, and to make it easier to digest this thread, can we summarise the answer to the original question(s):

1.the question relating to the last note of 
sample1 s
is cleared up:

```
<note pname="f" accid.ges="s"/>
```

the absence of `@accid` meaning that there's no visible accidental, and the presence of `@accid.ges="s"` meaning that the pitch is raised.

2.
sample2
The accidental in the second measure is performed _and_ visible, albeit not beside the note it belongs to. Raff's proposal intended to clear this one up:

```
<note><accid accid="s" place="above"/></note>
```

pe-ro:
Yep, that summarizes it.

ahankinson:
For timing, changes to the documentation (Guidelines) can be made now. Changes to the _schema_, including schematron, should wait until the next release please.

PR #262, opened 2015-10-01
New: <line /> takes additional attributes

ahankinson:
This commit adds additional attributes to line, including @dur, @dur.ges, @tstamp, @tstamp2, @staff, @layer and @place

Includes a test case.

### DIFF
--- a/customizations/mei-all.odd
+++ b/customizations/mei-all.odd
@@ -4104,30 +4104,28 @@
             </attDef>-->
           </attList>
         </classSpec>
-        <!-- Include SVG, issue #53: add @prefix to explicitly reference line in MEI namespace -->
-        <elementSpec ident="line" module="MEI.usersymbols" mode="replace" prefix="mei_">
-          <desc>A line that cannot be represented by a more specific element.</desc>
+        <classSpec ident="att.line.anl" module="MEI.shared" type="atts" mode="add">
+          <desc>Analytical domain attributes.</desc>
+        </classSpec>
+        <classSpec ident="att.line.log" module="MEI.shared" type="atts" mode="add">
+          <desc>Attributes for describing the logical behaviour of a line.</desc>
           <classes>
-            <memberOf key="att.common.anl"/>
-            <memberOf key="att.color"/>
-            <memberOf key="att.common"/>
-            <memberOf key="att.facsimile"/>
+            <memberOf key="att.controlevent"/>
+            <memberOf key="att.duration.additive"/>
             <memberOf key="att.startendid"/>
-            <memberOf key="att.typed"/>
+            <memberOf key="att.timestamp2.musical"/>
+          </classes>
+        </classSpec>
+        <classSpec ident="att.line.vis" module="MEI.shared" type="atts" mode="add">
+          <desc>Attributes for describing the visual appearance of a line.</desc>
+          <classes>
+            <memberOf key="att.color"/>
+            <memberOf key="att.placement"/>
             <memberOf key="att.visualoffset"/>
             <memberOf key="att.visualoffset2"/>
             <memberOf key="att.xy"/>
             <memberOf key="att.xy2"/>
-            <memberOf key="model.graphicprimitiveLike"/>
           </classes>
-          <content>
-            <rng:zeroOrMore>
-              <rng:choice>
-                <rng:text/>
-                <rng:ref name="model.textphraseLike.limited"/>
-              </rng:choice>
-            </rng:zeroOrMore>
-          </content>
           <attList>
             <attDef ident="form" usage="opt">
               <desc>Visual form of the line.</desc>
@@ -4184,6 +4182,36 @@
               </datatype>
             </attDef>-->
           </attList>
+        </classSpec>
+        <classSpec ident="att.line.ges" module="MEI.shared" type="atts" mode="add">
+          <desc>Attributes for describing the performed components of a line.</desc>
+          <classes>
+            <memberOf key="att.duration.performed"/>
+          </classes>
+        </classSpec>
+
+        <!-- Include SVG, issue #53: add @prefix to explicitly reference line in MEI namespace -->
+        <elementSpec ident="line" module="MEI.usersymbols" mode="replace" prefix="mei_">
+          <desc>A visual line that cannot be represented by a more specific; i.e., semantic,
+            element.</desc>
+          <classes>
+            <memberOf key="att.common"/>
+            <memberOf key="att.facsimile"/>
+            <memberOf key="att.line.anl"/>
+            <memberOf key="att.line.ges"/>
+            <memberOf key="att.line.log"/>
+            <memberOf key="att.line.vis"/>
+            <memberOf key="att.typed"/>
+            <memberOf key="model.graphicprimitiveLike"/>
+          </classes>
+          <content>
+            <rng:zeroOrMore>
+              <rng:choice>
+                <rng:text/>
+                <rng:ref name="model.textphraseLike.limited"/>
+              </rng:choice>
+            </rng:zeroOrMore>
+          </content>
           <remarks>
             <p>The starting point of the line may be identified in absolute output coordinate terms
               using the <att>x</att> and <att>y</att> attributes. The attributes in the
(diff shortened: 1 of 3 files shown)


pe-ro:
Any thoughts on whether a `<constraintSpec>` similar to the following should be added to `<line>`?

```
<constraintSpec ident="line_start-_and_end-type_attributes_required" scheme="isoschematron">
  <constraint>
    <sch:rule context="mei:line">
      <sch:assert test="@startid or @tstamp or @tstamp.ges or @tstamp.real">Must have one  of the attributes: startid, tstamp, tstamp.ges or tstamp.real</sch:assert>
      <sch:assert test="@dur or @dur.ges or @endid or @tstamp2">Must have one of the attributes: dur, dur.ges, endid, or tstamp2</sch:assert>
    </sch:rule>
  </constraint>
</constraintSpec>
```

Line adds a new wrinkle that similar controlevent-like elements don’t present in that its start and end points can be defined in page coordinate terms.  So, `@x` and `@y` would have to be added to the set of start-type attributes and `@x2` and `@y2` would need to be added to the set of end-type attributes.  Even so, `<line>` may _not_ need to have this constraint – it could be perfectly legitimate just to record the existence of a line without specifying any details about its location.  I’m conflicted, hence the question.  @lpugin, any thoughts on how this might affect rendering?

Also, I’m not sure how your test of the new line attributes actually passed as `<line … staff=”2”/>` shouldn’t pass without a preceding `<staffDef n=”2”/>` somewhere.

## 

p.

ahankinson:
I'm OK with adding the schematron rule. That sounds awesome.

I think it passed because the validator does not process schematron rules. For that we would need a two-pass validation process, where we extract the schematron rules from the schema and then validate the file with these rules. 

It looks like there are existing [stylesheets](http://www.schematron.com/implementation.html) for doing this. Would this be useful to add to the build script?

pe-ro:
Yes, it would be useful to check validity against the schematron rules.

th-we:
@pe-ro If `<line>` is used inside a `<symbolDef>`, all those start/end attributes won't make sense. Maybe use this instead?

```
<constraintSpec ident="line_start-_and_end-type_attributes_required" scheme="isoschematron">
  <constraint>
    <sch:rule context="mei:line">
      <sch:assert test="@startid or @tstamp or @tstamp.ges or @tstamp.real or ancestor::mei:symbolDef">If used outside a symbolDef element, must have one  of the attributes: startid, tstamp, tstamp.ges or tstamp.real</sch:assert>
      <sch:assert test="@dur or @dur.ges or @endid or @tstamp2 or ancestor::mei:symbolDef">If used outside a symbolDef element, must have one of the attributes: dur, dur.ges, endid, or tstamp2</sch:assert>
    </sch:rule>
  </constraint>
</constraintSpec>
```

Question is whether user symbols are at all of any relevance or whether they should be deprecated in favour of something better that is yet to be developed.

raffazizzi:
@ahankinson it would be awesome to have the schematron scripts run during the build. We're finding ourselves relying on Schematron quite a bit!

th-we:
To _really_ cover that `<symbolDef>` case, there's problably more Schmematron required than I suggested above:

```
<constraintSpec ident="line_start-_and_end-type_attributes_required" scheme="isoschematron">
  <constraint>
    <sch:rule context="mei:line">
      <sch:assert test="@startid or @tstamp or @tstamp.ges or @tstamp.real or (ancestor::mei:symbolDef and @x and @y)">
        If used outside a symbolDef element, must have one  of the attributes: startid, tstamp, tstamp.ges or tstamp.real.
        If used inside a symbolDef element, must have both attributes: x2 and y2.</sch:assert>
      <sch:assert test="@dur or @dur.ges or @endid or @tstamp2 or (ancestor::mei:symbolDef and @x2 and @y2)">
        If used outside a symbolDef element, must have one of the attributes: dur, dur.ges, endid, or tstamp2.
        If used inside a symbolDef element, must have both attributes: x2 and y2</sch:assert>
    </sch:rule>
  </constraint>
</constraintSpec>
```

ahankinson:
I've updated the test to pass schematron validation, which is in a separate pull request ( #265 ).

ahankinson:
I can integrate further Schematron rules, but I'll let you guys figure out what's best. OK?

pe-ro:
`@startid`, `@endid`, `@x` and `@y` can be used in both symbolDef and music contexts. Even though it's a little more verbose, to make things simpler, I prefer two rules -- one for each context:

```
<constraintSpec ident="line_start-_and_end-type_attributes_required" scheme="isoschematron">
  <constraint>
    <sch:rule context="mei:line[ancestor::mei:symbolDef]">
      <sch:assert test="@startid or (@x and @y)">When used within the symbolDef context, must have either a startid attribute or x and y attributes.</sch:assert>
      <sch:assert test="@endid or (@x2 and @y2)">When used within the symbolDef context, must have either an endid attribute or both x2 and y2 attributes.</sch:assert>
     </sch:rule>
     <sch:rule context="mei:linenot([ancestor::mei:symbolDef)]">
       <sch:assert test="@startid or @tstamp or @tstamp.ges or @tstamp.real or (@x and @y)">When used outside the symbolDef context, must have a startid, tstamp, tstamp.ges or tstamp.real attribute or both x and y attributes.</sch:assert>
      <sch:assert test="@dur or @dur.ges or @endid or @tstamp2 or (@x2 and @y2)">When used outside the symbolDef context, must have an endid, dur, dur.ges, or tstamp2 attribute or both x2 and y2 attributes. 
      </sch:assert>
    </sch:rule>
  </constraint>
</constraintSpec>
```

th-we:
@pe-ro Something went wrong with the second `@context` - `not(` is in the wrong place.

Anyway - how would you sensibly use `@startid` in a `<symbolDef>`? In my understanding, `<symbolDef>` is for describing a graphical rendering, so it needs to be precise. Where would a `@startid` and `@endid` point to and how would you determine precise start and endpoints from the pointed to element? I think the only sensible elements to point to are other elements inside the same `<symbolDef>`, which would be `<anchoredText>`, `<curve>`, `<line>` and `<symbol>` (and lots of stuff that could be found inside `<anchoredText>`).

pe-ro:
Sorry, I mis-copied the assertion.  The constraint should be --

```
<constraint>
  <sch:rule context="mei:line[ancestor::mei:symbolDef]">
    <sch:assert test="@startid or (@x and @y)">When used in the symbolDef context, must have either a startid attribute or x and y attributes.</sch:assert>
    <sch:assert test="@endid or (@x2 and @y2)">When used in the symbolDef context, must have either an endid attribute or both x2 and y2 attributes.</sch:assert>
  </sch:rule>
  <sch:rule context="mei:line[not(ancestor::mei:symbolDef)]">
    <sch:assert test="@startid or @tstamp or @tstamp.ges or @tstamp.real or (@x and @y)">When used in the score context, must have a startid, tstamp, tstamp.ges or tstamp.real attribute or both x and y attributes.</sch:assert>
    <sch:assert test="@dur or @dur.ges or @endid or @tstamp2 or (@x2 and @y2)">When used in the score context, must have an endid, dur, dur.ges, or tstamp2 attribute or both x2 and y2 attributes.</sch:assert>
  </sch:rule>
</constraint>
```

You're absolutely right -- the goal was to make it possible position the line relative to other things (lines, text, symbols) _within the same symbol definition_.  Additional assertions could enforce that condition.  You want to give it a go?

th-we:
@pe-ro When you say _relative_, what do you mean by that? What would be the point of origin for each of the elements inside the `<symbolDef>`? The point specified by `@x`/`@y` (which would default to 0 if absent)?

Is it worth opening another issue for discussing `<userSymbol>` in general?

pe-ro:
`<symbolDef>` is a member of att.coordinated, meaning that it defines its coordinate system using `@ulx, @uly, @lrx, and @lry` attributes.  At least one entity (line, anchoredText, etc.) has to have `@x` and `@y` coordinates, but the location of other "objects" may be defined in terms of that first one.  This provides the ability to say "put a line from here to there and place some text relative to the line".  Some things are just easier to "draw" that way, but another advantage is that at some point in the future, if the coordinates of the line are treated as parameters, the position of the text can change without having to directly modify its x and y coordinates.  Said another way, a symbol definition can be a "template" -- that is, a general, re-usable procedure for drawing a symbol.

Creating another issue is fine with me.

ahankinson:
OK, I've added the constraint. Although it goes against the contributing guidelines, I've talked it over with Perry and I'll merge this one now.

PR #263, opened 2015-10-01
Updating guidelines for accid.ges

lpugin:
Clarifying the use of accid.ges following discussion #261

### DIFF
--- a/source/specs/mei-source.xml
+++ b/source/specs/mei-source.xml
@@ -2451,7 +2451,7 @@
       <desc>Attributes for capturing momentary pitch inflection in the gestural domain.</desc>
       <attList>
         <attDef ident="accid.ges" usage="opt">
-          <desc>records the performed pitch inflection when it differs from the written
+          <desc>records the performed pitch inflection including when it is identical to the written
             accidental.</desc>
           <datatype>
             <rng:ref name="data.ACCIDENTAL.IMPLICIT"/>


ahankinson:
Is this still the desired wording?

zolaemil:
As far as I understand, there's a consensus that this wording would be better. 

However we haven't reached a conclusion on how to address the consequences: 
1. forget prima practica
2. allow both practices, but require an explicit statement which practice is in use in a particular customization.

As Jo reminds us, MEI is a framework, and one of it's greatest power is documentation, so I too lean towards the second option.

lpugin:
Unless we can clearly identify the advantage of having two practices I would suggest to forget the prima practica, at least for now. We can still adopt it later but with the addition of an encoded description of it in <encodingDesc> as suggested in the discussion. This pull request is fixing the inconsistency we currently have between the documentation and all our examples.

ahankinson:
I will wait for Perry's changes to be merged into develop before merging this one.

lpugin:
I guess this can now be ignored since @pe-ro has clarified the use of accid.ges

zolaemil:
@lpugin do we have a link to the other thread?

lpugin:
https://github.com/music-encoding/music-encoding/pull/261

pe-ro:
Yeah, I think adding `@visible` on `<accid>` and `<artic>` isn't necessary at this point.

PR #265, opened 2015-10-04
New: Build script now checks schematron rules as well

### LINKED ISSUE #264: Schematron validation in build process
Related to #262 

Schematron validation should be integrated into the build process. I'm willing to do this, but would appreciate some guidance (@kepper, @pe-ro, @raffazizzi, @th-we ), especially around getting a command-line (i.e., no oXygen) system for validating.

I've extracted the attached schematron rules (***) from the latest `mei-all.rng` using saxon-he. I can't seem to get it to validate without spitting out a bunch of warnings:

```
Warning: on line 1112
  The attribute axis starting at an attribute node will never select anything
Warning: on line 1112
  The child axis starting at an attribute node will never select anything
Warning: on line 1183
  The attribute axis starting at an attribute node will never select anything
Warning: on line 1183
  The child axis starting at an attribute node will never select anything
```

etc.

I've used `probatron` and `xmllint`. `xmllint` failed rather spectacularly:

```
XPath error : Invalid expression
not(normalize-space(.) eq '')
                       ^
schematron.sch:18: element assert: Schemas parser error : Failed to compile test expression not(normalize-space(.) eq '')
```

**\* You will need to rename .txt to .xml or .sch, since GH doesn't like .xml extensions 

[schematron.txt](https://github.com/music-encoding/music-encoding/files/6965/schematron.txt)

ahankinson:
When the schemas are built, the schematron rules are extracted and transformed into an XSLT that is stored in the build/schematron directory

When the tests are run, saxon is used to apply the validation XSLT to the MEI files. A custom validation checker xquery file is then used to extract any errors from the schematron validation file and displays them in the console.

Fixes #264
Refs #262


ahankinson:
OK, I think this is ready for review. You can see the results in the TravisCI output:

https://travis-ci.org/music-encoding/music-encoding/builds/83616237

ahankinson:
@raffazizzi could you have a look at this and merge it if it looks OK?

Thanks

raffazizzi:
Tested locally and it works great! merging

PR #268, opened 2015-10-14
Changed ornament to ornam

### LINKED ISSUE #253: Create a generic ornament element

raffazizzi:
This commit changes the new element ornament to the shorter ornam. This is in keeping with other abbreviated element names. Related attribute classes att.ornament.\* were changed as well to att.ornam.*.

Fixes #253

### DIFF
--- a/source/specs/mei-source.xml
+++ b/source/specs/mei-source.xml
@@ -6181,7 +6181,7 @@
         </attDef>
       </attList>
     </classSpec>
-    <classSpec ident="att.ornament.log" module="MEI.shared" type="atts">
+    <classSpec ident="att.ornam.log" module="MEI.shared" type="atts">
       <desc>Logical domain attributes.</desc>
       <classes>
         <memberOf key="att.controlevent"/>
@@ -6191,7 +6191,7 @@
         <memberOf key="att.timestamp2.musical"/>
       </classes>
     </classSpec>
-    <classSpec ident="att.ornament.vis" module="MEI.shared" type="atts">
+    <classSpec ident="att.ornam.vis" module="MEI.shared" type="atts">
       <desc>Visual domain attributes.</desc>
       <classes>
         <memberOf key="att.color"/>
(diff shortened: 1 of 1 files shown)


ahankinson:
@pe-ro if you give your OK I'll merge this into both develop and develop-perry.

PR #296, opened 2016-03-24
guideline generation xslt

kepper:
This XSLT and the readme.md documentation may be used for custom.music-encoding.org


ahankinson:
I don't think we can merge this because it contains unrelated changes -- the chapter about "cpMark" should be moved to a different pull request.

ahankinson:
Closing -- new guidelines are generated in a different way.

PR #300, opened 2016-04-21
Added string datatype to data.HEADSHAPE

raffazizzi:
This was added to make explicit that the `<valList type='semi'>` should accept other string values for the note.head attribute.

I'm using two "pure ODD" recently added elements to solve this problem: `<alternate>`and `<dataRef>`.
This allows me to indicate that the datatype should either allow the list of attribute values or string. 

Caveat: the current TEI schema doesn't validate this yet, but it's correctly processed by the latest release of TEI XSLTs.


raffazizzi:
I think this was solved by @pe-ro in 3.0.0. Close?

PR #303, opened 2016-05-16
Text instruct elem spec

ironchicken:
This pull request implements the element `<textInstruct>` which encodes a textual instruction in music notation. These are portions of text written on the score which give instructions to the performer. Such text plays a similar role to features such as dynamic markings and tempo markings, but the composers (or transcribers or editors) felt that a non-standard instruction was necessary.

`<textInstruct>` can appear in a `<section>`, `<layer>`, `<staff>`, or `<measure>` and may also be part of a `<lem>` or `<rdg>`. It may contain any of the `model.transcriptionLike` or `model.annotLike` elements.

It's also in the `model.rdgPart.text` class. Another commit alters this class adding it to the content model of `<lem>` and `<rdg>`.

### DIFF
--- a/source/specs/mei-source.xml
+++ b/source/specs/mei-source.xml
@@ -20207,6 +20207,31 @@
         </rng:zeroOrMore>
       </content>
     </elementSpec>
+    <elementSpec ident="textInstruct" module="MEI.shared">
+      <desc>(text instruction) Encodes a playing or performance
+      direction which is given in prose text and which cannot be
+      captured by more specific elements such as &lt;tempo&gt;.</desc>
+      <classes>
+        <memberOf key="att.common"/>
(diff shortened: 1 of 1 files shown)


raffazizzi:
Isn't [&lt;dir>](http://music-encoding.org/documentation/2.1.1/dir/) sufficient for this?

ironchicken:
I have to admit I overlooked `<dir>` when I came up with this. It seems that `<dir>` might be OK. But perhaps there may be some other opinions on this?

ahankinson:
@ironchicken is `<dir>` sufficient? If so, I will close this.

ironchicken:
@ahankinson yes, `<dir>` is sufficient for my purposes. Let's close.

PR #304, opened 2016-05-16
Pure odd

raffazizzi:
This Pull Request refactors the MEI ODD into PureODD, a new version of ODD that does not rely on RelaxNG elements.
It also includes the XSLTs used for the refactoring (the ODD was then edited by hand).

Note:
- There are two content models that could not yet be defined in PureODD: `data.BEZIER` and `data.BULGE`. 
- tei_odds.rng was replaced with the content of tei_all because the current tei_odds released by the TEI is actually outdated.
- the new ODD is actually not valid because of `<content/>`. It should be valid but there's currently a bug in the schema (see https://github.com/TEIC/TEI/issues/1450)
- tests fail because of new restrictions to the `@rend` attribute and other regex patterns. At least I think that's the case. Help wanted! :)


raffazizzi:
Obsolete since development of pureODD has continued elsewhere. Closing.

raffazizzi:
Reopened with new up-to-date conversion and scripts

raffazizzi:
After discovering some issues related to datatypes in PureODD, we're lowering the priority of this. Possibly after the next release. In the meantime action on @raffazizzi to fix what's fixable on the TEI side.

Major issues noted:
- lack of complex support for combining non-string datatypes (e.g. lists of numbers)
- bugs in parsing datatypes containing lists of attribute values (e.g. data.ARTICULATION)

ahankinson:
Does this mean we have to stick to a certain version of the stylesheets?

raffazizzi:
@ahankinson technically not because the use of RNG in ODD has not been deprecated, but we should keep an eye open for bugs.

ahankinson:
Outdated change -- closing for now.

PR #422, opened 2017-03-15
Use proper node name markup inside <desc> and <remarks>

th-we:
Replaces literal "`<>`" with `<gi>` and literal "`@`" with `<att>`.
Also corrects `<gi>fingering</gi>` to `<gi>fing</gi>`.


pe-ro (approved):
Looks fine.

Issue #415, opened 2017-03-06
Scordatura example/unusual key signatures


craigsapp:
Here is some interesting notation for encoding:

<img width="947" alt="screen shot 2017-03-06 at 10 03 18 am" src="https://cloud.githubusercontent.com/assets/3487289/23622693/790ce0f0-0254-11e7-934a-782baa4e33af.png">

J.S. Bach cello suite no. 5 in C minor, BWV 1011, Breitkopf & Härtel/Klengel, *c*1900.

(http://ks.imslp.net/files/imglnks/usimg/9/97/IMSLP36902-PMLP04291-Bach-BWV1007-1012klengel.pdf)
 

The cellist can optionally tune the top string down a whole tone to play the top staff (scordatura), or use the regular tuning by reading the bottom staff.

MEI can handle this pretty well, with liberal use `@pname.ges`/`@oct.ges`/`@accid.ges`.  A tricky item is the key signature of the top staff.  How would that be encoded?  The problem is that it is a super-position of the regular key signature (C minor) and a transposed key signature (D minor) which applies only to the top string).  This results in the `@pname="a"` of the key signature having two different states: the lower A being flat for C minor, and the upper being natural for D minor.

`keySig@sig.mixed` seems close to what is needed, but it only seems to encode a horizontal sequence of accidentals, without a way to encode the accidentals vertically aligned.  Semantically it is two key signatures (3-flats and an irregular one with a-natural), so one possibility would be a way of encoding two keySig elements and instructions on how to overlap them.

This is also somewhat related to key signatures such as:

<img width="344" alt="screen shot 2017-03-06 at 10 32 07 am" src="https://cloud.githubusercontent.com/assets/3487289/23623800/36de2c94-0258-11e7-8423-a4265ad53e6e.png">

How are these encoded in MEI?

pe-ro:
For the 'cello notation use --
```
<keySig>
  <keyAccid pname="b" oct="2" accid="f"/>
  <keyAccid pname="e" oct="3" accid="f"/>
  <keyAccid pname="a" oct="2" accid="f"/>
  <keyAccid pname="a" oct="3" accid="n"/>
</keySig>
```
The other examples work the same way.

There's no info about how to place the individual accidentals, but I'd venture to guess that accidentals on the same pitch name in different octaves should be vertically aligned.

craigsapp:
[...]

PR #458, opened 2017-08-30
Modifying Schmatron rule for keyAccid in order not to require oct

lpugin:
In some cases it can be useful to have `<keyAccid`> specifying only the `@pname` and not the `@oct`. That way, the octave can be left undetermined and calculated according to the clef and the standard practice for key signatures. 

With the requested change, the following will be valid:
```xml
<keySig>
   <keyAccid accid="s" pname="f"/>
   <keyAccid accid="s" pname="c" enclose="brack"/>
</keySig>
```

### DIFF
--- a/source/specs/mei-source.xml
+++ b/source/specs/mei-source.xml
@@ -17655,9 +17655,9 @@
       <constraintSpec ident="Check_keyAccidPlacement" scheme="isoschematron">
         <constraint>
           <sch:rule context="mei:keyAccid">
-            <sch:assert test="(@x and @y) or (@pname and @oct) or @loc">One of the following is
-              required: @x and @y attribute pair, @pname and @oct attribute pair, or @loc
-              attribute.</sch:assert>
+            <sch:assert test="(@x and @y) or @pname or @loc">One of the following is
+              required: @x and @y attribute pair, @pname attribute, or @loc attribute.
+            </sch:assert>
           </sch:rule>
         </constraint>
       </constraintSpec>


pe-ro:
In the example you provided, what would you expect the value of `@key.sig` to be?  Would it be "2s" or "mixed" or something else?

lpugin:
No `@key.sig` because we have something that cannot be represented with `@key.sig`. (Hence the use of `<keySig>`.)

pe-ro:
Does the use of `<keySig>` preclude the presence of `@key.sig`?

lpugin:
I would not recommend to have both. Similarly I would not recommend to have `@accid` on `<note>` when an `<accid>` child is provided. Is this any different?

pe-ro:
Actually, your example is a standard key signature.  The example you provided me yesterday was not --
```
<keySig>
   <keyAccid accid="s" pname="f"/>
   <keyAccid accid="s" pname="g"/>
</keySig>
```
This is a non-standard key signature that presumably follows the standard layout for key signatures.

pe-ro:
So, do you want to add a Schematron rule to disallow the presence of both?

pe-ro:
For a non-standard key signature, where/how does one indicate that the standard layout for the accidentals is to be followed?

lpugin:
> So, do you want to add a Schematron rule to disallow the presence of both?

Do you mean both `@key.sig` and `<keySig>` ? Or both `@oct` and `@pname` ?

pe-ro:
> Do you mean both `@key.sig` and `<keySig>` ? Or both `@oct` and `@pname` ?

I mean `@key.sig` and `<keySig>`.

pe-ro:
While we're changing stuff, should we change the `@key.sig` value "mixed" to "nonstan" to indicate any non-standard key signature, not just ones with a mix of sharps and flats?

pe-ro:
"mixed" doesn't apply in the cases we're discussing, but "nonstan" would.

lpugin:
> I mean @key.sig and <keySig>

This seems to be a general discussion we should have about double encoding. So I would not disallow both without this general discussion to be conducted.

> For a non-standard key signature, where/how does one indicate that the standard layout for the accidentals is to be followed?

By not providing `@oct`. (E.g., a sharp for pname "f" has to be placed on the top line with a G-2 clef.)

pe-ro:
I'll let this sit for a day or two and if no one objects, then I'll merge it.

lpugin:
We can change it. BTW, in what case would you use `@key.sig="mixed"` (or "nonstan"). It seems to be not a very rich piece of information.

pe-ro:
It was intended for analytical purposes; that is, when you don't care what the key signature looks like, only that it's non-standard with respect to the circle of fifths.  It could also be used to signal a required `<keySig>`, but the changes you're suggesting don't fit this use, since you'd allow `<keySig>` for both standard and non-standard key signatures.

lpugin:
Is a key signature with some accidentals in bracket really standard? This is debatable.

pe-ro:
It depends on the meaning of the bracket.  If it indicates an editorially supplied addition to an otherwise standard key signature and the result is a standard key signature, then I think it can be called "standard".  In this instance, the bracket is a visual detail that signals a semantic difference between this key signature and some other.  Maybe `@enclose` is a not-so-good way of marking a key signature with editorial emendations.  Perhaps the content of `<keySig>` should allow members of att.transcriptionLike (add, corr, damage, etc.)  These elements would also have to allow `<keyAccid>`, of course.  For example,
```
<keySig>
  <keyAccid accid="s" pname="f"/>
  <supplied>
    <keyAccid accid="s" pname="c"/>
  </supplied>
</keySig>
```
We'd also need a Schematron rule to disallow content of `<supplied>` other than `<keyAccid>`.

These are significant changes, but they provide better capture of semantics.

lpugin:
> These are significant changes, but they provide better capture of semantics.

Yes, I like this. Allowing editorial elements to be used would be very useful and avoid entire scoreDef to be duplicated when two source have differences (e.g., different clefs). But we better discuss this somewhere else.

rettinghaus:
So `@oct` is a *visual* attribute? Meaning, if you want to encode a piece from a 18th century where the first sharp is double printed (with standard G clefs):

```
<keySig>
   <keyAccid accid="s" pname="f" oct="4"/>
   <keyAccid accid="s" pname="f" oct="5"/>
</keySig>
```

So what if you have a 20th century piece where a sharp only has meaning within a single octave? 
(Not to say that there's no way to encode, that the sharps from upper example share the same horizontal position.) 

`note@oct` basically has a visual *and* logical meaning, whereas `keyAccid@oct` not necessarily does. 
I don't know if there is a way to fix this. Perhaps an attribute `oct.range` could hold information about to which octaves an key accidental applies?

craigsapp:
Is there a way to distinguish (or need to distinguish) these two cases:

screen shot 2017-08-31 at 6 55 04 am

pe-ro:
@craigsapp: Well, that was the original reason for requiring `@oct` on `<keySig>`; that is, to force the encoder to say exactly which octave the accidental applies to as well as where it should be rendered.  For example --
```
<keySig>
  <keyAccid pname="f" oct="5"/>
  <keyAccid pname="f" oct="4"/>
</keySig>
```
`<keyAccid>` doesn't currently allow `@ho`, but I suppose it could.  I'm not certain exactly how it would work, though.

Frankly, it's unclear to me how making `@oct` optional helps.  It seems to me that it muddies the water.  How should --
```
<keySig>
  <keyAccid pname="f" oct="5"/>
  <keyAccid pname="f"/>
</keySig>
```
Or --
```
<keySig>
  <keyAccid pname="f"/>
  <keyAccid pname="f" oct="5"/>
</keySig>
```
be interpreted?  And how are these different from --
```
<keySig>
  <keyAccid pname="f"/>
</keySig>
```
More discussion is desperately needed.

lpugin:
> Well, that was the original reason for requiring @oct on <keySig>; that is, to force the encoder to say exactly which octave the accidental applies to as well as where it should be rendered. 

Sorry, I am lost. Are you saying that 
```xml
<scoreDef key.sig="1s">
```
is not equivalent to 
```
<scoreDef>
   <keySig>
      <keyAccid oct="5" accid="s" pname="f"/>
   </keySig>
</scoreDef>
```
because in the second case the accidental applies only to the 5th octave?

pe-ro:
Whether they are equivalent or not is open to debate.  Often the point of mixed or non-standard key signatures is to indicate that the accidental applies only in the given octave, say, F5 but not F4, but this is not always the case.  Likewise, accidentals in mixed key signatures, say a flat on B4 and a sharp on F5, may or may not apply in all octaves depending on the composer's wishes.  I suppose there are probably cases where even a standard key signature of 1 sharp on F5 could be interpreted to apply only in octave 5.  This "maybe/maybe not" quality of key signatures is probably best captured using a separate attribute.

In any case, `@key.sig="1s"` is a CMN short cut; that is, the standard CMN conventions apply -- the accidental is rendered based on the clef and applies to all octaves.

lpugin:
> This "maybe/maybe not" quality of key signatures is probably best captured using a separate attribute.

Yes, I think so. Having an additional optional attribute to specify to what octave(s) an accid applies as suggested by @rettinghaus seems to be a good direction to me. If the attribute is omitted, then I would assume that here too the standard CMN convention applies, meaning that the accid applies to all octaves. (And if oct attribute is omitted, then standard CMN convention applies too and it is rendered based on the clef, as suggested by the current PR.)

craigsapp:
> More discussion is desperately needed.

The main goal is (or should be)  to make it easier to encode editorial accidentals in key signatures.  It is common in 17th-century music to drop a flat from the key signature, and modern editors want to put it in, but also indicate that they added it rather than that it was in the original.   If the octave has to be specified in the `<keySig>` this is a hassle to encode, as well as a hassle to switch clefs (such as switching between the original clef and a modern clef).  In these cases, the standard circle-of-fifths placement for the accidentals is kept, and only parameters for adding brackets around the accidentals at the end of the key signature is needed.

For such cases, it might be nicer to have a parameter similar to the short-form of the key signature, something like:

```
<scoreDef>
   <keySig sig="3f" edit="1"/>
</scoreDef>
```

for

screen shot 2017-08-31 at 8 38 36 am

and

```
<scoreDef>
   <keySig sig="3f" edit="2"/>
</scoreDef>
```
for

screen shot 2017-08-31 at 8 38 32 am

For doubled accidentals, perhaps something like:

```
<scoreDef>
   <keySig sig="1s" doubled="true"/>
</scoreDef>
```

for

screen shot 2017-08-31 at 8 42 26 am

There can only be two positions on a staff for accidentals, so the octave placement should be easy to automatically calculate (and if someone wants `<keySig sig="1f" doubled="true">` on a treble clef, then they would have to use a more verbose encoding to say which of the two equidistant accidentals from the staff should be used.

In other words circle-of-fifths key signatures should rely on the existing short-form representation, and additional parameters for adding brackets on the leftmost accidental(s) or doubling at the octave could be added.   This would perhaps reduce the need for dropping `@oct` from the verbose description of key signatures, since the verbose form would presumably be for cases that the circle-of-fifths system does not apply and octave placement is necessary (and would therefore be difficult to automatically switch to a new clef layout).

-----------------

For reference, here are the positions of the accidentals in 4 clefs (the shape of the key signatures are all similar except for tenor clef placement of sharps:

screen shot 2017-08-31 at 8 16 45 am

--------------

> So what if you have a 20th century piece where a sharp only has meaning within a single octave?
(Not to say that there's no way to encode, that the sharps from upper example share the same horizontal position.)

There was one Bach example I posted in issue  https://github.com/music-encoding/music-encoding/issues/415 where the different octaves had different chromatic states due to scordatura.

In this case the top F has a natural in the key signature.  This is how some theoretical 20th century piece should encode the accidentals: they would need to cancel out the accidental on octaves since standard key signature syntax applies to pitch classes not to the specific octave that the accidental is placed.  If they do not want to see the natural sign in the key signature, then they need to add `invis="true"` for that particular accidental.

lpugin:
@craigsapp scordatura is a full topic that I feel needs to be discussed and developped in MEI. How to represent the key signatures in that context is a little bit a chicken and egg problem. It is good to have it in mind but I would not mix up issues here.

Regarding octaves in key signature, it seems to me that the starting point is to assume that an accidental given in a key signature applies to all octaves because this is how they traditionally work. (When an accidental is duplicated in older sources this is generally for reading purposes and not because otherwise the accidental would not apply.) For this reason, it seems logical to me that a `keyAccid` in `keySig` applies to all octaves, if not specified otherwise. Not only this is logical but also the opposite is problematic. If we agree that we can have non standard key signatures where an accidental applies to all octaves, then in that case we would need to explicitly provide all the octaves. I am not sure how would that work.

This means that we need to make the distinction between encoding the octave(s) where the `keyAccid` appears visually, and the octave(s) to which it applies. This PR does not modify any thing about the second aspect, which still has to be discussed and fixed. The PR only assumes that `@oct` on `keyAccid` represent the visual position, and that this information is not indispensable. I do not see where the problem is with this step.

Now this
```
<keySig>
  <keyAccid pname="f"/>
  <keyAccid pname="f" oct="5"/>
</keySig>
```
does not make much sense to me. If `@oct` is provided, then I would expect it to be provided on all `keyAccid`.

lpugin:
One example (taken from http://www.academicjournals.org/journal/ERR/article-full-text/59DFA8158144)

apaydinli2

I would suggest to encode this as
```xml
<scoreDef >
    <keySig>
        <keyAccid accid="f" pname="e"/>
        <keyAccid accid="s" pname="f"/>
    </keySig>
    <staffGrp>
        <staffDef clef.shape="G" clef.line="2" n="1" lines="5" />
        <staffDef clef.shape="G" clef.line="2" n="2" lines="5" />
        <staffDef clef.shape="G" clef.line="2" clef.dis="8" clef.dis.place="below" n="3" lines="5" />
        <staffDef clef.shape="F" clef.line="4" n="4" lines="5" />
    </staffGrp>
</scoreDef>
```

pe-ro:
I'm not dismissing Craig's suggestions, I'm just pointing out that adding a bracketing mechanism to the short-form encoding method conflates this method and the explicit encoding of all accidentals in the key signature.  This contradicts the use of `<supplied>` and its brethren, which is the way transcriptional material is handled in other parts of an encoding.

Regarding "doubling", what's supposed to happen if there's more than one doubled accidental?  For example --
```
<scoreDef>
   <keySig sig="3s" doubled="true"/>
</scoreDef>
```
Does this mean that only those accidentals that fall inside the staff should be doubled?  What if the source material has a sharp on C4 (which is not outside the realm of possibility)?  To do this, `@doubled` would have to permit selection of individual accidentals, perhaps a non-sequential set, which brings up the problem of parsing a complex value in `@doubled`.

I propose that one should either 1) use the short-hand, CMN method --
```
<scoreDef>
  <keySig sig="3f"/>
</scoreDef>
```
in which all the usual rules apply, or 2) explicitly encode every accidental, whether the need is for more information for a "standard" key signature or for some extended kind.

This approach starts from the assumptions that 
```
<keySig sig="3f"/>
```
and 
```
<keySig>
  <keyAccid accid="f"/>
  <keyAccid accid="f"/>
  <keyAccid accid="f"/>
</keySig>
```
are equivalent (I reserved the right to reverse myself 😁) and that in performance an accidental applies to all octaves.  The presence of `@sig` and child `<keySig>` elements are mutually exclusive.

Additional info on/around `<keyAccid>` can qualify, even nullify, this equivalence.  For instance, in the following example, `<supplied>` is used to indicate the editorially supplied A flat --
```
<scoreDef>
   <keySig>
      <keyAccid accid="f"/>
      <keyAccid accid="f"/>
      <supplied>
         <keyAccid accid="f"/>
      </supplied>
   </keySig>
</scoreDef>
```
while 
```
<scoreDef>
   <keySig>
      <keyAccid accid="f"/>
      <supplied>
         <keyAccid accid="f"/>
         <keyAccid accid="f"/>
      </supplied>
   </keySig>
</scoreDef>
```
means both the 2nd and 3rd flats are supplied.  (Whether they are bracketed or not is a rendering decision. They could just as easily be rendered in a different color.)

The addition of `@pname` and `@oct` provides information about the *location* of the accidentals.  "Doubled" accidentals therefore must be explicitly encoded --
```
<scoreDef>
   <keySig>
      <keyAccid accid="f" pname="b" oct="4"/>
      <keyAccid accid="f" pname="e" oct="5"/>
      <keyAccid accid="f" pname="e" oct="4"/>
      <keyAccid accid="f" pname="a" oct="4"/>
   </keySig>
</scoreDef>
```
This is certainly more verbose, but more clear, I think.  It also makes each accidental individually addressable; that is, each one can take new / different attributes (`@ho`, for example) that pertain only to it.

The presence of `@oct` doesn't say anything about which octave the accidental applies to in performance.  That can be addressed with an attribute on either 1) `<keySig>` if all accidentals apply in the same way, or 2) on `<keyAccid>` if they don't.  Adding a Schematron rule that forces all accidentals to have the attribute if any one of them does would force the encoding to be 100% explicit.

Using this approach, doubling and bracketing can be combined easily --
```
<scoreDef>
   <keySig>
      <keyAccid accid="f" pname="b" oct="4"/>
      <keyAccid accid="f" pname="e" oct="5"/>
      <supplied>
        <keyAccid accid="f" pname="e" oct="4"/>
      </supplied>
      <keyAccid accid="f" pname="a" oct="4"/>
   </keySig>
</scoreDef>
```
I think vertically stacking accidentals on the same pitch name (subject to any attributes on the individual accidental) still applies.

pe-ro:
@lpugin types faster than I do!

But I don't see anything in his proposed encoding that contradicts my suggestions, although I would add `@oct` to each of the key signature accidentals --
```
<scoreDef>
    <keySig>
        <keyAccid accid="f" pname="e" oct="5"/>
        <keyAccid accid="s" pname="f" oct="5"/>
    </keySig>
    <staffGrp>
        <staffDef clef.shape="G" clef.line="2" n="1" lines="5" />
        <staffDef clef.shape="G" clef.line="2" n="2" lines="5" />
        <staffDef clef.shape="G" clef.line="2" clef.dis="8" clef.dis.place="below" n="3" lines="5" />
        <staffDef clef.shape="F" clef.line="4" n="4" lines="5" />
    </staffGrp>
</scoreDef>
```
I think `@oct` is necessary because the order of the accidentals doesn't follow the circle of fifths.

lpugin:
All you said make sense, and I support the idea the `@oct` should be given on all `keyAccid` if given. I also support the idea of using `supplied`. But I do not agree that `@oct` is necessary. On the contrary, if it is required, then I would expect the example to be encoded
```xml
<scoreDef >
    <staffGrp>
        <staffDef clef.shape="G" clef.line="2" n="1" lines="5" >
            <keySig>
                <keyAccid accid="f" pname="e" oct="5"/>
                <keyAccid accid="s" pname="f" oct="5" />
            </keySig>
        </staffDef>
        <staffDef clef.shape="G" clef.line="2" n="2" lines="5" >
            <keySig>
                <keyAccid accid="f" pname="e" oct="5"/>
                <keyAccid accid="s" pname="f" oct="5"/>
            </keySig>
        </staffDef>
        <staffDef clef.shape="G" clef.line="2" clef.dis="8" clef.dis.place="below" n="3" lines="5" >
            <keySig>
                <keyAccid accid="f" pname="e" oct="4"/>
                <keyAccid accid="s" pname="f" oct="4"/>
            </keySig>
        </staffDef>
        <staffDef clef.shape="F" clef.line="4" n="4" lines="5" >
            <keySig>
                <keyAccid accid="f" pname="e" oct="3"/>
                <keyAccid accid="s" pname="f" oct="3"/>
            </keySig>
        </staffDef>
    </staffGrp>
</scoreDef>
```
because the octave changes according to the clef. This is the problem, and it creates additional problems with clef changes.

lpugin:
PS The fact that the accidentals do not follow the circle of fifths makes `@pname` necessary, but not `@oct`

pe-ro:
We're very close to agreement 😀.  But without `@oct` --
```
<scoreDef >
    <keySig>
        <keyAccid accid="f" pname="e"/>
        <keyAccid accid="s" pname="f"/>
    </keySig>
    <staffGrp>
        <staffDef clef.shape="G" clef.line="2" n="1" lines="5" />
        <staffDef clef.shape="G" clef.line="2" n="2" lines="5" />
        <staffDef clef.shape="G" clef.line="2" clef.dis="8" clef.dis.place="below" n="3" lines="5" />
        <staffDef clef.shape="F" clef.line="4" n="4" lines="5" />
    </staffGrp>
</scoreDef>
```
how is the location of the accidentals supposed to be determined?  In the original you provided, they're  in the usual locations, but what if they were on E4 and F4 instead?  Perhaps the answer is, as you suggest, to move the encoding of the key signature to the staff definition.

Which may open up a new tack -- Should `@pname` and `@oct` only be allowed `<keyAccid>` when it's a descendant of `<staffDef>`?  Hmmm..... This would "link" the clef (which is usually provided in `<staffDef>`) and the key signature.  Whaddaya think?

lpugin:
Pretty close indeed. Though.. ;-) I would suggest checking that `@oct` is given on all `keyAccid` if provided, and the same for `@pname`. That is, it is has to be given on `keyAccid` if provided. In other words, respectively, you cannot hardcode the octave for some accidentals but not all of them, and you cannot partially assume the circle of fifths.

Now I am not sure about requiring `@oct` on `keyAccid` within `<staffDef>` because `<keySig>` might be moved down to `<staffDef>` for another reason. But again, we ensure consistency as much as possible. What about that?

lpugin:
> Should `@pname` and `@oct` only be allowed `<keyAccid>` when it's a descendant of `<staffDef>`

Do you mean not allow them when a descendant of `<scoreDef>`? I think `@pname` alone should be allowed.

pe-ro:
Yes, if `@oct` is specified on any `<keyAccid>`, then it has to be specified on all of them.

> Do you mean not allow them when a descendant of `<scoreDef>`?

I meant to say that we could disallow `scoreDef/keySig/keyAcccid/@oct`, but permit `staffDef/keySig/KeyAccid/@oct`. This would force non-standard key signatures down to `<staffDef>`. Non-standard key signatures at the score level are difficult to handle due to the fact that they (the key signatures) depend on the clef, which is provided at the staff level.

Standard key signatures would continue to be possible at both levels.

axgeertinger:
Sorry to arrive (too) late at the party :-) 
Laurent just made me aware of this discussion. Just wanted to say that making `@oct` optional sounds good to me. The requirement that either it should be provided on all `<keyAccid>` or none also seems reasonable. 
I have a piano piece with a harmonic F minor key signature, that is, flats on b, a, and d, and they definitely apply to all octaves. I would find it misleading having to encode the exact placement of the accidentals for every staff depending on the clef. When omitting `@oct` I'd expect the renderer to find convenient positions for them.

PR #468, opened 2017-10-10
Making fig member of att.horizontal/verticalAlign

lpugin:
This PR makes it possible to encode the horizontal and vertical alignment of a `<fig>` without having to in enclose it in a `<rend>`.


pe-ro (approved):
Looks good to me.

Issue #518, opened 2018-05-07
make proport part of att.color


rettinghaus:
as `<mensur>` allows `@color`, `<proport>` should, too.

Issue #517, opened 2018-05-07
make ligature part of att.color


rettinghaus:
One can color individual (child) notes in a ligature, but having an *figura obliqua* there are no individual (rendered) notes anymore.

PR #522, opened 2018-05-28
Remove remnants of ligatureSpan; fix testfiles

rettinghaus:
* fixing travis build
* removing remains of ligatureSpan
* fixed some testfiles


ahankinson:
Thanks for this, Klaus.

This PR seems to conflate several fixes. In the future, it would be much better to separate them out into individual fixes for discrete problems. For example, making changes to the gitignore or travis builds should not be in the same PR as changes to the schema. Changes to tooling are much more likely to get merged quickly than changes to the spec.

In the future, could you also please follow the [contributing guidelines](https://github.com/music-encoding/music-encoding/blob/develop/CONTRIBUTING.md#commit-messages) for formatting your pull request and commit messages? "several fixes" does not help us when understanding the impact of your changes. I did review your PR earlier, but because I couldn't determine the impact and extent of your changes from your messages, and didn't have time to go through the individual commits, I was not able to properly assess the full impact of your contribution. 

You can request a review for your PR from any one of the core committers, and not just @pe-ro . You can do this by addressing your review to the 'Core Contributors' team under the 'reviewers' section above.

rettinghaus:
@ahankinson I'll keep it in mind. Should I try to split this up, or is it ok for this time?

.gitignore
```diff
@@ -1,3 +1,5 @@
-/tests/.DS_Store
-.DS_Store
+# build
 /build
+
+# macOS
```
ahankinson, line 4: Is there a reason for this change?

ahankinson:
Were you intending to fix all of the tests? If so, I don't think they're passing yet

.gitignore
rettinghaus: Just to make clear to exclude all `.DS_Store` files (which is `macOS` specific).

ahankinson: Sure, but they were already excluded, weren't they?

rettinghaus: Yes, but why then make an extra statement for `/tests/.DS_Store`?

rettinghaus:
Didn't updated all the test files. Would be good if they were (re-)reviewed by others, too.

.gitignore
ahankinson: Historical reasons, I suppose.

(BTW, this is a good example where a helpful commit message would clear things up quickly: "update gitignore" --> "Removed redundant paths in gitignore.")

ahankinson:
I see @pe-ro has already committed fixes for #517 and #518 that is different from yours. What is the status of this?

rettinghaus:
I fixed the merge conflicts which arose because of remaining attributes for `<ligatureSpan>`, which was removed with 39d5e46.

ahankinson:
So the fix for #517 and #518 are no longer in this PR?

rettinghaus:
No. Removed the comments.

pe-ro:
Something went wrong somewhere, presumably in merging this pull request.  There are a number of features, `@aboveorder`, etc. that I added over the last several months/years that now appear to be gone!

I just got a new copy of the repo, so I don't think it's a synch problem between me and the repo.

I'll need some hand holding to get back to where I was.  :-(

ahankinson:
:(

Hold on... I'll revert these changes. Sorry.

ahankinson:
I've rolled back to your previous commit. Could you please pull and see if that fixes things?

pe-ro:
Will do, but I'm at home now, without my laptop, so it'll have to wait until tomorrow.

pe-ro:
Everything appears normal on my desktop at home, but it's not my primary computer.

rettinghaus:
`@aboveorder` etc are still in MEI.shared.xml and weren't affected by my changes!

pe-ro:
It seems I was having GitHub / SourceTree problems.  So, I blew away my local files (again), remembering to check the "recurse sub-directories" button this time, and got a fresh clone of the repository.  I seems like everything is working again.

pe-ro:
"ligatureSpan" needs to be removed as a possible value of `@aboveorder` and `@beloworder` and "bracketSpan" added in its place.  Is this what you were attempting to do, @rettinghaus?  Should we re-try Klaus' PR?

pe-ro:
I just removed the remaining ligatureSpan stuff and replaced "ligatureSpan" with "bracketSpan" in the declaration of data.STAFFITEM.cmn.  So, I there we're back on track there.  The other changes in @rettinghaus' PR can be re-submitted.

ahankinson:
Yes -- Klaus, could you please re-submit your fixes to the test files in a separate, dedicated PR? Thanks.

rettinghaus:
see #527 #528 #529

Issue #521, opened 2018-05-22
let breaks be beginnings 


rettinghaus:
As TEI moved from line/page/etc. "breaks" to "beginnings", MEI probably should follow the lead. That should cause less confusion in encodings.

pe-ro:
Fixed by a2529bef36bc59e140ad0fa0eaac50a2962a690c

PR #530, opened 2018-06-09
Rename break elements

rettinghaus:
This PR renames break elements to underline their expected position in the encoding. 

Fix for #521

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -3807,7 +3807,7 @@
     </classes>
   </classSpec>
   <classSpec ident="model.lbLike" module="MEI.shared" type="model">
-    <desc>Groups elements that function like line breaks.</desc>
+    <desc>Groups elements that function like line beginnings.</desc>
     <classes>
       <memberOf key="model.milestoneLike.text"/>
       <memberOf key="model.textPhraseLike.limited"/>
@@ -3864,7 +3864,7 @@
     <desc>Groups elements that collect separate performer parts.</desc>
   </classSpec>
   <classSpec ident="model.pbLike" module="MEI.shared" type="model">
-    <desc>Groups page break-like elements.</desc>
+    <desc>Groups page beginning-like elements.</desc>
     <classes>
       <memberOf key="model.milestoneLike.music"/>
       <memberOf key="model.milestoneLike.text"/>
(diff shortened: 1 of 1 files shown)


ahankinson:
I would like to get a confirmation of this change from one other core committer (noted @raffazizzi) before merging since there does not seem to be extensive discussion on the issue. A thumbs-up would do (@kepper? @pe-ro?)

pe-ro:
This is fine with me.

ahankinson:
In reviewing this again, why is `@func` being removed from `<lb />`? Was there any discussion of this?

(Clarifying info in commit messages would be helpful!)

rettinghaus:
Sorry, I thought it would be self-explanatory. `@func` was used to state 
> whether the line break follows a single line or a line group.

As it shouldn't "follow" either anymore, it seems superfluous now.

pe-ro:
I agree with @rettinghaus that, if we switch to a "beginning" model instead of a "break" one, then `@func` is superfluous.  However, this may pose a problem for compatibility with MusicXML, since `@func` was added to facilitate capture of MusicXML `<end-line>` and `<end-paragraph>` elements.

ahankinson:
What’s a “MusicXML”? 😀

pe-ro:
MusicXML = an early attempt at MEI.  :grinning:

rettinghaus:
In this so-called MusicXML it 
> comes from RP-017 for Standard MIDI File Lyric meta-events. 

So it seems to me there should be a MIDI-event-attribute on `<l>` or `<verse>` in MEI for that. But this should be discussed somewhere else.

pe-ro:
@rettinghaus, backward compatibility between MEI v. 3 and v. 4 isn't important in this case?  :grinning:

ahankinson:
Can we keep `@func` and just re-work the description to indicate that it's the beginning of a single line (default?) or a line-group?

rettinghaus:
@pe-ro backward compatibility is less important than consistency. But where probably everyone uses `@key.sig` I doubt many people used this `@func` knowingly. 

@ahankinson Why use an extra attribute, when the "function"/position is made clear in using `<lg>`?

I really think, it should be discussed if we need the "Standard MIDI File Lyric meta-events" in MEI, and how to encode them. The current approach seems a bit fishy to me.

craigsapp:
> I really think, it should be discussed if we need the "Standard MIDI File Lyric meta-events" in MEI, and how to encode them. The current approach seems a bit fishy to me.

`<syl>` functions pretty close to what a Lyric meta-event does, so I cannot imagine the need for a duplicate encoding as `<syl>` and a Lyric meta event.  The Lyric meta event is used for karaoke machines, and there can only be one "verse".

pe-ro:
I say we drop `@func`.  Everyone agreed?  Good!  :grinning:

rettinghaus:
@pe-ro feel free to merge this PR :smile:

PR #537, opened 2018-06-24
Support for "divisio"

giulianodibacco:
Adds support for divisio (14th-c Italian mensurations) their values and signs. Approved by the Mensural IG. First Git pull request by GDB please excuse any errors.


ahankinson:
Thanks, Giuliano. It would be helpful to see some examples of this in use so we can review it for inclusion. Do you have some small snippets of MEI that can help us see how it might be used?

The appropriate place for this might be in the development version of the Guidelines. If you (or others from the Mensural IG) would like help with this, to know what to edit and how, please let us know.

http://music-encoding.org/guidelines/dev/content/mensural.html

rettinghaus:
I would prefer to split `data.MENSURATIONSIGN` into `data.MENSURATIONSIGN.trecento` and `data.MENSURATIONSIGN.cinquecento` (or something similar), to make clear where these values are coming from and to ease customization.

ahankinson:
@rettinghaus do you see any overlap between the values, e.g., the same value used to mean different things in trecento v. cinquecento?

I think we need to balance between specification and usability. If we make too many distinctions without any actual differences it unnecessarily complicates the spec.

rettinghaus:
It's just that these values affect different repertoires. It's a bit like trying to combine `data.METERSIGN` and `data.MENSURATIONSIGN` because the values are the same. 
Am I making any sense here @giulianodibacco ?

lpugin:
I agree with @ahankinson we should not make any distinction if there is no need to. That is, only if there is clearly a need for disambiguation. Adding sub-categories will introduce additional problems for notations a the are in-between or overlapping them.

lpugin:
> It's a bit like trying to combine data.METERSIGN and data.MENSURATIONSIGN because the values are the same.

But they are not the same, are they?
http://music-encoding.org/guidelines/dev/data-types/data.metersign.html
http://music-encoding.org/guidelines/dev/data-types/data.mensurationsign.html

rettinghaus:
@lpugin no, of course not. I meant merely the "C" sign, that *looks* identical.

lpugin:
> I meant merely the "C" sign, that looks identical.

This is where I see a need for disambiguation.

ahankinson:
I see several opportunities for improvement:

1. "O" and "C" are uppercase, while other values are lower-case; this should probably be standardized to lower case. If this is the case, it will conflict with the value for Divisio octonaria.

2. However, "O" and "C" are only values because the symbols "look" like "O" and "C". I'm wondering if this should instead be `tp` (tempus perfectum) and `ti` (tempus imperfectum), which would follow the pattern of the other values.

3. Many of these values are duplicated (Divisio senaria imperfecta, Divisio senaria gallica (imperfecta), Divisio senaria perfecta, etc.). There should only be one value for each. It should probably follow the two-letter convention, so the single-letter values should be dropped.

4. All values should be standardized to two-letter values.

ahankinson (changes requested):
I see several opportunities for improvement:

"O" and "C" are uppercase, while other values are lower-case; this should probably be standardized to lower case. If this is the case, it will conflict with the value for Divisio octonaria.

However, "O" and "C" are only values because the symbols "look" like "O" and "C". I'm wondering if this should instead be tp (tempus perfectum) and ti (tempus imperfectum), which would follow the pattern of the other values.

Many of these values are duplicated (Divisio senaria imperfecta, Divisio senaria gallica (imperfecta), Divisio senaria perfecta, etc.). There should only be one value for each. It should probably follow the two-letter convention, so the single-letter values should be dropped.

All values should be standardized to two-letter values.

pe-ro (changes requested):
I agree that duplicate descriptions should be removed and values of `@ident` standardized to 2 letters.  Disambiguation between 13th vs. 14th c. notations can be added later.

giulianodibacco:
@ahankinson Yes we have thought about providing examples for the guidelines but I thought that first we need the markup approved. Yes please guide me/us on how to edit and how.

@ahankinson and @pe-ro. Let me clarify. The Trecento system of "divisiones" does not follow the traditional modus/tempus/prolatio. Hence the need of the new attribute `divisio`. In the analytical domain there are several possible ways to divide the breve in this system, to be expressed with standardized values in `data.DIVISIO`.  In the actual sources, however, twelve signs are used to indicate which divisio applies at any give time. These signs are lowercase letters, one or two. These are the new values for `data.MENSURATIONSIGNS`. We cannot 'make standard' the latter: the purpose is to encode the one/two letters found in the sources: so we are not introducing a duplication but representing signs that have the same meaning. Perhaps the confusion originated from the way I used the <desc> tag, where I meant to remind the encoder that an "i" or a "si" found in a source would mean "senaria imperfecta" in both cases, and "sg" or "g" which stand for "senaria gallica" ultimately mean the same thing. The standardization that you are calling for is taken care by using `@prolatio="senariaimp"` which is also used to supply the information when no sign is used at all. (I see that this may be confusing: will rephrase and create a new pull request). See the attached images with some examples of the one/two-letter signs.

About O and C I don't see neither the need or the opportunity to change them really. O/C and the new q/d/o/si/sp etc. belong to two different mensuration systems. We already have  `@tempus` to encode a perfect/imperfect tempus in the French system (and in subsequent late-medieval and Renaissance music). O and C are the most common signs for tempus perfectus and tempus imperfectus and the only ones supported by MEI at the moment but other equivalent signs exist in French sources, and we will need to take care of them in the near future.  If I understand it correctly, the introduction of "o" would create a duplication with "O". It would be too easy to say that uppercase and lowercase are not really the same. However, should O and C have to be changed, the best alternative would be still to describe the shapes because this is the purpose here, and the best alternatives would be "circulus" and "semicirculus" (as music treatises of the time describe them).

@rettinghaus and others: the new values for `data.MENSURATIONSIGNS` apply only to the earliest sources of the repertory in black notation from 14th-century Italy. The "divisiones" are used in alternative to the usual modus/tempus/prolatio. When the former are used, the latter are not, and vice versa. There are cases of pieces existing in both versions. So our idea was that the usage could be regulated by schematron. But I don't see any problem in creating a special `data.MENSURATIONSIGN.trecento` if this is useful.

image

image

image

pe-ro:
@giulianodibacco, OK, I understand now -- you're capturing signs, not semantics, hence the changes are in `data.MENSURATIONSIGN`.  :grinning: 

At some point it may be useful to "compose" `data.MENSURATIONSIGN` out of several other repertoire-specific declarations, such as `data.MENSURATIONSIGN.trecento`, `data.MENSURATIONSIGN.cinquecento`, etc.  But, it's not necessary right now.

I'm satisfied.

rettinghaus:
> At some point it may be useful to "compose" data.MENSURATIONSIGN out of several other repertoire-specific declarations, such as data.MENSURATIONSIGN.trecento, data.MENSURATIONSIGN.cinquecento, etc. But, it's not necessary right now.

So I'm satisfied, too.

ahankinson:
Hm. I'm not satisfied. 

The data stored in the `divisio` attribute [will be in the logical domain](http://music-encoding.org/guidelines/dev/attribute-classes/att.mensural.shared.html) (via other attribute classes, such as `att.mensur.log,` `att.staffDef.log`, etc.), which effectively means it will affect the processing of the score, rather than capturing 'annotations'. I know that the line between these is sometimes very fine, but my personal litmus test for the logical domain is whether it affects the interpretation of rendering or playback.

If I understand correctly, it seems to me that in this case you are trying to do two things: one is capture an annotation from the source, and the other is to instruct the interpreter (human or software) in how to divide the breve. Whether or not the letters used to indicate 'senia imperfecta' in the source are variously 'i' or 'si' in different sources is irrelevant in this context. When encoding attributes in the logical domain we would want a single value that indicates a way of interpreting the notation. Whether this is done with the values 'si' or 'senia-imperfecta' or 'flibflab' is irrelevant -- what matters is that we have a single value to represent a logical division of the breve in a particular way so that we could, for example, build software to play it back with the correct interpretation.

Whether it's an 'i' or 'si' in the original source can be captured using other means in MEI.

The 'O' vs. 'o' is a processing question -- it **will** cause problems with any software interpreter and implementation, and I think we would be doing a disservice to everyone to accept it as part of the spec. O and o are not the same, sure, but there are any number of processing steps that MEI may go through for indexing or rendering that may, as a standard course of action, lowercase all values. Generally it's very bad practice to make the distinction between two things to be one of case only (especially for letters like 'O' and 'C' where the difference between the upper and lower cases is only about size and not shape, unlike, say 'b', 'B').

The suggested change of 'O' and 'C' to 'circulus' and 'semicirculus' is fine (great, actually), but the question there is one of consistency of implementation: in that case, we should then change 'si' to 'senia-imperfecta' instead, since it follows the pattern (which is that you expand the name instead of trying to abbreviate it).

On a processing note: To make changes you don't need to submit a new pull request -- you can just edit the file on your branch and commit your changes directly. It will show up as a new 'proposal' that amends the first.

giulianodibacco:
Hello 
@ahankinson 

Thanks for the food for thought! and the instructions. 

Honestly, we (the Mens IG) thought that this spec would be totally in line with what is already in place in MEI for the Franconian/Renaissance system. Yes you are right: 

> you are trying to do two things

but with two different sets of attributes, on the same lines of what is already in place in the Mensural module. With the purpose to discover what I/we do not understand of your criticism, please accept more blathering about signs and their meaning:

- In the Franconian system, **mensurations** (the actual organization of durations) are encoded considering separately their components/orders of division: modusmaior, modusminor, tempus and prolatio:
`<mensur modusmaior="3">` or `="2"`
`<mensur modusminor="3">` or `="2"`
`<mensur tempus="3">` or `="2"`
`<mensur prolatio="3">` or `="2"`
Typically you need at least two of the attributes above to define the final combination in place, for example: `<mensur tempus="3" prolatio="2">`.
For pieces written in the Italian mensural dialect, the new attribute `divisio` would take care of the whole thing. Here no modus, tempus or prolatio apply so that the markup `<mensur divisio="senariaperf">` describes more or less the same situation where a breve is divided in three and then in two, for six resulting semibreves (the difference with the French system is that the French would call the resulting figures six minims, and the intermediate three semibreves, and individual durations are handled differently).

So, Yes, by all means the attribute `divisio` will affect the processing of the score! A "quaternaria" is usually transcribed in modern time 2/4, a "senaria perfecta" in 3/4, a "senaria imperfecta" in 6/8, a "novenaria" in 9/8, and so on, and durations are differently affected in a way similar to what happens within the four Franconian mensurations. In the `divisio` attribute the necessary standardization has been introduced, so that yes, each of the proposed values **are** unique in their meaning, for both the encoder and the processor. 

- About **signs**: MEI encodes mensuration signs separately, if and when they are present in the score. At the moment MEI supports only the classic four main signs of the Franconian system, which, however, are encoded not by means of four unique attribute values but by four combinations of two attributes:
`<mensur sign="O" dot="false">` where false is default and can be omitted
`<mensur sign="C" dot="false">` where false is default and can be omitted
`<mensur sign="O" dot="true">`
`<mensur sign="C" dot="true">`
As I mentioned, these are not the only symbols used to represent the relation tempus/prolatio. In a future post we will present a proposal for inclusion of other symbols that are used to represent the same and other relations in the four levels of division (squares with 2-3 strokes, strokes with 2-3 dots etc.). 
Anyway, in the Italian system, which is much simplified (no modus, tempus or prolatio) the same variability in the use of signs applies: scores are showing different signs (abbreviations, in this case: and references such as "this is the French senaria" cannot just be eliminated because of their historical significance).

You said: 
> in the logical domain we would want a single value that indicates a way of interpreting the notation

we have:
- `<mensur  tempus="3" prolatio="2">`
- `<mensur divisio="senariaperf">`

but I am not happy with considering mensuration signs as mere "annotations"... (all of them?).  The encoder may not know (or not have the time) to discover the `divisio` of the piece, and this is the same for a reversed C with a slash, or a square with 2 vertical strokes. I would say that processors should be instructed somehow to recognize the signs as well, including those with equivalent meaning... 

But here we are outside the domain of my knowledge...! (What we are asking is to have at least a mechanism to encode these...)


**About the 'O' vs. 'o' problem**  
> it will cause problems with any software interpreter and implementation, and I think we would be doing a disservice to everyone to accept it as part of the spec.

of course we don't want that to happen! so thanks for pointing this out. Let me propose something that may avoid the need of changing from "O/C" to "circulus/semicirculus".
 
As you can see from the pics I attached to an earlier message, the signs that we find in the scores are almost always preceded and followed by a dot. There was some discussion in MEI-Mens-L whether we should keep the dots. I was among those who advocated the need to simplify the attribute values, and in the end we decided to remove the unnecessary dots since they are mere graphic details. However, if this is useful to avoid processing difficulties, we can re-insert the dots and have `.q.` `.si.` `.d.` `.o.` etc. as attribute values.

The revision that I uploaded reflects the above suggestion. In my revision I also added one missing value and added some clarifications.

ahankinson:
Hi Giuliano,

Just to be clear: I'm not questioning the *need* for this -- I absolutely agree that it is needed. My concerns are around the implementation that you are proposing.

The logical domain controls the functioning of the notation: Note durations, pitches, etc. When proposing an alteration to the logical domain, it is important to separate out the _functional behaviour_ from the _visual evidence_. If we were to design a notation encoding system that conflates the two, we would be forever creating new values from on the slightest visual differences.

Although it's an imperfect metaphor, consider if we were to have, in two different fonts, a cursive letter "t" and a roman letter "t". Functionally, they are the same: they have a "tuh" sound, but visually they are very different. Creating attributes based on the visual appearances called "cursive-t" and "roman-t" would then require us to also have "book-t," "grotesk-t", etc, which would lead to an explosion of attribute values.

It is the same here. Whether the visual evidence shows "i" or "si" is irrelevant in the functional domain, since the behaviour is the same. I certainly don't consider the mensuration signs to be "merely" annotations, but I think the visual evidence must be separated from the functional behaviour. 

We could use any sort of 'flag' we want to indicate the functional behaviours: zero could be tempus perfectum, 1 could be tempus imperfectum, 2 could be senaria imperfectum, etc. That we try to keep attribute values closely linked to their names (e.g., "4" for a quarter note, "2" for a half note, etc.) is for the benefit of the humans who will be reading and implementing it.

So, what I'm suggesting is that given this is a _functional_ change, we have a couple options:

1. Standardise the same functional behaviours to a consistent pattern in `<mensur>`. I would recommend two letters ("si") but you could spell it out hyphenated ("senaria-imperfectum"). You then capture the actual visual evidence in the `<mensuration>` tag in the header, e.g., `<mensuration divisio="si">.i.</mensuration>`. 
2. If you wish to persist with 'i' and 'si', you can use the existing `<dir>` element, e.g., `<dir>.s.i.</dir>` already, without needing these changes. A word of warning, however: This will make accurate software writing, parsing, and rendering far more difficult, as we cannot standardise the values of this, so we cannot write software "to spec". I would not recommend this course of action.

rettinghaus:
So why is the sign attribute part of the logical domain? The sign itself doesn't have to carry a meaning. And the values "O" and "C" are just imitating the look of the used sign. "O" and "C" are a pretty straight forward encoding method nevertheless. I think `@sign` and `@dots` should be in the visual domain.

ahankinson:
If I am understanding correctly, it’s the fact that multiple visual signs can have the same functional meaning.

bwbohl:
@ahankinson as you've been involved in this discussion, could you take a lead on sorting out whether these things are addressed in MEI 4.0?

giulianodibacco:
It is really time to revive this discussion, which I abandoned for lack of technical knowledge. I can see the point for having a sign (any sign) in the logical domain when the meaning is one and exclusive, but in the case of mensuration signs we have evidence of multiple signs used to indicate the same thing in different contexts and from different notational cultures. We know that mensuration signs may or may not be written down, we know that they have to be considered as a sort of visual aids inserted by a scribe to facilitate interpretation. (In some cases their presence or absence in a source can lead on hypotheses on the expertise of the scribes/performers.) So my original point was that indeed there should be a way to univocally recognize the mensuration -- some "logical" attributes of <mensur> that have univocal values (and these are tempus & prolatio, to which I suggested adding prolatio). Then the encoded original may also contain a sign (any sign) that I would like to represent "as is" and where it appears. Further comments? @rettinghaus @ahankinson @kepper @martha-thomae

martha-thomae:
I agree with @giulianodibacco. And I think that this is a good argument to move the mensuration signs into the visual domain (this includes attributes such as `@sign`, `@dot`, `@slash`, `@num` and `@numbase`), while the attributes that encode the semantics of the mensuration (`@modusmaior`, `@modusminor`, `@tempus`, `@prolatio`, and the newly added `@divisio`) should remain in the logical domain. Comments?

This is what I was going to include in my next pull request (which is ready to go), but we can make those changes here, Giulianno, if you prefer that.

annplaksin:
@martha-thomae I agree with having attributes about mensuration signs in the visual domain and attributes describing the mensuration in the logical domain. 
I would see this as well as a part of the solution of my issue concerning proportion signs: https://github.com/music-encoding/mensural-ig/issues/4

giulianodibacco:
I wonder whether `@num` and `@numbase` require some prudence. If we move these as well into the visual domain, how do we represent the actual (logical) measurement of time? (I may have misunderstood the point, hence my post on MEI-Mens-L)

giulianodibacco:
About the examples required for the Guidelines, the ones I posted a while ago in this thread would be good but I am afraid I don't think we can get images in the public domain from the sources using this notation. We need to ask what the policy and the requirements of MEI are at the moment.

martha-thomae:
> About the examples required for the Guidelines, the ones I posted a while ago in this thread would be good but I am afraid I don't think we can get images in the public domain from the sources using this notation. We need to ask what the policy and the requirements of MEI are at the moment.

Yes, I was thinking on these images for illustrating that part. But you are right, I won't use them if I don't know for sure they are public domain. I think that for now, it is alright to not have images for this. But the images that are still needed in the guidelines are the ones for stems (notes with different forms of stems, and notes with two stems).

martha-thomae:
> I wonder whether `@num` and `@numbase` require some prudence. If we move these as well into the visual domain, how do we represent the actual (logical) measurement of time? (I may have misunderstood the point, hence my post on MEI-Mens-L)

True! I didn't consider that. You are right. Maybe `@num` and `@numbase` should not be moved, since they are used by many other elements.

PR #541, opened 2018-07-07
remove mensur and propert from scoreDef/staffDef

rettinghaus:
This PR removes `mensur` and `proport` (with respective attributes) from `scoreDef`/`staffDef` as discussed and approved at the Mensural Interest Group's meeting at College Park. 

Two main reasons: 
`<meter>` and the respective short hand notation is used on `scoreDef`, because meter changes happen between two measures in measured music. As mensural music is un-measured, there is basically no need to keep it on `scoreDef`. 
When the mensur changes however, the new `mensur` has to be encoded in the stream (i.e. the `layer`). So the traditional approach would be something like this: 
```xml
<staffDef mensur.sign=„O" mensur.dot="true" mensur.slash="1" mensur.color="red" />
<section>
   <staff n="1">
       <layer>
	    <note />
	    <note />
	    <note />
	    <mensur loc="3" sign="C" />
	    <note />
	    <note />
	    <note />
	</layer>
   </staff>
</section>
```
But basically there is no point for the encoding of the first mensuration sign to be distinct from the following ones. 
The same encoding would make more sense (and raise consistency) this way: 
```xml
<section>
   <staff n="1">
       <layer>
	    <mensur sign="O" dot="true" slash="1" color="red" />
	    <note />
	    <note />
	    <note />
	    <mensur loc="3" sign="C" />
	    <note />
	    <note />
	    <note />
	</layer>
   </staff>
</section>
```
Additionally this decreases the "zoo of attributes" on the def elements.


giulianodibacco:
This is very useful and very logical. We had almost no discussion at the IG meeting because it looked immediately as the best option to have <mensur> where it happens in the stream. It would be good to include in the guidelines two (real) examples: 1) two mensuration signs in the stream (a change of mensuration) of the kind @rettinghaus exemplified. And 2) an example with no mensuration signs (i.e. where the mensura is given by the context). I think that this pair of examples will reinforce the concept. -- But then I have a big question here: are then mensuration signs part of the logical, or visual domain? The question arose some time ago because there are cases where we have more than one sign meaning the same mensuration. See the discussion in #537

kepper (approved):
this has been properly reviewed in the context of #634

kepper:
thanks @martha-thomae for your work on this!!!

martha-thomae:
Thanks to you for all the help and patience with a novice in ODD-related stuff.

bwbohl:
am I right to assume tat this is one of the PRs breaking backward compatibility, ergo making the new MEI release 5.0?

martha-thomae:
Yes, you are right. This is supposed to go to MEI version 5.0.

Issue #582, opened 2019-03-29
Create MEI 4.0.1 release


ahankinson:
I think we need to do a critical bugfix release to push the fix in #581 out to master.

bwbohl:
is there anything else that might be bugfixed on the same go? Maybe we should revisit all open issues

wergo:
May I point to #577 that I raised some time ago. It seems to be a quick fix (if I did not misunderstand). Thanks

wergo:
May I additionally point to this discussion https://github.com/rism-ch/verovio/issues/774. 

`<pedal>` elements should be allowed to have `@vgrp` attributes, just as `<dynam>` and `<hairpin>` are.

bwbohl:
Hi @wergo thanks for spotting this! It pretty obviously seems to be an oversight. Although we will not fix it in v4.0.1, as we want  the critical bugfix asap, we will fix this in a soon to come bugfix release.

ahankinson:
Fixed.

PR #584, opened 2019-04-04
first beta of MEI Basic

kepper:
this adds a draft for MEI basic, based on our discussions in Oxford. It's still not complete and needs significant consideration of:

- how much metadata should be included
- make sure controlevents have one method for placement only (but potentially a different for every type)

We could continue to develop in this branch, or merge it into develop.


ahankinson (approved):

customizations/mei-basic.xml
```diff
@@ -0,0 +1,405 @@
...
+                    <constraint>
+                        <sch:ns xmlns:sch="http://purl.oclc.org/dsdl/schematron" prefix="mei"
+                            uri="http://www.music-encoding.org/ns/mei"/>
+                        <sch:ns xmlns:sch="http://purl.oclc.org/dsdl/schematron" prefix="xlink"
+                            uri="http://www.w3.org/1999/xlink"/>
+                    </constraint>
+                </constraintSpec>
+                
+                <moduleRef key="MEI"/>
+                <moduleRef key="MEI.cmn" except="beamSpan bend bracketSpan meterSig meterSigGrp mSpace oLayer ossia oStaff tupletSpan"/>
```
lpugin: not sure about `meterSig` being excluded.

lpugin (approved):

kepper:
The spreadsheet isn’t addressing everything, but we made a statement there about meterSig to be excluded. I think the idea was to encourage people to use scoreDef between measures instead. Basic should exclude meters changing in the middle of a measure (no clue how this would work anyway…)

lpugin:
> The spreadsheet isn’t addressing everything, but we made a statement there about meterSig to be excluded. I think the idea was to encourage people to use scoreDef between measures instead. Basic should exclude meters changing in the middle of a measure (no clue how this would work anyway…)

Makes sense. Is `keySig` excluded as well? It should be

kepper:
If I remember correctly (in need of a power outlet to verify…) the argument to keep keySig was that keys may actually change in the middle of a measure. However, we should make sure that it’s not used inside scoreDef…

ahankinson:
I would like to see us address #582 first before merging this PR. MEI Basic shouldn't land as part of 4.0.1.

kepper:
We now have a first beta of MEI Basic. It hasn't been tested to great detail yet, though, so please have a look. I'm particularly curious to hear from @rettinghaus and @lpugin if it has everything that's needed for rendering with Meiler and Verovio, respectively.

kepper:
Dear @music-encoding/core-contributors, dear all, I finally managed to get around with the draft of MEI Basic, as we've discussed in the past. Please have a look and see, I surely missed some aspects. It is pretty restrictive, but is build to allow integration of additional features in the future without causing many disruptions -> very often, we picked elements, but disallowed all advanced attributes on them for the time being. For the `@meiversion`-attribute, the allowed value is `4.0.1-rc1+basic`. I like the `+basic` part at the end (which is a possible part in semantic versioning), because it allows us to indicate MEI profiles in the schema. We may even have multiple profile combinations (something like cmn + mensural), and control them through a regex, instead of a value list as we have right now. However, that'll be a separate issue to discuss…

kepper:
btw., could someone more familiar with Travis look into the build failures? They don't seem to originate from problems in the repo, but in the Travis setup. @ahankinson, are you able to help out?

martha-thomae:
> btw., could someone more familiar with Travis look into the build failures? They don't seem to originate from problems in the repo, but in the Travis setup. @ahankinson, are you able to help out?

I think the Travis issue is the one I fixed on the pull request for the mensural module (see https://github.com/music-encoding/music-encoding/pull/634/commits/4fbb8315501a2532d0b77086956c9a7b7865b6f3). I was the first one who got the building error, so I added this fix in the PR so that, when accepted, it would build fine for all. (I don't know Travis, this was with the help of someone from my Lab that does. It built ok after that commit.)

lpugin (commented):
`@stem.dir` missing on `<chord>`
`@fontname` missing on `<rend>`
`@accid` is allowed on `<note>` but not `@accid.ges`
`@vgrp` missing on `<dynam>` (I would still consider basic, but maybe not everybody would agree)
`@oloc` and `@ploc` on `<rest>` (ditto)

lpugin:
For evaluating it with MNX, we should use the test files provided in their converter https://github.com/w3c/mnxconverter/tree/master/tests
(We could easily have an importer in Verovio for them. To test with MEI basic we should add an option to set a specific schema in the MEI output, but this is straighforward)

PR #586, opened 2019-04-10
added support for @stem.sameas

### LINKED ISSUE #551: Stem notes in different layers together
It is a common practice in music engraving when two parts or instruments are notated on a common staff they are usually treated as separate entities, i.e. with individual stems pointing in opposite directions. Once they reach the interval of the octave or more, however, the parts are stemmed together, as can be seen in the example (Mozart, K. 425/1) below. The traditional approach in MEI would be to code the notes on the first beat in measure 1 individually in two layers while the notes on the last beat of measure 1 and on the first beat of measure 2 are coded with the `chord` element (with some `space `elements in the second layer). However, these are only chords in appearance (or pseudo chords) but not real chords, since the notes belong to different parts/instruments. 

stemming_example_k425

```
<staff xml:id="staff_31425" n="1">
	<layer xml:id="layer_31431" n="1">
		<note xml:id="note_31437" pname="g" oct="5" dur="4" tstamp="1"/>
		<rest xml:id="rest_31443" dur="4" tstamp="2"/>
		<rest xml:id="rest_31449" dur="4" tstamp="3"/>
		<beam xml:id="beam_31450">
			<chord xml:id="chord_31450" dur="8" stem.dir="down">
                                <note xml:id="note_31455" pname="d" oct="6" tstamp="4"/>
				<note xml:id="note_31494" pname="d" oct="5" tstamp="4"/>
			</chord>
			<chord xml:id="chord_31499" dur="8">
				<note xml:id="note_31461" pname="d" oct="6" tstamp="4.5"/>
				<note xml:id="note_31500" pname="d" oct="5" tstamp="4.5"/>
			</chord>
		</beam>
	</layer>
	<layer xml:id="layer_31470" n="2">
		<note xml:id="note_31476" pname="b" oct="4" dur="4" tstamp="1"/>
		<space xml:id="space_31482" dur="4" tstamp="2"/>
		<space xml:id="space_31488" dur="4" tstamp="3"/>
		<space xml:id="space_31494" dur="8" tstamp="4"/>                                                            
		<space xml:id="space_31500" dur="8" tstamp="4.5"/>
	</layer>
</staff>
```

The easiest way to fix this (and to avoid 'false' chords) would be to have both parts encoded in two layers throughout the whole measure and have an attribute that allows the notes on the last beat to be stemmed together.
What comes to mind is the attribute `stem.with `which, however, - despite its name - has a different function and definition; it works, so far, only between staves and not between layers. Therefore, we propose adding a suitable attribute, e.g. in `note` in layer 2, which could be called `commonStem`; its value could be either the xml:id in the corresponding `note` in layer 1 or just the xml:id of the `layer`. 
Here is the same example as above in a ficticious coding:
```
<staff xml:id="staff_31425" n="1" >
	<layer xml:id="layer_31431" n="1">
		<note xml:id="note_31437" pname="g" oct="5" dur="4" tstamp="1"/>
		<rest xml:id="rest_31443" dur="4" tstamp="2"/>
		<rest xml:id="rest_31449" dur="4" tstamp="3"/>
		<note xml:id="note_31455" pname="d" oct="6" dur="8" tstamp="4"/>
		<note xml:id="note_31461" pname="d" oct="6" dur="8" tstamp="4.5"/>
	</layer>
	<layer xml:id="layer_31470" n="2">
		<note xml:id="note_31476" pname="b" oct="4" dur="4" tstamp="1"/>
		<rest xml:id="rest_31482" copyof="#rest_31443" tstamp="2"/>
		<rest xml:id="rest_31488" copyof="# rest_31449" tstamp="3"/>
		<note xml:id="note_31494" pname="d" oct="5" dur="8" tstamp="4" commonStem="#note_31455"/>
		<note xml:id="note_31500" pname="d" oct="5" dur="8" tstamp="4.5" commonStem="#note_31461"/>
	</layer>
</staff>

```
The condition for the applicability of this method is that both notes that are stemmed together must have the same timestamp.
This solution has the advantage of maintaining the integrity of the part and the identity of part and layer. It would thus make automatic part extraction much easier.
Therefore, we suggest to either modify the `stem.with` attribute or to add a new attribute here tentatively called `commonStem`.

frakel:
- extends att.stems class with @stem.sameas which is needed to correctly encode orchestral scores 
- adds support for notes that share one stem but are located in different layers
- fixes #551


ahankinson:
As this is a new feature, not a bugfix I would like to defer this for merging until after we handle #582 .

ahankinson:
The indents and formatting of the contribution seem to be different than the surrounding style; could you please re-format to match the existing style?

ahankinson (approved):

lpugin:
Can this be merged?

kepper (changes requested):
In general, I agree. There is only one thing that isn't clear to me (yet), and that I would like to see clarified in the documentation at least: How does this relate to stems across staves? There should be at least a reference to the concepts for that, and a specific hint that `@stem.sameas` isn't intended to cover that use case. Or, if it is intended to cover that as well, the Schematron probably needs to be changed. All that could perhaps be blamed to my ignorance, but I still think others may run into the same issue…

lpugin:
@frakel : could you clarify this?

frakel:
It does not relate in any case to stems across staves as `@stem.sameas` is meant to work only in one staff with two layers. 
It has absolutely nothing to do with crossing staves, as the two layers with the corresponding notes have to be in the same staff. `@stem.with` covers a totally different issue. 
I can add this to the description if necessary.
(sorry for procrastinating but I was away and unfortunately I am ill now)

kepper:
I would actually like to see a sentence about `@stem.with` in the `<desc>` of this. I would also love to have a paragraph about these two in the guidelines. @frakel, could you please add something as soon as you're back? Thanks :-)

CONTRIBUTING.md
```diff
@@ -62,7 +62,7 @@ Do **not** simply reference another issue or pull request by number in the summa
 
 The details of the commit go in the body. Specifically, the body should include the motivation for the change.  For commits that fix bugs you should contrast behavior before the commit with behavior after the commit.
 
-### Message Summary
+### Message Footer
```
bwbohl, line 65: This change has nothing to do with the changes addressed by the issue, please undo this one.

bwbohl (changes requested):
Basically I do approve, except for the one change in CONTRIBUTING.md and considering the change to the <desc> is made according to @kepper 's request.

Thanks you very much!

bwbohl:
Now that we bugfixed MEI 4 I think we can move on with this, as Part of MEI 4.1

craigsapp:
How is incorporating this feature going?  This would be very useful for encoding parts sharing the same staff in orchestral scores.  I.e., the parts can be stored in two layers in the data allowing the parts to be easily separated (useful for part extraction), but the visual syntax could be a single layer with chords when the rhythms of the two parts match.  So in other words, it would be useful to have this feature available for those working on orchestral or other large scores so that they can encode optimally (and not have to recode it later :-).

frakel:
I added now some lines in the documentation as @kepper asked. 
@bwbohl: As the original fork does no longer exist I hope you can nonetheless merge just the MEI.shared.xml file and ignore any other change I made in that branch. Otherwise I guess I would have to make a new PR from another fork ...

kepper:
we need to resolve the conflicts, but other than that, this seems fine with me

kepper:
closed by https://github.com/music-encoding/music-encoding/commit/8ed927144666dcfc57246bcf90736065038ebb61. Thanks @frakel for your work on this, and apologies for the long delay…

PR #591, opened 2019-05-13
Fix: custos and clef elements should be within layer

vigliensoni:
With the current MEI Neume schema, custos and clef elements are expected 
to be within syllable. I asked MEI Neume IG and this seems to be an unintended 
mistake in the schema rather than something they need.

When trying to validate a basic MEI neume file, a few warnings are returned in 
regard to clef and custos elements not allowed as children of syllable.

```
jing mei-Neumes.rng example-neume.mei
```

```
example-neume.mei:14:35: error: element "clef" not allowed here;
expected the element end-tag or element "add", "anchoredText",
"annot", "app", "cb", "choice", "colLayout", "corr", "curve",
"damage", "del", "gap", "handShift", "line", "midi", "orig", "pb",
"reg", "restore", "sb", "sic", "subst", "supplied", "syllable" or
"unclear"

example-neume.mei:24:36: error: element "custos" not allowed here;
expected the element end-tag or element "add", "anchoredText",
"annot", "app", "cb", "choice", "colLayout", "corr", "curve",
"damage", "del", "gap", "handShift", "line", "midi", "orig", "pb",
"reg", "restore", "sb", "sic", "subst", "supplied", "syllable" or
"unclear"
```

If, on the contrary, custos and clefs are children of syllable, validation passes.

This issue is related to the replace happening on https://github.com/music-encoding/music-encoding/blob/develop/customizations/mei-Neumes.xml#L113-L120, which this PR fixes.

### DIFF
--- a/customizations/mei-Neumes.xml
+++ b/customizations/mei-Neumes.xml
@@ -111,13 +111,13 @@
         <classSpec ident="model.layerPart.cmn" module="MEI.shared" type="model" mode="delete"/>
 
         <!-- Disallow events within layer -->
-        <classSpec ident="model.eventLike" module="MEI.shared" type="model" mode="replace">
+        <!-- <classSpec ident="model.eventLike" module="MEI.shared" type="model" mode="replace">
           <desc>groups event elements that occur in all notational repertoires.</desc>
           <classes>
             <memberOf key="model.syllablePart"/>
             <memberOf key="model.rdgPart.critapp"/>
           </classes>
-        </classSpec>
+        </classSpec> -->
 
         <!-- Disallow neume events within layer; only permitted in mei-all customization -->
         <classSpec ident="model.eventLike.neumes" module="MEI.neumes" type="model" mode="replace">
--- a/customizations/mei-Neumes.xml
+++ b/customizations/mei-Neumes.xml
@@ -127,6 +127,18 @@
             <memberOf key="model.rdgPart.critapp"/>
           </classes>
         </classSpec>
+        
+        <!-- Allow custos and clef to appear within layer; compensate side effect of removing membership in model.layerPart from model.eventLike -->
+        <elementSpec ident="custos" module="MEI.shared" mode="change">
+          <classes mode="change">
+            <memberOf key="model.layerPart.neumes" mode="add"/>
+          </classes>
+        </elementSpec>
+        <elementSpec ident="clef" module="MEI.shared" mode="change">
+          <classes mode="change">
+            <memberOf key="model.layerPart.neumes" mode="add"/>
+          </classes>
+        </elementSpec>
 
       </schemaSpec>
     </body>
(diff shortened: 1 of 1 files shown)


kepper (commented):
The proposed solution seems a bit over the top – yes, we may need clefs and custos in layer, but we don't want other events like notes in there. What happens here is that membership in model.layerPart is removed, and membership in model.rdgPart.critApp is added. We probably still need the latter, it's just that the first takes away too much. I haven't tried it yet, but perhaps the smallest possible change would be to make custos and clef members of model.eventLike.neumes – then, they should find a different way into layer. I'll investigate…

customizations/mei-Neumes.xml
```diff
@@ -111,13 +111,13 @@
         <classSpec ident="model.layerPart.cmn" module="MEI.shared" type="model" mode="delete"/>
 
         <!-- Disallow events within layer -->
-        <classSpec ident="model.eventLike" module="MEI.shared" type="model" mode="replace">
+        <!-- <classSpec ident="model.eventLike" module="MEI.shared" type="model" mode="replace">
```
kepper, line 114: it shouldn't be necessary anymore to comment lines 114 to 120. Instead, this will have side effects on editorial markup, as it effectively removes model.rdgPart.critapp.

vigliensoni: That is indeed correct, with the changes you suggested to allow custos and clef to appear within layer:

https://github.com/vigliensoni/music-encoding/blob/414d632919632d2f6c2dd583ff867660ce6af35d/customizations/mei-Neumes.xml#L131-L141

This is no longer required

vigliensoni: @kepper The change you propose actually allows to have `clef` and `custos` elements inside `layer`, which is the expected behaviour. However, it does not warn about having these elements inside a `syllable`. 

The expected behaviour should be to return a warning:

`error: element "clef" not allowed here; expected the element end-tag or element "add", "anchoredText", "annot", "app", "cb", "choice", "colLayout", "corr", "curve", "damage", "del", "div", "ending", "expansion", "gap", "handShift", "line", "orig", "pb", "reg", "relation", "relationList", "restore", "sb", "scoreDef", "section", "sic", "staff", "staffDef", "subst", "supplied" or "unclear"`

kepper: ok, will look, but probably not before Tuesday…

kepper: Hi @vigliensoni, apologies for the extreme delay :-( I've tried to address your issue in https://github.com/vigliensoni/music-encoding/pull/2. Could you please have another look for potential side effects? We can still get this into the schema on October 29 (assuming that today's ODD Friday will be too soon to resolve everything). Again, apologies for dropping that ball…

kepper (approved):
thanks for checking this again, @vigliensoni :-) I'm glad we can get this in now 👍 Thanks for both your work on this, and your patience ;-)

Issue #590, opened 2019-05-10
key signatures: written vs. gestural


rettinghaus:
Perhaps I missed something, but I don't understand why 
`scoreDef@key.accid` is of `data.ACCIDENTAL.GESTURAL`
whereas `keySig@accid` is of `data.ACCIDENTAL.WRITTEN`.

lpugin:
@rettinghaus how changing this would break compatibility? In other words, 4.1 or 5.0?

rettinghaus:
I don't think this would break anything in existing encodings. So more like `4.0.2`?

bwbohl:
I think it is not really a mistake. When comparing the two data types, it becomes obvious that `data.ACCIDENTAL.GESTURAL` eliminates graphically distinct but logically identical values. The `<scoreDef>` as a basic _analytical_ definition of the score uses the gestural set for that reason. Whereas `<keySig>`(when occurring in `<layer>`) is part of the graphical domain. I understand that this somehow seems a bit awkward due to the name of the data type.

@rettinghaus is correct to assume changing the allowed values in `scoreDef@key.accid` to `data.ACCIDENTAL.WRITTEN` would not cause backward compatibility breaks.

This seems to bring up the whole domain discussions again (never mind, that's something good ;-) and the question would be what domain `<keySig>` in `<staffDef>` belongs to?!

bwbohl:
maybe this could be tackled on the way with #1010

bwbohl:
how is this related to #1170?

musicEnfanthen:
I guess the change in #1170 made this issue obsolete. Do you agree @rettinghaus?

rettinghaus:
@musicEnfanthen on the contrary. And @bwbohl's assumption is wrong. `scoreDef@key.accid` and `keySig@accid` are both in the analytical domain, because they both are additional indicators for the tonic, there is nothing visual about this. 

I'll prepare a PR to resolve this.

rettinghaus:
@riedde why was this moved to Backlog?

Issue #525, opened 2018-06-06
Harmonize key.sig et al. attributes on scoreDef, staffDef, key and keySig


craigsapp:
Since `staffDef@key.sig.showchange` is changing to `staffDef@keysig.showchange` in MEI 4, it is probably best to change `staffDef@key.sig` to `staffDef@keysig` for consistency.

There are `<key>` and `<keySig>` elements:

http://music-encoding.org/guidelines/v3/elements/key.html
http://music-encoding.org/guidelines/v3/elements/keysig.html

`<key>` should be used for analytical purposes, and `<keySig>` should be for print purposes, so it is best that `staffDef@key.sig.showchange` is changed to `staffDef@keysig.showChange` since it refers to the printed key signature.  Likewise `staffDef@key.sig` should change to `staffDef@keysig` for the same reason.

`<keySig>` has `@sig.showchange`.  Perhaps this should be changed to `@showchange` to also maintain a uniform style?

http://music-encoding.org/guidelines/v3/elements/keysig.html

rettinghaus:
Do you mean to use `@keysig` `@keypname` `@keyaccid` `@keymode` etc on `staffDef`, or `@keysig.sig` `@keysig.pname` `@keysig.accid` `@keysig.mode`? 
(Only) the latter would make some sense, as they are shorthands for `keySig@pname` etc.

ahankinson:
I've reverted the commit and re-opened this issue for discussion.

lpugin:
I guess this change was motivated by another one, namely `@key.sig.showchange` to `@keysig.showchange` and `@key.sig.show` to `@keysig.show`. Maybe these should be reverted as well?

craigsapp:
There are two elements to keep in mind:

(1) `<key>` "Key captures information about tonal center and mode."
http://music-encoding.org/guidelines/v3/elements/key.html
"This element is used exclusively within bibliographic descriptions. Do not confuse this element with keySig, which is used within the body of an MEI file to record this data."

(2) `<keySig>` "Written key signature."
http://music-encoding.org/guidelines/v3/elements/keysig.html

[...]

PR #598, opened 2019-06-05
revert name from KEYFIFTHS to KEYSIGNATURE

### LINKED ISSUE #592: data_KEYSIGNATURE vs data.KEYFIFTHS
82e2648261eaf6afecc32e976789dd67fbab36bc seems to be part of an unfinished revision which we discussed in #525. Same with #590. 

I think we should revert these changes for 4.0.2 and (if we manage to clear things up) move them to 5.0.0.

rettinghaus:
fixes #592

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -1530,7 +1530,7 @@
       <rng:data type="time"/>
     </content>
   </macroSpec>
-  <macroSpec ident="data.KEYFIFTHS" module="MEI" type="dt">
+  <macroSpec ident="data.KEYSIGNATURE" module="MEI" type="dt">
     <desc>Indicates the location of the tonic in the circle of fifths.</desc>
     <content>
       <rng:data type="token">
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -1298,7 +1298,7 @@
       <attDef ident="sig" usage="opt">
         <desc>Written key signature.</desc>
         <datatype maxOccurs="unbounded">
-          <rng:ref name="data.KEYFIFTHS"/>
+          <rng:ref name="data.KEYSIGNATURE"/>
         </datatype>
       </attDef>
     </attList>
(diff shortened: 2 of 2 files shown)


source/modules/MEI.xml
```diff
@@ -473,8 +473,7 @@
     </content>
   </macroSpec>
   <macroSpec ident="data.COLORNAMES" module="MEI" type="dt">
-    <desc>List of SVG-recognized color keyword names described at
-      https://www.w3.org/TR/SVG/types.html#ColorKeywords.</desc>
+    <desc>List of named colors from CSS Color Module Level 4.</desc>
```
bwbohl: This change actually doesn't belong here

source/modules/MEI.xml
```diff
@@ -2567,8 +2575,8 @@
     </content>
     <remarks>
       <p>Instrument names are taken from the list at <ref
-        target="http://www.midi.org/techspecs/gm1sound.php"
-        >http://www.midi.org/techspecs/gm1sound.php</ref>. </p>
+        target="https://www.midi.org/specifications/item/gm-level-1-sound-set"
```
bwbohl: nothing to do with issue

source/modules/MEI.xml
```diff
@@ -920,6 +922,12 @@
         </valItem>
       </valList>
     </content>
+    <remarks>
+      <p>Color names are taken from the list at <ref
```
bwbohl: nothing to do with issue

bwbohl (changes requested):

bwbohl:
All in all I'm not sure whether discussion of #525 and #590 and #592 are really at a point that can be considered decided once and for all. I don't like the idea of changing it back and forth. I think we have to come to a definite decision on those issues before merging.

bwbohl:
Maybe we should even try to involve MEI-L on this?

source/modules/MEI.xml
rettinghaus: something got mixed up

rettinghaus:
@bwbohl changing it back and forth wasn't the idea here. The intention is to revert it, until a definite decision on all those matters has been made.

rettinghaus:
@lpugin perhaps you want to approve this?

lpugin:
@bwbohl ?

kepper:
FWIW, I'm also not convinced to revert at this point. I'd like to have a plan before we take any further steps, even if they're only backwards…

Issue #524, opened 2018-06-04
More elements with @color


craigsapp:
It would be useful if all visible elements could have `@color`

    http://music-encoding.org/guidelines/v3/attribute-classes/att.color.html

Of particular interest is `<harm>`.  Also `<syl>` to color individual syllables in a `<verse>`.  Also applying to `<dir>` would be particularly useful.

rettinghaus:
The original idea was to have `@color` on the `<rend>` subelement for textual elements,
see #337.

craigsapp:
Thanks!  That should work for most everything I will want to color.

Example:

<img width="317" alt="screen shot 2018-06-04 at 12 19 55 am" src="https://user-images.githubusercontent.com/3487289/40903604-be9d7c08-678c-11e8-94a4-c40a67e4d062.png">

MEI data:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="4.0.0">
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title />
            </titleStmt>
            <pubStmt />
        </fileDesc>
        <encodingDesc>
            <appInfo>
                <application isodate="2018-06-04T00:18:09" version="2.0.0-dev-4b57efb-dirty">
                    <name>Verovio</name>
                    <p>Transcoded from Humdrum</p>
                </application>
            </appInfo>
        </encodingDesc>
        <workDesc>
            <work>
                <titleStmt>
                    <title xml:id="title-L1" analog="humdrum:Xfi" type="translated">extract -s 1,0,0</title>
                </titleStmt>
            </work>
        </workDesc>
        <extMeta>
            <frames xmlns:humxml="http://www.humdrum.org/ns/humxml">
                <metaFrame n="0" token="!!!Xfilter: extract -s 1,0,0" xml:id="loc0">
                    <frameInfo>
                        <startTime float="0" />
                        <frameType>reference</frameType>
                        <referenceKey>Xfilter</referenceKey>
                        <referenceValue>extract -s 1,0,0</referenceValue>
                    </frameInfo>
[...]

Issue #346, opened 2016-09-19
add color to meterSig


rettinghaus:
`@color` is missing on `<meterSig>`
(whereas `<mensur>` has it)

pe-ro:
Will fix in the next release.

PR #611, opened 2019-09-10
allow color on key signatures

rettinghaus:
This PR makes  `att.keyAccid.vis` and `att.keySig.vis` member of `att.color`.

### DIFF
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -752,6 +752,7 @@
     <desc>Visual domain attributes.</desc>
     <classes>
       <memberOf key="att.altSym"/>
+      <memberOf key="att.color"/>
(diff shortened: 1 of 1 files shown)


rettinghaus:
@bwbohl could you check and approve?

ahankinson:
Is there an issue or some other explanation of the reason for this change available?

rettinghaus:
I didn't open another issue for this, as I thought it would be self-explanatory. 
Refers to issues #346 #517 #518 #524

ahankinson:
Great, that's all I needed -- just some reference so that we have a record of why the change is being requested.

ahankinson (approved):

rettinghaus:
Thanks!

PR #612, opened 2019-09-11
add turkish accidentals

rettinghaus:
This PR adds (very much needed) [turkish accidentals](https://www.smufl.org/version/latest/range/arelEzgiUzdilekAeuAccidentals/) as values to `data.ACCIDENTAL.WRITTEN`. 
To keep it nice and clean, I've split up `data.ACCIDENTAL.WRITTEN` in `data.ACCIDENTAL.WRITTEN.basic`, `data.ACCIDENTAL.WRITTEN.extended`, and the new `data.ACCIDENTAL.WRITTEN.aeu`.
The new values are identical to those used in [Lilypond](http://lilypond.org/doc/v2.18/Documentation/notation/turkish-classical-music.html).

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -79,6 +79,30 @@
         <valItem ident="3qs">
           <desc>3/4-tone sharp accidental.</desc>
         </valItem>
+        <valItem ident="bms">
+          <desc>Büyük mücenneb (sharp).</desc>
+        </valItem>
+        <valItem ident="kms">
+          <desc>Küçük mücenneb (sharp).</desc>
+        </valItem>
+        <valItem ident="bs">
+          <desc>Bakiye (sharp).</desc>
+        </valItem>
+        <valItem ident="ks">
+          <desc>Koma (sharp).</desc>
+        </valItem>
+        <valItem ident="kf">
+          <desc>Koma (flat).</desc>
+        </valItem>
+        <valItem ident="bf">
+          <desc>Bakiye (flat).</desc>
+        </valItem>
+        <valItem ident="kmf">
+          <desc>Küçük mücenneb (flat).</desc>
+        </valItem>
+        <valItem ident="bmf">
+          <desc>Büyük mücenneb (flat).</desc>
+        </valItem>
       </valList>
     </content>
     <remarks>
@@ -2980,7 +3004,7 @@
       <constraint>
         <sch:rule context="@place">
           <sch:assert
-            test="not((some $token in tokenize(normalize-space(.),' ') satisfies 
+            test="not((some $token in tokenize(normalize-space(.),' ') satisfies
             $token =('below','above','between','within')) and count(tokenize(normalize-space(.),' ')) gt 1)"
             >Other values not permitted when 'above', 'below', 'between' or 'within' is
             present.</sch:assert>
@@ -3081,7 +3105,7 @@
       <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:space="preserve" valid="feasible">
 <note slur="i1 i2"/>
 <note slur="t1"/>
-<note slur="t2"/>   
+<note slur="t2"/>
         </egXML>
     </exemplum>
   </macroSpec>


lpugin:
I am not convinced by the use of so short abbreviations. This makes the encoding very cryptic. Why not use values such as "aeu.bakiye.sharp"? We could even go further and say that we use abbreviations only for the ones listed in data.ACCITENDAL.GESTURAL, and have longer names for the other values in data.ACCIDENTAL.WRITTEN. This would mean deprecating some of the values there, but I think it would make the list clearer.
I am also not convinced by splitting this into sub data types because nothing make use of the sub types separately. This creates a significant overhead when processing the schema. (It is currently a problem with the LibMEI version used for Verovio and requires the grouping data type class to be written by hand instead of having everything generated automatically.)

source/modules/MEI.xml
```diff
@@ -8,6 +8,21 @@
     <desc>Data type definitions.</desc>
   </moduleSpec>
   <macroSpec ident="data.ACCIDENTAL.WRITTEN" module="MEI" type="dt">
+    <desc>Description of direction with respect to an imaginary compass.</desc>
```
ahankinson: I don’t understand this description. What imaginary compass?

ahankinson (changes requested):

ahankinson:
I’m ok with the abbreviations, since they tend to conform with established methods (fq, ss, etc.)

How do the written and performed (gestural) accidentals interact?

Can we also expect a note in the guidelines about these values?

kepper:
Another thought: Are there any other things which are necessary for turkish classical music? If so, it might be helpful to put that into a separate module, so that one could more easily turn support for these features on and off. If it's just about eight additional accidentals, this might be too much overhead, though…

lpugin:
> I’m ok with the abbreviations, since they tend to conform with established methods (fq, ss, etc.)

I am questioning the method, or at least how it is applied here. My understanding is that very short abbreviations are used when the understanding is obvious. For example `@pname="c"`. It think it works for `s` abbreviating `sharp`, but I do not think it works for `b` abbreviating `Bakiye (sharp)`.  In other words, the encoding is not human readable (at least for me) because I have to look that the Schema documentation to understand what is meant by `c`, `b`, etc.
My feeling is that it actually already does not work for the list we have in data.ACCIDENTAL.WRITTEN and I am proposing to take this opportunity to refactor it.

> How do the written and performed (gestural) accidentals interact?
> Can we also expect a note in the guidelines about these values?

I fully agree this needs to be clarified. I am in the midst of writing a function for converting the values and I have no idea what to do with triple sharp and flat values when converting from written to gestural...

source/modules/MEI.xml
rettinghaus: copy and paste error, sorry!

rettinghaus:
@kepper I agree. Perhaps @annplaksin could say something about that?
@lpugin The abbreviations may not be straight forward, but the ones already in use aren't much better. Also the new ones have been used in other context, so why re-inventing the wheel?  
@ahankinson I'm aware of the gestural values, but haven't thought this through right now. And of course there will be an entry in the guidelines.

lpugin:
@rettinghaus 
>  the ones already in use aren't much better.

Well, this is exactly what I am saying: we should fix them.

> Also the new ones have been used in other context, so why re-inventing the wheel?

Maybe, but mixed with the one we have makes them very difficult to understand. At the very least they should be prefixed, something like "aeu.b", "aeu.c", etc.

You did not comment on the data sub typing aspect. I think having more explicit names (e.g., prefixes) makes it unnecessary.

sertansenturk:
Hi there,

Just wanted to give my 5 cents as someone who worked on Turkish makam music scores in the past. I'm not part of the MEI community, so please ignore my comments if they do not make sense.

I don't have much to add to the naming discussion. However, if you would like to carry on with the abbreviations, I believe they can be made more consistent by changing abbreviations of `Küçük mücenneb` to `km` and `koma` to `k`.

@kepper These are other systems (attempts?) to describe the music, and some of them use different accidentals. I am rusty on that topic, but I can check it back with you all if it is useful for the discussion. Nevertheless, AEU is the common way to notate so I would not bother to work on other classical systems before implementing the [Turkish folk accidentals](https://w3c.github.io/smufl/gitbook/tables/turkish-folk-music-accidentals.html), the other conventional Turkish music notation (my apologies if they are already part of MEI). 

Happy to see that the accidentals are finally arriving!

ahankinson:
Hi @sertansenturk! Your comments are always welcome here. Thank you for contributing!

kepper:
Hi @sertansenturk, glad to have you in this discussion, and your input is highly appreciated :-) I wasn't trying to make a distinction between turkish classical music and turkish folk music – I know so little about these kinds of music that I'm certainly not qualified for any kind of statement, and will surely mix up things. What I tried to do – and that question still stands – is to see if we can cut down the [list](https://music-encoding.org/guidelines/v4/data-types/data.accidental.written.html) of allowed accidental values. We already have 22 allowed values, where most users interested in CWMN will only need 3. MEI already has a mechanism to selectively disable parts of the specification, which aren't of interest to the user, called _modules_. We use those modules for specific repertoires like CMN, tablature or mensural notation, and we're always willing to enrich MEI with more modules for additional repertoires and notation styles. I was wondering if there are more specifics to turkish music than just a small set of accidentals, because I'm not sure if that already justifies the overhead. But at the same time, I do see some value in moving all the existing quarter-tone accidentals into a separate module, so that users can easily turn support for quarter-tones on or off. Do you (or anyone else) has an opinion on that, or could clarify if there are additional specifics to turkish music notation that should be considered? Thanks!

rettinghaus:
@sertansenturk Yes, thank you for your input. 
@lpugin @kepper Do I understand this correct? Sub types are bad and separate modules is good? Isn't it possible to switch of sub types? I do think it would be great to easily turn support for quarter-tones (and other microtones) on or off, that's why I've split them up.

lpugin:
> Do I understand this correct? Sub types are bad and separate modules is good? 

I would not put it like that. Sub types are good. However, as long as they do not belong to distinct modules or they are not used separately by attributes classes, then they basically bring nothing but complications. So I would suggest to stay away from them in such cases. Eventually, when we encounter cases where it make sense to split them because distinct sub groups of values will be used by some schema instances, then we can split them on a case-by-case basis. I guess that if this becomes the norm, then somebody will be willing to find a proper way to handle them with LibMEI / C++ ;-)

bwbohl:
>We already have 22 allowed values, where most users interested in CWMN will only need 3

That's why  we have to teach people not to use mei-all and where the modularisation proposed by @rettinghaus comes in very handy as it allows easy customisation. I would like to introduce more such modularisation, this would also allow a more specific schemata.

And certainly there's more to Turkish or Osman music than just the accidentals. We should encourage @annplaksin and her colleagues from the CMO project (https://corpus-musicae-ottomanicae.de/content/index.xml) to submit a respective module.

annplaksin:
Sorry for the late answer, but I was away from my desk yesterday.
First of all, thank you @rettinghaus for starting this discussion and all the others for participation.  

Since there is currently a lot of stuff to do in our project, I didn't find the time to consolidate my thoughts on integrating the support of turkish music into MEI. But as the topic just arose, I want to give a short report on my current outlines:

I would prefer the integration of spezialized models, since there is in fact more necessary than just the aeu accidentals -- but yes, they are mainly used in our edition. So, we need:
* aeu accidentals in data.ACCIDENTAL.WRITTEN
* aeu accidentals in data.ACCIDENTAL.GESTURAL (since sibmei already exports following notes of the same pitch as a note with accidental within a measure as @accid.ges)
* Support of our key signatures, which consist of more than one kind of accidental on differenct locations. I am aware his could be done right now with `<keyAccid>`, but I just want to mention it explicitly

And yes @sertansenturk is right, there are also special accidentals for folk music, I think [those](https://www.smufl.org/version/latest/range/turkishFolkMusicAccidentals/) are used, but I am not 100% sure right now. I need to check this...

Also, it is necessary to know, that turkish music is not measure-based per se, even if it could be transcribed into measures. It rests upon rhythmic modes, called usûller (sg. usûl). For a proper representation of the repertoire, I would prefer to model those structures.

And as well, in MEI.header we would need `<makam>` / `<maqam>` and `<usul>` instead of `<meter>` and `<key>` for cataloguing expressions and works.

Apart from a module which deals with the structual features of turkish music, independent of a notation system, I am also thinking about creating a model for the notation system, our editors currently work with, called Hampartsum notasi. Since turkish music does not depend on one notation system, in my opinion, these should be separate modules...

And in the matter of transcribing music, there is also one topic I am still not sure on how to model it. A part of our edition is the so called pitch set, which gives a short summary on how the pitches are transcribed -- pitch symbol in Hampartsum -> pitch in staff notation -- which is in fact some visual kind of a key-value list. My first guess was, to put it somewhere into `<encodingDesc>`, but on how this should be done I currently don't have any ideas.

Since this comment just grew way to long, I would like to use this as a kind of a proposal to think this more through. And because I am not the expert in turkish music in our project, it would be wise to have some of those experts involved.  
But yes, if someone of you want to start a MEI.makam module by implementing the aeu accidentals, that would be a great start. :-)

kepper:
Hi @annplaksin, thanks for these thoughts. With all these arguments out, I would suggest to just add the new accidentals to the existing class for the time being, and postpone a more substantial support for turkish music with proper models to MEI5. We all agree that the list of accidentals is already too long and needs more structure. At the same time, these structures seem to cause problems to libmei and subsequently to Verovio. I'm quite sure we also all agree that this must not restrict us from improving our models, but at the same time, I don't want to rush into problems. If someone wants a restricted version of MEI, it's not that much more complicated to turn off some attribute values than it is to turn of some nested data types – those who can do one can probably do the other as well, and for those who don't it doesn't make a difference. This is not an ideal solution, but perhaps sufficient for now?

kepper:
and, of course, we're expecting a workshop on M3 (the MEI Makam Module…) to be hosted in Münster – so it'll be M4?!? ;-)

annplaksin:
@kepper : Yes, it seems fine to me to postpone these questions to a coming MEI5 since I am now trying to focus on reworking the Sibelius export because of MEI4. But it's good, to have outlined my thoughts already, so I will hopefully not forget them...
Yes, I would like to do a workshop too... we just have to see if this will be M3 or M4... ;-)

rettinghaus:
Thanks for the input. I renamed the values and added it directly to the list. 
The values follow now a more stringent and consistent approach:

> Küçük mücenneb (sharp) -> kms
> Koma (flat) -> kf

@sertansenturk @annplaksin are you ok with this?

To keep the splitted version I created a "develop-maqam" branch as a starting point for a later MEI-magam model.

sertansenturk (approved):

sertansenturk:
Glad to be of some help. To the best of my knowledge, everything @annplaksin has pointed out is accurate (in particular, I agree with using `makam` and `usul` in the header). The Turkish folk music symbols she points out in smulf are also the correct set. In my opinion, between AEU and folk music accidentals, you would cover "99%" of the pitch-related requirements from Turkish traditional musics. The rest will be historical/experimental notations like Hampartsum, ebced, Yarman etc. 

I have approved the PR. If there are any more questions now or in the future, I would gladly try to answer. Likewise feel free to tag me in any issues or PRs in the future, where I may be of some help. Finally, I would also like to suggest MEI people to contact Kemal Karaosmanoglu and/or Baris Bozkurt, who are still actively working with Turkish music, in such cases.

annplaksin:
Thanks for your feedback @sertansenturk 
Involving Kemal Karaosmanoglu and Baris Bozkurt is a wonderful idea. It seems like I will be able to speak to them next week.

kepper:
where are we on this? Could someone who's more involved (@rettinghaus, @annplaksin?) get us (or at least me) a short wrap-up? I'm somewhat lost in all that wonderful discussions ;-) Do we have the new values separated somehow? Even if we don't go for a separate module (which I guess I still prefer in the long run), we should have separate classes / data types here… Is this the case, and do we have a consensus on the contents? Please help me out…

rettinghaus:
@kepper as far as I can see everything is fine here, but this is of course just the first step. 
For MEI5 we should have everything separated in a `makam` module. But until we decide to split up `data.ACCIDENTAL.WRITTEN`, as I proposed originally, all the accidentals here are in one list.

kepper:
Hi @rettinghaus – I know I'm sometimes slow, and I've probably taken over the habit of our good old friend @pe-ro to disagree with myself ;-), but I think your original proposal to split up `data.ACCIDENTAL.WRITTEN` is certainly better than putting everything in one basket – even though we agree that it's only a first step… Sorry :-)

rettinghaus:
closing this in favour of #720

PR #616, opened 2019-09-17
add single value

rettinghaus:
This adds a missing "single" value to `@wordpos`.
Shouldn't this be a datatype, too?

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -3163,6 +3163,9 @@
           <valItem ident="m">
             <desc>(medial) neither first nor last syllable.</desc>
           </valItem>
+          <valItem ident="s">
+            <desc>(single) single syllable.</desc>
+          </valItem>
           <valItem ident="t">
             <desc>(terminal) last syllable.</desc>
           </valItem>


rettinghaus:
refers to this discussion: rism-ch/verovio/issues/1143

lpugin:
Sorry, but this does not really make sense to me. Could you provide an existing example where you would need this? Thanks!

bwbohl:
Well, I can see the use: there are indeed words consisting of single syllables and when you intend to have a consistent markup (all syllables indicating their wordpos) a value indicating this is necessary

kepper (approved):
I think the current solution with `@wordpos="i"`, `@wordpos="m"`, `@wordpos="t"` and `no @wordpos` doesn't allow to differentiate between a single-syllable word and no information about syllable positions being provided. Since MEI avoids default values in most cases, I think it improves consistency to add a value to identify single-syllable words.

rettinghaus:
Especially if you follow the argumentation by @craigsapp that `@wordpos` is necessary and `@con` only overrules the word connector given by `@wordpos`. 

Take the following example:
grafik

```xml
<verse>
  <syl con="d">we</syl>
</verse>
```
wouldn't work this way, so you have to encode: 
```xml
<verse>
  <syl wordpos="s" con="d">we</syl>
</verse>
```

lpugin:
I assume it should be "we be-seech" and not "we-be seech". So I would encode it like this:
```xml
<choice>
  <sic>
    <verse>
       <syl wordpos="i" con="d">we</syl>
    </verse>
  </sic>
  <corr>
    <verse>
      <syl>we</syl>
    </verse>
  </corr>
</choice>
```
And corresponding encoding for "be" and "seech"

rettinghaus:
@lpugin That way you create a new word ;-) The connector is wrong, not the syllable.

lpugin:
> That way you create a new word ;-) 
 
Well, not really. The new word it created by the connector, not me... ;-) More seriously, this lyric setting makes no sense. It is basically wrong. So I understand you might be interested to re-encode it and re-engrave it. But if you want to do so, I would recommend not to use an infrastructure that relies on strong semantic encoding, precisely because semantic encoding is there to avoid such things. Maybe you can find a way to represent this in MEI that you will consider semantically satisfactory, but I do not think we should tweak Verovio for it to be able to render that encoding in order to produce something that is obviously wrong. I think we have enough to do with what is correct. ;-)

ndubo:
If we are serious about @wordpos, we should fix this case first: 
In Italian, but I assume in other (Romance) languages as well, there may be two (sometimes even more) syllables on one note, one of the at the end of a word, the other one at the beginning (see third note). 

wordpos_01

In the present case it is "-da" (which is wordpos="t") and "I-" (which is wordpos="i"). The correct coding therefore should be:
```

<note dur="8" oct="4" pname="f" tstamp="3" xml:id="note_7719">
	<verse xml:id="verse_7725">
		<syl con="s" wordpos="t" xml:id="syl_7731">da</syl>
		<syl con="d" wordpos="i" xml:id="syl_7732">I</syl>
	</verse>
</note>

```
(Alternatively, in the first syllable con="t" or "c" is possible, but I would not want MEI or Verovio to force me using any of those).

The output of this code in Verovio is: 

wordpos_02
  
It works correctly only with the following, where the wordpos for a combined "-da I-" is "m" which is simply wrong:

```
<note accid.ges="f" dots="1" dur="8" oct="4" pname="b" tstamp="1" xml:id="note_8127">
	<verse xml:id="verse_8133">
		<syl con="d" wordpos="i" xml:id="syl_8139">me</syl>
	</verse>
</note>

``` 
I think fixing this should have priority over any automatism that connects `@wordpos` and `@con`.
I support keeping them seperate as expressed by @kepper in issue #1143.

PR #620, opened 2019-10-02
Add a skew attribute to att.coordinated

JRegimbal:
This pull request adds a new optional attribute to `att.coordinated` called skew which takes a value given in degrees. The number of degrees is equivalent to the angle of a vertical shear.

Its purpose is to address issues that arise in encoding handwritten manuscripts. Specifically discrepancies between the specified location in a `<facsimile>` and the calculated location by pitch and octave, line number, etc. The following is an example to illustrate this. With a staff's position defined only using the upper-left and lower-right points, the following can occur.
image
Since in the source the staff slants down, lower staff lines near the left and upper staff lines near the right significantly differ in position from the corresponding source staff lines. The problem continues for elements within the staff. Keeping pitch the same as the source results in the positioning like this:
image
Keeping location the same as the source results in the calculated pitch being incorrect. Shifting the neume to the right up to the correct location results in its pitch being E2, not D2.
With this rectangular bounding box for the staff, the location and pitch of elements cannot agree and both be correct encodings of the source.

The following is how the staff looks with the same upper-left and lower-right points with a skew applied:
image
The addition of one value for skew allows the location and pitch of elements to both be correct and in agreement.

More examples can be seen using [Neon](http://ddmal.music.mcgill.ca/Neon/). Skewing is available as one of the new experimental features.

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -690,7 +690,10 @@
         </datatype>
       </attDef>
       <attDef ident="skew" usage="opt">
-        <desc>Indicates the skew of the bounding box.</desc>
+        <desc>Indicates the skew of the bounding box. This represents the offset in degrees from
+          a typical 90-degree rectangle. By default the offset is 0-degrees. A positive angle
+          indicates that the upper-right y coordinate is above the upper-left y coordinate while
+          a negative angle indicates the opposite.</desc>
         <datatype>
           <rng:ref name="data.DEGREES"/>
         </datatype>
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -2191,7 +2191,7 @@
           notehead in a counter-clockwise fashion, while negative values produce clockwise
           rotation.</desc>
         <datatype>
-          <rng:ref name="data.DEGREES"/>
+          <rng:ref name="data.ROTATION"/>
         </datatype>
       </attDef>
       <attDef ident="head.shape" usage="opt">
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -227,7 +227,7 @@
       convenient typology of annotation suitable to the work in hand; e.g. annotation, gloss,
       citation, digression, preliminary, temporary, etc.</desc>
     <!-- Some attributes defined in att.controlEvent (att.timestamp.logical, att.timestamp.gestural,
-          att.staffIdent, and att.layerIdent) are provided here directly instead of making annot a 
+          att.staffIdent, and att.layerIdent) are provided here directly instead of making annot a
           member of att.controlEvent. -->
     <classes>
       <memberOf key="att.alignment"/>
@@ -687,6 +687,12 @@
         <desc>Indicates the lower-left corner x coordinate.</desc>
         <datatype>
           <rng:data type="nonNegativeInteger"/>
+        </datatype>
+      </attDef>
+      <attDef ident="skew" usage="opt">
+        <desc>Indicates the skew of the bounding box.</desc>
+        <datatype>
+          <rng:ref name="data.DEGREES"/>
         </datatype>
       </attDef>
     </attList>
@@ -1480,7 +1486,7 @@
         </datatype>
       </attDef>
       <!-- @llength not necessary:
-            @llength implies we know the direction of the vector which we 
+            @llength implies we know the direction of the vector which we
             can't know without establishing an end point, which in turn makes
             @llength redundant. -->
       <attDef ident="lsegs" usage="opt">
@@ -2185,7 +2191,7 @@
           notehead in a counter-clockwise fashion, while negative values produce clockwise
           rotation.</desc>
         <datatype>
-          <rng:ref name="data.ROTATION"/>
+          <rng:ref name="data.DEGREES"/>
         </datatype>
       </attDef>
       <attDef ident="head.shape" usage="opt">
@@ -5875,7 +5881,7 @@
       <memberOf key="att.bibl"/>
       <memberOf key="model.incipLike"/>
     </classes>
-    <!-- Can XSLT be used within content to "select" an incipit from the 
+    <!-- Can XSLT be used within content to "select" an incipit from the
         already-encoded MEI transcription in the music element?
         <rng:ref name="macro.XSLT"/> -->
     <content>
@@ -7089,7 +7095,7 @@
       <constraint>
         <!-- See http://vocab.org/frbr/core for more-precise entity-to-entity constraints -->
         <sch:rule
-          context="mei:relationList/mei:relation[parent::mei:work or parent::mei:expression or             
+          context="mei:relationList/mei:relation[parent::mei:work or parent::mei:expression or
           parent::mei:source or parent::mei:item]">
           <sch:assert
             test="matches(@rel, 'hasAbridgement') or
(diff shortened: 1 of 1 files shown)


rettinghaus (commented):
Looks good, but I'm wondering why the existing attributes `@lrx`, `@lry`, `@ulx`, and `@uly` do not imply a skew already?

Anyhow the description has to be adapted then:
> This attribute class records the position of a feature within a two-dimensional coordinate system.

JRegimbal:
Well currently I've seen every bounding box interpreted from `@lrx`, `@lry`, `@ulx` and `@uly` having what would be a skew of 0 degrees. It's just those same coordinates could have various other skews. For example this box
image
would have the same upper left and lower right points as this box
image
but all that differs is the skew.

rettinghaus:
I don’t know why, but I assumed it would always be a rectangle.

ahankinson:
Thanks for opening this @JRegimbal!

To me skew implies a parameter that transforms a rectangle (90 degree corners) into a parallelogram, while I think what you are describing is rotation?

I can see how the two could be used in your use case: a parallelogram rotated around a point. But perhaps they are two different properties?

It would also be good to clarify where the point of rotation is. I’m assuming that you have your rotation origin in the upper-left corner, but that assumption should be documented.

ahankinson:
Could you please provide some encoding  examples with the visual equivalents?

It would also be good to ensure this gets written into the guidelines with the examples

caitlin-hutnyk:
Here are two examples of skew. The staff is shown first with the image, then without so that the visual change is more obvious. The snippet for the zone element in the file's <surface> is shown, then the <sb> with which it is associated.

In this example the lines are all drawn at a slight angle:

Screen Shot 2019-10-09 at 12 11 09 PM
Screen Shot 2019-10-09 at 12 11 18 PM

```
<zone lrx="3288" lry="4260" skew="-0.694987" ulx="650" uly="4038" xml:id="m-dd2d6916-3722-4ef6-aaee-e540730ffcd8" />
```

```
<sb facs="m-dd2d6916-3722-4ef6-aaee-e540730ffcd8" n="8" xml:id="m-ec78e758-94e5-4c13-aa5b-fa6a0551019b">
```

With the skew added the lines rendered by Verovio match the lines in the image file almost perfectly.

In this example the page was not flat when the photos were taking, leading to the staff actually being curved. Still, having skew here is a huge improvement from just rendering the staff straight.
The staff with the skew added:

Screen Shot 2019-10-09 at 12 18 04 PM
Screen Shot 2019-10-09 at 12 18 19 PM

```
<zone lrx="3769" lry="1581" skew="1.138322" ulx="1152" uly="1439" xml:id="m-5635a0c4-edfa-4de9-bc84-e15983bf3a27" />
```

```
<sb facs="m-5635a0c4-edfa-4de9-bc84-e15983bf3a27" n="4" xml:id="m-f4410048-ce3d-423d-b24f-52f8a99d825e">
```

The way that the staff looks without skew, even after some editing attempts:

Screen Shot 2019-10-09 at 12 19 37 PM

All of these screenshots were taken in [Neon](http://ddmal.music.mcgill.ca/Neon/).

ahankinson:
Thanks @jacob-hutnyk . Could you also please post some fragments of the MEI so we can see what that looks like as well?

caitlin-hutnyk:
@ahankinson Yes, sorry about that. I've edited the snippets into my original comment.

ahankinson:
So the skew value is 1 degree off from an assumed 90 degree angle?

caitlin-hutnyk:
> So the skew value is 1 degree off from an assumed 90 degree angle?

Exactly. Skew = 0 or no skew value would imply the zone is a rectangle (90 degree angles). Positive skew would imply that the urx > ulx (the upper right point of the bounding box is above the upper left point), and negative skew means urx < ulx (the opposite).

ahankinson:
OK. That description should be added to the PR in the `<desc>` block, to aid in interpretation of the values.

kepper:
FWIW, in the past we agreed that MEI itself should offer only plain rectangles, and delegate all other shapes to SVG instead. There is also a corresponding `<remarks>` for this in the Guidelines: "Scalable Vector Graphics (SVG) markup may be used when allowed by the graphicLike model." This doens't mean we can't change our minds here, but at least I want to bring up our old argument…

ahankinson:
A parallelogram is still a rectangle. :)

JRegimbal:
@ahankinson Added to the `<desc>` block for skew. Let me know if anything is missing. For the guidelines, should that be written after changes to the schema are done or should we be doing that now? Sorry for the delay.

JRegimbal:
Could someone provide a bit more clarification on what we should do at this point? Thank you.

kepper:
@JRegimbal, apologies for the late reply – too many things, both MEI and beyond right now. Our priority right now is to get Guidelines and specs in sync. Basically, we should not move forward with the specs until the documentation has caught up. However, if you can provide corresponding documentation (in a separate PR to the guidelines repo), we can still merge this together. In the future, we will only accept and implement changes which will address both Guidelines and Schema. For now, it would be highly appreciated if you could not only write something about the skew thing, but also make sure this fits into the surrounding parts of the guidelines, and that these parts also reflect the current state of the schema. If you need assistance on that, please get in touch…

rettinghaus:
As this element comes from TEI I would plead to use `@rotate`: 
https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-zone.html

The intention here seems to skew the rendering, which shouldn't be declared within  the `zone` or `facsimile` elements. `staffDef` and `staffGrp` seem to be more appropriate for that, but that is to be discussed elsewhere.

JRegimbal:
Sorry for the late reply!

@kepper I've tried to add guidelines concerning `@skew` in this pull request to the guidelines: music-encoding/guidelines#142. I've tried to keep it general and show how adding a value for `@skew` would change a typical `zone` without it.

@rettinghaus I see your point, but I think that TEI's `@rotate` and this `@skew` are different transformations that do similar but different things. Applying a skew to a rectangle results in a parallelogram while rotating it results in a rectangle at an angle to the coordinate system.

rettinghaus:
`@rotate` is used when a zone's orientation differs from the "normal orientation of the parent surface". What you want to do is apply a skew to the rendering to match a somewhat rotated zone on the facsimile. So there is indeed a difference. 
A good example on how to use `@rotate` is given in here:
https://www.digitalmanuscripts.eu/wp-content/uploads/sites/6/2017/09/05-Digital-Facsimiles-EP.pdf
And as @kepper pointed out: 
> in the past we agreed that MEI itself should offer only plain rectangles

The skew approach leads to situations where lower points are higher than upper points (which makes no sense). So you always have to process the skew attribute to make sense out of the data. This makes it harder to handle than a rotate attribute. If you want to be more flexible with the form you want to describe, you should specify more points from the polygon and use `@points` (which is also not available in MEI yet). Again, look up the example given above. 

Trying to re-invent the wheel simply isn't always a good idea.

JRegimbal:
Ah alright I understood `@rotate` to actually be a rotation of region itself rather than how its contents are to be understood. I interpreted the TEI description to mean that the bounds described by the zone were either previously rotated or should then be rotated, and that processing would still be required.

Now knowing that I can understand why `@rotate` would likely work here.

lpugin:
> Now knowing that I can understand why @rotate would likely work here.

I agree that `@rotate` would work, but don't you feel that `@points` would be a better way to go?

JRegimbal:
I'm not sure. I feel like `@rotate` is actually quite similar to the result I wanted to get through `@skew`. `@points` provides a lot of flexibility but AFAIK its contents would still have the normal orientation of the surface without `@rotate`. And if an element was at such an angle that representing it with a rectangle wouldn't make sense, it would need `@rotate` if I'm understanding things right. So I guess it seems like `@rotate` would be a good starting point.

JRegimbal:
I've got changes reflecting the `@rotate` attribute in [this branch](https://github.com/JRegimbal/music-encoding/tree/rotate) and the documentation over [in this branch](https://github.com/JRegimbal/guidelines/tree/rotate). @rettinghaus Should I make separate PRs for these now?

rettinghaus:
I'm totally ok with those now. It's up to you to update this PR, or close it and start a new one. The guidelines need a separate PR anyway.

JRegimbal:
Okay! For the sake of neatness I'll open two new PRs and link both back to this one.

JRegimbal:
Closing in favor of the changes in PR 674.

PR #625, opened 2019-10-11
add divisio element

rettinghaus:
supersedes #618

### DIFF
--- a/source/modules/MEI.neumes.xml
+++ b/source/modules/MEI.neumes.xml
@@ -204,7 +204,7 @@
     </classes>
     <attList>
       <attDef ident="form" usage="opt">
-        <desc>Identifies the type of division. The <att>altsym</att>,
+        <desc>Identifies the form of division. The <att>altsym</att>,
           <att>glyph.name</att>, or <att>glyph.num</att> attributes may be used to specify the
           appropriate symbol.</desc>
         <datatype maxOccurs="unbounded">
--- a/source/modules/MEI.neumes.xml
+++ b/source/modules/MEI.neumes.xml
@@ -192,6 +192,35 @@
       <memberOf key="att.alignment"/>
     </classes>
   </classSpec>
+  <classSpec ident="att.divisio.log" module="MEI.neumes" type="atts">
+    <desc>Logical domain attributes.</desc>
+    <classes>
+      <memberOf key="att.alignment"/>
+      <memberOf key="att.layerIdent"/>
+      <memberOf key="att.partIdent"/>
+      <memberOf key="att.plist"/>
+      <memberOf key="att.staffIdent"/>
+      <memberOf key="att.targetEval"/>
+    </classes>
+    <attList>
+      <attDef ident="form" usage="opt">
+        <desc>Identifies the type of division. The <att>altsym</att>,
+          <att>glyph.name</att>, or <att>glyph.num</att> attributes may be used to specify the
+          appropriate symbol.</desc>
+        <datatype maxOccurs="unbounded">
+          <rng:data type="NMTOKEN"/>
+        </datatype>
+        <valList type="semi">
+          <valItem ident="minima"/>
+          <valItem ident="maior"/>
+          <valItem ident="maxima"/>
+          <valItem ident="finalis"/>
+          <valItem ident="virgula"/>
+          <valItem ident="caesura"/>
+        </valList>
+      </attDef>
+    </attList>
+  </classSpec>
   <classSpec ident="model.eventLike.neumes" module="MEI.neumes" type="model">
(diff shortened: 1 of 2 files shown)


source/modules/MEI.neumes.xml
```diff
@@ -192,6 +192,35 @@
...
+      <memberOf key="att.alignment"/>
+      <memberOf key="att.layerIdent"/>
+      <memberOf key="att.partIdent"/>
+      <memberOf key="att.plist"/>
+      <memberOf key="att.staffIdent"/>
+      <memberOf key="att.targetEval"/>
+    </classes>
+    <attList>
+      <attDef ident="form" usage="opt">
+        <desc>Identifies the type of division. The <att>altsym</att>,
```
lpugin: Shoudn't this be changed to "the form of the division" ?

ahankinson:
Uh.... why can't we continue the discussion on the previous PR?

rettinghaus:
We can. This is just a public proposal.

ahankinson:
I strongly disagree with this approach. This is not the correct way to go about asking for changes to the proposal. The change of the name to 'divisio' is *not* agreed, and presuming such is impolite and not conducive to community conversation.

Reviewers review proposals. They do not suggest their own because they don't like the original.

ahankinson:
Thank you.

craigsapp:
> I strongly disagree with this approach. This is not the correct way to go about asking for changes to the proposal. The change of the name to 'divisio' is not agreed, and presuming such is impolite and not conducive to community conversation.
> 
> Reviewers review proposals. They do not suggest their own because they don't like the original.

Your instructions are not complete: what is the proper thing to do if a reviewer wants a change, or to give a counter proposal?   In other words, your comments above are not overly constructive as well since you should also state what the proper thing to do as well as what is improper.  Surely the purpose of a reviewer is not to rubber stamp submissions; otherwise, why have reviewers?  Of course, the submitter should be allowed response to any proposed changes and the ability to veto changes.  But also the general procedure should be to discuss possible solutions before they are implemented, and I do not see much recently of that happening in the issues to this repository.  I am most definitely not doing that on the MEI-L mailing list since graphic music examples are not archived (and also are not displayed well in the emails that I get from the list) and I want to see a historic paper trail of the discussions towards implementation of things which github issue threads are better at handling than mailing lists.

Lilypond calls them divsio:
     http://lilypond.org/doc/v2.18/Documentation/music-glossary/divisio
which is likely where @rettinghaus  gets his proposed name for them.

So mostly it depends on if you want to use the English or Latin name for them.  The Mensural working group members have a high preference for Latin names for mensural elements.  Neume names are all in Latin:
     https://music-encoding.org/guidelines/v3/content/neumes.html
So it is not unreasonable to suggest "divisio" as the name of the element.  In the 1896 first edition of the liber usualis they are called "pauses" in French:
http://www.ccwatershed.org/media/pdfs/15/02/19/17-20-23_0.pdf

Screen Shot 2019-10-11 at 10 23 06 PM

A 1924 edition of the Liber usualis calls them "linea" in Latin, although a bit vague on what all of them would be called in general:
http://www.lakewoodsound.com/quilisma/pdf_files/liber_full/00_liber_usualis_pp_001_100.pdf

Screen Shot 2019-10-11 at 10 41 29 PM

rettinghaus:
Thank you, Craig. Fan fact: I haven't looked it up in Lilypond (this time). It's something I overheard when I was a student some hundred years ago … And as you noticed:

> Neume names are all in Latin

So it seemed strange to me to not use a latin name. "Pause" is still used today, so it could be confused with "rest".

craigsapp:
I found the lilypond graphic examples of the items in question:

http://lilypond.org/doc/v2.18/Documentation/notation/typesetting-gregorian-chant#divisiones

Screen Shot 2019-10-12 at 9 20 52 AM

Notice lilypond uses "maior", so even "major" is in Latin (which I think is also true for mensural module spellings for major).

Caesura has a CMN analog, which is graphically different (two slashes rather than one):
     https://music-encoding.org/guidelines/v4/elements/caesura.html
Should the same name be used for chant, but use a different graphic display depending on the context?

The virgula would be good to add if it is not already included.

ahankinson:
Hi @craigsapp , @rettinghaus 

I think the messages here demonstrate why it's a bad idea for this conversation to continue in a second pull request. The valuable feedback you are providing, @craigsapp, is not being communicated back to the original poster, and the conversation has forked. For those who do not follow GitHub as closely as others, this puts them at a disadvantage.

The original PR came from the Neumes IG, and I expect them to want to have oversight over any changes being proposed. 

Could you please move the conversation back there? Thanks.

PR #626, opened 2019-10-13
Improve multiRest

rettinghaus:
Working on the guidelines I noticed, that the description of the `@block` attribute on `multiRest` is a bit unclear (and contained an outdated reference). I tried to clear it up. 

Additionally (to the `@block` attribute) it seems to be missing the `@num.visible` attribute to control whether the number above is shown or not. 

Also it was missing a way to control the color (refers to #524).

### DIFF
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -1125,6 +1125,7 @@
       <memberOf key="att.staffLoc"/>
       <memberOf key="att.staffLoc.pitched"/>
       <memberOf key="att.typography"/>
+      <memberOf key="att.width"/>
     </classes>
     <attList>
       <attDef ident="block" usage="opt">
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -1121,6 +1121,7 @@
     <classes>
       <memberOf key="att.altSym"/>
       <memberOf key="att.extSym"/>
+      <memberOf key="att.numberPlacement"/>
       <memberOf key="att.staffLoc"/>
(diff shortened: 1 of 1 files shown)


rettinghaus:
I wonder if it also should be part of `att.width` to address the following: 
800px-Multirests_H-bars
800px-Multirests_narrow_H-bars

kepper (commented):
I think `@width` makes sense, but then it must be documented that makes sense only in combination with `@block`. Perhaps, we should even control that with Schematron. Also, I have to admit I've never heard the term "Kirchenpause" before, but I don't use Capella ;-)

kepper:
@rettinghaus: Shall we sneak in `@width`here? And is there any way to give a longer explanation of the word "Kirchenpause"? Would be sufficient to explain it in the Guidelines, I guess…

rettinghaus:
@kepper MEILER myself and I would like to see `@width` in there. 
And "Kirchenpause" is kinda explained in the Guidelines already. We could improve it, but I would wait, until schema and Guidelines are merged again. 
In the description perhaps it would be better to add "symbols": "that are combinations of longa, breve and semibreve rest symbols."
Should I remove f7ec383 here in favour of #719?

kepper:
Yes, we should remove duplication upfront. Revising the documentation when everything is in one place again seems reasonable in this case, we should just make sure we don't forget. Thanks!

rettinghaus:
dropped `@color` in this PR and added `@width` instead.

kepper (approved):
perfect, thanks 👍 . Ready for the ODD Friday today…

PR #630, opened 2019-10-26
remove custos from sb

### LINKED ISSUE #629: custos and sb
The guidelines say `sb` is 

> An empty formatting element that forces musical notation to begin on a new line.

Hence it may contain a `custos`. As costodes only appear at the end of lines, it does not make sense to put it in an empty (!) element, that starts (!) a new line.

rettinghaus:
fixes #629


kepper (approved):
good catch, thanks a lot for this :-)

PR #637, opened 2019-11-23
allow custodes with accidentals

### LINKED ISSUE #635: custos with accid
In CMN `custos` is a "substitute" for `note`, so it should allow `accid` as subelement.
Example from Mattheson's *Grosse General-Baß-Schule* (upper system):
example

rettinghaus:
Closes #635

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -768,6 +768,7 @@
   <classSpec ident="att.custos.log" module="MEI.shared" type="atts">
     <desc>Logical domain attributes.</desc>
     <classes>
+      <memberOf key="att.accidental"/>
       <memberOf key="att.pitched"/>
     </classes>
     <attList>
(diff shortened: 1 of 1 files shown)


kepper (approved):
Thanks for this, makes perfect sense. I was thinking if it's necessary to have `.ges`-attribues as well, but it does not, since a `<custos>` is always just a visual hint, and never performed. It might become important when the custos is the last sign on a page, and the following page is missing, but this seems so special that we should safely ignore it.

rettinghaus:
@kepper shouldn't `custos` also be part of `att.timestamp.logical` and `att.timestamp.gestural`?

kepper:
certainly not `att.timestamp.gestural` – like I said, a custos is never performed, and so it should not have a performed timestamp. The situation seems less clear for `att.timestamp.logical`. I would say that it's not needed – a custos is always the last thing in there, and that information should be sufficient to calculate its timestamp. I believe this is the main difference to other things like clefs or so, which may actually be placed anywhere inside a layer. A custos only makes sense when it comes last, or directly precedes a `<sb/>` / `<pb/>`. An encoded timestamp may contradict that, and cause problems. We all know that almost everything can be found in music sources, but without at least one example at hand that requires it, I'm inclined to not add timestamps to custos…

PR #640, opened 2019-12-20
fix midi value issue

### LINKED ISSUE #638: Values for MIDI volume
`@midi.volume` should conform to `data.MIDIVALUE_PERCENT`, but there is no such datatype.

rettinghaus:
closes #638

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -2599,11 +2599,20 @@
     </content>
   </macroSpec>
   <macroSpec ident="data.MIDIVALUE_PAN" module="MEI" type="dt">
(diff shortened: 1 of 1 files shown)


kepper (approved):
Basically, the change is correct and fine. I wonder, though, why / if we need to give percentages at all, as MIDI itself doesn't use percentages. Yes, they're easier to read, but will people use this beyond directly interfacing with MIDI files? We may want to revise this independent of the current PR…

craigsapp:
Yes, there is not great need for encoding MIDI volumes as percentages since the volume values in MIDI can only be 0 to 127 and will never change  (at least for the MIDI 1.0 standard).  So the only function of percentages would be to make it easier for people to read.

PR #647, opened 2020-02-22
lv ties don't need duration information

th-we:
`@dur` and `@dur.ges` are not even allowed on `<lv>`, so don't enforce them.

### DIFF
--- a/source/modules/MEI.cmn.xml
+++ b/source/modules/MEI.cmn.xml
@@ -1782,8 +1782,6 @@
         <sch:rule context="mei:lv">
           <sch:assert test="@startid or @tstamp or @tstamp.ges or @tstamp.real">Must have one of the
             attributes: startid, tstamp, tstamp.ges or tstamp.real.</sch:assert>
-          <sch:assert test="@dur or @dur.ges or @endid or @tstamp2">Must have one of the attributes:
-            dur, dur.ges, endid, or tstamp2.</sch:assert>
         </sch:rule>
       </constraint>
     </constraintSpec>


rettinghaus (approved):

kepper (approved):
good catch – thank you…

lpugin:
But don't we want to require `@endid` or `@tstamp2`?

th-we:
@lpugin I think lv ties don't (always) have a definitive logical endpoint, do they?  It might be up to interpretation where to stop the sound – or if it just sounds until it dies out. Therefore I wouldn't force the encoder to decide this.

bwbohl:
Yeah, good catch!!

I think Thomas is right, not to require a defined endpoint of laisser vibrer ties. Besides being up to the interpretation, it’s also up to the engraver how to render the tie…

In fact one could even argue, that it’s not really a tie, it just looks like one and maybe one should have a designated element to encode lv?

kepper:
Ahem, there is a specific element called `<lv>` for this, see https://music-encoding.org/guidelines/v4/elements/lv.html. The question is if the `<sch:assert>` in question should be dropped entirely, or if it should just be simplified. Right now, this assert enforces to use either `@dur` or `@dur.ges` (which aren't available on `<lv>`), or `@endid` or `@tstamp2` (which are available). I believe it might be useful to encode a `@tstamp2` in some situtions. I'm less sure about `@endid`, but let's put that aside. In any case, I think encoders should be safe to _not_ provide this kind of information at all, and rely on a standard rendition instead. So yes, dropping the complete assert seems the most adequate solution to me…

bwbohl:
> Ahem, there is a specific element called `<lv>` 

sorry… of course… thinking in two worlds at the same time makes blind and dumb sometimes :-~

+1 to drop it entirely

PR #650, opened 2020-03-01
drop default on system.leftline

### LINKED ISSUE #649: specify scope of system.leftline
The guidelines say about `@system.leftline` on `<scoreDef>`:
> Indicates whether the staves are joined at the left by a continuous line. The default value is "true".

To me it's not completely clear what should happen to one-staff systems. 
Normally single staves do not have a starting bar line, yet the default value is "true". If this attribute only applies to multi-staff systems, a single staff *never* could be drawn with a line at the beginning of the system. 
We should drop the un-MEI-ish default value and better say 

> Indicates whether the system starts with a continuous line connecting all staves.

rettinghaus:
fixes #649

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -3194,8 +3194,8 @@
     <desc>Attributes that capture system layout information.</desc>
     <attList>
       <attDef ident="system.leftline" usage="opt">
-        <desc>Indicates whether the staves are joined at the left by a continuous line. The default
-          value is "true". Do not confuse this with the heavy vertical line used as a grouping
+        <desc>Indicates whether the system starts with a continuous line connecting all staves,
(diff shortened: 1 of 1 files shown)


bwbohl (approved):

kepper:
There's nothing wrong with this PR, so no worries. Also, this is more a reminder to myself than a complaint. However, I think we agreed at some point that such PRs can be commented and prepared here at any time, but that ultimately we wanted to have a short discussion during our ODD Fridays / Thursdays to take final decisions, so that these decisions become a little more transparent and less individual. We've never fully established that, and may need to think about how to implement that properly (perhaps we could use labels like "needs decision", or Milestones, but we shouldn't create too much overhead). Again, nothing wrong here, we just need to think about how to do this properly…

PR #655, opened 2020-03-15
split and extend modes

### LINKED ISSUE #627: data.MODE is incomplete
`data.MODE` is incomplete, e.g. all the plagal modes are missing. 
I would like to add them. 
Question is if all the Gregorian modes should go into `data.MODE.gregorian` as part of the mensural module. @music-encoding/mensural-ig

rettinghaus:
This PR adds missing values in `data.MODE` and splits them up into `data.MODE.cmn`, `data.MODE.gregorian`, and `data.MODE.extended`, of which the last one is a semi-open list.
closes #627

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -2619,6 +2619,16 @@
   <macroSpec ident="data.MODE" module="MEI" type="dt">
     <desc>Modes.</desc>
     <content>
+      <rng:choice>
+        <rng:ref name="data.MODE.cmn"/>
+        <rng:ref name="data.MODE.gregorian"/>
+        <rng:ref name="data.MODE.extended"/>
+      </rng:choice>
+    </content>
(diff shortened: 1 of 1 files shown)


kepper (approved):
sounds very reasonable to me. I first thought whether there's a benefit in putting `data.MODE.cmn` into the `MEI.cmn` module (etc.), but I think it's not, as it would eventually make it impossible to encode a piece in phrygian mode (or any other…) in CMN. I also remember that there've been issues with nested data types in libmei, but IIRC, those have been resolved (@lpugin??). Unless someone complains last minute, seems ready to go to me… thanks, @rettinghaus!

lpugin:
>  I also remember that there've been issues with nested data types in libmei, but IIRC, those have been resolved 

Yes (until something else breaks it)

PR #634, opened 2019-11-16
Changes in the mei-Mensural.xml customization to make it compliant with the decisions made by the mensural-ig

martha-thomae:
Three main changes:

1. New attribute `@dur.quality` for `<note>` elements to indicate whether the note is perfect / imperfect / altered / etc. and, therefore, be comprehensive in the durational information of the note (given that in mensural notation, the note shape is not enough to indicate the length of a note). The new attribute can have the following six values: `perfecta`, `imperfecta`, `altera`, `maior`, `minor`, and `duplex`. The last three are used in Ars antiqua: `maior` and `minor` are used for semibreves and `duplex` is used for longae.

2. New values for the attribute `@dur` for the `<rest>` elements: two-breve rests ("2B") and three-breve rests ("3B"). 

3. New `<stem>` element to be used as children of the `<note>` element. Notes can have more than one stem. Therefore the introduction of the `<stem>` element as a children of `<note>`, allowing to associate multiple stem elements to a single note, each stem with its own form, direction, etc. Attributes of `<stem>`: `pos`, `length`, `form`, `dir`, `flag.pos`, `flag.form`, `x`, and `y`.

Two small changes:

4. Change in the description of the `@dur` attribute

5. Added a new stem-related attribute `@stem.form` to be encoded within the `<note>` element.

These are all (but one) of the changes that were discussed and approved by the mensural-ig in the 2018 meeting. The one change I did not include has to do with the encoding of the mensuration sign, because (based on the notes) further discussion is needed. 
**[The discussion can be seen in the attached file.]**
[MEI-Mensural IG - Approved MEC 2018 Pending Implementation.pdf](https://github.com/music-encoding/music-encoding/files/3853621/MEI-Mensural.IG.-.Approved.MEC.2018.Pending.Implementation.pdf)

PR #657, opened 2020-03-20
Support for encoding Italian Trecento Notation (Mensural Module)

martha-thomae:
The arrival of this pull request was stated in https://github.com/music-encoding/music-encoding/pull/634#issuecomment-585416648, and it is meant to replace **PR #537**
- Add `@divisio` attribute for encoding the different divisions of the breve used in Italian Trecento notation. 
- Add Schematron rule since this attribute substitutes the use of `@tempus` and `@prolatio` (in French notation).
- Add other values into `@sign` for encoding mensuration signs used in Italian Trecento notation.
- Moved attributes related to mensuration signs into the visual domain (these are: `@dot`, `@sign`, and `@slash`). 
Therefore, the attributes that encode the semantics of the mensuration (`@mdousmaior`, `@modusminor`, `@tempus`, `@prolatio`, and `@divisio`, which encode the divisions / relative_values of the notes) are separated from the attributes encoding the mensuration signs (`@dot`, `@sign`, and `@slash`). This is useful because some pieces don't have a mensuration sign, but there is an implied mensuration still. Additionally, the meaning of some mensuration signs varies through time and between geographical regions. And, for Italian Trecento mensuration, multiple symbols can represent the same mensuration (have the same semantics).

### DIFF
--- a/source/modules/MEI.mensural.xml
+++ b/source/modules/MEI.mensural.xml
@@ -134,6 +134,12 @@
           <rng:ref name="data.TEMPUS"/>
         </datatype>
       </attDef>
+      <attDef ident="divisio" usage="opt">
+        <desc>Describes the divisions of the breve in use in 14th-century Italy.</desc>
+        <datatype>
+          <rng:ref name="data.DIVISIO"/>
+        </datatype>
+      </attDef>
     </attList>
   </classSpec>
   <classSpec ident="att.note.anl.mensural" module="MEI.mensural" type="atts">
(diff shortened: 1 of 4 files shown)


martha-thomae:
## Examples:

While all the following examples are valid for the elements `<mensur>`, `<staffDef>`, and `<scoreDef>`, I am only providing examples for `<mensur>` since the changes requested in **PR #541** will eliminate the use of these attributes within `<staffDef>` and `<scoreDef>`.

### Diviso values allowed now:
divisio_values

### No simultaneous use of `@divisio` and the pair of attributes `@tempus` and `@prolatio`:
Both options allow to encode the division of the breve, but the former is used for Italian Trecento notation and the latter for French notation.
divisio_XOR_tempusORprolatio

### Sign values allowed now:
sign_values

kepper (approved):
this has been reviewed in the context of #634

PR #670, opened 2020-05-29
Mensural source

martha-thomae:
This PR replaces https://github.com/music-encoding/music-encoding/pull/634. It includes the exact same changes, but the changes are done in the source files (rather than in the customization file as in https://github.com/music-encoding/music-encoding/pull/634).


kepper (approved):
this has been properly reviewed in #634 already

PR #671, opened 2020-05-30
fix removed attribute

rettinghaus:
`@ref` has been replaced by `@altsym`, fixing this in the remarks.

What I'm not sure about is the meaning of `@startid`. How is that supposed to be used, if `<symbol>` cant't be used as a free controlEvent?

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -7985,7 +7985,7 @@
         coordinate terms using the <att>x</att> and <att>y</att> attributes or relative to another
         element using the <att>startid</att> attribute. Attributes in the att.visualOffset class may
         be used to record horizontal, vertical, or time offsets from the absolute coordinates or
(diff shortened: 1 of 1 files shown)


pe-ro:
The `<symbol>` element occurs in 2 different contexts --

1. within the 'music' context where it marks a graphical symbol that can't be/isn't drawn based on the markup; and
2. within the symbol definition context (i.e. inside `<symbolDef>`) where a custom symbol can be defined

The `@startid` attribute is only used in the symbol definition context.  It's necessary when attributes from the classes att.visualOffset.ho, att.visualOffset.to, and/or att.visualOffset.vo are used to define a symbol's position relative to another symbol.

I'll concede that the name 'startid' is somewhat confusing.  Perhaps 'relativeTo' would be a better name?  I'll also admit that the Schematron rules associated with `<symbol>` should be rewritten in order to better enforce these expectations.

bwbohl (approved):

PR #678, opened 2020-06-20
BREAKING(modules): More clarity in typography and measurement units

th-we:
Motivation: Unclarities and misleading naming.

1.  The naming `data.MEASUREMENTABS` and `data.MEASUREMENTREL` suggested relative vs. absolute units, however they are actually unsigned vs. signed units.
2. In [`data.TEXTRENDITIONPAR`](https://music-encoding.org/guidelines/v4/data-types/data.textrenditionpar.html) (used by `@rend`), it was not clear what the implied units were. With this pull request, the (freshly renamed) `data.MEASUREMENT(UN)SIGNED` are used and the units `ch`, `em` and `ex` are introduced. This allows specifying the intended unit and makes `vu` the implied unit if none is specified.
3. Percentages for letter spacing are odd. The base value for letter spacing is 0 and any percentage thereof is still 0.

The reason why I made letterspacing and lineheight attributes of their own is because this way, I don't have to duplicate the existing data type patterns (and maybe their documentation) multiple times. Or can this be solved in a better way? In any case, `letterspacing="0.25" lineheight="120%"` is more processing friendly than `rend="letter-spacing(0.25) line-height(120%)"`.


th-we:
BTW, is it possible to define the signed types by referencing the unsigned types rather than duplicating the patterns?

rettinghaus (commented):
The renaming makes sense and is great. I'm less confident about extracting the font attributes.  And why adding the new `data.MEASUREMENTFONT(UN)SIGNED` data types? Wouldn't adding `em` etc. to the regex patterns in the existing ones be sufficient?

th-we:
`em` etc. only makes sense in the text context, I think. In HTML, these units are usable anywhere, but in HTML, in any context, there is a defined text size and font family. Defining a page size in terms of `em`, for example, seems odd to me.

I'm not per se against keeping line-height and letter-spacing as part of `@rend`, but I don't see a striking argument against making them attributes either. If there was a way to re-use the data types with the `@rend` approach, that would be great.

rettinghaus:
`em` does not only make sense in the text context, as SMuFL glyph sizes are all based on relations to text sizes: https://w3c.github.io/smufl/gitbook/specification/scoring-metrics-glyph-registration.html

kepper (commented):
I agree with @rettinghaus – the name changes are sound, while the introduction of `@letterspacing` and `@lineheight` seems less striking. In general, I think MEI needs a well-thought mechanism for styling, whcih is consistent across the schema. Both attributes could be a part of that, but I'm slightly concerned that we might jump the gun here, and run into compatibility issues later on. As it's already possible to encode this with `//rend/@rend` (via [data.TEXTRENDITIONPAR)](https://music-encoding.org/guidelines/v4/data-types/data.textrenditionpar.html), this would be syntactic sugar only, which I'm inclined to refuse for now. Other opinions?

kepper:
We should take a final decision on this today. Personally, I'm in favor of including https://github.com/music-encoding/music-encoding/pull/678/commits/42cc58a187d757e305d599890025b543f476049f, but dropping https://github.com/music-encoding/music-encoding/pull/678/commits/00ac37289f9ab532ccc65541cf809cf3662b8a36. Other thoughts?

kepper:
Dear @th-we, we discussed the situation during the ODD Thursday on Oct 29, and we agreed that we'd like to include https://github.com/music-encoding/music-encoding/commit/42cc58a187d757e305d599890025b543f476049f, but skip https://github.com/music-encoding/music-encoding/commit/00ac37289f9ab532ccc65541cf809cf3662b8a36 until we have a more complete picture for styling in / with MEI. Could you please remove the second commit from this PR? This is not to say that we don't like the idea in general, we just don't want to rush into a specific solution with two values of `@rend`, without considering all the others at the same time… Thanks, and best regards :-)

kepper (dismissed):
everything sounds reasonable, according to @th-we, @fujinaga, @musicEnfanthen, @pe-ro, @kepper…

source/modules/MEI.shared.xml
```diff
@@ -3558,6 +3558,22 @@
           <rng:ref name="data.FONTWEIGHT"/>
         </datatype>
       </attDef>
+      <attDef ident="letterspacing" usage="opt">
+        <desc>Indicates letter spacing (aka tracking) in analogy to the CSS letter-spacing
```
```suggestion
        <desc xml:lang="en">Indicates letter spacing (aka tracking) in analogy to the CSS letter-spacing
```

source/modules/MEI.shared.xml
```diff
@@ -3558,6 +3558,22 @@
...
+          property.</desc>
+        <datatype>
+          <rng:ref name="data.MEASUREMENTTYPOGRAPHYSIGNED"/>
+        </datatype>
+      </attDef>
+      <attDef ident="lineheight" usage="opt">
+        <desc>Indicates line height in analogy to the CSS line-height property.</desc>
```
```suggestion
        <desc xml:lang="en">Indicates line height in analogy to the CSS line-height property.</desc>
```

source/modules/MEI.xml
```diff
@@ -1862,28 +1862,67 @@
...
+      data.MEASUREMENTUNSIGNED, in which only positive values are allowed, both positive and negative
       values are permitted.</desc>
     <content>
       <rng:data type="token">
         <rng:param name="pattern">(\+|-)?\d+(\.\d+)?(cm|mm|in|pt|pc|px|vu)?</rng:param>
       </rng:data>
     </content>
   </macroSpec>
+  <macroSpec ident="data.MEASUREMENTFONTUNSIGNED" module="MEI" type="dt">
+    <desc>Measurement expressed relative to properties of the current font, in analogy to the
```
```suggestion
    <desc xml:lang="en">Measurement expressed relative to properties of the current font, in analogy to the
```

source/modules/MEI.xml
```diff
@@ -1862,28 +1862,67 @@
...
+      respective CSS length units. Unlike data.MEASUREMENTFONTUNSIGNED, only positive values are
+      allowed.</desc>
+    <content>
+      <rng:data type="token">
+        <rng:param name="pattern">\d+(\.\d+)?(ch|em|ex)?</rng:param>
+      </rng:data>
+    </content>
+  </macroSpec>
+  <macroSpec ident="data.MEASUREMENTFONTSIGNED" module="MEI" type="dt">
+    <desc>Measurement expressed relative to properties of the current font, in analogy to the
```
```suggestion
    <desc xml:lang="en">Measurement expressed relative to properties of the current font, in analogy to the
```

source/modules/MEI.xml
```diff
@@ -1862,28 +1862,67 @@
...
+      respective CSS length units. Unlike data.MEASUREMENTFONTUNSIGNED, both positive and negative values
+      are allowed.</desc>
+    <content>
+      <rng:data type="token">
+        <rng:param name="pattern">(\+|-)?\d+(\.\d+)?(ch|em|ex)?</rng:param>
+      </rng:data>
+    </content>
+  </macroSpec>
+  <macroSpec ident="data.MEASUREMENTTYPOGRAPHYUNSIGNED" module="MEI" type="dt">
+    <desc>Measurements used for typographical features. Unlike data.MEASUREMENTTYPOGRAPHYSIGNED, only
```
```suggestion
    <desc xml:lang="en">Measurements used for typographical features. Unlike data.MEASUREMENTTYPOGRAPHYSIGNED, only
```

source/modules/MEI.xml
```diff
@@ -1862,28 +1862,67 @@
...
+      positive values are allowed.</desc>
+    <content>
+      <rng:choice>
+        <rng:ref name="data.MEASUREMENTFONTUNSIGNED"/>
+        <rng:ref name="data.MEASUREMENTUNSIGNED"/>
+      </rng:choice>
+    </content>
+  </macroSpec>
+  <macroSpec ident="data.MEASUREMENTTYPOGRAPHYSIGNED" module="MEI" type="dt">
+    <desc>Measurements used for typographical features. Unlike data.MEASUREMENTTYPOGRAPHYSIGNED, both
```
```suggestion
    <desc xml:lang="en">Measurements used for typographical features. Unlike data.MEASUREMENTTYPOGRAPHYSIGNED, both
```

source/modules/MEI.xml
```diff
@@ -1862,28 +1862,67 @@
...
     <content>
       <rng:data type="token">
         <rng:param name="pattern">(\+)?\d+(\.\d+)?(cm|mm|in|pt|pc|px|vu)?</rng:param>
       </rng:data>
     </content>
   </macroSpec>
-  <macroSpec ident="data.MEASUREMENTREL" module="MEI" type="dt">
-    <desc xml:lang="en">Measurement expressed in real-world (<abbr>e.g.</abbr>, centimeters, millimeters, inches, points,
+  <macroSpec ident="data.MEASUREMENTSIGNED" module="MEI" type="dt">
+    <desc xml:lang="en">Measurement expressed in real-world (e.g., centimeters, millimeters, inches, points,
```
rettinghaus:
```suggestion
    <desc xml:lang="en">Measurement expressed in real-world (<abbr>e.g.</abbr>, centimeters, millimeters, inches, points,
```

kepper:
thanks @rettinghaus and @musicEnfanthen for fixing this so quickly 👍

source/modules/MEI.xml
```diff
@@ -1862,28 +1862,67 @@
       </rng:data>
     </content>
   </macroSpec>
-  <macroSpec ident="data.MEASUREMENTABS" module="MEI" type="dt">
-    <desc xml:lang="en">Measurement expressed in real-world (<abbr>e.g.</abbr>, centimeters, millimeters, inches, points,
+  <macroSpec ident="data.MEASUREMENTUNSIGNED" module="MEI" type="dt">
+    <desc xml:lang="en">Measurement expressed in real-world (e.g., centimeters, millimeters, inches, points,
```
```suggestion
    <desc xml:lang="en">Measurement expressed in real-world (<abbr>e.g.</abbr>, centimeters, millimeters, inches, points,
```

musicEnfanthen (approved):

kepper (approved):
we're all happy now in Berlin…

musicEnfanthen:
Thanks everyone!

PR #681, opened 2020-06-25
Move rotate attribute to att.coordinated

JRegimbal:
This makes `@rotate` apply to all elements with the `att.coordinated` class and not just the `zone` element. This expands it to the `surface` and `symbolDef` elements.

See #674 for more information on `@rotate`.

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -227,7 +227,7 @@
       convenient typology of annotation suitable to the work in hand; e.g. annotation, gloss,
       citation, digression, preliminary, temporary, etc.</desc>
     <!-- Some attributes defined in att.controlEvent (att.timestamp.logical, att.timestamp.gestural,
-          att.staffIdent, and att.layerIdent) are provided here directly instead of making annot a 
+          att.staffIdent, and att.layerIdent) are provided here directly instead of making annot a
           member of att.controlEvent. -->
     <classes>
       <memberOf key="att.alignment"/>
@@ -688,6 +688,19 @@
         <datatype>
           <rng:data type="nonNegativeInteger"/>
         </datatype>
+      </attDef>
+      <attDef ident="rotate">
+        <desc>
+          Indicates the amount by which the contents of this zone have been rotated clockwise, with respect
+          to the normal orientation of the parent surface. The orientation is expressed in arc degrees.
+        </desc>
+        <datatype minOccurs="1" maxOccurs="1">
+          <rng:ref name="data.DEGREES"/>
+        </datatype>
+        <defaultVal>0</defaultVal>
+        <remarks>
+          <p>This attribute is based on the TEI attribute of the same name.</p>
+        </remarks>
       </attDef>
     </attList>
   </classSpec>
@@ -1480,7 +1493,7 @@
         </datatype>
       </attDef>
       <!-- @llength not necessary:
-            @llength implies we know the direction of the vector which we 
+            @llength implies we know the direction of the vector which we
             can't know without establishing an end point, which in turn makes
             @llength redundant. -->
       <attDef ident="lsegs" usage="opt">
@@ -5865,7 +5878,7 @@
       <memberOf key="att.bibl"/>
       <memberOf key="model.incipLike"/>
     </classes>
-    <!-- Can XSLT be used within content to "select" an incipit from the 
+    <!-- Can XSLT be used within content to "select" an incipit from the
         already-encoded MEI transcription in the music element?
         <rng:ref name="macro.XSLT"/> -->
     <content>
@@ -7079,7 +7092,7 @@
       <constraint>
         <!-- See http://vocab.org/frbr/core for more-precise entity-to-entity constraints -->
         <sch:rule
-          context="mei:relationList/mei:relation[parent::mei:work or parent::mei:expression or             
+          context="mei:relationList/mei:relation[parent::mei:work or parent::mei:expression or
           parent::mei:source or parent::mei:item]">
           <sch:assert
             test="matches(@rel, 'hasAbridgement') or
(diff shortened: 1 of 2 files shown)


kepper (changes requested):
I think in general it's better to move this up to att.coordinated. I'm not perfectly sure if I'd prefer to see it in a separate sub-class or not. Normally, I would, but considering where this occurs, I think it's fine to put placement and rotation in one class. What I'm more concerned with is the description. @JRegimbal's clarification in a18fee061b0e5c69f4b8410aca68cfbc60a9b1fd helped, but didn't fully fix it in my perception. Ideally, we should have an image somewhere in the documentation, both in the Guidelines and here as a remark (compare with https://music-encoding.org/guidelines/v4/data-types/data.accidental.written.html). It should be absolutely clear what the center of the rotation is, and if the two points encoded by ulx/uly and lrx/lry are already rotated (so that width and height of that rect can't be derived directly anymore, but one could add att.dimensions to compensate). I'd really like to see more explicit statements about all this before putting it in, even though it's certainly going in the right direction :-)

JRegimbal:
My impression of how `@rotate` would apply to the other two elements is that, like in `zone`, the bounding box defined by ulx/uly/lrx/lry would not be impacted by rotation but would inform the user or software that the content should be interpreted as rotated.

So for something like `<surface rotation="45">...</surface>` that would mean that the entire surface described (including any zones in it) should by default be interpreted as being captured at a 45° angle. Similarly `<symbolDef rotation="15">...</symbolDef>` would mean that the symbol being defined should be interpreted as rotated 15° from whatever it's normal orientation would be. But in both cases the x-y plane would be unaffected.

Does this seem like a sensible interpretation to you @lpugin and @kepper? If so I can add another commit making this explicit.

source/modules/MEI.shared.xml
```diff
@@ -689,6 +689,19 @@
           <rng:data type="nonNegativeInteger"/>
         </datatype>
       </attDef>
+      <attDef ident="rotate">
+        <desc>
+          Indicates the amount by which the contents of this zone have been rotated clockwise, with respect
```
bwbohl: The description is still referring to `mei:zone` if `@rotate` is to be moved to `att.rotated`this should be generalised

source/modules/MEI.shared.xml
```diff
@@ -7079,7 +7092,7 @@
       <constraint>
         <!-- See http://vocab.org/frbr/core for more-precise entity-to-entity constraints -->
         <sch:rule
-          context="mei:relationList/mei:relation[parent::mei:work or parent::mei:expression or             
+          context="mei:relationList/mei:relation[parent::mei:work or parent::mei:expression or
```
bwbohl, line 7096: It's nice that this PR fixes those trailing whitespaces but for the sake of reviewing it would be better not to include such changes ;-)

bwbohl (changes requested):
Unfortunately I missed the addition of `@rotate` in the first place… At the moment I'm in eager need of some real world examples for `@rotate`. As of the name  (picking up the question is it the zone or the content that are being rotated) why not pick @`rotated` which makes it clear in the current facsimile context. (then why limit it to theses facsimile elements? What about born digital documents?)
Concerning `symbolDef` I'd see more sense in adding`@rotate(d)` to `symbol`as otherwise one would need multiple `symbolDef` just to encode a normal and rotated version of a symbol.

source/modules/MEI.shared.xml
JRegimbal: Oh thanks nice I missed that!

source/modules/MEI.shared.xml
JRegimbal: Ah sorry about that my editor is very aggressive with trailing whitespace.

lpugin:
Anything wrong with this PR? (This is currently blocking the update of the Schema within Verovio)

bwbohl:
need a small fix in the documentation I'd say, sorry for the delay: https://github.com/music-encoding/guidelines/pull/178

lpugin:
How does it look now?

kepper (approved):
Apologies for the long wait – this is definitely ready for merge, just as https://github.com/music-encoding/guidelines/pull/178# (to be done tomorrow at the ODD Friday…)

rettinghaus:
Could we revert this particular change? (Yes, I know I'm late to the party.) I don't see the sense in having `@rotate` on `surface`. The encoding always follow the main text direction. You would not encode music/text upside down, just because you're too lazy to rotate the folio in front of you.  
Same for `symbolDef`. Why would you define a symbol in a rotated state? If one symbol seems to be rotated on the page you would still use `zone`, and that seems to be sufficient.

ahankinson:
The intention was to be able to capture the offset of the surface from the main text direction. If an image was rotated by 2° during digitization there was no way to capture this such that the notation and the surface could be aligned. You could capture width and height.

It's explained here: https://github.com/music-encoding/music-encoding/pull/620

ahankinson:
In fact, looking back at that you agreed with this, and actually suggested `@rotate`?

rettinghaus:
Then a zone of the size of the surface is rotated. (Or `graphic`, not the `surface` itself). 
I agreed with #674. 

Connected to this is actually #971.

ahankinson:
`graphic` points to an image file, which cannot be rotated by definition, since all image files assume 90° corners. 

`zone` is an area of the surface. We do not allow nested zones, and I agree with the reasons why. It complicates the parsing for very little benefit.

`surface` is the abstract space onto which a `graphic` is projected, and allows the co-ordinates of the `zones` to be addressed. It allows for a global rotation such that all zones on that surface may be adjusted per this co-ordinate space, if the user desires. 

If it helps you can think of a `surface` as a specialized `zone` -- naming it differently means we can enforce a single layer of nesting, since otherwise we would be forced to always allow `zone` within `zone` to arbitrary depths.

PR #696, opened 2020-07-27
Allow ossia staves below the regular staves

th-we:
We should allow `<staff>` and `<oStaff>` to be in any order because ossia staves can appear below their "parent" staves.

grafik


kepper (approved):
that was clearly a bug – thanks for spotting and fixing it 👍

PR #698, opened 2020-07-28
Add schematron rules for dur.quality values.

martha-thomae:
- Not allowing `"duplex"` quality to be used in any other note than a `"longa"`.
- Not allowing `"maior"` and `"minor"` quality values to be used in any other note than a `"semibrevis"`.

This would be what I expect with this PR:
dur_quality

### DIFF
--- a/source/modules/MEI.mensural.xml
+++ b/source/modules/MEI.mensural.xml
@@ -171,6 +171,24 @@
         </datatype>
       </attDef>
     </attList>
+    <constraintSpec ident="check_duplex_quality" scheme="isoschematron">
+      <constraint>
+        <sch:rule context="mei:note[@dur.quality='duplex']">
+          <sch:assert test="@dur='longa'">
+            Duplex quality can only be used with longas (in Ars antiqua).
+          </sch:assert>
+        </sch:rule>
+      </constraint>
+    </constraintSpec>
+    <constraintSpec ident="check_maiorminor_quality" scheme="isoschematron">
+      <constraint>
+        <sch:rule context="mei:note[@dur.quality='maior' or @dur.quality='minor']">
+          <sch:assert test="@dur='semibrevis'">
+            Maior / minor quality can only be used with semibreves (in Ars antiqua).
+          </sch:assert>
+        </sch:rule>
+      </constraint>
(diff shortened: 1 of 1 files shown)


kepper (commented):
The schematron should do what you expect. On second thought, I think its quite clever to use the notes with `@dur.quality` as context, as this seems more specific than `@dur`. So, in general, this is an approval. However, I wonder if we need the reference to Ars antiqua – is this different in Ars nova? If so, would the proposed solution still support it? If not, why mention it specifically here?

kepper (changes requested):
well, I guess my last comment should have been a request for changes instead ;-) A clarification on the Ars antiqua reference would be much appreciated…

martha-thomae:
> However, I wonder if we need the reference to Ars antiqua – is this different in Ars nova? If so, would the proposed solution still support it? If not, why mention it specifically here?

Yes, in Ars antiqua you have these 'maior' and 'minor' values for semibreves and the 'duplex' value for longas, but in Ars nova that doesn't happen anymore (the semibreves become 'altera' or 'regular', and the duplex longa is replaced by the maxima). 

The problem is that I cannot check whether the piece is in Ars antiqua or Ars nova because this would be something that you encode on `@notationSubtype`, which can have any string value. So it is something that I cannot add in the constraint of the schematron rule (I just added it as a text), but at least I can help people avoid using these quality values (maior, minor, and duplex) in other note values (e.g., brevis, minima, semiminima, etc.)

@kepper: Should I remove the 'Ars antiqua' mention? I think it makes it more specific, even though I cannot check for that.

bwbohl:
@martha-thomae could you please try building the schema, and see whether the schematron constraint makes it into the RNG because this one (as you pointed out) is defined within the classSpec, whereas all others aren't

kepper (commented):
checked everything if it compiles properly, as requested by @bwbohl – it does 👍 Ready to be merged tomorrow, I'd say :-)

kepper (approved):

martha-thomae:
> checked everything if it compiles properly, as requested by @bwbohl – it does 👍 Ready to be merged tomorrow, I'd say :-)

It does! I checked it last week :)

PR #703, opened 2020-08-14
Allow 100% with trailing fractional zeroes

th-we:
Values like "100.00%" were not accepted, but values like "90.00%" were. So let's make it consistent. Verovio sometimes generates these values when converting from MusicXML.

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -3000,7 +3000,7 @@
     <desc>Positive decimal number between 0 and 100, followed by a percent sign "%".</desc>
     <content>
       <rng:data type="token">
-        <rng:param name="pattern">(([0-9]|[1-9][0-9])(\.[0-9]+)?|100)%</rng:param>
+        <rng:param name="pattern">(([0-9]|[1-9][0-9])(\.[0-9]+)?|100(\.0+)?)%</rng:param>
       </rng:data>
     </content>
   </macroSpec>
@@ -3008,7 +3008,7 @@
     <desc>Positive decimal number between -100 and 100, followed by a percent sign "%".</desc>
     <content>
       <rng:data type="token">
-        <rng:param name="pattern">(\+|-)?(([0-9]|[1-9][0-9])(\.[0-9]+)?|100)%</rng:param>
+        <rng:param name="pattern">(\+|-)?(([0-9]|[1-9][0-9])(\.[0-9]+)?|100(\.0+)?)%</rng:param>
       </rng:data>
     </content>
   </macroSpec>


kepper (commented):
I'm inclined to leave this as is. Technically, those numbers are the same, and so I do see the argument. However, I think it's a bad habit of software to generate unnecessary trailing zeros – I frequently run into things like `@tstamp="1.00000"`, which unnecessarily bloat files and makes them harder to read than necessary. If we'd accept this change, we would legitimate such software behavior, and put the burden of cleaning up to the ones responsible for data maintenance. I would like to hear more arguments (@rettinghaus, @lpugin ?), but at this point, I'm inclined to leave it as is.

lpugin:
I think if we don't want `100.00%` then `90.00%` should definitely not be valid. (However, I don't think deciding how numbers should be formatted is the role of the schema.)

kepper:
I can see some value in values (pun intended) like `100.0%`, `90.0%` or `tstamp="1.0`, as they indicate precision, but beyond the first zero, there's little to gain with this argument. I'm happy to follow if others prefer more open values – it's what we have right now – and we can even allow trailinzg zeros on the 100, if that's what the community needs. I just wanted to bring up a different perspective, because that's what I typically see on that end ;-) Let's chat tomorrow about this – either solution is fine with me :-)

lpugin:
I agree there is no gain, but at the same time I don't see the need to 'clean' them since in terms of data `1.0` and `1.00` are the same.

pe-ro:
Since the decimal portion is optional, allowing it after a value of "100" makes little difference.  But, it's an easy change to make and affords a high degree of consistency.  I agree with @lpugin that the schema should not enforce rules with regard to precision.

rettinghaus:
I think it would be an improvement in consistency if there were values above `100%`  allowed. 
Having `100.00%` personally leads me to the assumption, that `100.56%` is possible too. 
So I see no improvement in encoding quality here (especially not for manual encoding).

lpugin:
> Having `100.00%` personally leads me to the assumption, that `100.56%` is possible too.

Then you will see that `100.56%` is actually an invalid value, which is what the schema stands for.

You can turn it the other way around. It leads you to the assumption that 99.99% is possible too, which is good. But reading guidelines is probably better than relying on assumptions ;-)

kepper:
And, eventually, we're talking about a datatype called `data.PERCENT.LIMITED`, which has a `<desc>`ription as follows: 

> Positive decimal number between 0 and 100, followed by a percent sign "%".

We should not allow it to go over 100% ;-)

kepper (approved):
discussions at the ODD Friday are in favor of adding this.

Issue #140, opened 2015-05-28
Make att.typography and rend/@rend CSS compatible


ahankinson:
_From [zupf...@googlemail.com](https://code.google.com/u/113984963995030839842/) on December 20, 2012 06:20:00_

From the description in the specs, I get the impression that the current att.typography set of attributes seems to be modeled after mup's concept of fonts. This is not a concept that is shared by a lot of other applications. Therefore, I suggest to move towards CSS, like we did for color.

The most important issue I currently see is @fontname. While the attribute name makes me think I can use something like "Arial" here, this is already covered by @fontfam. Instead, "Mup-acceptable values: rom, ital, bold, boldital" are suggested. But "ital" is already covered by @fontstyle and "bold" by @fontweight. "rom" is basically equivalent to "normal" (the specs even say 'normal' = 'rom' in Mup), which is also covered by @-fontstyle.

@-rend also has some issues. It makes non-mutually-exclusive features like small caps and underline mutually exclusive.

I'd therefore suggest consequently following CSS, using the properties
- font-family
- font-style
- font-variant
- font-size
- font-weight
- text-decoration

We could continue using the existing attribute names
- fontfam
- fontsize
- fontstyle
- fontweight

and add names for the missing properties, or we could use the CSS property names as attribute names. What we should do in any case is use the same data as CSS.

CSS to my knowledge however does not cover all aspects that @-rend covers, namely:
- circle/box (the CSS border property doesn't seem approprate here because it's not abstract enough)
- sub/sup (requires two properties: font-size and vertical-align)
- dblunderline

This needs to be resolved somehow. (BTW: What is the meaning of the values "quote", "bslash" and "fslash"? Those seem to be names of single characters and no rendering information.)

_Original issue: http://code.google.com/p/music-encoding/issues/detail?id=140_

ahankinson:
_From [kep...@edirom.de](https://code.google.com/u/113787537940489926112/) on December 20, 2012 05:14:28_

[...]

PR #704, opened 2020-08-17
Introduce <smufl> element for multi-character smufl strings

th-we:
As per [a suggestion](https://github.com/music-encoding/guidelines/pull/145#issuecomment-591927828)  by @kepper, this introduces a way of encoding SMuFL strings like:

```xml
<!-- "fzp" -->
<dynam><smufl>&#xE522;&#xE525;&#xE520;</smufl></dynam>
```

I split [att.typography](https://music-encoding.org/guidelines/v4/attribute-classes/att.typography.html) for this as `@fontstyle` and `@fontweight` are not applicable to SMuFL text (there is no non-italic variant of a forte, no oblique variant of a time signature numeral, and no bold variant of quarter note).

Concerning `@fontfam` and `@fontname`, I do not understand the difference between the two and we have to either deprecate one or document clearly what the difference is.  If I interpret [this](https://github.com/music-encoding/music-encoding/issues/140#issue-82036031) correctly then the origin seems to be historic and `@fontname` was originally intended for mup-specific purposes that are now covered by `@fontweight` and `@fontstyle`. @pe-ro already [stated in 2012](https://github.com/music-encoding/music-encoding/issues/140#issuecomment-106500294) that `@fontname` is "not useful anymore".

I can't find much other discussion about those attributes. (Is that because there is none or is that because the GitHub issue search possibly ignores code blocks, things with `@` or in backticks?)

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -3436,6 +3436,13 @@
   </classSpec>
   <classSpec ident="att.typography" module="MEI.shared" type="atts">
     <desc>Typographical attributes.</desc>
+    <classes>
+      <memberOf key="att.typography.font"/>
+      <memberOf key="att.typography.style"/>
+    </classes>
+  </classSpec>
+  <classSpec ident="att.typography.font" module="MEI.shared" type="atts">
+    <desc>Attributes for fonts and font sizes</desc>
     <attList>
       <attDef ident="fontfam" usage="opt">
         <desc>Contains the name of a font-family.</desc>
@@ -3449,6 +3456,11 @@
           <rng:ref name="data.FONTNAME"/>
         </datatype>
       </attDef>
+    </attList>
+  </classSpec>
+  <classSpec ident="att.typography.style" module="MEI.shared" type="atts">
+    <desc>Attributes for font style and weight</desc>
+    <attList>
       <attDef ident="fontsize" usage="opt">
         <desc>Indicates the size of a font expressed in printers' points, i.e., 1/72nd of an inch,
           relative terms, e.g., "small", "larger", etc., or percentage values relative to "normal"
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -7217,6 +7217,44 @@
         be used instead of the <gi scheme="MEI">rend</gi> element.</p>
     </remarks>
   </elementSpec>
+  <elementSpec ident="smufl" module="MEI.shared">
+    <desc>Contains music text characters in SMuFL encoding.</desc>
+    <classes>
+      <memberOf key="att.color"/>
+      <memberOf key="att.common"/>
+      <memberOf key="att.horizontalAlign"/>
+      <memberOf key="att.typography.font"/>
+      <memberOf key="att.verticalAlign"/>
+      <memberOf key="model.rendLike"/>
+    </classes>
+    <content>
+      <rng:text/>
+    </content>
+    <constraintSpec ident="check_smufl" scheme="isoschematron">
+      <constraint>
+        <sch:rule context="mei:smufl">
+          <sch:assert
+            test="matches(., '[&#x266D;-&#x266F;&#x1D100;-&#x1D126;&#x1D129;-&#x1D1BE;&#x1D1C1;-&#x1D1D4;&#x1D1DE;-&#x1D1E8;&#xE000;-&#xF8FF;]+')"
+            role="warning">May only contain codepoints from the Basic Multilingual Plane Private Use
+            Area as well as musical symbols from Unicode.</sch:assert>
+        </sch:rule>
+      </constraint>
+    </constraintSpec>
+    <attList>
+      <attDef ident="rotation" usage="opt">
+        <desc>A positive value for rotation rotates the text in a counter-clockwise fashion, while
+          negative values produce clockwise rotation.</desc>
+        <datatype>
+          <rng:ref name="data.DEGREES"/>
+        </datatype>
+      </attDef>
+    </attList>
+    <remarks>
+      <p>Similar to <gi scheme="MEI">symbol</gi> elements with <att>auth</att> set to "smufl", but
+      for a string of multiple characters. Typical use-case are dynamics for which there is no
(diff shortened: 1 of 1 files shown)


th-we:
An alternative and less "radical" approach would be to define a generic font family [similar to CSS's](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#%3Cgeneric-name%3E) `serif`, `sans-serif`, `monospace` etc.  I'd suggest `@fontfam="smufl-text"`. Like this, `<rend fontfam="smufl-text">` could be used instead of `<smufl>`. However, `<smufl>` is more semantic, `<rend>` is more visual.

rettinghaus:
In your example using `<smufl>` within `<dynam>` isn't "dynam" providing sufficient semantics? 

And I'm still not 100% sure about the reason behind this. The goal is to say specifically (within a textual context) to use musical glyphs from the used music (SMuFL) font. So what would happen, if you have a renderer that doesn't rely on SMuFL? Or what would you do, if you want music glyphs within text from another SMuFL font? Would you then use `<rend>` within `<smufl>`?
Also I think `<smufl>` is a bit too technology specific in naming. Enabling `<rend>` to do this properly still sounds better to me.

th-we:
Then what about introducing `rend/@encoding.auth`, in analogy to [`symbol/@glyph.auth`](https://music-encoding.org/guidelines/v4/attribute-classes/att.extsym.html)? This would signal the renderer to switch from a text font to a SMuFL font. A more specific font can then be specified using `@fontfam` if desired.

In addition to the predefined value `smufl` we should probably also have `unicode` so that in nested `<rend>`s we can switch back the default.

Concerning non-SMuFL based renderers:  They would have to map the SMuFL glyphs to glyphs they use.

kepper (commented):
I wouldn't necessarily split up att.typography into multiple classes, but instead remove `@fontname` (#716). Even though I apparently triggered this discussion, I think @rettinghaus's arguments are good, and maybe something like `<smufl>` is over the top. So, at this point, I would follow the proposal of a predefined font family `smufl-text`. I don't think we need a value `unicode`, but could perhaps mention something like `unset` (as in CSS). However, all that would be done in documentation only, and wouldn't require changes to the specs (unless we switch from the `token` datatype to a semi-closed list, but that seems too drastic again…). Thoughts?

th-we:
I find semi-closed lists perfect for this purpose.

kepper:
Ok, but how would that look like? Something like the following would allow to use any other font as well, and would be largely compatible with CSS: 

``` xml
<valList type="semi">
    <valItem ident="serif">
        <desc>This region should be rendered with a default serif font.</desc>
    </valItem>
    <valItem ident="sans-serif">
        <desc>This region should be rendered with a default sans-serif font.</desc>
    </valItem>
    <valItem ident="monospace">
        <desc>This region should be rendered with a default monospace font.</desc>
    </valItem>
    <valItem ident="smufl">
        <desc>This region should be rendered with a SMuFL-compliant font and uses the corresponding Unicode codepoints.</desc>
    </valItem>
    <valItem ident="smufl-text">
        <desc>This region should be rendered with a SMuFL-compliant font for texts and uses the corresponding Unicode codepoints.</desc>
    </valItem>
    <valItem ident="unset">
        <desc>This region should be rendered with a default font. This is used to revert more specific font declarations on ancestor elements.</desc>
    </valItem>
</valList>
```

kepper:
During the ODD Thursday on Oct 29, the last proposal was considered to be good. @kepper will prepare an implementation and put it up here.

bwbohl:
what's the status of this, can we bring it up on the next ODD meeting?

bwbohl:
Although I get into the discussion rather late, here's my two cents…

In general I think this is a good idea, nevertheless I think there's a granularity mismatch between serif, sans-serif – monospace – smufl, smufl-text.

While the first two describe font-features that can also be encountered in both monospace or smufl fonts, the counterpart to a monospaced typeface would be a proportional typeface, while smufl or smufl-text indicate a specific usage of the character map designed for musical purposes. Somehow this all doesn't really fit together or am I missing something?

th-we:
@bwbohl You're right, `@fontfam` is not the cleanest solution. [What about introducing `@encoding.auth`](#issuecomment-679030146) – or just allow the existing `@glyph.auth` on more elements?

ahankinson:
Assuming `<dynam>` might be used by parsers to extract some machine readable text given raw values, how about:

```xml
<dynam>
   <rend glyph.auth="smufl">&#xE522;&#xE525;&#xE520;</rend>
</dynam>
```

The only change would be to add `@glyph.auth` (and probably the other `@glyph...` attributes) to `<rend>`, which is probably OK?

bwbohl:
What about not having to wrap this in rend, because, from my perspective, rend is a maximum unspecified last resort way of doing things. Either we could use `glyph` in `dynam` or put the attributes directly on `dynam`, couldn't we?

th-we:
@bwbohl Do you mean `<symbol>`, which already works inside `<dynam>`? Or do you suggest introducing a `<glyph>` element? Maybe the conclusion of this long-winded discussion might be to stick with `<symbol>`?

Putting the attributes on `<dynam>` causes problems for "mixed" dynamics like "subito pp".

lpugin:
> Maybe the conclusion of this long-winded discussion might be to stick with `<symbol>`

Does it mean you would do
```xml
<dynam>
   <symbol>&#xE522;</symbol><symbol>&#xE525;</symbol><symbol>&#xE520;</symbol>
</dynam>
```
?

I much prefer
```xml
<dynam>
   <rend glyph.auth="smufl">&#xE522;&#xE525;&#xE520;</rend>
</dynam>
```

I agree that putting the attributes on `<dynam>` would not work.

ahankinson:
Discussion at MEI Developers Meeting (24.05.2023): Closing this issue in favour of adding `@glyph.auth` to `<rend>`. A new PR is incoming.

bwbohl:
```xml
<dynam>
   meno <rend glyph.auth="smufl">&#xE522;&#xE525;&#xE520;</rend>
</dynam>
```

In this example 'meno' would be regular text and the contents of the rendered element rendered according to the SMuFL-table

PR #717, opened 2020-10-15
Remove fontname attribute

### LINKED ISSUE #716: deprecate fontname
As @th-we pointed out in https://github.com/music-encoding/music-encoding/pull/704#issue-469038961, we still do have both `@fontfam` and `@fontname`. The latter is a very old artifact, and there is no corresponding attribute / value in TEI or CSS. Both are defined in the same attribute class [att.typography](https://music-encoding.org/guidelines/v4/attribute-classes/att.typography.html), which means they are always available in the same spaces. There seems absolutely no reason to preserve both, and I recommend to go with `@fontfam`, and drop `@fontname`. Taking this out will change "forward compatibility", but an automatic conversion seems easy, so that shouldn't be an issue.

kepper:
fixes #716

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -3444,12 +3444,6 @@
           <rng:ref name="data.FONTFAMILY"/>
         </datatype>
       </attDef>
-      <attDef ident="fontname" usage="opt">
-        <desc>Holds the name of a font.</desc>
-        <datatype>
-          <rng:ref name="data.FONTNAME"/>
-        </datatype>
-      </attDef>
       <attDef ident="fontsize" usage="opt">
         <desc>Indicates the size of a font expressed in printers' points, i.e., 1/72nd of an inch,
           relative terms, e.g., "small", "larger", etc., or percentage values relative to "normal"


rettinghaus:
This breaks backwards compatibility and should go into MEI5.

kepper:
During ODD Thursday on Oct 29, we decided that this should be merged only if there are other breaking changes in the specs already. If not, it will be set aside for later consideration.

kepper:
one question to @lpugin: Does Verovio support both `@fontname` and `@fontfam`? There is at least one example in the Guidelines that uses `@fontname`, and it would be good to know if this would break more than just backwards compatibility ;-)

lpugin:
Verovio uses `@fontname`, but this is easy to change. Since Verovio is currently aligned with the develop version of MEI we can break compatibility without problem. Thanks for asking.

Regarding the example we can see, but I would suggest to drop it in the new version of the guidelines. In the end it would probably be simpler to have only examples images generated with Meiler. We can see this next week.

bwbohl:
probably should be closed in favour of #897 ? Could you clarify?

bwbohl:
closed in favour of #897

PR #722, opened 2020-10-22
Add plica element & fix stem element

martha-thomae:
This PR adds support for a `<plica>` element to be used as child of `<note>`. The plica element has two attributes (which are shared with the already existing `<stem>` element): `@dir` and `@len`. A `<note>` element can have only one child `<plica>`.

This PR also contains some corrections regarding the `<stem>` element and its attributes:
- Removed the line `<memberOf key="model.noteModifierLike"/>` from the `<stem>` element classes since, in addition to allow `<stem>` as a child of `<note>` (which was the intention), it also allowed `<stem>` to be a direct child of `<layer>`. Instead, I added `<stem>` directly into the content of `<note>`.
- Substitute the `@length` attribute of `<stem>` by `@len` (to be more consistent with MEI).
- Add `@color` attribute.

For both stem and plica, added corresponding `anl`, `log`, `ges`, and `vis` attribute classes.

**Acknowledgement:**
Thanks to @pe-ro for helping me getting this PR ready. He noticed issues with it that are fixed now.

### DIFF
--- a/source/modules/MEI.mensural.xml
+++ b/source/modules/MEI.mensural.xml
@@ -510,7 +510,7 @@
     <constraintSpec ident="Check_plica" scheme="isoschematron">
       <constraint>
         <sch:rule context="mei:plica">
-          <sch:assert test="count(../mei:plica) &lt;= 1)">Only one plica is allowed.</sch:assert>
+          <sch:assert test="count(../mei:plica) &lt;= 1">Only one plica is allowed.</sch:assert>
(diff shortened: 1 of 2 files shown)


kepper (approved):
We should trust in the work of our IGs, and @pe-ro has already had a look from a technical perspective – ready to go…

kepper:
thanks @martha-thomae 👍

PR #723, opened 2020-10-22
Add fermata attribute for mensural notes

martha-thomae:
The fermata attribute (found in the `att.fermataPresent` class) is only available for notes in CMN.

In this PR, I add the missing class `att.note.anl.mensural` (the logical, gestural, and visual corresponding classes already exist), and make it a member of `att.fermataPresent`, similar to how `att.note.anl.cmn` is a member of `att.fermataPresent`. Then include the newly introduced `att.note.anl.mensural` class in the classes of `att.note.anl`.

### DIFF
--- a/source/modules/MEI.analytical.xml
+++ b/source/modules/MEI.analytical.xml
@@ -351,6 +351,7 @@
       <memberOf key="att.intervalMelodic"/>
       <memberOf key="att.melodicFunction"/>
       <memberOf key="att.note.anl.cmn"/>
+      <memberOf key="att.note.anl.mensural"/>
       <memberOf key="att.pitchClass"/>
       <memberOf key="att.solfa"/>
     </classes>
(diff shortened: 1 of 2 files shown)


craigsapp:
I do not have a problem with allowing `<fermata>`, but note that it is often called [corona](https://en.wikipedia.org/wiki/Mensural_notation#Corona) in mensural music.  with the link saying:

>  It is similar to the fermata in modern notation. 

Perhaps there is no functional difference between them, in which case using fermata as the name for the mensural version is no problem. (I am not aware of any difference at all, at least in the music of Josquin).

Perhaps some mensuralists could comment on if there is any real difference between them?

craigsapp:
Jesse Rodin cannot think of any difference between a fermata and a corona, so it should be safe to assume that they are equivalent.

martha-thomae:
Hi @craigsapp! Thank you for checking on this :)

kepper (approved):
sounds good to me

PR #726, opened 2020-11-01
Update pattern for humdrum recip values

rettinghaus:
See discussion on https://github.com/rism-ch/verovio/issues/1796


bwbohl (approved):
This PR updates the definition of `@dur.recip` to match the current Humdrum definition (in place for 10 years!!)

kepper:
thanks for the good catch :-)

PR #728, opened 2020-11-10
remove empty att.midi.vis

rettinghaus:
This PR removes the `midi.vis` attribute class, because it is empty and not clear what should be _visual_ about midi information.

### DIFF
--- a/source/modules/MEI.midi.xml
+++ b/source/modules/MEI.midi.xml
@@ -357,7 +357,6 @@
     <classes>
       <memberOf key="att.common"/>
       <memberOf key="att.midi.log"/>
-      <memberOf key="att.midi.vis"/>
       <memberOf key="att.midi.ges"/>
       <memberOf key="att.midi.anl"/>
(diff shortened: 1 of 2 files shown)


bwbohl:
@kepper do you know any reason why there is att.midi.vis class which actually is never being referenced from the MEI Source? If I'm not mistaken there is nothing »visual« about MIDI after all!

kepper (approved):
We have quite a number of "empty" attribute classes, which are there mostly for consistency reasons – they could be useful sometime in the future, and probably have been auto-generated to large parts. I'm still a bit undecided because of the consistency argument, but I agree that it's rather unlikely to say something visual about the `<midi>` element ("Container for elements that contain information useful when generating MIDI output."). I think it's fine to add this back in when it becomes necessary ;-)

kepper:
as per the discussions on ODD Friday, we take this one out, but this is not to be understood as general tendency to remove unused stuff…

Issue #672, opened 2020-06-02
enclosed noteheads


rettinghaus:
Values in `data.NOTEHEADMODIFIER.list` doubles what can be addressed with `@enclose`. 
Or does `enclose` here means "the whole note is parenthesized/bracketed"? I personally never have witnessed the latter before. 
If it is possible to be more strict in choices of encoding (and to be more consistent) I'd propose to drop the values `paren` and `brack` from `data.NOTEHEADMODIFIER.list`.

pe-ro:
`enclose` does indeed mean "the whole note is parenthesized/bracketed" as opposed to the values in `data.NOTEHEADMODIFIER.list` of `paren` and `brack` which indicate that only the note head is enclosed by these symbols.

I can't readily produce an example of the former, but I think it makes more sense to anticipate its occurrence than to deny the possibility of its existence.  Brackets, i.e. `[ and ]`, are often used to indicate that the stuff between the brackets is optional.  It's not too difficult to imagine an optional note, e.g. one supplied by an editor.  Of course, this is best represented semantically, using editorial markup `<supplied>`.  However, the naive encoder, an OMR program for example, can't make this determination _a priori_, leading to the need to deal with the notation visually rather than semantically.

rettinghaus:
I have thought about this some more. 
If `enclose` means "the whole note is parenthesized/bracketed" there will be problem when it comes to rendering when the stem size is changed.
![stems](https://user-images.githubusercontent.com/7693447/112626679-cf27d280-8e30-11eb-86e1-bed0cfba5fff.png)
How should that be handled?

On the other hand: what is the "whole note" within in chord? 
Obviously the stem is a part of the chord: 
![chord04](https://user-images.githubusercontent.com/7693447/112626678-cf27d280-8e30-11eb-85c3-ad8e9131e8fe.png)
So enclose in this context would mean "enclose the notehead AND its dots AND its accidentals". 
![chord02](https://user-images.githubusercontent.com/7693447/112626683-cfc06900-8e30-11eb-9de2-e739d73b4511.png)
While the `NOTEHEADMODIFIER` will give this:
![chord03](https://user-images.githubusercontent.com/7693447/112626676-cdf6a580-8e30-11eb-9370-2e44f1d6f500.png)
Does it even make sense? 
And what about adjacent notes in a chord? 
![chord01](https://user-images.githubusercontent.com/7693447/112626681-cfc06900-8e30-11eb-9888-98d110228978.png)

[...]

PR #729, opened 2020-11-10
make non-textual elements part of att.enclosingChars

### LINKED ISSUE #473: Support enclosed dynamics and hairpins
To support enclosed dynamics and hairpins; that is, those surrounded by square brackets or a box, described by Gould on pp. 249, 597, and 599, make `<dynam>` and `<hairpin>` members of att.enclosingChars and extend data.ENCLOSURE to include the value "box".  This change will also make it possible to enclose the current members of att.enclosingChars (accid, ambNote, artic, chord, cpMark, keyAccid, note, and rest) by a box.  For the sake of completeness, the value "none" should also be added so that it's possible to say explicitly that the object is _not_ enclosed.

rettinghaus:
This PR adds the ability to encode parentheses and brackets around non-textual glyphs in MEI, as seen in the following examples:
image
image
image
image
image

Closes #473

### DIFF
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -348,6 +348,7 @@
     <classes>
       <memberOf key="att.altSym"/>
       <memberOf key="att.color"/>
+      <memberOf key="att.enclosingChars"/>
       <memberOf key="att.extSym"/>
       <memberOf key="att.typography"/>
       <memberOf key="att.visibility"/>
@@ -510,6 +511,7 @@
     <classes>
       <memberOf key="att.altSym"/>
       <memberOf key="att.color"/>
+      <memberOf key="att.enclosingChars"/>
       <memberOf key="att.extSym"/>
       <memberOf key="att.placementRelStaff"/>
       <memberOf key="att.typography"/>
@@ -1070,6 +1072,7 @@
     <classes>
       <memberOf key="att.altSym"/>
       <memberOf key="att.color"/>
+      <memberOf key="att.enclosingChars"/>
       <memberOf key="att.extSym"/>
       <memberOf key="att.typography"/>
     </classes>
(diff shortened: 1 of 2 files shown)


rettinghaus:
I'm wondering if we should change the regex pattern for `meterSig@count` (and add one for `@unit`) to allow parentheses and brackets for each?

kepper:
With this addition, we're dealing with both the visual and the logical domains. First and foremost, it is about the brackets that are to be found in the score. And then those brackets have a specific meaning. Both aspects are connected, but should not be considered the same thing. So my first reaction was to say that we should wrap those elements with a `<supplied>` and be done with it. Then, I noticed that there's nothing on `<supplied>` to specifiy the type of rendition (adding `att.enclosingChars` to `<supplied>` might be another related issue therefore…). And then again, sometimes those brackets may have a different meaning, or our favorite naive encoder may not feel comfortable to call it supplied. So, in order to allow a full separation of domains, I think it's necessary to allow it on those elements. Haven't looked up the full list yet, and would like to reflect about it at today's ODD Friday, so not approving it for now – but that doesn't mean I'm against it…

bwbohl:
As has become clear during the [2021-03-26 ODD Meeting](https://github.com/orgs/music-encoding/projects/2#column-13107157), this issue to some degree is connected to #672

kepper:
I'll have a closer look on the consequences of this. In addition, it seems important to have proper examples and documentation for this, probably split between chapters 2 and 11. I'll try to have a rpoposal ready by next month.

rettinghaus:
rebased in order to resolve conflicts

kepper (approved):
There is consensus that this is necessary and helpful. We do need documentation for this, and that documentation probably needs to be spread out across multiple chapters. A separate ticket will be opened by @rettinghaus for this. We're including the schema changes anyway to move this forward now. Thanks everyone!

PR #732, opened 2020-11-12
fixed tables

aseipelt:
deleted not transformable information

### DIFF
--- a/source/docs/06-neumes.xml
+++ b/source/docs/06-neumes.xml
@@ -121,13 +121,9 @@
                         <cell>Non-connected</cell>
                      </row>
                      <row>
-                        <cell>:---------:</cell>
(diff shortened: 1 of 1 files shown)


kepper (changes requested):
Hi @aseipelt, thanks so far. Could you please turn those new values into valid CSS, as usually preferred by the TEI? So `left` should be `text-align: left;` etc. Again, many thanks…

kepper:
I just looked at https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.global.rendition.html, and it should be slightly different: either `rend="align(left)"` or `style="text-align: left;"`. In that case, I prefer to use `@style`, which translates more easily…

kepper (approved):
perfect, thank you 👍

PR #740, opened 2020-12-04
Feature local build

lpugin:
Enable the GH Action to deploy the Schema and the Guidelines


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,131 @@
...
+name: Deploy Schema and Guidelines
+
+on:
+  push:
+    branches: [ develop ]
+
+env:
+  # schema
+  SCHEMA_REPO: ${{ github.repository_owner }}/schema
+  SCHEMA_BRANCH: feature_local-build
```
kepper: Is this the correct branch?

kepper (commented):
Not sure about the branch on the `schema` repo. And I haven't tested this locally (need to familiarize myself with this first…)

.github/workflows/deploy.yml
lpugin: Well spotted. I changed it.

build.xml
```diff
@@ -0,0 +1,175 @@
...
+        <basename property="odd.basename" file="${customization.path}" suffix=".xml"/>
+        <echo>building HTML guidelines for ${odd.basename} from ${customization.path}</echo>
+        <java classname="net.sf.saxon.Transform" classpath="${basedir}/lib/xerces/oxygen-patched-xerces-21.1.0.2.jar;${basedir}/lib/saxon/saxon9he.jar">
+            <arg value="-s:${dir.dist.schemata}/mei-source_canonicalized.xml" />
+            <arg value="-xsl:${basedir}/utils/guidelines_xslt/odd2html.xsl" />
+            <arg value="-o:${dir.dist.guidelines}" />
+            <arg value="-xi:on"/>
+            <arg value="output.folder=${dir.dist.guidelines}/"/>
+            <arg value="git.sha=git.sha"/>
+            <arg value="branch=develop"/>
```
kepper: I assume those are the parameters passed into the XSLT, right? If that is the case, `output.folder` is correct, but `git.sha` is expected to be `hash`, and `branch` is probably supposed to be `version`, even though the value given here doesn't really fit (I'd expect something like `4.1.0-dev` or similar). @bwbohl, could you verify? We can still decide where to change things, but I'd like to understand first…

kepper (commented):
Hi Benni, could you have a look at this?

build.xml
lpugin: I assume we need to pass it to the ant script. I can add this later if you don't mind.

bwbohl: Forgot about the sha in the first place, thx for adding. I thought $version is supposed to be dev, v4, v3 etc like in the guidelines XSLT and in ANT is used to name the dist subfolder

bwbohl: ok fixed git sha in ant and XSLT relation, still have to submit it to the ant call from the GHAction, though.
For the CI, if ant is called the following way:

```
ant init
ant -lib lib/saxon/saxon9he.jar -Dgithub.sha=1234567
```

`1234567`will be submitted to the XSLT as value of the $hash parameter

If it is called without the parmeter, e.g.

```
ant init
ant -lib lib/saxon/saxon9he.jar
```

Ant will determine the local git sha and dirty state and submit it as value of the XSLT's $hash.

bwbohl: I wonder about the XSLT's $version parameter and if we actually shouldn't stick to the semantic versioning concept of adding the commit sha behind the last stable version, e.g.:
`4.0.1+e429146`

or even add the develop branch-name as "pre-release info": 
`4.0.1-feature_local-build+e429146`
Of course one could argue that that the GitHub SHA already serves as a pointer to the correct branch, nevertheless this would make the version more human-readable and could come in handy with pull requests.

`git describe --tags` also allows to output the number of commits since a tag, e.g.: 
`v4.0.1-269-ge429146`
Which would give a hint on how different the version is from the last release.

or `git describe --tags --dirty` even giving dirty state if any file has been modified locally:
`v4.0.1-269-ge429146-dirty`

Nevertheless this does not fully comply to http://semver.org and should be turned into something like:
`v4.0.1-269+g.e429146`

resp.
`v4.0.1-269+g.e429146-dirty`

@music-encoding/core-contributors : any opinions?

lpugin: Where is this version number then used? My understanding was that we wanted to have somewhere a way to go back to the commit of the source from which the guidelines where built from. We definitely need this, but I am fine to have more if you think we need to as long as we have that and that the output folders unchanged (v4, v5, dev).

(as I said, we can always improve this later but I think for now it would be good if we can merge this it in order to have it run so the users have back what we had before)

kepper: The `version` parameter is only partially necessary. When not provided, I will extract the version from the `@meiversion` attribute. The problem there us we need to make sure this is maintained properly. The version number found there is automatically processed into the `/v4/`-style folder names. We may need to agree on some strict naming conventions for the dev version, because there it's not equally easy to get that name. 

The `hash` parameter (that @bwbohl passes in as `git.sha` – I'm fine with either way) is needed to add a link to the Git repo. Here, we just need to take a decision about the name of that parameter. Benni, your choice…

bwbohl: concerning the branch arg, we don't really need this at the moment, as the XSLT isn't using it; and version is populated by the XSLT from the `@meiversion`defaultVal

bwbohl (approved):

PR #742, opened 2020-12-08
Fix local build

bwbohl:


utils/guidelines_xslt/odd2html/guidelines.xsl
```diff
@@ -129,6 +129,9 @@
     <xsl:template match="tei:gi" mode="guidelines">
         <xsl:variable name="text" select="string(text())" as="xs:string"/>
         <xsl:choose>
+            <xsl:when test="@scheme = 'TEI' and doc-available('https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-' || $text || '.html')">
+                <a class="link_odd_elementSpec" href="https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-{$text}.html">tei:<xsl:value-of select="$text"/></a>
+            </xsl:when>
             <xsl:when test="$text = $elements/@ident">
```
bwbohl, line 135: @kepper this seems to be the problem as e.g. 'moduleRef' is not included in `$elements/@ident`

kepper: as mentioned on slack, this is triggered by the two `<gi scheme="TEI">` elements in here: https://github.com/music-encoding/music-encoding/blob/85b9a03258ddb826bf1816712633568f90cbae31/source/docs/14-integration.xml#L122. However, the lines you brought up from the XSLT resolve that perfectly well on my machine. Only thing I can imagine is that the `doc-available()` function fails in that context, as it may not have access to outside ressources. We could drop that link entirely, or drop the test whether that page exists.

kepper: or, of course, try to make `doc-available()` work as expected (assuming that this indeed is the showstopper…)

bwbohl: yes it indeed is the showstopper ;-/
if I remove doc-available from the test and put it in a message it returns false, even running locally on ubuntu 20.04.1

bwbohl: unparsed-text-available returns true though

bwbohl: @kepper are you ok to change that?

kepper:
thanks @bwbohl 👍  🎉

utils/guidelines_xslt/odd2html/guidelines.xsl
kepper: This is changed, no? https://github.com/music-encoding/music-encoding/blob/e02b5fe1f700ff3df0b1e57ceafc2ec8af39e024/utils/guidelines_xslt/odd2html/guidelines.xsl#L132 
Am I missing something (perfectly possible…), or is this just a delayed message?

PR #743, opened 2020-12-08
fix TEI link-check in building guidelines HTML

bwbohl:
due to unclear reasons doc-available fails on the TEI website using the commandline saxon via ant; this commit switches to unparsed-text-available function which returns true


bwbohl:
closes #742

PR #746, opened 2020-12-08
remove schemata from repository

### LINKED ISSUE #677: Schema update via Travis or GitHub Actions
It seems that the schema available in https://github.com/music-encoding/music-encoding/tree/develop/schemata are not up to date. Isn’t it expected to have them build / updated on every commit to develop?

bwbohl:
the build artifacts are now automatically published at https://github.com/music-encoding/schema to avoid confusion they have to be deleted here

closes https://github.com/music-encoding/music-encoding/issues/677


kepper (approved):
They should go out of here, yes. The links over at https://github.com/music-encoding/music-encoding.github.io/blob/master/resources/schemas.md#mei-40-schemas are already pointing to the new location, so this is good to go, I believe…

Issue #675, opened 2020-06-16
Update meiversion list


lpugin:
The develop version of the schema restricts /mei@meiversion to 4.0.0 and 4.0.1. This is not accurate and it would probably be wise to have a way to distinct files validated against a develop version of the schema that it not version 4.0 anymore. (This is currently a problem for Verovio because it cannot make the distinction between the type of files being passed to it. However, this is probably a problem not only for Verovio, which is the reason why I am opening an issue.)

kepper:
since we're looking at a v5 at some point, I'd recommend to allow something like a regex expressing something like `>=5.0.0-0 <5.0.0` (see https://jubianchi.github.io/semver-check/#/%3E%3D5.0.0-0%20%3C5.0.0/5.0.0-0). Eventually, this means we'd have to increment the last digit with each commit, which seems rather impractical. Not sure yet what to do…

ahankinson:
Would it make sense to set the value of meiversion to just the major, or possibly major.minor? “5” or “5.0”?

kepper:
I would say we should set it to something pre-5, but post-4. However, I'm not sure how where this should be validated – in the data, or in an application. Should we have only one single value in the schema for pre-5, or shall we be more specific? I think for all subsequent releases of 5, we need to be specific, but one should perhaps not expect a huge amount of consistency of an explicit dev version, right?

lpugin:
Something like "5.0.0-dev" would be appropriate in my opinion. I don't think we can reasonably have a version number (5.0.0-0) for each commit, even though strictly speaking this would be better.

lpugin:
Anybody opposed to "5.0.0-dev"? If not I am happy to prepare a PR for this.

rettinghaus:
If I'm not wrong, we're still fixing bugs for a minor release. @bwbohl? @kepper?

lpugin:
I think you are wrong. We already had several PR merged that introduced modifications (mostly additions) to the schema.

rettinghaus:
But we have a `MEI 4.1.0` milestone and a `MEI 5.0.0` milestone.

lpugin:
Correct. As far as I can see we have no corresponding branches, and 66 issues with no milestone. Things become quite difficult to track....

bwbohl:
> Correct. As far as I can see we have no corresponding branches, and 66 issues with no milestone. Things become quite difficult to track....

[...]

PR #747, opened 2020-12-08
Set `@meiversion`default value to 4.0.1-dev

bwbohl:


lpugin:
Version number should be `4.1.0-dev` or `5.0.0-dev` depending on @kepper analysis of the current status of the branch. See https://github.com/music-encoding/music-encoding/issues/675#issuecomment-734853336

source/modules/MEI.shared.xml
```diff
@@ -1855,7 +1855,7 @@
     <attList>
       <attDef ident="meiversion" usage="opt">
         <desc>Specifies a generic MEI version label.</desc>
-        <defaultVal>4.0.1</defaultVal>
+        <defaultVal>4.0.1-dev</defaultVal>
```
lpugin, line 1858: Can we have a default value that is not in the closed list? This does not make sense to me.

bwbohl: yeah you're right, missed that

lpugin: Since 4.0.1-dev is already out it should be 4.1.0-dev or 5.0.0-dev, shoudn't it?

kepper: @lpugin is right, but we'll always have that problem – our `dev`-version will always be named in relation to something that comes after the thing we know the name of. The only solution I see at this point is to add all potential successors to the list of allowed values, so basically something like
```xml
<valList type="closed">
  <valItem ident="4.0.2-dev">
    <desc>Potential bugfix release 4.0.2</desc>
  </valItem>
  <valItem ident="4.1.0-dev">
    <desc>Potential minor release 4.1.0</desc>
  </valItem>
  <valItem ident="5.0.0-dev">
    <desc>Potential first release of MEI 5</desc>
  </valItem>
</valList>
```
While we work on the schema, we can escalate those levels – as long as we have only bugfixes, our `defaultVal` will be `4.0.2-dev`. When we have sufficiently complex changes, we will take out the 4.0.2 and 4.1.0 options one after another. When we do a release, we just pick the smallest value still available. It seems arguable to keep the 4.0.0 and 4.0.1 releases in the list, as long as we don't break compatibility. At the same time, I like to be strict about the dev version… Thoughts?

kepper: Regarding the question of whether we do a `4.1.0` or a `5.0.0` now – I'm still in favor of having a clean 5 now. Yes, the changes aren't dramatic, but there are some things which aren't allowed anymore, beyond bugfixes. The only thing that let's me hesitate is that we don't have the PDF yet, and I would prefer to have that together with the next major release. We could delay the release by a couple of weeks now, or we do a 4.1 now, where I feel less urged to have the PDF in place. However, this is dangerously close to http://sentimentalversioning.org/, and I certainly don't want to get there. Actually, I like the idea of an automated comparison with the previous release to decide about the severity of changes. I did that with the tool that generated https://music-encoding.org/archive/comparison-4.0.html, and it spotted a small number of breaking changes. In order to avoid sentiments in this, I think that even a single such change on an element that's barely ever used has to trigger a new major release. Comments?

lpugin: I don't think potential bug fixes should be included in the develop branch. What we currently have in develop should never become 4.0.2. If we make a bug fix, we take master and then integrate the bugfix from master into develop.

bwbohl:
evidence shows that 5.0.0-dev is the way to go ;-)

PR #748, opened 2020-12-10
bump meiversion to 5.0.0-dev

bwbohl:
bump default and allowed values of `@meiversion` to 5.0.0-dev

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -1855,13 +1855,10 @@
     <attList>
       <attDef ident="meiversion" usage="opt">
         <desc>Specifies a generic MEI version label.</desc>
-        <defaultVal>4.0.1</defaultVal>
+        <defaultVal>5.0.0-dev</defaultVal>
         <valList type="closed">
-          <valItem ident="4.0.0">
-            <desc>First release of MEI 4</desc>
-          </valItem>
-          <valItem ident="4.0.1">
-            <desc>Bugfix Release 4.0.1</desc>
+          <valItem ident="5.0.0-dev">
+            <desc>Development version of MEI 5.0.0</desc>
           </valItem>
         </valList>
       </attDef>


kepper (approved):
This is what we have agreed on. The changes do require a 5.0.0 release, and this sets the path to that…

kepper:
thanks @bwbohl 👍

rettinghaus:
@kepper you have agreed on that.

kepper:
We do have some changes that are not backwards compatible. Yes, they are not at the core of MEI, but who's to decide what that is? I firmly believe that even the slightest change requires us to do a major bump. As no contradictions to [such statements](https://github.com/music-encoding/music-encoding/pull/747#discussion_r539396532) were coming in, the Technical Chairs took the liberty to decide the issue. I apologize if this was too fast, but eventually, taking a decision was the only way to actually resolve https://github.com/music-encoding/guidelines/issues/195. And it was my understanding that during the dev workshop, we agreed that when we find breaking changes, that this would indeed trigger the 5.0.0. We did find such changes, and I don't think we should start to discuss how relevant they are. I am aware that this policy may end in a higher frequency of major releases, but is that really a problem? We could think about a different interpretation of the semver levels (major only for architectural changes, minor for breaking changes, bugfixes for bugfixes or some such), but even then I would say that the move back to everything in ODD is indeed a major change, even though it affects our workflows and structures more than it affects our resulting product… Again, apologies if I haven't listened sufficiently to the community – it was done with the best intentions…

bwbohl:
I fully second @kepper in this and of course this does not mean a huge delay in issuing the 5.0.0 release. We should closely inspect open issues. btw. the Milestones I created someday (e.g. https://github.com/music-encoding/music-encoding/milestone/8) were only to categorise whether an issue would be breaking or not, I should better have done that with labels, probably.

So my intention would not be to wait until https://github.com/music-encoding/music-encoding/milestone/8 is implemented but revisit ALL issues whether we want some of them implemented before the release

PR #720, opened 2020-10-16
split data.ACCIDENTAL.WRITTEN and add turkish accidentals

rettinghaus:
this PR supersedes #612
(see discussion there)

PR #772, opened 2021-02-22
Add turkish AEU accidentals to data.ACCIDENTAL.GESTURAL

annplaksin:
This is the sequel PR to #720 :
Since the aeu accidental values cannot be expressed with the already existing values of `data.ACCIDENTAL.GESTURAL`, it would be very helpful to add them as gestural values too. 

I splitted `data.ACCIDENTAL.GESTURAL` as it has been done for written accidentals in #720 and added the aeu accidentals. To avoid the duplication of data types, I renamed `data.ACCIDENTAL.WRITTEN.aeu` to `data.ACCIDENTAL.aeu`.

A corrected version of the aeu accidental value overview has been added as well.


kepper (approved):
didn't compile to see if the image links work, but other than that seems plausible – thanks for the work on that 👍

kepper:
thanks for your work on this, @annplaksin 👍

rettinghaus:
Sorry for me complaining about this so late, but even if AEU accidentals would have be the same in gestural and visual domain, I don't think we should combine them nto `data.ACCIDENTAL.aeu`, as this breaks current naming (and basically the current schema). I would prefer to keep `data.ACCIDENTAL.WRITTEN.aeu` and add `data.ACCIDENTAL.GESTURAL.aeu` instead.

annplaksin:
You have either a break of naming or redundancy. I asked @kepper before I made the pull request if this naming is okay. He didn't complain.

kepper:
I'm certainly not the the person to decide what is right or wrong here – MEI is a community thing, and we should move forward by consensus whenever possible. I think this is a legitimate question, though, and while I refuse to be any more relevant for the decision than anyone else, I think there are arguments that @annplaksin's implementation is actually consistent with existing MEI. 

In CWMN accidentals, we see a certain amount of duplication of _visual_ values: `x` is two half-tones up, as is `ss`. For _gestural_ purposes, there's no difference between them, though. This difference legitimates to have two separate data types for the two domains. In the list suggested by @annplaksin, there is no similar duplication of values. Accordingly, we don't need to split this up by domain. I think this is conforming to other situations in MEI. The [data.DURATION.cmn](https://music-encoding.org/guidelines/dev/data-types/data.DURATION.cmn.html) data type is used by both data.DURATION (applies to `@dur`) and data.DURATION.gestural (applies to `@dur.ges`) – as the values apply to both categories alike. 

I think we all agree that MEI is a "grown" model that isn't perfectly consistent across all of its parts. In this particular situation, I think we're safely within the "usual range", though – but this is definitely something to discuss and negotiate…

PR #773, opened 2021-02-24
Add 'dur.quality' attribute to 'space' element

martha-thomae:
The space element is a placeholder that uses duration values to fill an incomplete layer. Because of this, for mensural notation, it should also include the 'dur.quality' attribute (since both 'dur' and 'dur.quality' are needed to convey duration of mensural notes).

### DIFF
--- a/source/modules/MEI.mensural.xml
+++ b/source/modules/MEI.mensural.xml
@@ -179,10 +179,24 @@
           </sch:assert>
         </sch:rule>
       </constraint>
+      <constraint>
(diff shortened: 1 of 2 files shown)


martha-thomae:
Tested and working properly: `@dur.quality` is valid when using the MEI Mensural (and MEI All) schema, and invalid when using MEI CMN.

bwbohl:
2021-02-25 ODD meeting: agreed to merge.

PR #777, opened 2021-03-10
add senza misura to data.METERSIGN

### LINKED ISSUE #731: Add senza misura X to data.METERSIGN
We have no easy way to encode the "X" sometimes used to indicate a _senza misura_ with `meterSig`. 
We should add a value "senza" (or something similar) to `data.METERSIGN`.

Example (from Gould):
senza

In SMuFl:
https://smufl-browser.edirom.de/index.html?glyphname=timeSigX

rettinghaus:
fixes #731

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -2020,7 +2020,7 @@
           <desc>Cut time; i.e. 2/2.</desc>
         </valItem>
         <valItem ident="open">
-          <desc>Senza misura</desc>
+          <desc>Open time signature, i.e., Senza misura. See Gould pp. 611–615.</desc>
         </valItem>
       </valList>
     </content>
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -2019,7 +2019,7 @@
         <valItem ident="cut">
           <desc>Cut time; i.e. 2/2.</desc>
         </valItem>
-        <valItem ident="senza">
+        <valItem ident="open">
           <desc>Senza misura</desc>
         </valItem>
       </valList>
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -2019,6 +2019,9 @@
         <valItem ident="cut">
           <desc>Cut time; i.e. 2/2.</desc>
         </valItem>
+        <valItem ident="senza">
+          <desc>Senza misura</desc>
+        </valItem>
       </valList>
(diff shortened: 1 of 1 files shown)


source/modules/MEI.xml
```diff
@@ -2019,6 +2019,9 @@
         <valItem ident="cut">
           <desc>Cut time; i.e. 2/2.</desc>
         </valItem>
+        <valItem ident="senza">
+          <desc>Senza misura</desc>
```
ahankinson: How does this differ from the `invis` value in `data.METERFORM`?

rettinghaus: Please see #731. It is meant to specifically encode [this symbol](https://smufl-browser.edirom.de/index.html?glyphname=timeSigX):
 


See Gould pp. 611–615

ahankinson: Ah, Ok. I was going by this:

https://usermanuals.musicxml.com/MusicXML/Content/EL-MusicXML-senza-misura.htm

Given that the other options in this are the "cut" and "common", perhaps something like "open" (from the SMuFL description "Open time signature") or just "x"? (not that I'm necessarily a fan of single-letter attribute values, but it is the most descriptive)  

Given that the MusicXML interpretation is an invisible time signature, but you definitely intend a visible one, maybe making that more explicit is a good thing. We can still keep "senza misura" in the description to make it clear...

ahankinson: Also this example, where the time signature has been hidden:

https://repertoire-explorer.musikmph.de/en/product/holst-gustav-17/

rettinghaus: I agree. `open` is a better value here.

ahankinson: Only in music notation does "X" mean "open", but it is what it is. :) 

Thanks!

source/modules/MEI.xml
```diff
@@ -2019,6 +2019,9 @@
         <valItem ident="cut">
           <desc>Cut time; i.e. 2/2.</desc>
         </valItem>
+        <valItem ident="open">
+          <desc>Senza misura</desc>
```
ahankinson:
```suggestion
          <desc>Open time signature, i.e., Senza misura. See Gould pp. 611–615.</desc>
```

ahankinson (approved):

PR #788, opened 2021-04-20
remove MIDI attributes from silent elements

rettinghaus:
This removes `att.instrumentIdent` from the _non-sounding_ layer elements `mSpace`, `rest`, `mRest`, and `multiRest`, and thus brings them in line with `space`.

### DIFF
--- a/source/modules/MEI.gestural.xml
+++ b/source/modules/MEI.gestural.xml
@@ -399,7 +399,6 @@
     <desc>Gestural domain attributes.</desc>
     <classes>
       <memberOf key="att.duration.gestural"/>
-      <memberOf key="att.instrumentIdent"/>
     </classes>
   </classSpec>
   <classSpec ident="att.mRpt.ges" module="MEI.gestural" type="atts">
@@ -412,14 +411,12 @@
     <desc>Gestural domain attributes.</desc>
     <classes>
       <memberOf key="att.duration.gestural"/>
-      <memberOf key="att.instrumentIdent"/>
     </classes>
   </classSpec>
   <classSpec ident="att.multiRest.ges" module="MEI.gestural" type="atts">
     <desc>Gestural domain attributes.</desc>
     <classes>
       <memberOf key="att.duration.gestural"/>
-      <memberOf key="att.instrumentIdent"/>
     </classes>
   </classSpec>
   <classSpec ident="att.multiRpt.ges" module="MEI.gestural" type="atts">
@@ -590,7 +587,6 @@
     <desc>Gestural domain attributes.</desc>
     <classes>
       <memberOf key="att.duration.gestural"/>
-      <memberOf key="att.instrumentIdent"/>
       <memberOf key="att.rest.ges.mensural"/>
     </classes>
   </classSpec>


bwbohl:
Makes perfectly sense, as there are no rests in MIDI. Thanks for spotting!

pe-ro:
But there can be notes with zero velocity, effectively making them rests.  😊

The ability to assign a rest to an instrument, even if nothing sounds, could be useful in situations where multiple instruments exist in a layer.  I can also imagine educational MIDI applications that would “sound out” rests.

--
p.

bwbohl:
But in the case of converting MIDI to MEI shouldn't the 0-velocity midi:note should still be a mei:note?
Vice versa, when converting MEI to MIDI (a format not containing any rests) shouldn't all the mei:rests be ignored programmatically?
The sounding out idea is an interesting one though...

pe-ro:
Comments below.

--
p.

craigsapp:
> But in the case of converting MIDI to MEI shouldn't the 0-velocity midi:note should still be a mei:note?

There is no 0-velocity MIDI note: the MIDI note-on command with zero velocity functions are a MIDI note-off message.

MEI notes are equivalent to two MIDI events: a Note-on message and a note-off message, so there is not a one-to-one mapping.   MEI notes have duration, but MIDI events are instantaneous, and the duration has to be inferred from the pairing of on/off events.  There are two models that can be selected for pairing: FIFO (first-in-first-out) and LIFO (last-in-first-out).  If there are more note ones than notes off for a sustained timbre, then you get a stuck note, and if there are more note offs and note ons, it does not really matter since the extra note-offs will not alter the sound.  I sometimes use an extra note-off at the end of a MIDI track to extend it longer to prevent the software synths from ending a previous note too soon and chopping off the reverb sound (works on some synths, but not all).

An exception is for General MIDI drum-track notes which are only supposed to have a note-on message and no note-off message.  The drum notes behave like triggers to start a recording of the percussion instrument it represents, and there is no note-off control for these recordings.

Also there are two types of note-offs: There are formal set of note-off messages starting with the hex digit `8`.  These messages have two parameters: (1) the key number to turn off and (2) how fast to turn off the note (release-velocity).  Most synthesizers (especially cheaper ones) do not pay attention to the release-velocity.  A second method for turning off a MIDI note is to use a note-on message with a 0 attack velocity.  This is not a note-on, but rather an alternate method of turning off MIDI notes.

If you want to encode rests in MIDI, using note-off commands would not be safe since it would likely alter the sound of the performed MIDI.  Something like MIDI file meta-messages would be more robust.

> Vice versa, when converting MEI to MIDI (a format not containing any rests) shouldn't all the mei:rests be ignored programmatically?

There is not a one-to-one mapping between MDI note events and MEI data, so there can be forms of MIDI data that cannot be converted into MEI.

rettinghaus:
> But there can be notes with zero velocity, effectively making them rests.

Sorry, but that doesn't make sense at all. As @craigsapp pointed out there is nothing like a 0-velocity MIDI note. Also you would need a pitch here, but the "sound" of the rest wouldn't change even if you change their position on the staff. What you are trying to do here looks to me more like an invisible note in MEI (which still has the `@instr`). 

> The assignment of rests to an instrument has uses beyond conversion to MIDI too.

Actually I can't think of any. But apart from my lack of understanding this, the concept of a notated rest is incompatible with MIDI. So even if you want/need to assign an instrument to a rest, why would you try to do this with a special MIDI attribute? If we had a general purpose instrument assignment method it would be something different, but we don't (and I don't see the pressing need for that). I think that this is something so special that you better should customize MEI for your needs.

pe-ro:
Proceed.

________________________________

kepper (approved):
I think it's fine to take these out in order to avoid confusion – the discussion has shown that this concept is rather "complex", and we might be better off to wait for a concrete use case to re-introduce this when necessary. However, I think MEI should be a place to reason about somewhat arcane and not immediately practical ideas, so it seems fine to me to continue the discussion…

PR #791, opened 2021-04-29
add back linkalign docs (fixes #769)

### LINKED ISSUE #769: Guidelines: Content from v4 not reflected in dev
Some content from the guidelines v4 is not reflected in the dev version. Especially the Linking Data chapter is only a rudiment of itself in dev:

https://music-encoding.org/guidelines/v4/content/linkingdata.html

vs.

https://music-encoding.org/guidelines/dev/content/linkingdata.html

Did not check another chapter so far, but would be good to automatically check (at least once) that dev uses the v4 status as starting point.

kepper:
this should do the trick for #769. Compiled locally, all looking good… – or at least as before ;-)

### DIFF
--- a/source/docs/13-linkingdata.xml
+++ b/source/docs/13-linkingdata.xml
@@ -87,7 +87,7 @@
       <body>
          <div xml:id="linkingdata" type="div1">
             <head>Linking Data</head>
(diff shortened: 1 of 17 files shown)


bwbohl (approved):

musicEnfanthen:
Thank you!

Issue #295, opened 2016-03-10
data.MIDITEMPO


pe-ro:
The datatype data.MIDITEMPO is defined incorrectly and its value range is too restrictive.  Therefore, the explanation for `@midi.tempo` is also incorrect.  Rather than "number of quarter notes per minute", it should be defined as "the number of microseconds per quarter note" (per the MIDI file specification at http://www.somascape.org/midi/tech/mfile.html).  At 120 beats-per-minute (and where a quarter note gets one beat), the value of data.MIDITEMPO would 500,000.

pe-ro:
To be included in the next revision --

```
<macroSpec ident="data.MIDITEMPO" module="MEI" type="dt" mode="delete"/>
<macroSpec ident="data.MIDIBPM" module="MEI" type="dt" mode="add">
  <desc>Tempo expressed as "beats" per minute, where "beat" is always defined as a quarter
  note, *not the numerator of the time signature or the metronomic indication*.</desc>
  <content>
    <rng:data type="positiveInteger">
      <!-- The range is taken from Mup. -->
      <!--<rng:param name="minInclusive">10</rng:param>
      <rng:param name="maxInclusive">1000</rng:param>-->
    </rng:data>
  </content>
</macroSpec>
<macroSpec ident="data.MIDIMSPB" module="MEI" type="dt" mode="add">
  <desc>Tempo expressed as microseconds per "beat", where "beat" is always defined as a
   quarter note, *not the numerator of the time signature or the metronomic
   indication*.</desc>
  <content>
    <rng:data type="positiveInteger"/>
  </content>
</macroSpec>
<!-- replace midi.tempo with midi.bmp and midi.mspb -->
<classSpec ident="att.miditempo" module="MEI.midi" type="atts" mode="replace">
  <desc>Attributes that record MIDI tempo information.</desc>
  <attList>
    <attDef ident="midi.bmp" usage="opt">
      <desc>Captures the number of *quarter notes* per minute. In MIDI, a beat is always
      defined as a quarter note, *not the numerator of the time signature or the
      metronomic indication*.</desc>
      <datatype>
        <rng:ref name="data.MIDIBPM"/>
      </datatype>
    </attDef>
    <attDef ident="midi.mspb" usage="opt">
      <desc>Records the number of microseconds per *quarter note*. In MIDI, a beat is always
       defined as a quarter note, *not the numerator of the time signature or the
       metronomic indication*. At 120 quarter notes per minute, each quarter note will last
[...]

PR #795, opened 2021-05-08
allow decimal values for midibpm

rettinghaus:
> With BPM, sometimes you have to deal with fractional tempos (for example, 100.3 BPM) if you want to allow a finer resolution to the tempo.

See http://midi.teragonaudio.com/tech/midifile/ppqn.htm.

Also see discussion on https://github.com/rism-digital/verovio/issues/1953.

Refers to https://github.com/music-encoding/music-encoding/issues/295.

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -2061,7 +2061,7 @@
       *not the numerator of the time signature or the metronomic indication*.</desc>
     <content>
       <rng:data type="decimal">
-        <rng:param name="minInclusive">0</rng:param>
+        <rng:param name="minExclusive">0</rng:param>
       </rng:data>
     </content>
   </macroSpec>
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -2059,10 +2059,12 @@
   <macroSpec ident="data.MIDIBPM" module="MEI" type="dt">
     <desc>Tempo expressed as "beats" per minute, where "beat" is always defined as a quarter note,
       *not the numerator of the time signature or the metronomic indication*.</desc>
-    <content>
-      <rng:data type="positiveInteger"/>
-    </content>
-  </macroSpec>
+      <content>
+        <rng:data type="decimal">
+          <rng:param name="minInclusive">0</rng:param>
+        </rng:data>
+      </content>
+    </macroSpec>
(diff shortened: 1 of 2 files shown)


source/modules/MEI.xml
```diff
@@ -2059,10 +2059,12 @@
...
-    <content>
-      <rng:data type="positiveInteger"/>
-    </content>
-  </macroSpec>
+      <content>
+        <rng:data type="decimal">
+          <rng:param name="minInclusive">0</rng:param>
+        </rng:data>
+      </content>
+    </macroSpec>
```
ahankinson: I think your indentation is off here?

rettinghaus: Thanks for the hint!

pe-ro:
What does a tempo value of 0 mean?

I believe 
```<rng:param name="minInclusive">0</rng:param>```
should be
```<rng:param name="minExclusive">0</rng:param>```

which would allow decimal values greater than 0, for example, '.000001'. :-)

rettinghaus:
@pe-ro I was asking me the same question, but actually this is also allowed on MusicXML's side. Excluding 0 here could then lead to an incorrect or invalid conversion from MusicXML. 
See https://usermanuals.musicxml.com/MusicXML/Content/ST-MusicXML-non-negative-decimal.htm

lpugin:
> @pe-ro I was asking me the same question, but actually this is also allowed on MusicXML's side. Excluding 0 here could then lead to an incorrect or invalid conversion from MusicXML.
> See https://usermanuals.musicxml.com/MusicXML/Content/ST-MusicXML-non-negative-decimal.htm

Maybe it can be dropped with a warning during the conversion from MusicXML?

kepper:
As discussed at the ODD Friday, we prefer to use the `<rng:param name="minExclusive">0</rng:param>`, so not allow a value of zero. Conversion from MusicXML could still output a value of zero, which will result in an error message about invalid data. As the zero is actually an error in the data, that seems correct. @rettinghaus volunteers to implement this accordingly. At the same time, we will address the same issue with `@meter.uni`, which also allows a value of zero – which doesn't make sense at all…

rettinghaus:
@kepper should be ready to merge now.

ahankinson (approved):

bwbohl (approved):

PR #798, opened 2021-05-17
fixes #797

### LINKED ISSUE #797: metaMark has incorrect att nesting
`att.metaMark.log` is a member of `att.timestamp2.gestural`, which is clearly a nesting error. As a consequence, `<metaMark>` has `@tstamp2.ges` and `@tstamp2.real`, but not `@tstamp2`. Replacing `.gestural` with `.logical` will do the trick here.

https://github.com/music-encoding/music-encoding/blob/bced1582192d2b24e74d904691b581546e8b0919/source/modules/MEI.edittrans.xml#L57

kepper:
this will allow `@tstamp2` on `<metaMark>`, and will disallow `@tstamp2.ges` and `@tstamp2.real`. The two latter might be useful, but this is clearly a bug and needs to be fixed as proposed. Adding the other two attributes should be done when there's a request for this.

### DIFF
--- a/source/modules/MEI.edittrans.xml
+++ b/source/modules/MEI.edittrans.xml
@@ -54,7 +54,7 @@
     <classes>
       <memberOf key="att.controlEvent"/>
       <memberOf key="att.startEndId"/>
-      <memberOf key="att.timestamp2.gestural"/>
+      <memberOf key="att.timestamp2.logical"/>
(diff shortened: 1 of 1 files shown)


ahankinson (approved):

kepper:
Agreed to merge at this month's ODD Friday

PR #799, opened 2021-05-20
schema: create agent in MEI shared

### LINKED ISSUE #555: publisher of a work
`<work>` allows the following elements as direct children: 
* arranger 
* author 
* composer 
* contributor 
* editor 
* funder 
* librettist 
* lyricist 
* sponsor

Why then is `<publisher>` not allowed? 
IMHO this is an inconsistency. While some elements totally make sense as children of work, others don't, e.g. why is `<editor>` allowed whilst `<publisher>` isn't?

I want to compile a simple straight-forward list of works containing just title, composer and publisher. 
The only way I see now is to put the publisher in work/biblList/bibl/publisher. Way to go?

doerners:
This PR adds the agent element to the schema as discussed in #555 and removes all the specialised elements related to persons (e.g. composer, librettist, arranger etc.)

fixes #555


doerners:
converted to draft since guidelines still need to be adapted

doerners:
Since this PR is affecting the [anyStart customization](https://github.com/music-encoding/music-encoding/blob/develop/customizations/mei-all_anyStart.xml#L64), does it have to be adapted manually?

musicEnfanthen:
We started to identify sections in the guidelines that need to be reworked for this PR:

- [chapter 3](https://music-encoding.org/guidelines/dev/content/metadata.html): multiple subchapters need a rewrite including revision of examples; some affected subchapters: 
  - 3.3. (most of the listings), 
  - 3.4.1.1(listing 43), 
  - 3.4.1.3, 
  - 3.4.1.4, 
  - 3.4.1.5, 
  - 3.4.3, 
  - 3.5.1 (listing 69), 
  - 3.6.1, 
  - 3.8.3 (listing 101,102)
- clarify the use of `agent` in connection with `respStmt`, i.e. allowed inside and outside? E.g.,

```xml
<titleStmt>
    <title>Festival jubilate : an electronic transcription</title>
    <agent role="composer">
        <persName>Amy Marcy Beach</persName>
    </agent>
    <respStmt>
        <resp>Encoded by</resp>
        <agent role="encoder">
            <name>Jane Doe</name>
        </agent>
    </respStmt>
</titleStmt>
```
- NB: there is no "encoder" in the MARC Relator List


- [chapter 9](https://music-encoding.org/guidelines/dev/content/textencoding.html): 
  - 9.2.8 (names): clarify "(optional) module" (namesdates) and "dedicated elements" (`persName`, `corpName`)
  - 9.2.11 (adresses)
  - 9.2.12 (bibl references)

source/modules/MEI.shared.xml
```diff
@@ -4922,58 +4877,32 @@
...
-      </rng:zeroOrMore>
-    </content>
-  </elementSpec>
-  <elementSpec ident="contributor" module="MEI.shared">
-    <desc>Names of individuals, institutions, or organizations responsible for contributions to the
-      intellectual content of a work, where the specialized elements for authors, editors, etc. do
-      not suffice or do not apply.</desc>
+  <elementSpec ident="agent" module="MEI.shared">
+    <desc>The name of an individual or corporate body who or which is related to any other entity. 
+      The nature of the relationship should be expressed within the <att>role</att> attribute.</desc><!-- TODO: check if speaker,audience, distributor are needed-->
```
ahankinson: Fix line length to match surrounding text.

source/modules/MEI.shared.xml
```diff
@@ -4922,58 +4877,32 @@
...
-            test="not(matches(normalize-space(lower-case(@role)),               '(arranger|author|composer|contributor|editor|funder|librettist|lyricist|sponsor)'))"
-            >The value of @role must not contain the name of another element available in this
-            context.</sch:assert>
-        </sch:rule>
-      </constraint>
-    </constraintSpec>
     <attList>
-      <attDef ident="role" usage="rec">
-        <desc>Used to specify the contributor's function.</desc>
+      <attDef ident="role" usage="req">
```
ahankinson, line 5044: Very few things in MEI are actually required. I would suggest this be `rec` (it is possible to have an agent and to not know what role they played!)

source/modules/MEI.shared.xml
```diff
@@ -4922,58 +4877,32 @@
...
+        <desc>Used to specify the agent's function.</desc>
```
ahankinson: There is no documentation here to suggest where the role values come from. I noticed someone mentioned MARC21 relator codes? 

Is there any value in having a corresponding `role.uri` attribute, similar to `glyph.uri`, so that we can point to a fully qualified URI for the role? e.g., from the list here: https://id.loc.gov/vocabulary/relators.html

There is also a set of URIs for `agentType` which may be useful to differentiate between a person, organization, or software agent: 

https://id.loc.gov/vocabulary/preservation/agentType.html

MEI is generally conservative about using the `type` attribute, but in this case perhaps:

```xml
<agent type="person" type.uri="http://www.loc.gov/premis/rdf/v3/Person" role="engraver" role.uri="http://id.loc.gov/vocabulary/relators/egr" />
```

Would that make sense?

ahankinson (changes requested):
See my earlier comments about changes.

source/modules/MEI.shared.xml
musicEnfanthen: Do you mean the comment? It will be removed later anyway .

musicEnfanthen:
@ahankinson Thank you for the helpful comments. The PR is implementing the approach developed and discussed within the metadata IG and described in #555 (esp. https://github.com/music-encoding/music-encoding/issues/555#issuecomment-808063170). As it stands, the idea was to use `role` attribute as a free text description, and `auth.uri` to "point to a fully qualified URI for the role". `<agent>` can then include a `<persname>`, `<corpname>` or `<name>` to specify the "type" of agent. Do you think another type attribution on `agent` itself would help? And would you prefer `role.uri` to `auth.uri`? 

On a more general note, the PR has been reset to draft status as some issues we encountered during implementation need discussion and support from the metadata IG again (see https://github.com/music-encoding/music-encoding/issues/555#issuecomment-850404391).

ahankinson:
So you might have:

```xml
<agent role="composer" auth.uri="http://id.loc.gov/vocabulary/relators/cmp" codedval="cmp" auth="marc21-relators">
   <persName auth.uri="https://isni.org/isni/0000000108628685">
         Bartholdy, Fanny C.
   </persName>
</agent>
```

There are a lot of attributes there that don't have any obvious relationship.

Does this also mean that for every time a person's role changes in a piece (composer and lyricist, bookseller and former owner, etc.) that you need to repeat the full `agent` block with the same `persName`?

source/modules/MEI.shared.xml
doerners: My understanding was that `agent` would be an element on the highest level of abstraction. So it seemed sensible not only to point out that there is an individual or group of individuals somehow related to e.g. a work but also to specify the nature of the relation. If one doesn't know the nature of the relation it should still be possible to insert free text for the `role` attribute denoting that the relation is unknown otherwise it should be considered best practice to use a relation specified e.g. in the marc relators list. Does this make sense?
However, I think it might be a good starting point for discussion within the community, since there might are some use cases where requiring the `role` attribute might not be ideal.

pe-ro:
In my opinion, `<agent>` isn't needed.  Its intended function can be encoded using `<respStmt>`.  `<respStmt>` has the added benefit of being able to associate multiple roles with a name or even with multiple names.  It would be unorthodox compared with traditional cataloging practice, but `<respStmt>` can be used as a replacement for the dreaded role-based elements of `<author>`, `<composer>`, etc.

bwbohl:
> In my opinion, `<agent>` isn't needed. Its intended function can be encoded using `<respStmt>`. `<respStmt>` has the added benefit of being able to associate multiple roles with a name or even with multiple names. It would be unorthodox compared with traditional cataloguing practice, but `<respStmt>` can be used as a replacement for the dreaded role-based elements of `<author>`, `<composer>`, etc.

just to add the corresponding XML example to @ahankinsons last encoding of `<agent>`with `<respStmt>`:

```xml
<respStmt xmlns="http://www.music-encoding.org/ns/mei" meiversion="4.0.1">
    <resp auth.uri="http://id.loc.gov/vocabulary/relators/cmp" codedval="cmp" auth="marc21-relators"/>
    <persName role="composer" auth.uri="https://isni.org/isni/0000000108628685">
        Bartholdy, Fanny C.
    </persName>
</respStmt>
```

And thank you @pe-ro for your conservative ;-) assessment of the proposal.

The example above is:
* *invalid* MEI, as `@meiversion`is not allowed on `<respStmt>`, I just added it for documentary purposes.
* *moved* `@role` from `<agent>` to `<persName>`( although I don't deem it adequate, but that'S another story ;-)
* *moved* all attributes from `<agent>`to `<resp>`

Summing up, the question we are dealing with is basically about sth. one could call *semantic nesting* (i.e. giving 'proper' names to the nodes and ensuring the semantics reflect proper attribution).
In this context I'd like to add some questions and maybe answers:

* Is there a problem with the semantics of the above `<respStmt>`example?
** the only thing I see, is my dislike of assigning a `@role` to a `<persName>`, because actually the role is being assigned to the person but I guess that's another story…

* as this issue arose from problems of encoding responsibilities in the context of `<work>` (cf. #555), I'd like to remind that the initial problem could be drilled down to the **unavailability of `<respStmt>` in `<work>` since MEI v4**. 

* What is the difference of the above `<agent>` and `<respStmt>`?

  1. The term 'agent' semantically states there is some entity taking action;
    
    * While the attribute `@role` semantically fits into this context – to put in in words: the entity described has a specified role in the context describe, e.g.
      * **agent:** *someone* is the composer of Festival Jubilate 
      * **agent content** *someone* has `persName` Amy Marcy Beach
    * the other attributes, though (`@auth.uri`, `@codedval`, and `@auth`) are used in the examples given throughout the discussion to describe the role, **not** the agent! Either they should be prefixed (as @ahankinson) pointed out, e.g. `@role.auth` or they sould be moved to their own element, which would bring us back to `<respStmt>`
    
  2. 'responsibility statement' is defined in the MEI Guidelines v4 as 
  
    >Transcription of text that names one or more individuals, groups, or in rare cases, mechanical processes, responsible for creation, realization, production, funding, or distribution of the intellectual or artistic content
    
    I do see two problems here:
    
    1. 'agent' is underspecified / under-defined, i.e. **TODO** we should come up with a clear definition of what 'agent' should capture
    2. 'respStmt' – from what I learned from the past discussion – might be overspecified as follows:
        * It's deemed to be a **transcription** this from my point of view might conflict with e.g. using normalized names in your encoding and might be a turndown for e.g. cataloguing work or linked data-driven encoding (which on the other side would not rely on the text content but the identifier of the entities.
        * If we take the description seriously it's limited to encoding the following responsibilities:
          >creation, realization, production, funding, or distribution
       
       I guess these two restrictions come from what's behind the "statement" part of `respStmt`, i.e. *statements* being something discovered in the original source's contents vs. *descriptions* capturing something the editor or encoder adds e.g. decoration description, physical description. But this might not hold true for all ???Desc-elements, e.g. figure description?
       
       **Breakball?:** If we do not want to change this definition of `respStmt` we do need something else for all the non-transcriptional cases.
       
        **nah…** But wouldn't it be strange to use different elements for encoding the composer of let's say a score giving his name and a sketch that does not give his name but can be clearly identified as being written by him by means of handwriting analysis? I think so.
        
        Moreover a closed definition of `respStmt`as described above would have to be changed for the context of administrative metadata such as `fileDesc`.
        
        So if we change it to a more open definition, e.g.:
        
        >~~Transcription of text that names~~**Captures one or more responsibilities of** one or more individuals, groups, or ~~in rare cases~~**even** mechanical processes, responsible for **e.g.** creation, realization, production, funding, or distribution of the intellectual or artistic content **described by its context**
        
        If we (re-)allow `respStmt` at all relevant positions, e.g. those where the specialised elements are allowed exclusively, paired with the above modifications in its definition, wouldn't all needs be satisfied I guess? Or is there something I'm missing?

ahankinson:
How does:

`<resp auth.uri="http://id.loc.gov/vocabulary/relators/cmp" codedval="cmp" auth="marc21-relators"/>`

and

`<persName role="composer"  auth.uri="https://isni.org/isni/0000000108628685">`

align? It seems that there are two places to record the same information ("role/responsibility")?

pe-ro:
Both

`<resp auth.uri="http://id.loc.gov/vocabulary/relators/cmp" codedval="cmp" auth="marc21-relators"/>`

and 

`<persName role="composer" auth.uri="https://isni.org/isni/0000000108628685">`

allow for the recording of the same information WRT to role/responsibility.  However, the `@role` attribute should be used in contexts outside of `<respStmt>`, for example, in descriptive prose where `<respStmt>` isn't permitted.  It shouldn't be used when `<persName>` is a child of `<respStmt>`.  A Schematron rule could be used to guide users toward proper usage of `@role`.

ahankinson:
I guess I was asking which one took precedence. If you had:

```xml
<resp auth.uri="http://id.loc.gov/vocabulary/relators/cmp" codedval="cmp" auth="marc21-relators"/>
<persName role="lyricist" auth.uri="https://isni.org/isni/0000000108628685">
...
```

that probably shouldn't be allowed.

pe-ro:
Precisely.  That's why I suggested using Schematron to disallow (or at least warn against) the use of `persName/@role` when `<persName>` is a child of `<respStmt>`.  This would force the encoder to use the attributes of `<resp>`.

rettinghaus (dismissed):

doerners:
Changing this PR from draft status to review status related to the last two comments ([1](https://github.com/music-encoding/music-encoding/issues/555#issuecomment-1693802313) and [2](https://github.com/music-encoding/music-encoding/issues/555#issuecomment-1694513365)) @pe-ro added in #555.

musicEnfanthen:
Thanks for re-preparing this PR @doerners ! 

Please be aware that we will not merge any breaking changes (which includes deletion of existing elements) before the next development workshop (about autumn 2024). At least, we'll try to avoid it. But this should give sufficient time to discuss this PR in preparation for MEI 6.

pe-ro:
Many thanks to @doerners for pursing the issue, but I don't support the PR because it doesn't go far enough in some respects *and* goes too far in others.  That being said, I agree that changes can be made to clarify the recording of responsibility in the bibliographic context.  To that end I suggest the following --

- modify the content model of `<titleStmt>` to permit only `<editor>` and `<creator>` (a new element in the vein of `<agent>`) .  These are "primary responsibilities" that traditional cataloging privileges by recording them alongside a bibliographic entity's title.  N.B. If this privileging isn't wanted, then creator and/or editor responsibilities can be deferred to `<respStmt>`.
- remove `<arranger>`, `<author>`, `<composer>`, `<contributor>`, `<funder>`, `<librettist>`, `<lyricist>`, and `<sponsor>` from model.respLikePart. This is how the model of `<titleStmt>` gets modified.
- delete `<author>` and `<composer>` everywhere.  These are replaced by `<creator>`.
- retain `<arranger>`, `<contributor>`, `<funder>`, `<librettist>`, `<lyricist>`, and `<sponsor>` for use in `<work>` and `<expression>`, which are less "bibliographic-record-like" and more "database-like" than bibliographic descriptions elsewhere (e.g., in `<fileDesc>`, `<bibl>`, `<biblStruct>`).
- revise model of `<respStmt>` to be more "TEI-like"; that is, permit one or more `<resp>` elements followed by one or more `<persName>`, `<corpName>`, or `<name>` elements - OR- one or more `<persName>`, `<corpName>`, or `<name>` elements followed by one or more `<resp>` elements.  This will disallow the free intermixing of names and responsibilities MEI has allowed up till now. While I suspect there hasn't be much abuse of the current situation yet, I'd like to eliminate the possibility while we can.

As illustrations I offer the following examples.  The first uses a traditional cataloging approach (useful when transcoding data from MARCXML, MODS, etc.) to capture the primary responsibility in `<creator>`.  "Secondary responsibilities" are captured in `<respStmt>`.  Note that `@auth.uri` on `<persName>` refers to a record for Fanny Mendelssohn, while the same attribute on `<resp>` refers to definition of "arranger".  When `@auth.uri` is used on `<creator>`, `@role` on `<persName>` is redundant, but `@role` might still be useful as an expansion of the URI on `<creator>`.  Vice versa, if `@auth,uri` is *not* present on `<creator>`, `@role` on `<persName>` is a human-readable substitute.
```xml
<titleStmt>
  <title>Confusion in C major</title>
  <creator auth.uri="http://id.loc.gov/vocabulary/relators/cmp">
    <persName role="composer" auth.uri="https://isni.org/isni/0000000108628685">
      Bartholdy, Fanny C.</persName>
  </creator>
  <respStmt>
    <resp auth="MARC relators" auth.uri="http://id.loc.gov/vocabulary/relators/arr"
      >arranger</resp>
    <persName>Willy Make It</persName>
  </respStmt>
</titleStmt>
```
The second example eschews traditional cataloging and puts all responsibilities in `<respStmt>`.  It also demonstrates how `@auth.uri` can be used on `<resp>` to standardize the myriad of responsibility statement formulations transcribed from books, scores, etc.
```xml
<titleStmt>
  <title>Confusion in C major</title>
  <respStmt>
    <resp auth.uri="http://id.loc.gov/vocabulary/relators/cmp" codedval="cmp"
      auth="marc21-relators">composer</resp>
    <persName role="composer" auth.uri="https://isni.org/isni/0000000108628685">
      Bartholdy, Fanny C. </persName>
    <resp auth="MARC relators" auth.uri="http://id.loc.gov/vocabulary/relators/arr"
      >Arranged for nose flute and kazoo by</resp>
    <persName>Willy Make It</persName>
  </respStmt>
</titleStmt>
```
My own PR to follow in the coming days.

pe-ro:
By the way, making the changes to model.respLikePart described above provides beneficial changes in `<analytic>`, `<byline>`, `<expression>`, `<seriesStmt>`, and `<titlePage>` (via titlePagePart).

PR #808, opened 2021-05-31
Fixed: Update PR labeller

### LINKED ISSUE #766: PR labeled fails on PRs from fork

ahankinson:
Updates the PR labeller to run on the PR target, rather than the pull request, to get around a permissions issue running on a fork.

Fixes #766

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -13,13 +13,11 @@ on:
       - edited
       - reopened
       - synchronize
-#  workflow_dispatch:
 
 jobs:
   label:
-
     runs-on: ubuntu-latest
-
+    if: github.repository_owner == 'music-encoding'
     steps:
     - uses: actions/labeler@v2
       with:
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -7,7 +7,7 @@
 
 name: PR Labeler
 on: 
-  pull_request:
+  pull_request_target:
     types:
       - opened
       - edited
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.repository_owner == 'music-encoding'
     steps:
-    - uses: actions/labeler@v2
+    - uses: actions/labeler@v3
       with:
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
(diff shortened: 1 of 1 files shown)


ahankinson:
Tagging @musicEnfanthen as well

musicEnfanthen:
Thanks @ahankinson for the PR. Looks good. Indeed, this labelling action seems to be one of the rare use cases where the official recommendation is to use pull_request_target.

BTW: Do we want to run the GH action only on music-encoding or also on all forks (i.e. all the forks get the labels, too). If only the former then we might want to include something like

`if: github.repository_owner == 'music-encoding'`

at line [22](https://github.com/music-encoding/music-encoding/blob/6f9ba917117b0757520c100f831f9796bebf525a/.github/workflows/pr-labeler.yml#L22).

ahankinson:
As far as I can tell, this will only run on forks if someone is targeting a pull request to that fork... that's what is meant by the 'target'. So I think the `github.repository_owner == 'music-encoding'` is implied, unless someone is maintaining a parallel MEI repo.

musicEnfanthen:
Can we exclude that some forks are working with Pull Requests to their own repos at their covenience? Or to put it the other way round: inserting the if clause would ensure that the gh action is never ever executed outside the organisational main repo. Cf. https://github.com/NixOS/nixpkgs/pull/113077/files#r575748234

(Actually, that is even more urgent for the the [deploy.yml](https://github.com/music-encoding/music-encoding/blob/develop/.github/workflows/deploy.yml))

ahankinson:
Added; let me know if it looks ok?

.github/workflows/pr-labeler.yml
```diff
@@ -7,19 +7,17 @@
...
-#  workflow_dispatch:
 
 jobs:
   label:
-
     runs-on: ubuntu-latest
-
+    if: github.repository_owner == 'music-encoding'
     steps:
     - uses: actions/labeler@v2
```
musicEnfanthen:
```suggestion
    - uses: actions/labeler@v3
```

musicEnfanthen (approved):

musicEnfanthen:
Yes, thank you! (Suggestion to update labeler version to v3)

bwbohl:
dear both, thanks for working on this and spotting the problem esp. with the deploy action! Could you submit a PR for that one, too?

bwbohl (approved):

bwbohl:
this ready to merge?

ahankinson:
Good from me.

PR #815, opened 2021-06-15
Add accid to neume components

yinanazhou:
Hi, we added both `accid` attribute and element to neume components. We have built the schema locally, and the validation succeeded. Please have a look at it. Thank you!

### DIFF
--- a/source/modules/MEI.analytical.xml
+++ b/source/modules/MEI.analytical.xml
@@ -329,6 +329,8 @@
   <classSpec ident="att.nc.anl" module="MEI.analytical" type="atts">
     <desc>Analytical domain attributes.</desc>
     <classes>
+      <memberOf key="att.accidental"/>
+      <memberOf key="att.articulation"/>
       <memberOf key="att.harmonicFunction"/>
       <memberOf key="att.intervalMelodic"/>
       <memberOf key="att.melodicFunction"/>
--- a/source/modules/MEI.analytical.xml
+++ b/source/modules/MEI.analytical.xml
@@ -330,7 +330,6 @@
     <desc>Analytical domain attributes.</desc>
     <classes>
       <memberOf key="att.accidental"/>
-      <memberOf key="att.articulation"/>
       <memberOf key="att.harmonicFunction"/>
       <memberOf key="att.intervalMelodic"/>
       <memberOf key="att.melodicFunction"/>
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -4164,6 +4164,7 @@
       <memberOf key="att.accid.ges"/>
       <memberOf key="att.accid.anl"/>
       <memberOf key="model.noteModifierLike"/>
+      <memberOf key="model.neumeComponentModifierLike"/>
     </classes>
     <content>
       <rng:empty/>


source/modules/MEI.analytical.xml
```diff
@@ -329,6 +329,7 @@
   <classSpec ident="att.nc.anl" module="MEI.analytical" type="atts">
     <desc>Analytical domain attributes.</desc>
     <classes>
```
ahankinson, line 331: An accidental is not part of the analytical domain.

source/modules/MEI.shared.xml
```diff
@@ -4164,6 +4164,7 @@
       <memberOf key="att.accid.ges"/>
       <memberOf key="att.accid.anl"/>
       <memberOf key="model.noteModifierLike"/>
```
ahankinson, line 4166: This is added in ‘shared’ which is not correct.

ahankinson:
Hi, thanks for your contribution! However, I think more work is needed for this change. There are several technical problems that I noted above. More importantly, I think it would be useful to have this looked over by the neume notation SIG prior to being submitted.

If this is a bug in the schema, please open an issue first describing the problem.

I am going to propose closing this PR, but this does not mean the change is rejected — it just means the current form and method of submitting it does not adequately describe the change, why it is needed, and whether others in the community agree it should be changed.

yinanazhou:
Hi @ahankinson, thanks for your reply. We need accidentals in neume notation because the manuscript _Salzinnes_ , which we are currently processing on Rodan, has flats. So we need to add accidental to neume component to make the MEI file valid.

source/modules/MEI.analytical.xml
martha-thomae: Hi @ahankinson. I was looking into the documentation for the `<note>` element (the equivalent to the `<nc>` element in CMN). It seems that the attribute `@accid` is coming from the [analytical domain (`att.note.anl`)](https://music-encoding.org/guidelines/dev/attribute-classes/att.note.anl.html). Is this correct? 

I recommended adding this attribute as part of the analytical domain because that seemed to be the case for notes (`att.note.anl` gets it from `att.accidental`; here `att.nc.anl` gets it from that same class).

source/modules/MEI.shared.xml
yinanazhou: hi, I'm raising an issue about the accidental requirements. But where is the right place to add that if not in the shared module?

source/modules/MEI.analytical.xml
ahankinson: OK; that makes a certain amount of sense for consistency. Still, it might make sense to use `att.accidental` directly on the element that requires it, instead of going through the `.anl` class. Let's see what others say.

source/modules/MEI.shared.xml
ahankinson: As I understand it you want `<nc>` to allow `<accid>` as a child? Then you would add a model class to the 'content' block of `nc` that contains the elements that you want. You would probably have to create your own model class -- see [`model.noteModifierLike`](https://music-encoding.org/guidelines/v4/model-classes/model.notemodifierlike.html) as an example.

If you add it directly to shared it creates an upward dependency on the 'neume' module -- generally the specific modules should always use and refer to things in shared, but not the other way around.

ahankinson:
OK. I still propose closing this and creating an issue for further discussion. There are significant changes that need to be made to this PR anyway, so when we finally land on a solution please feel free to open a new PR.

If you agree, please close it and we can continue the discussion in your issue. If not, I would like to wait to see what another reviewer thinks.

PR #831, opened 2021-06-23
Fix neumes image path

yinanazhou:
Corrected neumes image path. [Notice](https://github.com/music-encoding/music-encoding/discussions/829): The images will only show up when using **command lines** to build HTML files. To make the **Oxygen** method work, the image folder needs to be pasted to `music-encoding\source\web` manually.


ahankinson:
There are a lot of changes on this PR -- things like github actions PR labeller, etc.  It should not be merged without removing any unnecessary changes.

PR #837, opened 2021-07-12
Windows support for build.xml

annplaksin:
This PR solves the failing build process on Windows machines due to an incompatible file path to the defaultSource. I hope it still works on non-Windows machines. Please test!
(... and don't mind the rather clumsy implementation...)

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -28,6 +28,21 @@
         <pathelement path="${basedir}/lib/saxon/saxon9he.jar"/>
     </path>
     <property name="java.class.path" value="" classpathref="mei.classpath"/>
+    <loadresource property="windir">
+        <string value="${basedir}"/>
+        <filterchain>
+            <tokenfilter>
+                <replaceregex pattern="\\" replace="/"/>
+            </tokenfilter>
+        </filterchain>
+    </loadresource>
+    <condition property="canonicalizedSourcePath"
+        value="${dir.dist.schemata}/mei-source_canonicalized.xml"
+        else="file:/${windir}/${dir.dist}/schemata/${version}/mei-source_canonicalized.xml">
+        <not>
+            <os family="windows" />
+        </not>
+    </condition>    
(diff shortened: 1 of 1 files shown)


ahankinson:
Maybe you can explain a bit more how it works? I see some regex that seems to replace forward slashes for part of the path, but it doesn't replace them for all of the path?

That is, the value of `windir` seems to have its slashes replaced, but then it's joined in with others in the `canonicalizedSourcePath` value, which doesn't replace them?

rettinghaus:
Building the schema still seems to work on Mac.

annplaksin:
@ahankinson The problem on a Windows machine is reading the defaultSource in the odd2odd XSLT in the Stylesheets. It applies the same fix on the build file that has been explained in the instructions on how to build and RNG with Oxygen. To ensure the proper resolving of the path, you need to add the file protocol and turn the backslashes you get from `${basedir}` into slashes and add the file protocol. And I added the `canonicalizedSourcePath` since this problem only occurs when reading the file. Everything else works fine...
The two step solution is indeed clumsy, as mentioned. If someone has the skill and patience to make it more elegant, I'd be happy.

ahankinson:
Aha! gotcha. I understand now. Thank you.

Do we have someone else with Windows who can test this?

bwbohl:
First: many thanks for spotting and fixing the @annplaksin
Second: any chance you find some time to also apply the changes to the build-guidelines-html target?

Many thanks in advance!

annplaksin:
@bwbohl I am just checking the build-guidelines-html target. I'm not quite sure if the fix needs to be applied there too since it provokes an "Invalid relative URI" error. But while testing that, I get an "Output is a directory, but input is not" on my Windows and my Linux Subsystem. Maybe something is strange with the `${customization.path}`?
I don't know why as it seemed to run when I was creating this PR. Does it run on your machine?

PR #838, opened 2021-07-15
add enclose attribute to arpeg

rettinghaus:
This PR is a follow up to #729 and adds `@enclose` to `arpeg`, what was missed originally.


kepper (approved):
This seems pretty reasonable, thank you @rettinghaus

kepper:
as this has relatively minor consequences and is certainly solid, we're merging it despite the short time after review…

kepper:
thanks again ;-)

PR #839, opened 2021-07-15
Allow clef changes within fingered tremolos

rettinghaus:
This PR adds the possibility to have clef changes within tremolos. 

Not a common thing, but it could appear.
See https://github.com/rism-digital/verovio/issues/1804.

### DIFF
--- a/source/modules/MEI.cmn.xml
+++ b/source/modules/MEI.cmn.xml
@@ -1581,6 +1581,9 @@
       <rng:choice>
         <rng:group>
           <rng:ref name="chord"/>
+          <rng:optional>
+            <rng:ref name="clef"/>
+          </rng:optional>
           <rng:choice>
             <rng:ref name="chord"/>
             <rng:ref name="note"/>
@@ -1588,6 +1591,9 @@
         </rng:group>
         <rng:group>
           <rng:ref name="note"/>
+          <rng:optional>
+            <rng:ref name="clef"/>
+          </rng:optional>
           <rng:choice>
             <rng:ref name="chord"/>
             <rng:ref name="note"/>


kepper:
What is proposed is solid and should definitely be merged. However, let's think a bit more about other things that could go in between the components of an fTrem. If that is the case, we should have a more generic approach. The task now is to come up wiht examples for that…

ahankinson:
Does a clef change in a tremolo affect anything that comes after the tremolo?

kepper:
like any other clef, I would expect it to, yes…

ahankinson:
Do you need to have another clef *after* the tremolo to reset the clef change, or is the clef only local to the tremolo?

In other words: In the example given, a note that comes *after* the second "C" outside of the tremolo -- is it in bass clef, or treble clef?

rettinghaus:
For resetting you would need another clef after the tremolo, i.e. following notes are in treble clef.

ahankinson:
Ok, so the clef isn’t actually “in” the tremolo, it’s still on the staff?

rettinghaus:
@ahankinson fingered tremolos are a special form of beams, and like in beams a clef change affects all following notes within the same layer: 

```xml
<layer>
    <clef shape="F" line="4" />
    <note />
    <note />
    <beam>
        <note />
        <note />
        <clef shape="G" line="2" />
        <note />
        <note />
    </beam>
    <!-- still treble clef here -->
    <note />
    <note />
</layer>
```

bwbohl:
2022-02-24 ODD meeting: as soon as this is reviewed, we can merge it

lpugin (approved):
Even though this is probably quite rare, I do not see any reasons not to support it.

FWIW, I checked with Dorico and it perfectly supports it:

image

musicEnfanthen:
Thank you @lpugin for the review and @rettinghaus for preparing the PR

PR #842, opened 2021-07-23
Creating att classes for direct children at perfRes(List)

riedde:
- I moved the direct childs `@solo` and `@count` to attribute classes.
- To the class att.perfRes where `@solo`moved to, I added a new attribute `@adLib`. This new attribute was discussed in the metadata IG and on the metadata/cataloging mailing list. A consensus for this extension was found.
- Since transposing instruments can only be captured by text labels, I have added the att.transpostion to `<perfRes>` to give the possibility to encode transposing performance resources in a slightly more machine readable way


source/modules/MEI.header.xml
```diff
@@ -107,6 +107,35 @@
       </attDef>
     </attList>
   </classSpec>
+  <classSpec ident="att.perfRes" module="MEI.header" type="atts">
+    <desc>Attributes that define the characteristics and components of the performance resouce.</desc>
```
ahankinson: Typo: resouce

source/modules/MEI.header.xml
```diff
@@ -107,6 +107,35 @@
...
+    <attList>
+      <attDef ident="solo" usage="opt">
+        <desc>Marks this instrument or vocal part as a soloist. Do not use this attribute for a solo
```
ahankinson: > Do not use this attribute for a solo instrument which is not accompanied.

It would be helpful to indicate why they should not use it, and what they should use instead. (I understand this is from the original definition, but if we're doing renovations here it would be helpful to improve this so we don't have to wait a few more years for someone to fix it.)

source/modules/MEI.header.xml
```diff
@@ -107,6 +107,35 @@
...
+          instrument which is not accompanied.</desc>
+        <datatype>
+          <rng:ref name="data.BOOLEAN"/>
+        </datatype>
+      </attDef>
+      <attDef ident="adLib" usage="opt">
```
ahankinson: Was the more straightforward `optional="true|false"` considered? If so, why was `adLib` favoured instead?

source/modules/MEI.header.xml
```diff
@@ -1890,6 +1919,9 @@
       <memberOf key="att.bibl"/>
       <memberOf key="att.edit"/>
       <memberOf key="att.lang"/>
+      <memberOf key="att.perfRes"/>
+      <memberOf key="att.perfRes.count"/>
```
ahankinson: In reading the comments from Perry in the IG mailing list, he suggested that `adLib` be added to `<perfResList>` to support an optional group of instruments. Additionally, it seems like you can have multi-instrumental 'solos' (I looked for 'concertos for solo percussion' and found some cases where one player performs on multiple independent instruments), in which case it might be useful to allow `@solo` on perfResList as well. 

So maybe all the attributes on perfRes should *also* be supported on `perfResList`?

In looking at the remarks on `perfResList` (to see if this is a good idea) it says "The function of instrumentalists or vocalists is represented by the choice of perfRes and perfResList child elements. Arrangements are coded for the medium of the work being described, not for the original medium." 

So it's not immediately clear to me where you would use:

```
<perfResList>
   <perfRes>...
   <perfRes>...
</perfResList>
```

and when you would use:

```
<perfRes>
   <perfRes> ... </perfRes>
   <perfRes> ... </perfRes>
   <perfRes> ... </perfRes>
</perfRes>
```

Perhaps in clarifying this application, we could determine whether to allow the same attributes on both?

Ultimately, this is motivated by the observation that you're only breaking out the `.count` attribute group so you can support `@count` on both, but not `@adLib` or `@solo`; allowing the same attributes on both (subject to it actually be useful, of course) would simplify this since we would just need the `att.perfRes` group.

ahankinson (changes requested):
If more machine-readable attributes are being added to `@perfRes` (the `att.transposition` group) then it would be very helpful to clarify in what circumstances users should favour using this over similar mechanisms in `layerDef`, `scoreDef` or `staffDef`, all of which are members of the same class.

What I think we want to avoid is to have to support yet another place that looks like it's defining instruments for machine readability, since it will cause some confusion in the tool ecosystem: "I defined an instrument as a clarinet in Bb, but Verovio isn't displaying it correctly." Or "Why does the Sibelius plugin not use `perfRes` to describe the transposing instruments in my score?"

Clarifying this can be as simple as an application note or remark in the definition of the element that clarifies why you would use `perfRes` over the alternatives. For example, would it be useful to say this is a 'descriptive' rather than 'prescriptive' element (and explain what that means)?

source/modules/MEI.header.xml
riedde: I was trying to do minimal inversive changes but you are right. I was also struggling at this point.
```suggestion
        <desc>Marks this instrument or vocal part as a soloist. Use this attribute only to distinguish an instrument from others as a solo part. If an instrument is unaccompanied, it is implicitly clear that they are playing solo.
```

source/modules/MEI.header.xml
riedde: I'm sorry but I think I do not understand your question. Instruments can be a solo part or not (true|false) and also optional (ad libitum). So an instrument can be ad libitum or not (true|false).

source/modules/MEI.header.xml
riedde: Supporting this attribute classes in perfRes and perfResList would be fine for me. If then just one att.perfRes class exists I'm also fine with (you exactly described the problem I had).
The fact that it is unclear when using perfResList and perfRes nested is why I'm working on that. At the metadata group we are discussing at the moment if we can solve this by best practice suggestions or/and a better description in the guidelines.

riedde:
My perspective is e.g. a work catalogue which might be represented only by a meiHead. So if there is no music section or no body you do not have layerDef, scoreDef or staffDef where you can define that.
I would suggest that a remark on att.transposition could be a little hep, but also in the guidelines (perfMedium section) there should be a description for usage. (next step ;))

source/modules/MEI.header.xml
ahankinson: My question is: Why use `adLib` and not `optional` as the attribute name? 

It seems like every time you say what it is, you have to also explain that "ad libitum" means "optional":

"Marks this instrument or vocal part as optional (ad libitum)"
"...and also optional (ad libitum)..."

To me, this means that the word "optional" is actually more understandable in its function than the Latin form.

So I was wondering whether you considered naming it just `@optional` (with the same true or false values), and if so, why `@adLib` was chosen instead.

ahankinson:
> My perspective is e.g. a work catalogue which might be represented only by a meiHead. So if there is no music section or no body you do not have layerDef, scoreDef or staffDef where you can define that.

That's fine, but the opposite is not true: You *can* use `<perfRes>` in an MEI file that contains a `<music>` section, and it's in those instances when encoders will most likely get confused about where and when you should favour encoding an instrument definition using `<perfRes>` or any of the other choices.

A clear remark on the `<perfRes>` element definition that explains its purpose, and where you *should* use it, would probably help a lot of people decide how and when to use it.

source/modules/MEI.header.xml
ahankinson: >  If an instrument is unaccompanied, it is implicitly clear that they are playing solo.

Explicit is always better than implicit, particularly when it comes to encoding. Something more explicit in the documentation about it's positive use is probably more helpful than describing the cases when you *shouldn't* use it, e.g., 

"Use this attribute to identify the instrument(s) used by a soloist in an accompanied work, such as a concerto or vocal solo."

ahankinson: I'm also not sure it's *so* bad to mark an instrument as solo if it's the only one. Looking at a few pieces of sheet music, they generally say as much: "Music for solo saxophone" or "Solo percussion". Marking something as solo when there are no other instruments is redundant, but not particularly awful. What's more important is that there is a way to mark a solo instrument amongst other instruments.

rettinghaus:
@ahankinson The transposition information in the header and the body may not necessarily match, e.g. the composer may demand a _Cor anglais_, but in the encoded source this part is notated in C or not fully written out, because it is only a particell.

source/modules/MEI.header.xml
riedde: > "Use this attribute to identify the instrument(s) used by a soloist in an accompanied work, such as a concerto or vocal solo.
I'm also fine with this.

source/modules/MEI.header.xml
riedde: I used adLib because ad libitum is the technical term. In the description I used optional because in my opinion it is not the best way to describe a word using the word itsself.

riedde:
Another possibility could be to find a place for describing the instruments (some kind of instrDef). So the perfRes could be linked with that definition. But I think this would be a further step that may not be discussed and solved now.

gucl-mu:
I don't know if it quite fits, but here is a possible addition to the `perfRes`-definition: "Name of an instrument on which a performer plays, a performer's voice range, or a standard performing ensemble designation. `perRes` is intended to define the instrument/voice/performing_ensemble specification of a work in its ideal form or its different performance scenarios, but not the score or the layout of a music edition. E.g. a work may be encoded with horns in F in the MEI-header, but the editorial realisation is with horns in C." (Perhaps with a better example)
Actually, it would then probably be better to place a definition in `scoreDef` etc. as well. But that would be too big a change...

Perhaps the following could be helpful for the attributes: "For correct rendering and machine readability, use `@trans.diat` etc. in `scoreDef` and its childelements. For stand-alone headers, the information about singers and instrumentalists can also be stored in `perfRes`." 

I suggest that we discuss the exact wording of the definitions again in the next IG meeting.

riedde:
@gucl-mu yes, we should discuss this in august at the next metadata-ig meeting. But for now: I think it is better to keep the definition quite short and to give an extended example in the guidelines.

pe-ro:
Nested `<perfRes>` elements should be used for doublings.  `<perfResList>` marks a set of performance resources, each of which may or may not include doublings.  IMO, other than `@adlib`, the same attributes should not exist on `<perfRes>` and `<perfResList>`.  This should help guide users to the proper usage of `<perfRes>` and `<perfResList>`.

riedde:
> Nested `<perfRes>` elements should be used for doublings. `<perfResList>` marks a set of performance resources, each of which may or may not include doublings.

I didn't know this before, maybe we should write this down somewhere. @pe-ro could you give a very short example for a doubling?

bwbohl:
A short note on doubling is in included in https://music-encoding.org/guidelines/dev/content/metadata.html#headerWorkInstrumentation

bwbohl:
Could you please add additional explanations to the guidelines for the proposed changes

bwbohl:
And @riedde are there any new result from the last metadata IG meetings since you started implementing this?

riedde:
@bwbohl Thanks for the link! I will add the explanations to the guidelines before this pull request is ready to merge.
I just need a few weeks because the last metadata IG meeting was postponed. I'll come back to this soon!

source/modules/MEI.header.xml
ahankinson: I still think there is probably some problems with using `adLib` in the context of an instrument, since it could also mean 'improvised' or 'free' -- As much as I don't like just using dictionaries to argue the point, here is a screenshot of the dictionary entry for `ad lib`:

Screenshot 2021-11-11 at 11 45 10

So the question is whether the ad lib instrument is 'optional' (as in, the piece can be performed without it) or 'improvised' (the instrument is mandatory, but the player can improvise their own line).

riedde: I got your objection. But in my opinion it is still clear. At this point using `@solo` means that a soloist is needed for the performance and not that the part/voice contains solos. So why should `@adLib` mean here the way of performing and not (as it is with `@solo`) a description of the performance resource? Just a though. I am still in favor of using the technical term.

rettinghaus: I agree with @riedde

ahankinson: I think I was commenting on the fact that 'adlib' means "without previous preparation" instead of "optional". Surely an ensemble wouldn't have an 'adlib' instrument that shows up to play unannounced and without rehearsal. :-D

rettinghaus (dismissed):

rettinghaus:
Looks fine to me. Thanks!

kepper:
We discussed this extensively during today's ODD Friday. In general, this seems fine. However, there are multiple musical meanings of the words "ad libitum" in music (see [Wikipedia](https://en.wikipedia.org/wiki/Ad_libitum#Music_or_rhythm). With using `@adLib` here (or lower-case?), we should be clear about the use of the same attribute for those other purposes in the future, and how this might cause conflicts. We agree that the attribute should be in a separate attribute class to support it's use for those other cases, but then the question is if there can be a common description of this attribute for all these use cases. An alternative could be to use the word `@optional`,  which seems to make an identical statement _in this case_. In any case, the use of the attribute (no matter it's name) should be documented in a `<remark>` on `<perfRes>`. We would like the @music-encoding/ig-metadata to revise these arguments and bring this back up here.  Thanks @riedde and everyone else involved for bringing this forward 👍

source/modules/MEI.header.xml
```diff
@@ -107,6 +107,29 @@
       </attDef>
     </attList>
   </classSpec>
+  <classSpec ident="att.perfRes" module="MEI.header" type="atts">
```
bwbohl, line 110: Are you all sure about merging bot attributes of `perfRes` and `perfResList` into one attribute class that is being made available on both elements? This means `@adLib` and `@solo` will also be available o `perfResList`. I guess you spent some thought on this but could you please enlighten me?

bwbohl: Don't get me wrong, I like the idea to have a single definition of `@count` and reuse it wherever possible ;-)

riedde: The discussion about this starts [obove](https://github.com/music-encoding/music-encoding/pull/842#discussion_r675477539).
Thanks for your advice. When I review this now I'm not sure if it is right if `<perfResList>` can have `@solo`. This problem arises when we joined the attribute classes I modeled at first.
-> In the las ODD meeting we decided to rename `@adLib`and to move it to an own attribute class. So I'm going to think about this when I do the changes.

musicEnfanthen:
@riedde What is the status of this PR since your latest changes? Can it be reviewed again to be prepared for the next ODD meeting? Or are there still any open issues?

riedde:
Thanks @musicEnfanthen for the reminder. I noticed, that I answered the review comments but didn't click "resolve conversation". I'm going through in the evening and fix that. I also try to review the documentation about this.
It will be ready for the next ODD meeting!

riedde:
@musicEnfanthen I have resolved the open conversations and also updated the documentation and the code example (both contained an outdated naming). On my local machine I compiled the schema and the guidelines and as I can see everything works as I expect it. You can now review it if you like.

musicEnfanthen:
@ahankinson Since you requested changes for this PR, may I ask you to have a look again if all issues are adressed now?

musicEnfanthen:
@ahankinson May I ask you to have another look if all concerns are adressed now, or if this PR would need some more discussion?

source/modules/MEI.header.xml
```diff
@@ -107,6 +107,62 @@
...
+    <attList>
+      <attDef ident="adlib" usage="opt">
+        <desc>Marks an entity as not strictly defined.</desc>
+        <datatype>
+          <rng:ref name="data.BOOLEAN"/>
+        </datatype>
+      </attDef>
+    </attList>
+    <remarks>
+      <p>The technical term “ad libitum” has several meanings depending on the context in which it occurs. It could</p>
```
ahankinson: I think there is an unfinished sentence here.

source/modules/MEI.header.xml
```diff
@@ -107,6 +107,62 @@
...
+      <p>
+        <list type="gloss">
+          <head>Meanings of ad libitum</head>
+          <label>1</label>
+          <item>indicate an optional instrumental or vocal part or group or</item>
```
ahankinson:
```suggestion
          <item>indicate an optional instrumental or vocal part or group; or</item>
```

source/modules/MEI.header.xml
```diff
@@ -107,6 +107,62 @@
...
+          <label>2</label>
+          <item>marking a passage to be played freely or free in time or</item>
```
ahankinson:
```suggestion
          <item>mark a passage to be played freely or free in time; or</item>
```

source/modules/MEI.header.xml
```diff
@@ -107,6 +107,62 @@
...
+          <label>3</label>
+          <item>request or invite to improvise a passage of music or</item>
```
ahankinson:
```suggestion
          <item>request or invite a player to improvise a passage of music; or</item>
```

source/modules/MEI.header.xml
```diff
@@ -107,6 +107,62 @@
...
+          <label>4</label>
+          <item>indicate that the number repetitions can be set individually or spontaneously.</item>
```
ahankinson:
```suggestion
          <item>indicate that the number of repetitions can be set individually or spontaneously.</item>
```

source/modules/MEI.header.xml
```diff
@@ -1968,21 +2025,15 @@
...
-        <datatype>
-          <rng:ref name="data.BOOLEAN"/>
-        </datatype>
-      </attDef>
-    </attList>
+    <remarks>
+      <p><gi scheme="MEI">perfRes</gi> defines the specification of a work in its ideal form or its different performance scenarios, but not the score or the layout of a music edition.</p>
+    </remarks>
+    <remarks>
+      <p>In this context the attribute <att>adlib</att> marks a performance resource as not essential for the performance of a work or that the use  is optional.</p>
```
ahankinson:
```suggestion
      <p>In this context the attribute <att>adlib</att> marks a performance resource as not essential for the performance of a work or that the use of the performance resource is optional.</p>
```

source/modules/MEI.header.xml
```diff
@@ -1968,21 +2025,15 @@
...
+    </remarks>
+    <remarks>
+      <p>To indicate the tuning of an instrument, the attribute <att>trans.diat</att> can be used.</p>
```
ahankinson, line 2054:
```suggestion
      <p>To indicate the tuning of an instrument, the attribute <att>trans.diat</att> may be used.</p>
```

source/docs/03-metadata.xml
```diff
@@ -1160,7 +1160,7 @@
                            <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/header/header-sample063.txt" parse="text"/></egXML>
                         </figure>
                      </p>
-                     <p>Music for a single player should, however, never use the <att>solo</att> attribute.</p>
+                     <p>An ad libitum part, i.e. not essential for the performance of the work, may be marked with the <att>adlib</att> attribute.</p>
```
```suggestion
                     <p>An <hi rend="italic">ad libitum</hi> part, i.e. not essential for the performance of the work, may be marked with the <att>adlib</att> attribute.</p>
```

ahankinson: NB: Please check that I got the TEI markup right -- @rettinghaus ?

source/modules/MEI.header.xml
```diff
@@ -107,6 +107,62 @@
...
   </classSpec>
+  <classSpec ident="att.perfRes" module="MEI.header" type="atts">
+    <desc>Attributes that define the characteristics and components of the performance resource.</desc>
+    <classes>
+      <memberOf key="att.perfRes.basic"/>
+      <memberOf key="att.transposition"/>
+    </classes>
+    <attList>
+      <attDef ident="solo" usage="opt">
+        <desc>Use this attribute to identify the performance resource as a soloist especially in an accompanied work, such as a concerto or vocal solo.</desc>
```
ahankinson:
```suggestion
        <desc>Use this attribute to identify the performance resource as a soloist, especially in an accompanied work such as a concerto or vocal solo.</desc>
```

source/modules/MEI.header.xml
```diff
@@ -107,6 +107,62 @@
...
+      <attDef ident="count" usage="opt">
+        <desc>Indicates the number of performers.</desc>
+        <datatype>
+          <rng:data type="positiveInteger"/>
+        </datatype>
+      </attDef>
+    </attList>
+  </classSpec>
+  <classSpec ident="att.adlibitum" module="MEI.header" type="atts">
+    <desc>Attributes that define characteristics on an entity.</desc>
```
ahankinson: Is an 'entity' the same as a 'performance resource'? If so we should make this consistent; if not we should define what an "entity" is.

source/modules/MEI.header.xml
```diff
@@ -107,6 +107,62 @@
...
+    <attList>
+      <attDef ident="adlib" usage="opt">
+        <desc>Marks an entity as not strictly defined.</desc>
```
ahankinson: Again, we have the term "entity" -- needs definition or to be made consistent with previous statements.

ahankinson (changes requested):
I think it still needs some changes. Thanks!

source/modules/MEI.header.xml
doerners: In my understanding every performance resource can be considered an entity (of the bibliographic universe) but not every entity is necessarily a performance resource. So from that perspective "entity" is a broader/more generic concept. However, I agree that we should consider to align our terminology across descriptions so that it's clear what is meant.

source/docs/03-metadata.xml
riedde: @ahankinson `<rend rend="italic">` is MEI code, in TEI it should be `<hi rend="italic">`, I think, but please correct me if I'm wrong!

source/modules/MEI.header.xml
riedde: Yes, maybe this is to general. I used this words because I remember that we discussed that this attribute (and its class) should be so generic that this can also be used in other contexts in the future (e.g. to encode an 'ad libitum' section in the `<music>` section).
Would it be better to to say more precisely something like "Attributes that characterize the way of use of a resource"?

source/modules/MEI.header.xml
riedde: Thanks, might me a copy paste error.

riedde: Ah, no, it belongs to the list following. But this solution was bad, I'll fix that!

source/docs/03-metadata.xml
ahankinson: OK! I've made the change to the suggestion.

source/modules/MEI.header.xml
ahankinson: I wouldn't try to be too general; the application of the attributes in this class are specific to performance resources, as far as I am aware (that is, `adlib` will generally always apply to some sort of performer, regardless of the context in which it is used.)

So perhaps rather than trying to be generic in where it will be used, try to be specific in what it does. Something like: "Attributes that describe a performance resource as <hi rend="italic">ad libitum</hi> (optional)."

source/modules/MEI.header.xml
ahankinson: Further to my above comments, I think this attribute description tries to be too generic, to the point where it is not particularly descriptive. "Marks a performance resource as ad libitum (optional)."

source/modules/MEI.header.xml
riedde: Okay, sounds good! I'll integrate it.

source/modules/MEI.header.xml
riedde: As before, I agree :-) I missed that this class is part of the MEI.header module. You are absolutely right.

ahankinson (dismissed):

musicEnfanthen:
ODD meeting May 25, 2022: @kepper will have a second eye on the technical implementation. If everything is fine, this PR is ready to be merged afterwards.

musicEnfanthen:
@kepper Could you have a look at this PR before the next ODD meeting, please? Many thanks!

kepper (dismissed):
Discussed this with @riedde in person. I hereby officially agree :D

musicEnfanthen:
@ahankinson and @kepper Seems like @riedde 's last addition (https://github.com/music-encoding/music-encoding/pull/842/commits/fd9fd00a0b40f53357b60c4bffa115dd1fbf5616) dismissed your former approvals. Would you mind to re-approve (or re-check) to get back the 2 required approving reviews, so the PR can be merged?

ahankinson (dismissed):

ahankinson:
Done.

musicEnfanthen:
Thank you!

musicEnfanthen:
@ahankinson Since @riedde needed to solve some merge conflicts that dismissed the existing reviews, may I ask you to re-review if possible?

source/modules/MEI.header.xml
```diff
@@ -107,6 +107,62 @@
...
+    <attList>
+      <attDef ident="adlib" usage="opt">
+        <desc xml:lang="en">Marks a performance resource as ad libitum (optional).</desc>
+        <datatype>
+          <rng:ref name="data.BOOLEAN"/>
+        </datatype>
+      </attDef>
+    </attList>
+    <remarks xml:lang="en">
+      <p>The technical term “ad libitum” has several meanings depending on the context in which it occurs. Currently only the use within a performance resource is supported.</p>
```
ahankinson: The list below doesn't mention "performance resource" as one of the types, so I would either change this sentence to match the meaning from the list below, or change the list to mention a "performance resource." (I would personally go for the first, since I'm not sure it's very widely known what a "performance resource" is...)

Also, I would put this sentence after the list, not before it.

source/modules/MEI.header.xml
```diff
@@ -1989,21 +2046,15 @@
...
-        <datatype>
-          <rng:ref name="data.BOOLEAN"/>
-        </datatype>
-      </attDef>
-    </attList>
+    <remarks xml:lang="en">
+      <p><gi scheme="MEI">perfRes</gi> defines the specification of a work in its ideal form or its different performance scenarios, but not the score or the layout of a music edition.</p>
+    </remarks>
+    <remarks xml:lang="en">
+      <p>In this context the attribute <att>adlib</att> marks a performance resource as not essential for the performance of a work or that the use  is optional.</p>
```
ahankinson:
```suggestion
      <p>In this context the attribute <att>adlib</att> marks a performance resource as not essential for the performance of a work or that the use is optional.</p>
```

ahankinson (changes requested):
Just a couple of minor changes

source/modules/MEI.header.xml
riedde: the remarks have been added to declare that there are more dimensions in the meaning of `adlibitum`. After @kepper noticed that the content of the remarks lists only things that are not supported he asked me to clarify this. I think this will be lost if I skip the wording

ahankinson: I will try to rephrase my question. 

You are saying in this sentence that the word "ad libitum" has several meanings, and that only one of them applies here, "use within a performance resource"

Then you list four meanings of ad libitum, but none of the four mention "use within a performance resource" directly. 

So I am unsure which of the four meanings you list actually apply. I think it's the first, but it would really help if there was a consistent way of talking about it between the sentence and the list.

musicEnfanthen: Wording will be revised, thanks for the thorough reading @ahankinson

riedde: I got it. Done, thanks. Sorry for missunderstanding.

ahankinson (dismissed):

ahankinson:
Thanks!

source/docs/03-metadata.xml
```diff
@@ -1160,7 +1160,7 @@
                            <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/header/header-sample063.txt" parse="text"/></egXML>
                         </figure>
                      </p>
-                     <p>Music for a single player should, however, never use the <att>solo</att> attribute.</p>
```
musicEnfanthen, line 1163: Didn't we want to keep this as a warning?

riedde: I can't find it for now, but I'm sure that we discussed this. For me, this is 'just' a best practice suggestion. I we keep it I would suggest rephrasing it, e.g., `Music for a single player is not needed to be indicated as solo using the <att>solo</att> attribute.`

musicEnfanthen: Yes, i think we discussed it with the group in the Wannsee room. Maybe `Music for a single player does not have to be marked as solo with the <att>solo</att> attribute.`

riedde: @musicEnfanthen Done

musicEnfanthen (approved):

source/docs/03-metadata.xml
musicEnfanthen: Thank you!

musicEnfanthen:
@ahankinson may I ask you once more (hopefully for the last time) for your final OK?

musicEnfanthen:
@ahankinson , pinging you for final ok

ahankinson (approved):

musicEnfanthen:
Thank you!

musicEnfanthen:
What lasts long becomes good. Merry christmas, @riedde ;)

bwbohl:
many thanks to all!

Probably we should still add `gloss` es to the attribute definitions.

PR #843, opened 2021-07-23
Allow simple calculus in numerator of meterSig

rettinghaus:
This is a small change to allow simple calculations within the numerator of time signatures. 
We find numerators with multiplication, e.g. in Bartók's _Minor and Major_ from his _Mikrokosmos_:

Bartok-timeTimes

As we have all basic mathematical symbols within [SMuFL's Time signatures range](https://w3c.github.io/smufl/latest/tables/time-signatures.html), it seems the way to go to allow also subtraction and division in the numerator of time signatures.

As `@count` is in the logical domain, I chose to go with `*` over the (visual) `x` for multiplication.
(Especially nice to handle this then within Python with `eval()`.)


ahankinson:
You may also want to update the attribute description with additional examples, since it mentions additive, but not subtractive, multiplicative or divisive `@count` values.


You're brave, using `eval()` on user input! I look forward to writing my new composition with a time signature of `@count="import shutil; shutil.rmtree(\"/\")"`. I think I'll call it "Threnody for the victims of unsanitized input".

:-D

ahankinson:
PS: Also, I concur with "*" over "x", but mostly because we already use "x" to indicate sharps.

ahankinson:
PS: Also, I concur with "*" over "x", but mostly because we already use "x" to indicate sharps.

rettinghaus:
> You're brave, using eval() on user input! I look forward to writing my new composition with a time signature of @count="import shutil; shutil.rmtree(\"/\")". I think I'll call it "Threnody for the victims of unsanitized input".

I tried it, but this still isn't valid. Good thing all people only use valid MEI ;-)

ahankinson:
Yep, I have never ever seen any invalid MEI. Nope. Doesn’t exist. Lalalala not listening!

ahankinson:
Does it have to evaluate to a whole number? What does a `count=3/2` mean? Would a time signature of (3/2) / 4 be equivalent to 1.5 / 4?

rettinghaus:
No, the description says it has to evaluate to a decimal number. So (3/2) / 4 would semantically be equivalent to 1.5 / 4, but they would look different.

pe-ro:
But practically speaking, 1.5 / 4 is easier to handle if it's converted to 3/8.

ahankinson:
Oh sorry… That’s was ambiguous. The /4 was just to illustrate the “denominator” of the time signature. It’s not part of the `@count` value.

or am I the one getting confused? Looking again, I can see how 1.5/4 would be 3/8, and also (3/2)/4

pe-ro:
> PS: Also, I concur with "*" over "x", but mostly because we already use "x" to indicate sharps.

But since sharps don't occur within time signatures, I think it would be better to use "X" as it allows the encoder to encode what's on the page.  It also discourages the direct evaluation of `@count`, which I believe is a good thing.

rettinghaus:
> as it allows the encoder to encode what's on the page

But `@count` is in the logical domain, and on the page we read "three times two". 
And what if the publisher doesn't use a large X but a small one? A small x would then assume it to be printed on the numbers base line instead of a vertically centered position: 3x2. The next engraver maybe would prefer even a star as symbol. We could add every possible symbol there, but that wouldn't be very machine readable, would it?

kepper:
We discussed this during today's ODD meeting. In general, it seems clear we need such a thing. However, we think that the mixture of visual and logical domains need some more consideration here. An alternative could be to leave the final, calculated value in `@meter.count`, and have textual content in the `<meter>` element. That way, we could have transcriptional features as well, and wouldn't depend on user input calculations. We will push the topic for another month to have more room for discussions.

ahankinson:
Another possibility would be to have the fully evaluated value ("6" instead of "2*3") in the numerator and denominator, but simply use a different SMuFL glyph value to display it. 

That would address the concern with upper or lower-case, etc, and also provide an easy way to display the expression without needing to parse a string to create the time signature symbol.

bwbohl:
Although this is not directly connected to #848 I have the impression that we should come up with a similar solution for both cases.

rettinghaus:
@bwbohl You are right, that's why #818 is already part of the discussion #826. I added the PR there too.

bwbohl:
ODD meeting notes:

* only whole meter signature can point to glyph, e.g. SMUFL
* modern music meter signatures are an argument against putting the result of the calculus in the attribute e.g. π^e
* an appropriate solution for encoding both logical and visual (transcriptional) aspects is definitely needed
* a profound solution come up with equal solutions for this and #848
* **especially** see #826
* Merging this looks ok for us today, as it supports adding real-life values to the encoding, (although proper mechanisms and examples for a distinct encoding of visual and logical representations is still a desire)
* The guidelines should also be modified to provide some examples, after that is assured, we will leave a window of one week for comments before merging

bwbohl:
@music-encoding/core-contributors please review this pull request within the week, tanks ;-)

bwbohl (approved):

kepper:
We discussed this (again) extensively at today's ODD meeting. We are aware that the solution proposed is not fully satisfactory, and does leave open the question of separating logical and visual domains (and maybe more than that). However, this is about merging something into the dev version of MEI, it's not yet a release. We do agree that this still needs work, and that this future work may come to different conclusions, potentially invalidating all instances generated according to this proposal. We expect that such incompatibilities can be fixed, though, and so we will merge this one to collect some experience with this model. One change will be made by @rettinghaus before the merge, though: The regex expression is currently provided in both `att.meterSigDefault.log` and `att.meterSig.log`, which seems an unnecessary duplication. Instead, a new data type will be created for this, which will make maintenance much easier. After this change, the PR is ready to be merged. Thanks for everyone's contributions to this, and let's see where this will take us ;-)

ahankinson:
> However, this is about merging something into the dev version of MEI, it's not yet a release.

> We expect that such incompatibilities can be fixed, though, and so we will merge this one to collect some experience with this model.

Merging something into a specification generally means it becomes part of the specification. I'm not particularly comfortable with treating the spec as software in this way... 

Edit, to be more specific: We do not currently have a process of re-reviewing the rules that are in the spec AFTER they have made it in to the dev version. As such, it is highly likely that this change will be part of the next released version if it is merged now.

It is also unlikely to be made part of software applications until after it is released, so the process of 'trying it out' can only happen after it has been released. 

---

To be clear: At the moment I'm less concerned with this specific change, and more the way the changes are being handled. We have operated on a consensus model to date, and our discussions in reaching consensus have typically been recorded in GitHub so that we have an audit trail of decisions. 

Without minutes from the ODD meetings, but with decisions arising out of them, we have lost this audit trail and the consensus approach.

musicEnfanthen:
> The regex expression is currently provided in both att.meterSigDefault.log and att.meterSig.log, which seems an unnecessary duplication. Instead, a new data type will be created for this, which will make maintenance much easier. 

Looking at the proposed regex in detail, there is an issue and some questions:

**Issue**
The proposed regex is `\d+(\.\d+)?(\s*[\+\-\*/]\s*\d+(\.\d+)?)*`

`[\+\-\*/] ` will not allow for divisions but for +, -, * and ], the division sign needs to be escaped: `[\+\-\*\/]`

**Questions**
As it stands, the regex allows any integers, but also floating numbers (of any length) to be parts of the equation, as well as any number of spaces between numbers and math symbols. This would allow an input like this:

`2222.05678899 [...tons of tons of white spaces not displayed in markdown...]  -          5+    3/2+3-4`

Would we really need to allow a number in the millions followed by a quadrillion spaces in a time signature?
Would it be sufficient to restrict the length of numbers to something like {1,10} which is still a billion and the length of spaces to 1?

E.g., something like this: `\d{1,10}(\.\d{1,10})?(\s?[\+\-\*\/]\s?\d{1,10}(\.\d{1,10})?)*`, what would still allow for `1234567890.1234567890 + 1` (if that's ever needed). 

But probably it would be even sufficient to have a 4-digit integer part and a 1-digit floating part?
`\d{1,4}(\.\d?)?(\s?[\+\-\*\/]\s?\d{1,4}(\.\d?)?)*` allows for `1234.2 + 1`

I see that the regex was not introduced in this PR, only extended, but as @rettinghaus is about to include a datatype for it, it could be an opportunity to refactor it.

Maybe someone with more knowledge about regex than my home-glued snippets can confirm?

ahankinson (approved):

musicEnfanthen:
Thank you @rettinghaus. After thorough discussion at the ODD meeting 2022-01-28, we decided to add these numerators. The underlying issue (visual vs. logic) still goes deeper and needs ongoing follow-up discussion.

rettinghaus:
@musicEnfanthen Now that this is merged validation seems to work as expected on my side, but maybe you want to do a second check?

PR #846, opened 2021-07-28
Develop accents

### LINKED ISSUE #768: Soft accents (and combinations)
Is there a "native" way to encode the glyphs from https://w3c.github.io/smufl/latest/tables/articulation-supplement.html in MEI, or does one have to mix articulations with hairpins? 
Switching to `att.extSym` could solve it only in part, as you still not have an explicit semantic information.

rettinghaus:
This PR adds three new accent values to `data.ARTICULATION` as discussed in ODD meeting: 

- `acc-inv`
- `acc-long`
- `acc-soft`

First is to encode _inverted_ or _reversed_ accents, as seen in 19th-century vocal music:
Viardot Garcia

Second is to address "longer" accents, that may look like a hairpin: 
Chopin

Third is a combination of an inverted and a normal accent:
Brahms

Lortzing

SMuFL has added them as `articSoftAccentAbove` to the `Articulation supplement`:
articSoftAccentAbove

Closes #768

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -200,6 +200,15 @@
         <valItem ident="acc">
           <desc>Accent (Unicode 1D17B).</desc>
         </valItem>
+        <valItem ident="acc-inv">
+          <desc>Inverted accent.</desc>
+        </valItem>
+        <valItem ident="acc-long">
+          <desc>Long accent, used to indicate an elongated accent mark. It is the responsibility of the encoder to distinguish between accents and a hairpins.</desc>
+        </valItem>
+        <valItem ident="acc-soft">
+          <desc>Soft accent, see SMuFL Articulation supplement (U+ED40–U+ED4F).</desc>
+        </valItem>
         <valItem ident="stacc">
           <desc>Staccato (Unicode 1D17C).</desc>
         </valItem>
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -204,7 +204,7 @@
           <desc>Inverted accent.</desc>
         </valItem>
         <valItem ident="acc-long">
-          <desc>Long accent, used to indicate an elongated accent mark. It is the responsibility of the encoder to distinguish between accents and a hairpins.</desc>
+          <desc>Long accent, used to indicate an elongated accent mark. It is the responsibility of the encoder to distinguish between accents and hairpins.</desc>
         </valItem>
         <valItem ident="acc-soft">
           <desc>Soft accent, see SMuFL Articulation supplement (U+ED40–U+ED4F).</desc>


ahankinson:
I would be in favour of keeping the deprecated warning. It doesn't get in the way, and it could be useful for anyone who might be upgrading old MEI files.

ahankinson (approved):

ahankinson:
Actually -- looking at this again, I'm not sure about 'long accents'. The example you provided doesn't look particularly long, if you account for differences in handwriting style.

Do you have some more evidence that this is actually a "thing"? I didn't find anything in SMuFL for this.

rettinghaus:
_Every_ musicologist knows about this problem we try to address here, and we discussed it extensively during the ODD meetings. 
But I could collect examples the whole day long. Here are some more from Chopin, followed by Beethoven and Brahms:

36_Stockholm_05
Image05
image-2020-12-30-09-33-13-442
image
image

This matter is also discussed in several books. Here's a blog post about Schubert: 
https://www.henle.de/blog/en/2014/04/28/the-accent-question-in-schubert-an-old-theme-with-new-variations/
And some more on Brahms:
https://www.google.de/books/edition/Brahms_Studies/aqAPaRKxlGYC?hl=de&gbpv=1&dq=Brahms%20accent%20hairpins&pg=PA19&printsec=frontcover&bsq=hairpin%20looks%20like%20an%20accent
If you want go looking for yourself, there's plenty of Schumann and Liszt …

lpugin:
Thanks for providing some examples, @rettinghaus. However, I would like to point out that not _everybody_ in the MEI community is expected to be a musicologist. Quite on the contrary, one strength of MEI is to reach out in many directions. I am sorry to say that the tone of you message is disrespectful on that regard and it would be appreciable to avoid this in the future - even if the question asked has been discussed and answered somewhere else already. Thanks!

kepper:
Hi there. I definitely see the value of the inverted and soft accents (I didn't know the name _soft_, but there are so many more things that I don't know…). Having missed the last ODD meeting, I must say I'm less convinced by the long accent. Eventually, this comes down to the problem that in music notation, there is no absolute and clear distinction between an accent and a decrescendo hairpin. Actually, they even have a related meaning – start loud, decrease volume. I don't think it's safely possible to distinguish between just the two existing values, and have considered more than once how to encode something "choicy" for a combination of a [controlEvent](https://music-encoding.org/guidelines/dev/model-classes/model.controlEventLike.cmn.html) and a [noteModifier](https://music-encoding.org/guidelines/dev/model-classes/model.noteModifierLike.html) (spoiler: in current MEI, you can't…). Adding a third value "in the middle" could be helpful ("if you can't decide between accent and hairpin, use long accent"), but eventually I'd expect even bigger problems, because we now need guidance to distinguish between those three values. So, I'm not exactly opposed to this, but definitely have mixed feelings about it. By all means, I believe this needs coverage in the Guidelines, maybe using some of the examples shown here…

kepper:
After we've discussed this again during today's ODD meeting, we agree to understand this new value of `acc-long` as placeholder for something that can't be determined precisely. @rettinghaus has volunteered to provide coverage of this in the CMN chapter of the Guidelines, and ideally, we also integrate that somewhere in the editTrans chapter, dealing with uncertainty. Thanks everyone for moving this forward!

ahankinson:
I don't think this has been fully accepted by consensus, so I strongly disagree with merging this. I would like to see the change reverted. 

If something is unclear, there are other methods of encoding that in MEI (sic/corr, unclear, etc.). Were those options discussed?

ahankinson:
I was giving this issue a couple days to cool off, since @rettinghaus seemed to think my opinion on these accents was uninformed, and I needed to educate myself. Unfortunately, that meant that I missed the small window of time given to comment further.

I studied the examples given a few days ago -- I actually spent a couple hours on this issue trying to figure it out. The Henle blog quoted ends with this statement:

> The few examples show that length and direction alone are not enough for interpreting the > sign, but also, unfortunately, the musical context does not always lead to the de­sired clarity.
> Ideas on this old subject are therefore very welcome from your side!

The second, an excerpt from a book about Brahms, says:

>... the first hairpin looks like an accent sign (a problem Brahms corrected in mm. 1-3 and 9 but let stand here) ...

This says to me that:

1. There are certainly difficulties with interpreting "long" accents
2. There is no actual consensus on how they should be interpreted
3. Composers did not use them consistently to mean the same thing
4. Composers corrected them sometimes, but not other times

In this case, it seems that marking an "accent" as "long" simply means "the composer wrote it bigger". There is no actual agreed method of how these accents should be interpreted or rendered if they are encountered in an MEI file. Why did they write them bigger? How long? To the next note? Does `long` mean visually long? Or temporally long? If it stretches to the next note, does that mean to accent the next one too? In the manuscript examples that are given as evidence, did the composer actually correct them to be 'normal' sized when they went to be printed? Why or why not?

I'm sure *every* musicologist knows the answer to these questions too, so I would hope the answers are readily available and I can be further enlightened.

Nevertheless, given that these are actually a "thing" that people talk about, I was wondering if there might be alternative ways of expressing this that would be more specific in what is meant. Perhaps allowing some sort of `scale` on articulations to scale them so that they appear bigger. Or use `<rend>`? That would allow for more precise encoding of visually how long.

Or maybe `<artic>` elements need to allow start and end IDs? Then that would allow for encoding temporally how long.

Or perhaps there is a SMuFL code point for "Chopin-Long-Accent-that-looks-like-a-hairpin"?

If interpretation can vary, why not use the existing mechanisms in MEI to explicitly mark up the variance?

kepper:
Hi @ahankinson, despite having merged this, I do share your concerns. However, I think they're bigger than this issue, and have many more implications than covered here. But first, I _do not believe_ that this is something all musicologists are aware of. I'd be surprised if an editor experienced with music from the CWMN period hasn't come across this problem, but depending on the area of expertise, musicologists in general will easily find many interesting playgrounds, but not necessarily this one. And, as @lpugin has correctly pointed out, MEI is a mixed community with experts from many fields, so the expectation that something X is obvious to everyone is surely wrong, and we need to keep an eye on that in our communication. That said, I applaud your willingness to bring you up on speed on this – certainly not something we can take as granted, given the limited time most of us can dedicate to MEI work. 

Having taught this very problem in University classes for more than a decade, I think it's all but trivial, and I don't believe there is a single "correct" answer to the notational problem. It is my understanding (and this is from my perspective also what we agreed on during today's ODD meeting) that the proposed value shall be used as a kind of "undecided in-between". Admittedly, the name isn't indicating that properly, but we agreed that this needs to be made _very_ clear in the Guidelines (@rettinghaus volunteered to cover this in the CMN chapter). Maybe we rushed this by not waiting until we have that documentation – I apologize for that. 

Coming back to one of your questions about using existing MEI concepts to deal with the uncertainty: As I pointed out in [my earlier comment](https://github.com/music-encoding/music-encoding/pull/846#issuecomment-889022967), the problem is that `<hairpin>` is a controlEvent, while `<artic>` is a noteModifier. It is possible to mix them into a choice, but only because choices are allowed in so many places. Resolving the choice would always mean that one of the two things is incorrectly placed in the XML tree. Obviously, MEI lacks a concept like TEI's [`@exclude`](https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.global.linking.html) – we simply can't do choices that cross hierarchies. This is definitely an issue, but it is way bigger than the question of long accents. Until we have a solution for that, `acc-long` may serve as the best compromise we can get for now. But it's certainly not an elegant solution that I particularly like… Happy to consider alternatives, but _I_ don't see any :-(

PR #847, opened 2021-07-29
WIP: initial version of search

### LINKED ISSUE #749: search is broken
On our new static guidelines pages, the search function is broken. The old setup uses tipue search (https://github.com/michael-milette/Tipue-Search), which is a bit dated (last update on that Git repo was 4 years ago, but I'm not even sure it is this version that has been used), and also  introduces a dependency on jQuery. Apparently, tipue search is / was easy to set up with Jekyll (correct, @lpugin?). However, as those pages aren't built with Jekyll anymore, that's less of an argument now. It is certainly doable to reintroduce tipue search, but we may consider our options first. What we want is a client-side search function that takes a pre-generated index and lists guidelines pages which match. After some research, I see two other options for us: 

- Lunr.js (https://lunrjs.com/docs/index.html)
- Fuse.js (https://fusejs.io/)

The latter seems the more powerful, as it also brings fuzzy search. For both libraries, it should be fairly easy to adjust the index we currently generate. However, both options don't generate HTML to display search results – that's something we would have to rewrite. My guess is that it's a bit more work to integrate either Lunr.js or Fuse.js, but not by much. Personally, I don't like jQuery and would be willing to invest that extra time just to remove that dependency. With Fuse.js, we would get fuzziness, which could be nice, whereas with Lunr.js it seems pretty easy to set up different categories – searching for things like `annot`, we could prioritize the corresponding spec page before listing all chapters referencing the element and eventually bringing up all mentions of those characters somewhere in the text. Maybe this is also possible with Fuse.js (or tipue!), but I haven't seen a way there yet. 

I have no strong opinion on this, and actually going back to tipue would be perfectly fine with me. However, there might be an option to improve the user experience, so I'm wondering what others (@music-encoding/core-contributors and beyond) think about this? I would like to fix search soon, as this is the biggest remaining issue with the Guidelines…

kepper:
This is the first version of a new search based on [Fuse.JS](https://fusejs.io/). It allows fuzzy searching, which is generally helpful. It may require some more finetuning, though, to get the best possible results. The current options I use for Fuse are found [here](https://github.com/music-encoding/music-encoding/blob/0bc78b6e6d315eb666f82d683a87e76f53a6629b/utils/guidelines_xslt/odd2html/generateSearchIndex.xsl#L132-L153). The index itself is generated [here](https://github.com/music-encoding/music-encoding/blob/0bc78b6e6d315eb666f82d683a87e76f53a6629b/utils/guidelines_xslt/odd2html/generateSearchIndex.xsl#L20-L106). In summary, I have three levels of things I'm searching in: 
* **ident**
* **desc**
* **remarks**

For most parts of the Specs, it's fairly easy to see where these are coming from. The first field has a weight of 10, the second a weight of 2, the last a weight of 1. For **Guidelines Chapters**, *ident* is the title of the chapter (including the chapter number), *desc* is a list of all elements etc. referenced explicitly, and *remarks* is the full text of the chapter, including examples. Only for guidelines chapters I render a preview from the remarks, as everything else is either quite obvious or hard to render at all. All those parameters can be adjusted as necessary. 

The search index is generated alongside the web version of the guidelines and should work out of the box. Compilation times haven't changed for me, so this should be fairly efficient. The searchIndex.js that is generated weighs in at around 900kb, and is already quite compressed. We could think about doing this on the server instead, but I'm not sure this is really preferable. 

Of course I'd like to get feedback from others as well (@musicEnfanthen, @rettinghaus, …) – I just failed to set you up as reviewers ;-) 

I believe this is ready in general. That said, there is no rush to merge it – after tomorrow, I'll be offline for the coming two weeks ;-)

If merged, this fixes #749.


utils/guidelines_xslt/odd2html/generateWebsite.xsl
```diff
@@ -91,11 +93,15 @@
         <xsl:variable name="quot" as="xs:string">"</xsl:variable>
         <xsl:variable name="json.string" select="replace(replace(replace(xml-to-json($search.index),'\\/','/'),'\{', $newline || '{'),'(' || $quot || '(text|title|url|tags)' || $quot || ')',$newline || '  $1')" as="xs:string"/>
         
-        <xsl:result-document href="{$web.output}tipuesearch/search_content.js" method="text" indent="yes" omit-xml-declaration="yes" >
+        <!--<xsl:result-document href="{$web.output}tipuesearch/search_content.js" method="text" indent="yes" omit-xml-declaration="yes" >
             var tipuesearch = <xsl:value-of select="$json.string"/>;
+        </xsl:result-document>-->
```
rettinghaus: if this is broken anyway, why not remove it completely?

utils/guidelines_xslt/odd2html/htmlFile.xsl
```diff
@@ -96,11 +97,11 @@
                                         </div>
                                         
                                         <div class="searchbox">
-                                            <form action="/guidelines/dev/content/metadata.html"><!-- TODO: fix that link??? -->
-                                                <div class="tipue_search_group">
-                                                    <input name="q" id="tipue_search_input" pattern=".{{3,}}" title="At least 3 characters" required="" type="text"></input><button type="submit" class="tipue_search_button"><span class="tipue_search_icon">⚲</span></button>
-                                                </div>
-                                            </form>
+                                            
+                                            <form><div class="tipue_search_group">
```
rettinghaus: If tipue is gone, why keep the classes? I would prefer to have them renamed.

rettinghaus (approved):
Looks good to me!

utils/guidelines_xslt/odd2html/generateWebsite.xsl
kepper: I haven't had the time to do a proper cleanup – that's one of the reasons why I said there's no rush to merge this.

utils/guidelines_xslt/odd2html/htmlFile.xsl
kepper: Again, because of time constraints, I haven't done this yet. This still relies on a (potentially large) number of CSS rules from tipue, and cleaning this up may take some time. I agree it should be done, though.

ahankinson:
I just converted this to a draft to indicate that it's still a WIP.

kepper:
Thanks @ahankinson – draft is surely more appropriate at this point. I will put it on the agenda for the [next ODD meeting on August 26](https://github.com/orgs/music-encoding/projects/2#column-15347186). I can't promise that I'll have the CSS cleaned up by then, but maybe some more people will have had an opportunity to look at this.

kepper:
reviewed at ODD Thursday. Question (to myself): Is this going to work for MEI v4 as well?

bwbohl:
ODD meting notes:

* code cleanup still needed
* from the discussion I assume testing is also needed
* Proposal is: get code cleanup done, merge and let users test ;-)

kepper:
the code for the v4 guidelines is static (see https://github.com/music-encoding/guidelines/tree/master/v4), so this won't give search on older versions. While it's doable to adopt this, I suggest to test it on the dev guidelines in the wild first…

bwbohl:
2021-10-28 ODD meeeting:
* functional again
* old CSS stuff is removed now
* only working on current dev
** if we want it on v4 and earlier we would need to generate indices once and add them to the corresponding files in the schema repo

bwbohl (approved):

rettinghaus (approved):

rettinghaus:
@kepper do you want to keep the `search-guidelines` branch for additional things? Otherwise I'd recommend to delete it.

kepper:
let me see what I can accomplish this week. Would be nice to keep it as reference over there, but then remove it afterwards…

rettinghaus:
@kepper The search isn't working because `fuse.js` is missing. There are several possible ways to fix that, but I would leave this to you. If you like I'll open a new ticket for this.

kepper:
IIRC, this is an issue with moving things in the correct folder, so this is mostly an issue about the GitHub Actions workflows. Seems easiest to me to do this during a video call with someone more involved in this, maybe @musicEnfanthen, @bwbohl or you, @rettinghaus?

bwbohl:
that was what I tried to address in #871 but if I understand correctly I was mistaken in removing fuse.js, right? So we need the build.xml patch and KEEP fuse.js, correct?

Issue #686, opened 2020-07-05
"3" mensuration sign


craigsapp:
Related to issue https://github.com/rism-ch/verovio/issues/1528, how should a "3" mensuration sign be represented in MEI?

Here is an example:

<img width="802" alt="Screen Shot 2020-07-05 at 10 11 18 AM" src="https://user-images.githubusercontent.com/3487289/86537982-e6d8b880-bea7-11ea-9261-70e73c33dec9.png">

The top staff shows the original mensural music, and the bottom staff shows the meaning of the notation in CMN.  The "3"  sign means that three semi-breves in the following music take the same time as two semi-breves in the previous music.  In modern terms, the "3" indicates a region of triplet notes as illustrated on the bottom staff.   This is equivalent to using coloration, but "3" is typically used for longer regions of triplets than coloration.

In the [Josquin Research Project's](https://josquin.stanford.edu) score [database](https://github.com/josquin-research-project/jrp-scores), the "3" mensuration is the 4th most common mensuration sign, occurring 7% out of all mensuration signs:

| count	|	percent	|	mensuration	|
| ----- | ------------- | ------------- |
| 1729	|	52.02	|	Cut-C	|
| 583	|	17.54	|	O	|
| 331	|	9.96	|	C	|
| 232	|	6.98	|	3	|
| 144	|	4.33	|	C2	|
| 82	|	2.47	|	Cut-circle	|
| 42	|	1.26	|	O2	|
| 41	|	1.23	|	Cut-C-3	|
| 40	|	1.2	|	O/3	|
| 26	|	0.78	|	C3	|
| 18	|	0.54	|	C-dot	|
| 10	|	0.3	|	O3	|
| 8	|	0.24	|	3/2	|
| 8	|	0.24	|	Cut-circle-3/2	|
| 6	|	0.18	|	O-dot	|
| 6	|	0.18	|	Cut-C-2	|
| 4	|	0.12	|	2	|
| 3	|	0.09	|	O3/2	|
| 2	|	0.06	|	C-dot-3/2	|
| 2	|	0.06	|	Cut-circle-3	|
| 2	|	0.06	|	Cut-C/2	|
| 1	|	0.03	|	Cut-C/3	|
| 1	|	0.03	|	C2/3	|
| 1	|	0.03	|	C-reversed |
| 1	|	0.03	|	Cut-C-reversed	|
| 1	|	0.03	|	Cut-C-dot	|

[...]

PR #676, opened 2020-06-17
Revert "Merge pull request #541 from rettinghaus/develop-mensural"

martha-thomae:
This reverts commit 03e9478e6194984efd2422c06336fc80b43c49b4, reversing
changes made to c6faac8e132d296f2e13c9be3847c3d5b23ce992.

It reverts the removal of mensuration- and proportion-related attributes from `<staffDef>` and `<scoreDef>`, keeping the consistency with how meter and clefs are encoded in MEI.

PR #848, opened 2021-07-30
schema: Add encoding of numbers as parts of mensuration signs

annplaksin:
This PR implements the result of the discussion on mensuration and proportion signs of the Mensural IG.
For more insights and examples, see [here](https://docs.google.com/document/d/1M_CUqlvPNFiS-EI39DgggzwmX4VoLp0nvMXPp_rCbLY/edit#heading=h.gjdgxs).

If adds the attributes `@sign.num` and `@sign.numbase` or `@mensur.sign.num` and `@mensur.sign.numbase` for the encoding of numbers as parts of mensuration signs in the visual domain.
Due to technical reasons, currently "Z" is treated as number to avoid a lack of clarity on mensuration signs encoded in `@sign`.

To do:
* [x] Guidelines: Stress that num & numbase is not to be used as time signature (that is sign.num and sign.numbase), only to state general diminution or other general shifts in duration.
* [x] Guidelines: Add documentation about when to use `<mensur>` and when to use `<proport>`
* [x] Add remark to `data.MENSURATIONSIGNNUMBERS` that it allows tokens to deal with multiple numbers

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -2021,7 +2021,8 @@
       </rng:data>
     </content>
     <remarks>
-      <p>The glyph "Z" used in 17th century Spanish manuscripts is treated as number.</p>
+      <p>Allows for multiple tokens to deal with signs containing multiple numbers, like "O 2 2" in Hothby’s “Ora pro nobis” (Codex Faenza fol. 28r).</p>
+      <p>The glyph "Z" used in 17th century Spanish manuscripts is treated as number 3.</p>
     </remarks>
   </macroSpec>
   <macroSpec ident="data.METERFORM" module="MEI" type="dt">
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -2013,6 +2013,17 @@
       </valList>
     </content>
   </macroSpec>
+  <macroSpec ident="data.MENSURATIONSIGNNUMBERS" module="MEI" type="dt">
+    <desc>Number values as part of mensuration signs.</desc>
+    <content>
+      <rng:data type="token">
+        <rng:param name="pattern">[\d|Z]?[ [\d|Z]]*</rng:param>
+      </rng:data>
+    </content>
+    <remarks>
+      <p>The glyph "Z" used in 17th century Spanish manuscripts is treated as number.</p>
+    </remarks>
+  </macroSpec>
   <macroSpec ident="data.METERFORM" module="MEI" type="dt">
     <desc>Contains an indication of how a meter signature should be rendered.</desc>
     <content>
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -2022,7 +2022,7 @@
     </content>
     <remarks>
       <p>Allows for multiple tokens to deal with signs containing multiple numbers, like "O 2 2" in Hothby’s “Ora pro nobis” (Codex Faenza fol. 28r).</p>
-      <p>The glyph "Z" used in 17th century Spanish manuscripts is treated as number 3.</p>
+      <p>The glyph "Z" used in 17th-century Spanish manuscripts is treated as the proportion 3:2.</p>
     </remarks>
   </macroSpec>
   <macroSpec ident="data.METERFORM" module="MEI" type="dt">
--- a/source/docs/05-mensural.xml
+++ b/source/docs/05-mensural.xml
@@ -284,12 +284,13 @@
             </div>
             <div xml:id="mensuralProportions" type="div2">
                <head>Proportions</head>
-               <p>Proportions can also be indicated within the <gi scheme="MEI">staffDef</gi> element. The <att>proport.num</att> and <att>proport.numbase</att> attributes are available for encoding the numerator and the denominator of the proportion, respectively. There is also a <gi scheme="MEI">proport</gi> element that can be used as an alternative, with the corresponding <att>num</att> and <att>numbase</att> attributes.</p>
+               <p>Proportions can also be indicated within the <gi scheme="MEI">staffDef</gi> element. The <att>proport.num</att> and <att>proport.numbase</att> attributes are available for encoding the numerator and the denominator of the proportion, respectively. There is also a <gi scheme="MEI">proport</gi> element that can be used as an alternative, with the corresponding <att>num</att> and <att>numbase</att>.</p>
                <p>
                   <specList>
                      <specDesc key="proport"/>
                   </specList>
                </p>
+               <p>The time signatures of the actual proportion sign can be encoded with <att>sign.num</att> and <att>sign.numbase</att>, respectively <att>mensur.sign.num</att> and <att>mensur.sign.numbase</att> within <gi scheme="MEI">staffDef</gi>. <att>proport.num</att> and <att>proport.numbase</att>, respectively <att>num</att> and <att>numbase</att>, do not cover the visual information and should only be used to indicate the ratio of the given diminution or augmentation.</p>
             </div>
             <div xml:id="mensuralLigatures" type="div2">
                <head>Ligatures</head>
--- a/source/docs/05-mensural.xml
+++ b/source/docs/05-mensural.xml
@@ -291,6 +291,11 @@
                   </specList>
                </p>
                <p>The time signatures of the actual proportion sign can be encoded with <att>sign.num</att> and <att>sign.numbase</att>, respectively <att>mensur.sign.num</att> and <att>mensur.sign.numbase</att> within <gi scheme="MEI">staffDef</gi>. <att>proport.num</att> and <att>proport.numbase</att>, respectively <att>num</att> and <att>numbase</att>, do not cover the visual information and should only be used to indicate the ratio of the given diminution or augmentation.</p>
+               <div xml:id="mensuralProportionVsMensur" type="div3">
+                  <head>Proportion vs. mensuration</head>
+                  <p>Proportion signs, e.g. "3", can not only indicate a shift in duration but as well in mensuration. This may include the application of rules of imperfection and augmentation.</p>
+                  <p>While <gi scheme="MEI">mensur</gi> supports the explicit description of mensuration signs and note division levels, it should be used whenever there is an implicit or explicit change in the division levels or a mensuration or division sign is present. On the contrary, <gi scheme="MEI">proport</gi> only supports attributes describing time signatures and general diminution or augmentation. Therefore it is recommended to use <gi scheme="MEI">proport</gi> only for proportions that do not include any implicit change of mensuration.</p>
+               </div>
             </div>
             <div xml:id="mensuralLigatures" type="div2">
                <head>Ligatures</head>
--- a/source/docs/05-mensural.xml
+++ b/source/docs/05-mensural.xml
@@ -230,14 +230,14 @@
                      <specDesc key="att.mensural.shared" atts="modusmaior modusminor tempus prolatio"/>
                   </specList>
                </p>
-               <p>The mensur signs themselves can be encoded in the <att>mensur.sign</att> attribute with a possible value of "C" or "O". Its orientation can be encoded in the <att>mensur.orient</att> attribute, for example, with the value "reversed" for a flipped C sign. The number of slashes (up to 6) can be given in the <att>mensur.slash</att> attribute. There is also a <att>mensur.dot</att> attribute for indicating the presence of a dot through the boolean values "true" or "false".</p>
+               <p>The mensur signs themselves can be encoded in the <att>mensur.sign</att> attribute with a possible value of "C" or "O". Its orientation can be encoded in the <att>mensur.orient</att> attribute, for example, with the value "reversed" for a flipped C sign. The number of slashes (up to 6) can be given in the <att>mensur.slash</att> attribute. There is also a <att>mensur.dot</att> attribute for indicating the presence of a dot through the boolean values "true" or "false". To encode numbers accompanying a mensur sign, <att>mensur.sign.num</att> is used for numbers following the sign, <att>mensur.sign.numbase</att> could be used for numbers beneath the sign.</p>
                <p>
                   <specList>
-                     <specDesc key="att.mensural.vis" atts="mensur.sign mensur.dot mensur.slash mensur.orient"/>
+                     <specDesc key="att.mensural.vis" atts="mensur.sign mensur.dot mensur.slash mensur.orient mensur.sign.num mensur.sign.numbase"/>
                   </specList>
                </p>
                <p>
-                  <gi scheme="MEI">mensur</gi> elements can also be used instead of <gi scheme="MEI">staffDef</gi> and its attributes. In <gi scheme="MEI">mensur</gi>, the division levels are encoded with the previously mentioned <att>modusmaior</att>, <att>modusminor</att>, <att>tempus</att>, and <att>prolatio</att> attributes, while the attributes to indicate the mensur signs are: <att>sign</att>, <att>orient</att>, <att>slash</att>, and <att>dot</att>. <gi scheme="MEI">mensur</gi> can be a child of the <gi scheme="MEI">staffDef</gi> and <gi scheme="MEI">layer</gi> elements.</p>
+                  <gi scheme="MEI">mensur</gi> elements can also be used instead of <gi scheme="MEI">staffDef</gi> and its attributes. In <gi scheme="MEI">mensur</gi>, the division levels are encoded with the previously mentioned <att>modusmaior</att>, <att>modusminor</att>, <att>tempus</att>, and <att>prolatio</att> attributes, while the attributes to indicate the mensur signs are: <att>sign</att>, <att>orient</att>, <att>slash</att>, <att>dot</att>, <att>sign.num</att>, and <att>sign.numbase</att>. <gi scheme="MEI">mensur</gi> can be a child of the <gi scheme="MEI">staffDef</gi> and <gi scheme="MEI">layer</gi> elements.</p>
                <p>
                   <specList>
                      <specDesc key="mensur"/>
@@ -252,6 +252,7 @@
                   <specList>
                      <specDesc key="att.mensur.vis" atts="sign dot orient"/>
                      <specDesc key="att.slashCount" atts="slash"/>
+                     <specDesc key="att.signnumbers" atts="sign.num sign.numbase"/>
                   </specList>
                </p>
                <div xml:id="changeinmensuration" type="div3">
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -973,6 +973,7 @@
       <memberOf key="att.staffLoc"/>
       <memberOf key="att.typography"/>
       <memberOf key="att.slashCount"/>
+      <memberOf key="att.signnumbers"/>
     </classes>
     <attList>
       <attDef ident="dot" usage="opt">
@@ -1063,6 +1064,18 @@
           added for what we now call 'alla breve'.</desc>
         <datatype>
           <rng:data type="positiveInteger"/>
+        </datatype>
+      </attDef>
+      <attDef ident="mensur.sign.num" usage="opt">
+        <desc>Numbers following the base symbol in a mensuration sign.</desc>
+        <datatype>
+          <rng:ref name="data.MENSURATIONSIGNNUMBERS" />
+        </datatype>
+      </attDef>
+      <attDef ident="mensur.sign.numbase" usage="opt">
+        <desc>Base numbers following or beneath the base symbol in a mensuration sign.</desc>
+        <datatype>
+          <rng:ref name="data.MENSURATIONSIGNNUMBERS" />
         </datatype>
       </attDef>
     </attList>
@@ -1374,6 +1387,7 @@
       <memberOf key="att.extSym"/>
       <memberOf key="att.staffLoc"/>
       <memberOf key="att.typography"/>
+      <memberOf key="att.signnumbers"/>
     </classes>
   </classSpec>
   <classSpec ident="att.quilisma.vis" module="MEI.visual" type="atts">
@@ -1531,6 +1545,26 @@
           neume component with which it is associated.</desc>
         <datatype>
           <rng:ref name="data.EVENTREL"/>
+        </datatype>
+      </attDef>
+    </attList>
+  </classSpec>
+  <classSpec ident="att.signnumbers" module="MEI.visual" type="atts">
+    <desc>Attributes for recording numbers as part of the physical appearance 
+      of a mensuration or proportion sign.</desc>
+    <attList>
+      <attDef ident="sign.num" usage="opt">
+        <desc>Numbers following the base symbol in a mensuration sign or visual number 
+          in a proportion sign.</desc>
+        <datatype>
+          <rng:ref name="data.MENSURATIONSIGNNUMBERS" />
+        </datatype>
+      </attDef>
+      <attDef ident="sign.numbase" usage="opt">
+        <desc>Base numbers following or beneath the base symbol in a mensuration sign 
+          or base of a proportion sign.</desc>
+        <datatype>
+          <rng:ref name="data.MENSURATIONSIGNNUMBERS" />
         </datatype>
       </attDef>
     </attList>
--- a/tests/mei-Mensural/mensurations_proportions.mei
+++ b/tests/mei-Mensural/mensurations_proportions.mei
@@ -1,0 +1,101 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-model href="../../schemata/mei-Mensural.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
+<?xml-model href="../../schemata/mei-Mensural.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<mei meiversion="5.0.0-dev" xmlns="http://www.music-encoding.org/ns/mei">
+    <meiHead>
+        <fileDesc>
+            <titleStmt>
+                <title>mensuration and proportion sign tests</title>
+            </titleStmt>
+            <pubStmt/>
+        </fileDesc>
+    </meiHead>
+    <music>
+        <body>
+            <mdiv>
+                <score>
+                    <scoreDef>
+                        <staffGrp>
+                            <staffDef n="1" label="modus-com-tempore" clef.line="2" clef.shape="G" lines="5" 
+                                mensur.sign="O" mensur.sign.num="2 2" 
+                                modusmaior="3" modusminor="2" tempus="2" proport.num="4" proport.numbase="1" />
+                            <staffDef n="2" label="sesquialtera" clef.line="2" clef.shape="G" lines="5"
+                                mensur.sign="C" mensur.slash="1" 
+                                tempus="2" prolatio="2" proport.num="2" proport.numbase="1" />
+                            <staffDef n="3" label="proporcion-mayor-menor" clef.line="2" clef.shape="G" lines="5"
+                                mensur.sign="C" mensur.slash="1" mensur.sign.num="Z"
+                                modusmaior="2" modusminor="2" tempus="3" prolatio="2"/>
+                        </staffGrp>
+                    </scoreDef>
+                    <section>
+                        <staff n="1">
+                            <layer>
+                                <note dur="longa"/>
+                                <!-- Dufay example of O2 -->
+                                <mensur sign="O" sign.num="2" 
+                                    modusminor="3" tempus="2" num="2" numbase="1" />
+                                <note dur="longa"/>
+                                <!-- Cordier example of O2 -->
+                                <mensur sign="O" sign.num="2" 
+                                    modusminor="2" tempus="3" num="2" numbase="1"/>
+                                <note dur="longa"/>
+                                <!-- Busnoys -->
+                                <mensur sign="C" sign.num="3"
+                                    modusminor="2" tempus="3" num="2" numbase="1" />
+                                <note dur="longa"/>
+                                <!-- again Hothby as is <staffDef> -->
+                                <mensur sign="O" sign.num="2 2" 
+                                    modusmaior="3" modusminor="2" tempus="2" num="4" numbase="1" />
+                                <note dur="longa"/>
+                                <barLine form="dbl"/>
+                            </layer>
+                        </staff>
+                        <staff n="2">
+                            <layer>
+                                <note dur="semibrevis"/>
+                                <note dur="semibrevis"/>
+                                <!-- 3 as sesquialtera with implicit change of tempus -->
+                                <mensur sign.num="3" tempus="3" prolatio="2" num="2" numbase="1"/>
+                                <note dur="semibrevis"/>
+                                <note dur="semibrevis"/>
+                                <note dur="semibrevis"/>
+                                <mensur sign="C" slash="1" tempus="2" prolatio="2" num="2" numbase="1"/>
+                                <note dur="semibrevis"/>
+                                <note dur="semibrevis"/>
+                                <!-- Petrucci's sesquialtera sign -->
+                                <mensur sign="O" sign.numbase="3" tempus="3" prolatio="2" num="2" numbase="1"/>
+                                <note dur="semibrevis"/>
+                                <note dur="semibrevis"/>
+                                <note dur="semibrevis"/>
+                                <mensur sign="C" slash="1" tempus="2" prolatio="2" num="2" numbase="1"/>
+                                <note dur="semibrevis"/>
+                                <note dur="semibrevis"/>
+                                <!-- no implicit change of mensuration here! -->
+                                <proport sign.num="3" num="3" numbase="2"/>
+                                <note dur="semibrevis"/>
+                                <note dur="semibrevis"/>
+                                <barLine form="dbl"/>
+                            </layer>
+                        </staff>
+                        <staff n="3">
+                            <layer>
+                                <!-- propocion mayor in <staffDef> -->
+                                <note dur="semibrevis"/>
+                                <note dur="minima"/>
+                                <note dur="minima"/>
+                                <note dur="semibrevis"/>
+                                <!-- proporcion menor -->
+                                <mensur sign="C" sign.num="Z" tempus="2" prolatio="3"/>
+                                <note dur="semibrevis"/>
+                                <note dur="minima"/>
+                                <note dur="minima"/>
+                                <note dur="minima"/>
+                                <barLine form="dbl"/>
+                            </layer>
+                        </staff>
+                    </section>
+                </score>
+            </mdiv>
+        </body>
+    </music>
+</mei>


ahankinson:
`@num` and `@numbase` are defined as integer values, while `@sign.num` and `@sign.numbase` are defined as tokens. This is inconsistent. 

Were these differences noticed and discussed by the SIG?

Given the name, it would probably make sense for these values to actually be numbers instead of tokens.

kepper:
(side note from the ODD meeting: We discussed to split up the data type into multiple types, as seen on data.COLOR and other places)

annplaksin:
@ahankinson `@sign.num` and `@sign.numbase` containing tokens to deal with cases like "O 2 2" (example 2 in the linked document). While the new attributes should handle number parts of visual signs, it should be emphasized that `@num` and `@numbase` should be used for diminution or other general shifts in duration. Therefore, `@sign.num` and `@sign.numbase` are explicitly not meant to be just visual version of `@num` and `@numbase`.

I can add a remark to `data.MENSURATIONSIGNNUMBERS` making clear that it allows a token to deal with multiple numbers.

rettinghaus (dismissed):

ahankinson:
Thanks @annplaksin . I guess my question can be more succinctly put as: If they're not numbers, why call them `@num`? `@num` and `@numbase` are explicitly for encoding ratios, and a ratio of `C Z:1` doesn't really make sense.

So, there is a collision between what is expected for `@num` and `@numbase`, and what is expected for `@sign.num` and `@sign.numbase`. 

This is particularly confusing when they're used together on the `<mensur>` element:

```
<mensur sign.num="C Z" num="4">
```

What does this mean?

**Is there not another name we can use for this that would avoid this collision?** Or add a child element for `<mensur>` that is better able to describe what is on the page, rather than trying to encode everything in a single attribute that is named similarly to something else that's different?

kepper:
Hi @ahankinson, your argument matches closely an observations we had several times during today's meeting – it's causing trouble to mix visual and logical domain for meter signatures. Maybe we need to split them up. If I got that right, the "Z" is just a spanish convention for "3". With `@meter.sym` we already have an attribute that gives a visual rendition independent from the logical count, so this is not unprecedented…

source/modules/MEI.xml
```diff
@@ -2013,6 +2013,18 @@
...
+  <macroSpec ident="data.MENSURATIONSIGNNUMBERS" module="MEI" type="dt">
+    <desc>Number values as part of mensuration signs.</desc>
+    <content>
+      <rng:data type="token">
+        <rng:param name="pattern">[\d|Z]?[ [\d|Z]]*</rng:param>
+      </rng:data>
+    </content>
+    <remarks>
+      <p>Allows for multiple tokens to deal with signs containing multiple numbers, like "O 2 2" in Hothby’s “Ora pro nobis” (Codex Faenza fol. 28r).</p>
+      <p>The glyph "Z" used in 17th century Spanish manuscripts is treated as number 3.</p>
```
martha-thomae: Thanks for this PR, @annplaksin. A small correction regarding the interpretation of Z here, it should be: 
> The glyph "Z" used in 17th-century Spanish manuscripts is treated as the proportion 3:2.

Giving you either 3 _minims_ in the place of 2 (when following a `C` sign) or 3 _semibreves_ in the place of 2 (when following a cut-C `C|` sign).

pe-ro:
Rather than accounting for particular glyphs  within attributes of `<mensur>`, etc., they can be accommodated by allowing `<mensur>` to have textual content, as is already allowed by `<mensuration>`.  This would provide better separation between the visual symbols and their interpretation in the logical domain, which would still be captured in the attributes of `<mensur>`.

source/modules/MEI.xml
annplaksin: Thanks @martha-thomae !

annplaksin:
@ahankinson As far as I followed the explanations on Spanish proportions, "Z" is a glyph for "3:2". The question is, how to treat this. You can either take it as a sign or treat it (with a good portion of fancyfulness) as a number. The problem with the latter is obvious as it already raised this discussion. But the problems with treating it as a sign are: 
1. Currently `data.MENSURATIONSIGN` is very concise and works very well. With @kepper 's hint it's at least not as bad as I thought initially, but it still doesn't seem to improve `@sign` to allow multiple tokens.
2. (and structurally worse) When you allow more than one token within `@sign`, how should be made clear to which of the signs `@slash`, `@dot`, or `@orient` belongs? It's not possible anymore.

@pe-ro When allowing texutal content, did you implicate the visual content being put there? And if so, how could the different relative positions be stated? We have not only signs and numbers following each other, but on top of each other as well (see example 4 in the [linked document](https://docs.google.com/document/d/1M_CUqlvPNFiS-EI39DgggzwmX4VoLp0nvMXPp_rCbLY/edit#heading=h.gjdgxs)).

rettinghaus:
@martha-thomae If you have a good example at hand, maybe it would make sense to propose the "Z" as an addition to the [SMuFL](https://github.com/w3c/smufl) range [Medieval and Renaissance prolations supplement](https://w3c.github.io/smufl/latest/tables/medieval-and-renaissance-prolations-supplement.html) following https://github.com/w3c/smufl/pull/150?

annplaksin:
@rettinghaus See examples 7 and 8 in the linked document, they have been provided by @davidrizo.

pe-ro:
@annplaksin, I meant that visual domain info would go inside `<mensur>`.  One way to do this would be to promote `@num`, `@numbase`, and `@sign` to element status; e.g., `<num>`, `<numbase>`, and `<sign>`. Left-to-right position would be represented by encoding order; that is, `<sign>` followed by `<num>` and `<numbase>` would be different than `<sign>` preceded by `<num>` and `<numbase>`.  `<num>` and `<numbase>` would continue to indicate a "stacked" presentation; that is, num rendered above numbase.

Another possibility is to abandon these elements and employ a combination of  `<stack>` and `<symbol>` elements instead.  This more generic approach sacrifices the ability to record semantics within the content `<mensur>`, but offers greater control over visual rendition.  Of course, the meaning of the visual content in `<mensur>` could still be captured in `@num` and `@numbase`.

ahankinson:
I have taken some time to think about this issue. I find it a difficult one. First, I'll talk about a general discomfort with this, and then I'll get back to some specifics about the implementation here.

One the one hand, I do see the need to capture the visual aspect of the musical symbols, particularly in ancient notations where sometimes the visuals are a key element to deciphering the "meaning" of the notation, since the knowledge of the notation has conflicting interpretations, or is simply lost to time.

On the other hand, adding textual indications for precise visual clues will be a never-ending task, particularly for material that is primarily manuscript-based. If we add a "Z" value, does this mean that it can be rendered with any font? What if the font chosen to render it doesn't replicate the shape of the font used in the original? What if one set of copyists used a different shape "Z"? Or used a symbol that looks more like an "S"? Replicating the precise visuals is a very hard thing to do, **particularly** when the visuals are a key element to deciphering the meaning of the notation.

MEI isn't a visual encoding format -- there are much better tools out there for doing this, and MEI integrates with all of them to some extent. On the first level, there is SMuFL, where a glyph for indicating the general visual shape can be used, and you can point to that glyph for specific symbols.

MEI also has some rudimentary symbol drawing capabilities on its own, where you can 'create' a symbol (using the `symbolDef` functionality) and refer to it. This is generally OK for simple shapes.

If you want 'diplomatic facsimile' functionality for more complex shapes, MEI can embed SVG. This is good for precisely capturing an exact shape as a vector graphic, suitable for rendering (as closely as possible) the "original" look of the shape while still retaining the flexibility of positioning, layout, etc.

Finally, if you want to record the absolute appearance of the glyph, the graphic / zone bounding boxes can refer to a digital image of the original with a specific area of interest highlighted or selected.

All of these allow much more precise approaches to capturing the visual appearance of a symbol than trying to figure out how to render a "Z" based on a simple ASCII value. 

(Digressing a bit: The "Z" you see isn't actually the letter "Z" -- it's a symbol that means a ratio of `3:2` that just so happens to look like a "Z". This is fine from an encoding level -- you do need some ASCII character to represent the value, and we do this all the time with symbols like #, b, x, etc. -- but just because it has a **visual similarity** to the letter Z doesn't necessarily mean it has the **semantic meaning** of the letter Z. It's a bit like trying to encode the visuals of a C clef with the letter C! This is also why you rarely see these sorts of values in visual domain attributes; typically, the visual domain attributes are used for rendering hints (stem right / left, position offsets, colours, etc) than capturing specific visual shapes -- we don't have a "#" used in a visual domain attribute for accidentals, for example.)

---

I agree wholeheartedly with @pe-ro 's suggestions above. It seems like a lot of information is getting "stuffed" into attribute values. If these are complex symbols that need different attributes to accurately describe them, then breaking them out into elements makes sense. I'm still not crazy about them being called "num" and "numbase" since they're not numbers -- "@sign.upper" and "@sign.lower" maybe?

pe-ro:
@ahankinson, I think if we move the visual aspects of mensuration signs/symbols to elements, `@num` and `@numbase` can become actual numbers again.  They can express the logical, semantic numerical relationship represented by the mensuration sign, whether it actually contains number (-like?) glyphs or not.

annplaksin:
Please note, the following is my personal suggestion in this discussion while this PR implements the IG's agreed proposal.

@pe-ro Thanks for the further explanation. Actually, as I started the discussion in the IG, adding a new element for signs was my initial direction of thought, but I was not sure about handling the relative positions.

I thought a while about the suggestion and it seems convincing to me, although I would not put the visual domain info into textual content since attribute values can be controlled more easily. In the case of base mensuration signs, this has proven useful already.
How about something like this? 
* Adding an element `<sign>` or `<mensSign>` as a child element to `<mensur>` (better naming options are highly welcome as I am very bad at things like that)
* Make "Z" a value of `data.MENSURATIONSIGN` to allow it within `@sign`

And following that, I see two ways to proceed:

1. Either stick with `@sign.num` and `@sign.numbase`, allow only positive integers as values, and use a Schematron rule to make sure either `@sign`, `@sign.num` or `@sign.numbase` is used within `<sign>`. (Please note that I would not suggest something `.vis` to make clear, that it's more than a rendering hint.)
2. Or get rid of `@sign.numbase` and invent some attribute like `@stacked` that could contain "upper" or "lower" (again, naming options are welcome) and hope that we never come across stacks of more than two signs (applies to option 1 too). And of course, add Schematron to allow either `@sign` or `@sign.num`.

This would lead to two direct follow-up questions:

* Should the element only be used within `<mensur>` or should it be allowed within `<proport>` as well?
* Would it be necessary to use `<mensur>` within `<staffDef>`?

martha-thomae:
Hi! Sorry for arriving late to the party.

In the following comments I will be getting back to some of the previous messages.

@ahankinson:

> Thanks annplaksin . I guess my question can be more succinctly put as: If they're not numbers, why call them `@num`? `@num` and `@numbase` are explicitly for encoding ratios, and a ratio of `C Z:1` doesn't really make sense.
>
> So, there is a collision between what is expected for `@num` and `@numbase`, and what is expected for `@sign.num` and `@sign.numbase`.

The new `@sign.num` and `s@ign.numbase` are just meant for numbers: either a single number or a list of numbers. And, in this case, Z is also a "number"—standing for the ratio 3:2.
The `@sign.num` and `@sign.numbase` are meant to represent the numeric part (the numeric ratio) following the main mensuration sign (the `C` or `O` with their corresponding slashes and dots).

> This is particularly confusing when they're used together on the `<mensur>` element:
> 
> ```
> <mensur sign.num="C Z" num="4">
> ```
> 
> What does this mean?

Well, if using them together, `C` would never appear in the `@sign.num` because `C` is the main mensuration sign. This was misleading in the original document (sorry about that!). I am attaching [here the new document](https://docs.google.com/document/d/1d2Zu5O5gH0jMSwI8-ClpEYAfyBFErYraeWFjuYp2Kok/edit?usp=sharing) and the changes can be seen in pink highlighting (first page and end of the table).

Your example will look like: `<mensur sign=“C” sign.num=“Z” num=“4”/>`. And it will mean that the `Z` number has an effect of making the voice going four times as fast compared to the other voices (`num=“4”` encodes the actual effect of the proportion, the augmentation or diminutio of the voice).

> **Is there not another name we can use for this that would avoid this collision?** Or add a child element for `<mensur>` that is better able to describe what is on the page, rather than trying to encode everything in a single attribute that is named similarly to something else that's different?

It wouldn’t be encoding something else that is different (I think). It—the pair of `@sign.num` and `@sign.numbase`—will still encode just numbers, but visual numbers as part of a time signature. And `@num` and `@numbase` would encode their actual proportional effect (augmentation or diminution).

And I agree with @kepper that:
> With `@meter.sym` we already have an attribute that gives a visual rendition independent from the logical count, so this is not unprecedented…

@annplaksin also mentioned that:
> On Spanish proportions, "Z" is a glyph for "3:2". The question is, how to treat this. You can either take it as a sign or treat it (with a good portion of fancifulness) as a number. The problem with the latter is obvious as it already raised this discussion. But the problems with treating it as a sign are:
> 2. When you allow more than one token within `@sign`, how should be made clear to which of the signs `@slash`, `@dot`, or `@orient` belong? It's not possible anymore.

I completely agree! That is why we want to treat `Z` as a number (in `@sign.num`) rather than a sign (encoded in `@sign`) as shown in the [new document](https://docs.google.com/document/d/1d2Zu5O5gH0jMSwI8-ClpEYAfyBFErYraeWFjuYp2Kok/edit?usp=sharing).

martha-thomae:
@pe-ro:
> Rather than accounting for particular glyphs within attributes of `<mensur>`, etc., they can be accommodated by allowing `<mensur>` to have textual content, as is already allowed by `<mensuration>`.

I think this has been proposed before (someone can correct me if I am wrong), doing it kind of similar to Humdrum. Something like using `C|3/2` to encode the equivalent to `@sign=C @slash=1 @sign.num=3 @sign.numbase=2`. However, I think the reason why something like this was discarded had to do with being difficult to parse and not a very attractive solution for Verovio (again, please correct me if I am wrong).

martha-thomae:
@ahankinson:

Thanks for your detailed answer [here](https://github.com/music-encoding/music-encoding/pull/848#issuecomment-895288472). 

I was thinking, is the problem just that `Z` is being considered in `@sign.num`?

Because, as said before, that `Z` is actually standing in place of the ratio `3:2` (drawing 3 over 2 and combining the shape of these two numbers does look like a Z). So, maybe we can just not encode the `Z` at all, and just keep `@sign.num=3` and `@sign.numbase=2` and leave the rendering of this as a `Z` to something else(?)
This proposal is also included in the table on the [attached document]( https://docs.google.com/document/d/1d2Zu5O5gH0jMSwI8-ClpEYAfyBFErYraeWFjuYp2Kok/edit?usp=sharing) (see examples 7 and 8, second proposal).

Is this possible? I am not very familiar with SMuFL.

Will removing the `Z` issue solve the main issue completely? We still would allow for `sign.num` to encode a list of numbers (see example 2 of the attached [document]( https://docs.google.com/document/d/1d2Zu5O5gH0jMSwI8-ClpEYAfyBFErYraeWFjuYp2Kok/edit?usp=sharing)).

martha-thomae:
I know there is still another comment from Perry and the proposal given by Anna based on this, but I would rather wait to get an answer about the previous comments before getting into even more detail. Thanks!

martha-thomae:
> Will removing the `Z` issue solve the main issue completely? We still would allow for `sign.num` to encode a list of numbers (see example 2 of the attached [document](https://docs.google.com/document/d/1d2Zu5O5gH0jMSwI8-ClpEYAfyBFErYraeWFjuYp2Kok/edit?usp=sharing)).

If the issue is the **data type** of `@sign.num`, then we could do what was suggested by @kepper and split up the data type into multiple types. These types being:
- integer number
- list of tokens (even better if it is a list of integer numbers)
- and the value of `Z`

pe-ro:
@martha-thomae, when I said "textual content", I should have said just "content". See my earlier response about this.  Plain text content, as you've provided in your example, is an option, but structured content, using `<stack>`, `<num>`, `<numbase>`, etc. would be better; e.g., easier to parse..  Equally importantly, however, content would represent the visual signs, numbers, whatever as found in the document, while attributes would be used for the logical domain.  If the visual signs confirm the logical qualities, then the visual stuff doesn't have to be recorded as it's redundant.  Essentially, I'm saying that `@sign`, `@sign.num`, and `@sign.numbase`, and other visually-oriented attributes can be moved to the content of `<mensur>`.  This would have another advantage in the sense that the new elements could be allowed within editorial markup such as `<rdg>`, `<sic>`, `<corr>`, etc. to better capture scribal interventions.

martha-thomae:
Ok, I think I got it! Wanted to make sure because I did see your other comment too.
So, the idea is to use the `<mensur>` attributes to provide the logical-domain information and its child elements to provide the visual-domain information.

Ok, so taking a few of the examples in the document, would you say that the following encoding complies with your proposal @pe-ro ?

```xml
<mensur modusminor="2" tempus="3" num="2" numbase="1">
 <sign sign="C"/>
 <num num="3"/>
</mensur>
```

Screen Shot 2021-08-17 at 7 56 33 PM

(basically move the blue attributes into the children elements and keep the green ones as attributes)


```xml
<mensur modusmaior="3" modusminor="3" tempus="2" num="4" numbase="1">
 <sign sign="O"/>
 <num num="2"/>
 <num num="2"/>
</mensur>
```

Screen Shot 2021-08-17 at 7 56 39 PM


```xml
<mensur tempus="3">
 <sign sign="O"/>
 <numbase num="3"/>
</mensur>
```

Screen Shot 2021-08-17 at 7 56 58 PM


And I guess that in this case we will encode the `Z` as a sign rather than a number, so the encoding would be:

```xml
<mensur modusmaior=“2“ modusminor=“2“ tempus=”3” prolatio=“2“>
 <sign  sign=”C” slash=“1“/>
 <sign sign=”Z”/>
</mensur>
```

Screen Shot 2021-08-17 at 8 11 25 PM

-----

Here I have the same question as Anna: Would it be necessary to use `<mensur>` within `<staffDef>`?
`<staffDef>` usually provides the same attributes as `<mensur>` (and the other bunch of attributes to encode clefs, key signatures, and more)

Would it be bad to have `<sign>`, `<num>`, and `<numbase>` as direct children of `<staffDef>`?

martha-thomae:
> This would have another advantage in the sense that the new elements could be allowed within editorial markup such as `<rdg>`, `<sic>`, `<corr>`, etc. to better capture scribal interventions.

I like this point!

martha-thomae:
> If the visual signs confirm the logical qualities, then the visual stuff doesn't have to be recorded as it's redundant.

I am not completely convinced about this, @pe-ro. Because this implies that if I have a logical `@num` attribute in `<mensur>` and I have no `<num>` child, then that `@num` might be rendered visually in the mensuration sign. Am I right in this interpretation?

So, I have an example for this. The mensuration sign **cut C** (`<mensur sign="C" slash="1">`) makes the voice go twice as fast compared to the other voices, so it requires encoding a `@num=2` (logical domain). However, there is no actual `2` appearing in the mensuration sign. 
According to the quoted text above, since there is no `<num>`, then the visual domain must coincide with the logical domain and, therefore, a `2` must be rendered next to the main mensuration sign.

pe-ro:
I think this is going down the right path.  I'll check more closely tomorrow, after the margaritas have worn off. 😀

lpugin:
> Here I have the same question as Anna: Would it be necessary to use `<mensur>` within `<staffDef>`?
`<staffDef>` usually provides the same attributes as <mensur> (and the other bunch of attributes to encode clefs, key signatures, and more)
> 
> Would it be bad to have `<sign>`, `<num>`, and `<numbase>` as direct children of `<staffDef>`?

I think it would. `<sign>` and it would be much better to have them within a `<mensur>` because they "belong" to it. 

> I think this has been proposed before (someone can correct me if I am wrong), doing it kind of similar to Humdrum. Something like using C|3/2 to encode the equivalent to @sign=C @slash=1 @sign.num=3 @sign.numbase=2. However, I think the reason why something like this was discarded had to do with being difficult to parse and not a very attractive solution for Verovio (again, please correct me if I am wrong).

@pe-ro, let me correct you. Nothing is meant to be designed for Verovio. When parsing issues are brought here, there  are general parsing issues - assuming one purpose of MEI encodings is to be parsed for processing. I don't think Verovio is any worse in parsing MEI than most of other tools that do it, and if it does, it should be improved and by no means we should change anything in MEI for this.

What I did propose, however, is to use the same approach we have in many places elsewhere, namely the `.ges` variant of attributes. A reason for this proposal is that I am not fully convinced by the hard line drawn here between the logical and the visual domain. As far as I am aware, there is nowhere where we make such a distinction in MEI - simply because music notation is a visual system and because MEI primarily encodes this and to make it processable. But maybe am I wrong. 

Now if there is a consensus that the right direction is to do something like
```xml
<mensur modusmaior="3" modusminor="3" tempus="2" num="4" numbase="1">
 <sign sign="O"/>
 <num num="2"/>
 <num num="2"/>
</mensur>
```
that does not look great to me. Having `<num>` plus two types of `@num` looks suspicious. Also, I remember having heard that putting text that actually appears on the source within an attribute is a bad practice. Of course, there are many places in MEI where we do this, but mainly because we want the information to be structured and more easily processable (e.g., `<meterSig num="4" numbase="4">`. However, here `<sign>` and `<num>` would represent exclusively a visual element (if I understand correctly), and the processability of their content is much lower. Why not doing
```xml
<mensur>
  <sign>0</sign>
  <sign>2</sign>
  <sign>2</sign> 
</mensur> 
```

Or even better
```xml
<mensur>
  <sign>0 2 2</sign>
</mensur> 
```

We can then assume that two `<sign>` elements have to be stacked. 
```xml
<mensur>
  <sign>0</sign>
  <sign>3</sign>
</mensur> 
```
SMuFL codes would be used for specific symbols (e.g., Z)

ahankinson:
Hi @martha-thomae !

I won't quote it here, but this is responding to [your comment](https://github.com/music-encoding/music-encoding/pull/848#issuecomment-900630540) above. I'll start by laying out a few of my assumptions. I haven't  thought through @pe-ro and @lpugin 's comments, so it's probably best to assume that I'm talking about the original proposal, and not their most recent suggestions.

My main concern is that what is being encoded, and the motivations for its capture, is not entirely clear to me so I'm trying to fill in my own gaps of understanding. As I understand it, what this change is trying to do is to find a way of capturing how a mensuration sign appears visually, separately from how it functions on the interpretation of the music. 

Put another way, if you did not care about the "original appearance" of the mensuration sign, you could just ignore any of the `@sign.[something]` attributes, since any 'actionable' data that affects the timing and interpretation of the mensuration  (for, e.g., playback or rhythmic analysis) is already there encoded in the other attributes. So this is purely visual information, intended for a graphical notation renderer to "do something" with. Is this an accurate interpretation?

That interpretation is what leads me to wondering about what the envisioned purpose of this data will be. Why is this being encoded? What are the specific use cases that this solves?

Again, an assumption I made is that one of the primary use cases of capturing this data is to create a 'diplomatic facsimile' of the notation -- creating a digital rendering of the notation, but as close to the original appearance as possible. If this is the reason, then I think the solution outlined in the IG's proposal goes very far in trying to enumerate the possible values that might appear, but not far enough in describing precisely how they look, or interact with each other. There is a lot of grey area where it is not clear how this data should be interpreted to produce consistent results.

To try and illustrate what I mean further, let's take the example of the "C" mensuration value in `@sign`. This, of course, doesn't mean that it's actually a "C", but this is the **closest visual approximation** of the symbol that we have in ASCII. The actual symbol is an imperfect circle; for the perfect circle we use the "O". This is perfectly fine -- we need some mechanism to 'capture' this data, and as long as the relationship between "C" and 𝇋 is well understood, we can proceed with visually representing it.

Systems that render this know to substitute "C" with 𝇋, or "C" with a dot as 𝇊. These are formalized visual symbols that we can use to stand in place of the ASCII values that are used to encode it. Also, since they are in SMuFL, Verovio can use that to render the symbol easily -- there is a direct relationship between `<mensur sign="C">` and SMuFL glyph `U+E915`. 

Likewise, values like "Z" are not actually the letter "Z" -- they are symbols that look like the ASCII letter Z, so we are tempted to just type that on our keyboard and move on. However, that symbol actually "means" a ratio of 3:2 -- and, as I understand it, the ratio of "3:2" is already encoded, in the logical domain, using `@tempus` and `@prolatio` attributes. So "Z" is simply trying to encode the **appearance** of the sign, but, unlike the "C" and "O" values for mensuration signs, we don't have well-understood, formalized symbols to substitute for them. 

So if you provide a value of "Z" to Verovio, what should its behaviour be? Should it just render the letter "Z"? Does it matter what font is used? Or how big it is? Or whether it appears next to the mensuration sign? How big is it relative to that sign? You can substitute any other value and combination of `@sign.num` and `@sign.numbase` and you can probably see that it becomes very hard for a "generic" renderer to modify its behaviour based on imprecise, independent pieces of visual data, _particularly if the person doing the coding of the software has no visual reference to the original that the data is based on!_

So what is not immediately clear is how this same process will translate to all the various permutations of the `@sign`, `@sign.num`, and `@sign.numbase` for providing the same visual substitution methods. There does not seem to be any formalized interpretation of how these things will interact with each other when rendered. For example it seems that, `@sign.numbase` can occur both positioned under `@sign` (in the fourth example) or under `@sign.num` (eighth example, I think?). Also, in the third example given in the original document, why isn't `@sign="3"` used, instead of `@sign.num="3"`, since it's obviously the "main" mensuration sign? Can `@sign.num` appear on its own? Why isn't the 3 its own mensuration sign? Is it because "3" is a number, while "C" is not? What then about our friend "Z"?

Can the visual relationships of these different bits of data be fully enumerated to cover all cases so that we can formalize a range of glyphs to accurately represent all possibilities? My suspicion is no, since we're dealing with a corpus of music that covers a vast geographic and temporal area (hundreds of years?), so trying to standardize these appearances will always lead us to finding new cases where they wrote a symbol that we can't actually capture. I would be interested to hear contrary evidence, though.

There seem to be many, many different ways the `@sign.num` and `@sign.numbase` can appear. `@sign.numbase` seems to exist only to encode a number that appears, positionally, UNDER something else. It doesn't actually mean "the base of the numerical ratio" -- it means "the thing below another thing". `@sign.num`, in the same way, seems to mean "the thing above `@sign.numbase` OR the thing next to `@sign`." Again, this isn't really well defined. 

**If your goal is to capture a consistent visual representation, but you are using small chunks of separate data with imprecise positional and visual data encoded in it, it will be very difficult to get a consistent interpretation out of it.**

Meanwhile, MEI already has several more capable methods to precisely capture visual appearance, depending on how you want it used:

 1. Get a formalized variant of the sign in SMuFL and use the existing `<mensur glyph.num>` to reference it (also, using the unique codepoint values of the symbol means that you can have a precise indication of what sign was used, which would be a far better summary than trying to combine `sign/sign.num/sign.numbase` to determine the same); 
 2. Figure out how to use the `<symbol>` elements in MEI;
 3. Figure out how to draw and reference the shape in SVG, and use that;
 4. Use the existing `@facs` and zone functionality to point to an exact image of the symbol

So, to sum it all up:

 - The goal of this change seems to be to capture visual appearance of symbols separate from their functional interpretation, for the purposes of accurately capturing the original symbols used in the sources;
 - The proposal suggests encoding visual appearance using several attributes that hold ASCII characters, but the methods and procedures of interpreting these ASCII values **back** to a consistent visual appearance is not well-defined (a "lossless round-trip" from visual to MEI and back to the exact same visual);
 - MEI has several existing methods for capturing precise visual data; there may need to be some integration changes to make this happen (like adding symbols to SMuFL or ensuring SVG can be used for mensuration signs), but if the goal is to preserve accurate visual representations for diplomatic facsimiles, these mechanisms might be better suited for the goal;
 - `@sign.num` and `@sign.numbase` seem to refer to things positioned "over", "under", or "next to"  other things, and are not actually numerical or numerical bases. It would be good if we can make this more clear with the names of the things (elements or attributes) used to capture this data. Even better, it would also be design this change to fit in with all the other mechanisms in MEI used for capturing symbol positioning by, for example, elevating the different components to elements so that they can be independently placed and addressed, which I think is the direction the latest suggestions are taking.

*phew* Sorry this was so long-winded. I hope that is clearer?

pe-ro:
@lpugin, I don't remember claiming that anything in MEI is specifically designed with Verovio in mind or that Verovio is bad at parsing XML.  Of course you're right -- it's exactly the opposite on both points.  I'm sorry if my language wasn't clear.

As you point out, there are several options for the content of `<mensur>`.  I'm fine with any option that makes clear the distinction between the visual and logical/analytical domains.

```
<mensur>
  <sign>0</sign>
  <sign>3</sign>
</mensur> 
```
is fine.  Just to clarify -- in this scenario the space-separated content of `<sign>` would be rendered side-by-side, while multiple signs would be rendered top-to-bottom?  This is opposite from using `<stack>`.  For example,

```
<mensur>O 3</mensur>
```
would result in the circle and the 3 being rendered beside each other, while
```
<mensur>
  <stack>O 3</stack>
</mensur>
```
would place the circle atop the 3.  But the encoding order isn't significant, we just need to be clear about how it works.

Of course, `<symbol>` with an appropriate `@glyph` attribute could be used instead of the plain text 'O'.

The advantage of `<stack>` is that it offers control over the delimiter used between the items to be stacked as well as control over the left-to-right alignment of the stacked items.  Also, it already exists in MEI, so there's no need to invent or explain new conventions for the content of `<sign>` or even to create a new `<sign>` element.  In fact, the lack of a `<sign>` element is a good thing, I believe, in that it makes clear that there is (are?) no semantics in the content of `<mensur>` -- the content is present solely to describe the visual content of the document.

I believe this fits into the framework @ahankinson described above.  It's analogous to the situation with `<meterSig>` and `@meter...` attributes on `<staffDef>` and `<scoreDef>`.  The attributes  on `<staffDef>` are/can be interpretative of those on `<meterSig>`.  The difference between `<meterSig>` and `<mensur>` is that one uses attributes and the other uses (will use) content.  The main advantage of using content within `<mensur>` is that, unlike attributes, it can be wrapped with editorial and/or transcriptional markup.  For example,
```
<mensur>
  <stack>O <add hand="#h1">3</add></stack>
</mensur>
```
indicates that the '3' was added later in the scribal hand known as "h1", something that's (perhaps) more frequent and more important in the mensural repertoire than in CMN.

martha-thomae:
Thank you, @ahankinson! [This reply](https://github.com/music-encoding/music-encoding/pull/848#issuecomment-900984221) was very informative. I will be answering in the following comments:

> My main concern is that what is being encoded, and the motivations for its capture, is not entirely clear to me so I'm trying to fill in my own gaps of understanding. As I understand it, what this change is trying to do is to find a way of capturing how a mensuration sign appears visually, separately from how it functions on the interpretation of the music.

Yes, that is correct!

To give a bit of background:

Numbers can have two kinds of effects: 

- **proportional**, making the voice go faster/slower. We encode these in `num` and `numbase`.
- **mensuration change**, changing the values of `modusminor`, `tempus`, or `prolatio` (this is, changing the rhythmic relation between the different note values).

Right now `@num` and `@numbase` are being used both for capturing the proportional effect of the numbers (e.g., going twice as fast compared to other voices) and also to render them as if they were part of the mensuration sign. However, the presence of these numbers in the mensuration sign doesn't imply that they have a proportional effect as indicated above (e.g., the `@num=3 @numbase=2` normally don't imply that the voice is going 1.5 times faster, it has a "change in mensuration effect" in this case). And, in a similar way, sometimes you have a proportional effect being conveyed by other means (than numbers) in the mensuration sign (e.g., the use of a `@slash=1` means that the proportional effect is `@num=2 @nubmase=1`—going twice as fast—but there are no actual numbers in the sign).

We want to capture the numbers in the mensuration sign separately (with `@sign.num` and `@sign.numbase`) from their actual "proportional" (`@num` and `@numbase`) and "mensuration" (`@modusminor`, `@tempus`, and `@prolatio`) effect. The latter is already possible, the former isn't.

---
Here you can see the MEI encoding that I need now to have the sign `O22` displayed. With this encoding, I cannot indicate any proportional effect those `2 2` numbers have because I have already used `num` to indicate the numbers in the sign.

Screen Shot 2021-08-23 at 11 51 16 AM

---

**In summary, what we are suggesting is to move one of the effects `@num` and `@numbase`, the visual rendering, to a pair of visual domain attributes `@sing.num` and `@sign.numbase`. So that `@num` and `@numbase` can be used solely for encoding the actual proportion effect (logical domain) that these numbers have.**

---

> Put another way, if you did not care about the "original appearance" of the mensuration sign, you could just ignore any of the `@sign.[something]` attributes, since any 'actionable' data that affects the timing and interpretation of the mensuration (for, e.g., playback or rhythmic analysis) is already there encoded in the other attributes. So this is purely visual information, intended for a graphical notation renderer to "do something" with. Is this an accurate interpretation?

Yes, that is correct.

martha-thomae:
> So what is not immediately clear is how this same process will translate to all the various permutations of the `@sign`, `@sign.num`, and `@sign.numbase` for providing the same visual substitution methods. There does not seem to be any formalized interpretation of how these things will interact with each other when rendered. For example, it seems that @sign.numbase can occur both positioned under `@sign` (in the fourth example) or under `@sign.num` (the eighth example, I think?). 
>
> Can the visual relationships of these different bits of data be fully enumerated to cover all cases so that we can formalize a range of glyphs to accurately represent all possibilities? My suspicion is no, since we're dealing with a corpus of music that covers a vast geographic and temporal area (hundreds of years?), so trying to standardize these appearances will always lead us to find new cases where they wrote a symbol that we can't actually capture. I would be interested to hear contrary evidence, though.
> 
> There seem to be many, many different ways the `@sign.num` and `@sign.numbase` can appear. `@sign.numbase` seems to exist only to encode a number that appears, positionally, UNDER something else. It doesn't actually mean "the base of the numerical ratio" -- it means "the thing below another thing". `@sign.num`, in the same way, seems to mean "the thing above `@sign.numbase` OR the thing next to `@sign`." Again, this isn't really well defined.
>
> **If your goal is to capture a consistent visual representation, but you are using small chunks of separate data with imprecise positional and visual data encoded in it, it will be very difficult to get a consistent interpretation out of it.**

I think there might be some standard behind the scenes. I am afraid to make such a bold statement, but I think we can generalize the following:
- If `@sign.num` exist, `@sign.numbase` will be rendered below the `@sign.num`. 
- If there is no `@sign.num`, then `@sign.numbase` will be rendered below the `@sign`.

  Part of this proposal consisted of finding examples that behave otherwise and, so far, we did not find them.

- Regarding this proposal in particular: When a number appears by itself (`@sign.num`), we will encode it in `<mensur>` if it has an effect on the mensuration, and we will encode it in `<proport>` when the only effect regards the proportion (see comment above regarding the effects of these numbers).
 
> Also, in the third example given in the original document, why isn't `@sign="3"` used, instead of `@sign.num="3"`, since it's obviously the "main" mensuration sign? Can @sign.num appear on its own? Why isn't the 3 its own mensuration sign? Is it because "3" is a number, while "C" is not? What then about our friend "Z"?

The main mensuration sign is always `C` or `O` with the corresponding slashes or dots. Basically, all that we have allowed as valid values in [mensur.vis](https://music-encoding.org/guidelines/dev/attribute-classes/att.mensur.vis.html) as of now.

Numbers are not part of the main mensuration sign, they are usually a ratio—with or without an explicit denominator—that accompanies the main mensuration sign or modifies it. For example, when the `3` is not next to the mensuration sign, usually, it is acting as a modifier of the previous mensuration sign that appeared in the voice. 
Because of this, In this proposal, we are not including numbers as part of the main mensuration sign (in [mensur.vis](https://music-encoding.org/guidelines/dev/attribute-classes/att.mensur.vis.html)), but as `sign.num` and `sign.numbase` indicating a "visual time signature" that accompanies or modifies the main mensuration sign.

> What then about our friend "Z"?

Also a number modifying the `@tempus` or `@prolatio` attributes of `<mensur>`.

Something important to keep in mind is that, while the effect of that `Z` is encoded in the change in `tempus` or `prolatio`, those values of `tempus` (and `prolatio`) exist for other mensuration cases that do not involve `Z`.

I think that the `Z` is a problematic case. The other encoding proposed for it was `@sign.num=3 @sign.numbase=2` because that is what the `Z` sign is (a 3 over 2 merged into a single figure). We might encode it as those `@sign.num=3 @sign.numbase=2` and just use a SMUFL code to explicitly add it in the rendering as you suggested later in [1].

martha-thomae:
> So, to sum it all up:
> 
> * The goal of this change seems to be to capture visual appearance of symbols separate from their functional interpretation, for the purposes of accurately capturing the original symbols used in the sources;

True!

> * The proposal suggests encoding visual appearance using several attributes that hold ASCII characters, but the methods and procedures of interpreting these ASCII values **back** to a consistent visual appearance is not well-defined (a "lossless round-trip" from visual to MEI and back to the exact same visual);

I understand. I think the rules given above (and also found in the next point) would be enough for having a lossless round-trip.

> * MEI has several existing methods for capturing precise visual data; there may need to be some integration changes to make this happen (like adding symbols to SMuFL or ensuring SVG can be used for mensuration signs), but if the goal is to preserve accurate visual representations for diplomatic facsimiles, these mechanisms might be better suited for the goal;

We don't really want to capture precise visual data (how big the symbol and what type of 'Z'). Just to capture the fact that there are numbers next to the symbol (or below the symbol) and be explicit about the list of these numbers. Following rules like:
- If `@sign.num` exist, `@sign.numbase` will be rendered below the `@sign.num`. 
- If there is no `@sign.num`, then `@sign.numbase` will be rendered below the `@sign`.
- `@sign.num` can contain a list of numbers (and they are to be rendered one after the other).
(And we can forget about the `Z` thing if it is too problematic, we can do `@sign.num=3 @sign.numbase=2` and map that to a SMUFL code.)

> * `@sign.num` and `@sign.numbase` seem to refer to things positioned "over", "under", or "next to"  other things, and are not actually numerical or numerical bases. It would be good if we can make this more clear with the names of the things (elements or attributes) used to capture this data. Even better, it would also be design this change to fit in with all the other mechanisms in MEI used for capturing symbol positioning by, for example, elevating the different components to elements so that they can be independently placed and addressed, which I think is the direction the latest suggestions are taking.

Yes, that is the latest suggestion. However, I am a bit worried about backward compatibility. Lots of people will need to change their encodings coming from version 4 to version 5. The changes with `@sign.num` and `@sign.numbase` are minimal.

I hope I am not missing anything, @ahankinson. 
Thanks for all the options provided for capturing the exact visual appearance. I think all of them are indeed more appropriate. However, I think the important thing is not capturing the exact visual appearance but the fact that there are numbers next to the symbol (or below the symbol) and be explicit about the list of these numbers. (`Z` seems to be rather an exception, and I think that for that case, in particular, the other options you presented are more appropriate).

ahankinson:
Thanks! Just a quick question to respond to this:

> If @sign.num exist, `@sign.numbase` will be rendered below the `@sign.num`.
> If there is no `@sign.num`, then `@sign.numbase` will be rendered below the `@sign`.
> `@sign.num` can contain a list of numbers (and they are to be rendered one after the other).

Does this mean that nothing would be rendered if no `@sign.num` or `@sign.numbase` are present?

martha-thomae:
> Does this mean that nothing would be rendered if no `@sign.num` or `@sign.numbase` are present?

Oops!
Only the `@sign` (if present).

And, regarding the rendering of numbers. Yes, if no `@sign.num` or `@sign.numbase` are present, there will be no numbers being rendered.

martha-thomae:
Sorry @lpugin and @pe-ro, the misunderstanding about Verovio and the parsing of XML came from my side.

> Now if there is a consensus that the right direction is to do something like
> 
> ```
> <mensur modusmaior="3" modusminor="3" tempus="2" num="4" numbase="1">
>  <sign sign="O"/>
>  <num num="2"/>
>  <num num="2"/>
> </mensur>
> ```
> 
> that does not look great to me. Having `<num>` plus two types of `@num` looks suspicious. Also, I remember having heard that putting text that actually appears on the source within an attribute is a bad practice.

Thanks for pointing this out, @lpugin. 

I like the solution you propose here for rendering signs after one another and for stacking signs:

> Or even better
> 
> ```
> <mensur>
>   <sign>0 2 2</sign>
> </mensur> 
> ```
> 
> We can then assume that two `<sign>` elements have to be stacked.
> 
> ```
> <mensur>
>   <sign>0</sign>
>   <sign>3</sign>
> </mensur> 
> ```
> 
> SMuFL codes would be used for specific symbols (e.g., Z)

I still have to check @pe-ro's proposed solution using `<stack>`.

**One question: Even though this solution seems to solve the issue of having a distinction between the encoding of visual and logical domain numbers and also seems to solve the `Z` issue, what about backward compatibility?**

With the original proposal, people would only need to change their encodings regarding `num/numbase`. With this new proposal, they will have to re-encode the mensuration signs within the `mensur` element, which also means that all what they had encoded in `staffDef` won't work anymore. They will have to add a `mensur` child to the `staffDef` (which is a change we reverted a while ago [here](https://github.com/music-encoding/music-encoding/pull/676)—I cannot find the comment that justified the revert, will look for it). And then, move all the mensuration-related attributes of `staffDef` into the child `mensur` attributes or into its children depending on the nature of the attributes.

martha-thomae:
I have another question @lpugin and @pe-ro:

Considering the new proposal, either multiple `<sign>` elements within `<mensur>` or using the `stack` element. How would you handle the basic mensuration signs like **dotted circle** (currently encoded as `<mensur sign="O" dot="true">`) or **cut-C** (currently encoded as `<mensur sign="C" slash="1">`)?

pe-ro:
> How would you handle the basic mensuration signs like dotted circle (currently encoded as `<mensur sign="O" dot="true">`) or cut-C (currently encoded as `<mensur sign="C" slash="1">`)?

One would use either a reference to SMuFL code point --
```
<symbol glyphnum="#[SMuFL code point]"/>
```
a reference to a zone (optionally containing SVG markup that describes the symbol) 
```
<symbol facs="#[ID of zone element]"/>
```
or a reference to a symbol defined in an MEI `<symbolDef>` element 
```
<symbol altsym="#[ID of symbolDef element]"/>
```
depending on project requirements and/or resources.

Another possibility is to establish some kind of convention for textual content; e.g. where "O." =  circle-dot or "C|" = cut-C, but I don't recommend this.

martha-thomae:
@pe-ro, I see. Ok, these are also the options that @ahankinson mentioned before.

My question is, **aren't we getting too far away from the current specs?** I can imagine some resistance on the IG over this solution given how different it is from the way we have been encoding pieces so far. At the same time, it allows for more flexibility in our encodings. (Between these, I would go for the SMUFL or the symbol definition options, not the `@facs` one because we want the symbol getting rendered not to make a reference to the region where it is, what if there is no access to the facsimile?)

I guess I will give it some thought and we will discuss it in our next meeting. Thanks!

pe-ro:
> My question is, aren't we getting too far away from the current specs?

On the contrary, as @ahankinson points out, methods for dealing generically with signs/symbols are already present in MEI.  If anything, using attributes with ever-increasing specificity goes against the basic tenets of MEI.  Besides, backward compatibility for its own sake isn't useful -- it locks one into a specific approach just because "that's the way we've always done it" and stifles innovation.

Not all of the options of what the `<symbol>` element refers to are appropriate for all uses/users.  Since they all employ optional attributes, none is mandated.  Neither are they mutually exclusive, so, while it's not recommended, one could employ employ any combination of the three, including --
```
<symbol glyphnum="#[SMuFL code point]" facs="#[ID of zone element]" altsym="#[ID of symbolDef element] />
```

pe-ro:
In this case, "the way we've always done it" approach doesn't apply -- we're still trying to figure out how to "do it" -- and it obviously isn't extensible -- we're having trouble making it work for complex visual/semantic combinations.

ahankinson:
OK, so thanks to your comments (and patience!) I have a better understanding of what is happening.

I've been struggling a bit with trying to figure out why this "feels" different from other visual domain data that we encode. What I come back to is that we don't generally encode visual data *directly* -- we don't really ever say this thing is "visually" a 22. Instead, most visual domain attributes in MEI describe visual **alterations** of logical symbols: They adjust the visibility, position, colour, size, orientation, and specific fonts used to render them. So when we talk about the "visual domain" in MEI, we generally don't encode things to be displayed directly -- instead we talk about alterations to a presumed default.

But that's not quite right. Bear with me while I try to figure this out, but I thought I would put it out there to see what you think, and to collect counter-examples.

---

In any case, I think that if we are talking about encoding visual attributes, we will definitely need the ability to work with these values in a visual way. This means that any proposed solution should "tie in" to the other MEI mechanisms for visual placement, offsets, color etc. If we compare, for example, `[<mensur>](https://music-encoding.org/guidelines/dev/elements/mensur.html)` with `[clef]`(https://music-encoding.org/guidelines/dev/elements/clef.html), we can see that there are visual attributes, like `@color`, `@ho`, `@to`, `@visible`, etc. that are missing from `<mensur>`. 

At the same time, `<mensur>` does take `@glyph.uri` so we can encode the SMuFL codepoint, and it does take `@facs` so we could place it at a location on an image. However, it is not clear to me what a value in that attribute would mean, if we are also using the same element to encode *other* visual information in the `@sign.[something]` attributes. If I have (this is taken from your example given above):

```
<mensur glyph.num="U+E911" sign="O" sign.num="22">
```

The `@glyph.num` would apply to the mensuration sign, but it would not permit me to also have SMuFL values for the *other* graphical entities that are present in the rest of the mensuration sign; namely the "22" in `@sign.num` or any value in `@sign.numbase`.[*]

**Will the `@sign.num` and `@sign.numbase` visual renderings need to be independently visualized, organized, addressable, positioned, coloured, etc., separate from `@sign`?** Might you, for example, want to set the `@sign.num` or `@sign.numbase` to `visible="false"`? 

If so, then we should try not to overload our poor `<mensur>` element with too much data in attributes, when the values in those attributes will likely *also* need their own separate visual identity.

---

I feel I need to clarify that my objection to having `Z` in `@sign.num`. It isn't just the fact that it is not a number, and that it could be fixed by only allowing numbers in these attributes. If you need to have a `Z` value to specify something, then that is what you need. 

My objection is that `@num` and `@numbase` are *explicitly designed* to be computable values -- that is, you can reasonably assume to be able to do arithmetic with the values in them, since they must be positive integers. **My concern is that using the suffixes `.num` and `.numbase` for capturing visual data is inconsistent with the current uses of those attribute names for holding computable data.**

Put another way, assuming no other changes, I would propose that these attributes be called `@num.sign` and `@numbase.sign`, rather than the other way around. At least then we know that the data in these is **sign** data, rather than **numerical** data. 

Then you can put whatever you want in them, since there is no presumption that sign data holds anything but values meant to be shown in some way.

 > [*] As a side note, I am now realizing that we have a *second* current use of  `.num` attributes, namely the use of Glyph `.num` codepoints. So now we have three potential uses of `.num` with different data types: As an actionable number in `@num`, as a Glyph code point `@glyph.num` and the proposed `@sign.num`.

annplaksin:
> My main concern is that what is being encoded, and the motivations for its capture, is not entirely clear to me so I'm trying to fill in my own gaps of understanding. As I understand it, what this change is trying to do is to find a way of capturing how a mensuration sign appears visually, separately from how it functions on the interpretation of the music.

@ahankinson One additional remark on the reason for this proposal: It's about dealing with an ambiguous use of signs in original sources and conflicting theoretical statements on the meaning of signs. We find identical combinations of signs in a mensuration sign that lead to different mensural interpretations, like here:
grafik
grafik
Both sources use "O2", but they disagree on the question of which division level is addressed by the number and which is addressed by the main mensuration sign.
So, at least for me, this feature deals not only with the automatic rendering of a score but also with the explicit documentation of notational practice regarding the unstable relationship between a sign and its meaning.

annplaksin:
> **Will the `@sign.num` and `@sign.numbase` visual renderings need to be independently visualized, organized, addressable, positioned, coloured, etc., separate from `@sign`?** Might you, for example, want to set the `@sign.num` or `@sign.numbase` to `visible="false"`?

A bit of a guess: I don't have an idea why `@sign.num` or `@sign.numbase` should be set to `visible="false"` since they are visible parts of a mensuration sign, but having them in a different colour seems possible to me without knowing about an existing case.

> If so, then we should try not to overload our poor `<mensur>` element with too much data in attributes, when the values in those attributes will likely _also_ need their own separate visual identity.

Is this a slight nudge into the direction of moving things into the content of `<mensur>`?

> I feel I need to clarify that my objection to having `Z` in `@sign.num`. It isn't just the fact that it is not a number, and that it could be fixed by only allowing numbers in these attributes. If you need to have a `Z` value to specify something, then that is what you need.
> 
> My objection is that `@num` and `@numbase` are _explicitly designed_ to be computable values -- that is, you can reasonably assume to be able to do arithmetic with the values in them, since they must be positive integers. **My concern is that using the suffixes `.num` and `.numbase` for capturing visual data is inconsistent with the current uses of those attribute names for holding computable data.**
> 
> Put another way, assuming no other changes, I would propose that these attributes be called `@num.sign` and `@numbase.sign`, rather than the other way around. At least then we know that the data in these is **sign** data, rather than **numerical** data.

Interesting that you call it "**sign** data" and don't call the attributes `@num.vis`. As in the "O22" case, the numbers within the sign are quite far away from the double diminution that would be expressed by `@num/@numbase`. Therefore, this seems logical to me. In the context of your remarks of the visual domain as rendering hints, this is indeed very interesting... 🤓

bwbohl:
ODD meeting notes:

* is there a clear implementation path for this yet or is more discussion / decision needed?

benory:
Hi all! 

I'm a Ph.D. student at Stanford, developing a digital humanities project that uses the Verovio Humdrum Viewer to process kern files. (@lpugin suggested I ask my question here.) My project is a [database of music from the 1510s through the 1530s](https://github.com/benory/1520s-Project) in unreduced modern notation with the original mensural signs. Right now, I can only show the signs O, C, and Cut-C. I can't display sesquialtera signs (most commonly in my period: 3 and Cut-C3), and this is preventing me from uploading a large chunk of music. C2 is also an important sign for handful of works from the 1510s. I'm afraid I have less of a technical background than you all, but I wanted to ask: would the current proposal work for my use case?

Thanks so much,
Ben

ahankinson:
@benory have you looked at the Verovio test suite for Mensural notation? There seem to be examples there for your use case and how to encode them. (Specifically, e.g., using `<mensur num="2" prolatio="2" tempus="2" />` which seems to do what you want?)

https://www.verovio.org/test-suite.xhtml?cat=mensural

benory:
@ahankinson, thanks for your reply! I've now taken a look at the test suite examples, and two things come to mind. First, while the mensuration signs look very reasonable graphically, I want to be clear that my project uses mensuration signs as purely graphical symbols that the computer does not need to know the meaning of. The need for the signs is primarily to indicate tempo and metric information. Second—following my first point—I am not showing the music in mensural notation; I use mensural symbols together with modern notation. I imagine that it is possible to draw the symbols from this suite, once implemented, without using mensural notation?

ahankinson:
@benory Since this issue seems to concern some project-specific methodologies, rather than continue this conversation on this PR I would suggest taking it to the MEI Slack channel. Would that be OK?

craigsapp:
There is definitely no `3` mensuration currently available in verovio, as I have been asking for it for a while since it is very common in early Renaissance era music.  There has been a good amount of discussion about how it should be represented in MEI.  It seems that a consensus was gradually being approached.   @annplaksin, do you remember what that was? C.f. https://github.com/music-encoding/music-encoding/issues/686#issuecomment-818306754

The root of the problem is that `3` mensuration can mean slightly different things (perfecting the breve or semibreve subdivisions, but keeping the duration of the breve or semibreve subdivisions the same as the imperfect ones).  The main sticking point was what attributes should be used to describe these. `@num`/`@numbase` are the basic need, but you can see the discussion above about creating a new related pair of attributes for cases related to `3`.  

Here I describe some of the complications about which rhythm level to apply the `3` to (but it is almost always at the breve level and not the sembreve): https://github.com/music-encoding/music-encoding/issues/686#issuecomment-826392879.  In such cases, parameters such as `@prolatio.num`/`@prolatio.numbase` and `@tempus.num` and `@tempus.numbase` were proposed to disambiguate where the 3:2 proportion should be applied.  Here is the chart showing the three possible interpretations of "3" (in mensural music -- there is a "3" sign in later music, I think in CMN, that probably has a different meaning):


Screen Shot 2021-10-15 at 8 15 52 AM

The green notes show where the sequialtera from the `3` is being applied, i.e., where the 3 in the time of 2 proportion ("triplets") needs to be applied.

When the tempus is already divided into 3, such as in O mensuration, there is no ambiguity where the `3` should be applied: it can only mean at the prolatio level, going from 2 to 3 minims for the semibreve, but keeping the duration of the semibreve the same (i.e., creating triplets on the semibreve level).

The problem is that there are two possibilities for intepreting `3` mensuration signs when the tempus and prolation are both imperfect, since the `3` can apply to either one of them.  In actual music, it is the performers job to figure it out, and it will be based on if you see a lot of minims or a lot of semibreves in the music.  In all of the music in the JRP (about 1000 pieces), the only case is using the `3` to apply to the tempus, so prolatio-level sequialtera is probably quite rare, but it does happen.

In @benory's case the subtleties of the meaning are not important since the relative rhythms of mensural music have already been converted to the absolute ones of modern notation.  The mensural sign only conveys the metric and tempo information to the performer, so a purely graphical way of encoding mensuration signs (particularly `3`) is useful in these sorts of situations.

ahankinson:
> a purely graphical way of encoding mensuration signs (particularly 3) is useful in these sorts of situations.

What about embedded SVG in the MEI. If “musical meaning” is not necessary, then music encoding isn’t required, no?

craigsapp:
That will not work.  In the JRP I use the graphical mensuration symbols to categorize the music for analysis.  In CMN there is no longer any intrinsic rhythm information (other than to encode how to reverse the process back to mensural music), but it is a categorical symbol and has a meaning in that sense.   Here is an example of such use:

https://josquin.stanford.edu/analysis/rhythm

Screen Shot 2021-10-15 at 9 58 10 AM

This is a list of histograms of rhythm patterns (at the breve level) under various mensuration signs which in turn are used to analyze stylistic patterns in various composers.   Describing mensural music in terms of modern time signatures is of not much use for such analyses since there will be one modern time signature that could be more than one possibility of a mensuration sign.  Showing the mensuration sign also implies that no rhythmic reduction has been applied to the notes, which has be a common thing to do in 20th century editions of mensural music (dividing the durations by 1/2 or 1/4 to make the half note or quarter note the beat rather than the original use of the whole-note for the beat).

When a person looks at the music in modern notation, also seeing modern time signatures conveys no particularly useful information that can be worked back to the mensural notation.  In that sense, @benory's problem is not application specific.  For example the New Josquin Edition (published from 1980's to current):
     https://www.kvnm.nl/en/Publisher/The-New-Josquin-Edition
uses the same style: resolving the relative durations of mensural music into the absolute durations of CMN, but keeping the mensural signs:
     https://www.kvnm.nl/content/finder-upload/files/NJE%20informatie.pdf

Screen Shot 2021-10-15 at 10 05 14 AM


--------------------

Related to the display of mensural signs in CMN here is an example of a `3`  (mensuration?  as it seems to have a different meaning here than in the music from 1450-1550 where `3` cannot be the starting mensuration sign) from the Fitzwilliam's Virginal Book (music composed around 1600 give or take a few decades):

https://imslp.org/wiki/Fitzwilliam_Virginal_Book_(Various)
https://www.kvnm.nl/content/finder-upload/files/NJE%20informatie.pdf

Screen Shot 2021-10-15 at 10 26 27 AM

Screen Shot 2021-10-15 at 10 32 18 AM

https://en.wikipedia.org/wiki/Fitzwilliam_Virginal_Book

This is from an 1899 edition where they give the original mensuration sign along with a modern interpretation in brackets after it. (but why there is a 9/4 and 3/4 measure in 3/2, I do not know...).

Other example showing the original C-dot mensuration:

Screen Shot 2021-10-15 at 10 36 17 AM

martha-thomae:
@annplaksin, I will close this PR as I know we need to discuss this back with the Mensural IG and that we have already come up with some ideas over the last years (like, come on, this was in 2021). So let's regroup and then come up with a new PR / issue / GitHub discussion 🙂

PR #770, opened 2021-02-12
Improve placement encoding for artic and accid. Closes #761

lpugin:
* rename att.placement into att.placementRelStaff
* add att.placementRelEvent and make accid and artic member of it
* add att.placmenetOnStaff and make accid and artic member of it

Issue #761, opened 2021-01-27
Encoding positioning of articulations


lpugin:
Some articulation signs, such as accents, are sometimes placed outside the staff, and sometimes inside.

![image](https://user-images.githubusercontent.com/689412/106001810-b315e800-60b0-11eb-854f-db9e49f79c31.png)

![image](https://user-images.githubusercontent.com/689412/106001895-c7f27b80-60b0-11eb-9076-949cc90cd79b.png)

How would this distinction be encoded?

craigsapp:
The second example is probably more common, so that could be the default.   Possibly an option could be used for the first example's rendering.  Otherwise an attribute could be added to MEI that would mean do not place the articulation inside of the staff (unless there is already such a thing).

If there were two layers of music, and the first measure of the example was in the second layer, then the second example is the only possible rendering (otherwise it would be similar to a recent fixed problem for automatic placement of trills and mordents attached to notes in the second layer).

pe-ro:
Not sure what you're asking but perhaps this will help --

Currently, layout positioning of articulations is addressed by `@loc` or with a combination of `@oloc` and `@ploc` on each `<artic>` element.  There is no way to assign values for a group of articulation signs.

lpugin:
@pe-ro `@oloc` and `@ploc` could work in some cases but I think we would rather have something that leaves it relative to the note pitches. In other words, it would not be necessary to change it if the clef is changed or if the notes are transposed. So I am thinking of an attribute like `@outside="true|false"`. I am sure we can find a better name, though.

rettinghaus:
Wouldn't be `place="within"` explicit enough for this?

craigsapp:
Currently in verovio `artic@place="above"` is relative to the note, not the staff.  So `within` does not have any particular meaning in this context (i.e., what does "within" a note mean?).  There could be additional placement qualifiers: `above within` which would allow the articulation to be placed above the note, and allowed to be within the staff (but only if necessary based on the pitch of the note).

A general problem is that some articulations are allowed within the staff (staccatos), tenutos and accents are usually allowed in, marcatos are usually outside of the staff, while others are rarely within (bowings), so it would be encoding verbosity to have to add `within` on every single articulation that can be inside of the staff.

This could be a scoreDef/staffDef attribute that sets the default withinness of articulation types when no explicit `within` is given on the articulation.  The goal would be to have music that can be transposed without changing placement attributes.  For example this is useful for converting between sounding score and transposing parts.

[...]

Issue #369, opened 2016-12-29
whole-measure rest encoding


craigsapp:
How should/can the rests in this example be encoded:

<img width="358" alt="screen shot 2016-12-29 at 15 06 20" src="https://cloud.githubusercontent.com/assets/3487289/21556818/0e7d099a-cdd9-11e6-9e86-8eaae952d671.png">

In other words, how to encode that a "1" is shown between the staves.  Would this be a special case of `multiRest`?

Multi-rests probably have a different visual syntax, with duplicate numbers for each staff:

<img width="287" alt="screen shot 2016-12-29 at 15 09 07" src="https://cloud.githubusercontent.com/assets/3487289/21556833/3730f7d4-cdd9-11e6-8ae1-4cd95b6e86c6.png">

pe-ro:
>  Would this be a special case of `multiRest`?

No.  It would be better to use `<mRest>`.  

But, right now, `<mRest>` can't be numbered and there's no way to specify the location of the number.  (`<multiRest>` doesn't have any method of placing its number either.)

So, I suggest we --

1. make `mRest` a member of att.numbered -- this will allow the capture of the number using the `@num` attribute; and 
2. make `mRest` and `multiRest` members of number.placement. This provides the attributes `@num.place` and `@num.visible`.  `@num.place` will keep its current values: "above" and "below".

This will provide the capability to say where the number associated with `mRest` and `multiRest` should be rendered.  However, this assumes the same visual characteristics for both elements; that is, a number rendered above or below the rest *on each staff*.

For the special case of a full-measure rest with a number shown between staves, I think it would be better to use `<mRest>`, but **not** use the `@num.*` attributes, but `<dir>` instead --

```
<measure>
  <staff>
    <layer>
      <mRest/>
    </layer>
  </staff>
  <staff>
    <layer>
      <mRest/>
    </layer>
  </staff>
  <dir tstamp="2.5" place="between" staff="1 2">1</dir>
</measure>
```
I recognize that this is somewhat of a kludge (as "1" is not a performance direction), but since this example breaks tradition, it seems logical that there should be a special case in the markup too.  In another thread we decided to use `<dir>` for other things that aren't strictly performance directions (the Peter & the Wolf example), so I don't feel doing so here is particularly egregious.  A benefit of using `<dir>` is that its contents can be better controlled typographically; that is, using `<rend>`.

PR #853, opened 2021-09-24
Update description on att.numberPlacement

### LINKED ISSUE #852: Simplify att.numberPlacement
Do we need to have the attribute `@num.visible` in addition to `@num.place`? Shouldn't it be enough to say `@num.place="none"`?

rettinghaus:
Since this also affects multiRest the description should be adapted.

### DIFF
--- a/source/modules/MEI.cmn.xml
+++ b/source/modules/MEI.cmn.xml
@@ -875,17 +875,16 @@
     </attList>
   </classSpec>
   <classSpec ident="att.numberPlacement" module="MEI.cmn" type="atts">
-    <desc>Attributes that record the placement and visibility of numbers that accompany a bowed
-      tremolo or tuplet.</desc>
+    <desc>Attributes that record the placement and visibility of numbers that accompany a notational object.</desc>
     <attList>
       <attDef ident="num.place" usage="opt">
-        <desc>States where the tuplet number will be placed in relation to the note heads.</desc>
+        <desc>Indicates where the number is written in relation to its reference object.</desc>
         <datatype>
           <rng:ref name="data.STAFFREL.basic"/>
         </datatype>
       </attDef>
       <attDef ident="num.visible" usage="opt">
-        <desc>Determines if the tuplet number is visible.</desc>
+        <desc>Determines if the number is visible.</desc>
         <datatype>
           <rng:ref name="data.BOOLEAN"/>
         </datatype>


source/modules/MEI.cmn.xml
```diff
@@ -875,17 +875,16 @@
     </attList>
   </classSpec>
   <classSpec ident="att.numberPlacement" module="MEI.cmn" type="atts">
-    <desc>Attributes that record the placement and visibility of numbers that accompany a bowed
-      tremolo or tuplet.</desc>
+    <desc>Attributes that record the placement and visibility of numbers that accompany a natational object.</desc>
```
bwbohl: please fix typo notational

source/modules/MEI.cmn.xml
```diff
@@ -875,17 +875,16 @@
     </attList>
   </classSpec>
   <classSpec ident="att.numberPlacement" module="MEI.cmn" type="atts">
-    <desc>Attributes that record the placement and visibility of numbers that accompany a bowed
-      tremolo or tuplet.</desc>
+    <desc>Attributes that record the placement and visibility of numbers that accompany a notational object.</desc>
     <attList>
       <attDef ident="num.place" usage="opt">
-        <desc>States where the tuplet number will be placed in relation to the note heads.</desc>
+        <desc>Indicates where the number is written in relation to its reference object.</desc>
```
bwbohl, line 881: I think this is a correct modification reflecting MEI's transcriptional perspective

bwbohl (commented):
Generally this can be considered ready to merge, nevertheless the PR could also implement future discussion / results from #852

source/modules/MEI.cmn.xml
rettinghaus: done

bwbohl:
@music-encoding/core-contributors please add your comments concerning this PR and its potential relation to #852 , i.e.:

* should we merge this one
*  should we additionally remove `@num.visible` as it logically could be merged with `@num.place` by allowing a `none` value there

ahankinson:
1) Is there an issue that provides the motivation for this change? I notice that it's changing the target of the documentation (from "numbers that accompany a bowed tremolo or tuplet." to "notational object") but I can't find an issue that flags this. I'm not sure we have a defined concept of what a 'notational object' is?

2) -1 to remove `@num.visible`. Explicit behaviours of when to show / hide things are really handy (see the whole "when to show/hide an accidental" discussion -- `@accid.visible` would be handy here...). `@num.place=none` could either mean "don't show it" or "show it wherever you like".

kepper:
I may be late to the party, but I'm with @ahankinson on this. The consequences of removing `@num.visible` remind me of MusicXML's "yes-no-number" datatype – and that's clearly a road I don't want to go down…

ahankinson:
Further to my first point, the documentation for att.placement says: "Captures the placement of the item with respect to the staff with which it is associated."

Given that the datatype is `data.STAFFREL.basic`, it might be useful to modify the description to provide clarification that the placement relates to the **staff** and not to the **item**. 

So perhaps:

"Attributes that record the placement and visibility of numbers **on the staff** that **accompany an item**."

"Indicates where the number **for an item** is written in relation **to a staff**."

An example would also be handy to provide here...

lpugin:
Maybe we should also consider what has been discussed here https://github.com/music-encoding/music-encoding/issues/1604

rettinghaus:
These changes were made to reflect #626 after the discussion in #369.

In these cases the placement does not relate to the staff per se: 
tuplets
or an even weirder case:
tuplets2

And I think you are really in trouble if you want to encode something like this (Gould, p 319):

gould-tuplets

source/modules/MEI.cmn.xml
bwbohl: as @ahankinson minted out the undefinedness of `reference object` produces a problem here since data.STAFFREL.basic itself defines the staff as reference object.

bwbohl:
I think the examples ( the first two and a) above by @rettinghaus really do illustrate the problem that although – if allowing `data.STAFFREL.extended`(between | within) – the position can be expressed in the relation to the staff but not in relation to the tuplet itself, which (esp. in transcription work) would be more interesting to capture or describe.

Moreover, as @ahankinson mentioned the undefinedness of `reference object` produces a problem here since `data.STAFFREL.basic` itself defines the staff as reference object.
The actual problem is a modelling problem then, that could be resolved by switching to `data.EVENTREL.basic`, right?

It gets a bit more tricky with examples (b) and (c) though, as for what `above` or `below` are in those cases ;-)

ahankinson:
So it seems that `data.STAFFREL.basic` isn't really appropriate here? If `above` and `below` are specifically **for the staff** (https://music-encoding.org/guidelines/v4/data-types/data.staffrel.basic.html, "Above the staff", "Below the staff") then using it for above/below other things, like tuplets, or measure repeats, isn't really appropriate?

A couple options:

1. Change the name: `data.STAFFREL.basic` -> `data.ITEMREL.basic`, and use 'above', 'below' an item in the definition. Not sure what to do about `data.STAFFREL.extended`. This can then be used by anything that would be above or below anything else.
2. Keep `data.STAFFREL.basic` and create a new datatype, `data.ITEMREL.basic` with the same values.
3. Create a generic data type, `data.ABOVEBELOW`, and then make `data.STAFFREL.basic` and `data.ITEMREL.basic` refer to that as their datatype values.

Option 3 is a bit of an optimization on 2.

bwbohl:
2021-10-28:
* other things like fingering should also be considered in this discussion
* providing encoding using the mentions datatypes could help getting things clearer
* considering tuplets wouldn't it be more appropriate to encode whether the number is on the head-side of the note or the stem-side?

kepper:
This turns out to be a relatively complicated issue. Thing is that the "anchor point" for values from data.STAFFREL and data.EVENTREL are not clear from the values (which largely overlap). One way would be to have a separate attribute to specify that anchor (event or staff, could also deal with notehead vs. tail issues etc.), and then merge them. That way, we could probably keep compatibility with older instances, while doing a lot of architectural changes to the ODD. The only other way right now would probably be to prefix values in one of the two data types, which means that either Neumes or CMN files will get invalid… This needs more discussions ;-)

ahankinson:
Just removing the 'Ready to Merge' label to indicate the current status.

kepper:
We discussed this extensively again during today's ODD Friday, and after that long discussion, we noticed that it actually overlaps with #770, and the `@onstaff` attribute brought in there (see also #761). This means we "just" need to clean up the class setup with data.STAFFREL here, to get better names. This will change ODD structures only, but not affect the MEI surface "API". @rettinghaus and @kepper volunteer themselves to clean this up and prepare the PR accordingly. Thanks everyone involved :-)

bwbohl:
@kepper and @rettinghaus just a reminder: maybe you find some time before this week's ODD meeting to fix this

bwbohl:
@kepper and @rettinghaus any progress on this?

rettinghaus:
Changed the target of this PR.

kepper:
we're moving this into a branch on music-encoding to make it more accessible

kepper:
@rettinghaus and I had a discussion on this and hope to prepare something for review at next ODD whateverDay…

PR #854, opened 2021-09-27
Fix name

bwbohl:
Add middle initial to my name


bwbohl:
ups sth went wrong here closing…

Issue #668, opened 2020-05-25
`@cue` needs clarification


th-we:
The [documentation for `@cue`](https://music-encoding.org/guidelines/v4/attribute-classes/att.cue.html) just says

> Attributes that describe "cue-ness".

Can we define what cue-ness means?  Specifically, does this mean that a note is (logically) a cue from a different part, or does it mean that the note is (visually) cue size?  There are some contradictions here: 

* `@cue` is a logical domain attribute for [notes](https://music-encoding.org/guidelines/v4/attribute-classes/att.note.log.html), [chords](https://music-encoding.org/guidelines/v4/attribute-classes/att.chord.log.html), [rests](https://music-encoding.org/guidelines/v4/attribute-classes/att.rest.log.html) and [measure rests](https://music-encoding.org/guidelines/v4/attribute-classes/att.mrest.log.html).
* But it's also a visual domain attribute for [beams](https://music-encoding.org/guidelines/v4/attribute-classes/att.note.log.html) and [beamSpans](https://music-encoding.org/guidelines/v4/attribute-classes/att.beamspan.vis.html).

It should not be used for both those domains. While the original intent seems to be flagging logical cues (#433, #406), the actual usage seems to be focused on visual size:

* Verovio [imports](https://github.com/rism-ch/verovio/blob/d746305f29483d14dec8bd5a300ef9f7050b3d58/src/iomusxml.cpp#L2173) MusicXML's "logical" `<cue>` element the same way as the size-related ("visual") `type/@size='cue'` attribute instead of converting one to `@cue='true'` and the other to `@fontsize='small'`
* The only occurrence of [`@cue` in the sample encodings](https://github.com/music-encoding/sample-encodings/blob/master/MEI_4.0/Music/Complete_examples/Parker-Gillespie_ShawNuff.mei#L209) seems to use it for size (there is no part that the small notes "quote", it's just more music that is added to the leadsheet)
* `@cue` was suggested for [encoding small "cadenza-style"/rhythmically free-floating notes](https://github.com/music-encoding/music-encoding/issues/509)
* [`@cue` was added to beams](https://github.com/music-encoding/music-encoding/commit/8507516acfe8b07b37eb15e7de9721b1781e74c8), motivated by cases where [regular-sized and cue-sized notes share a beam](https://github.com/music-encoding/music-encoding/issues/558)

[...]

PR #859, opened 2021-10-05
Make layer cue-able

rettinghaus:
Usually in parts a whole layer gives the cue, so it seems to make sense to make `layer` part of `att.cue`. 
image
(Example taken from _Behind Bars_, p 572)


craigsapp (approved):
This is a useful feature to have.  The example music demonstrates that most cues in orchestral parts are encoded in layers, with a regular-sized full-measure rests in another layer to reinforce that the cue notes are not to be played by the performer.

kepper (changes requested):
This sounds perfectly reasonable, and I'm not requesting a change to the one line added here. But. There should be some documentation about this, like we just discussed at the ODD Friday. It needs to be checked how much actual documentation we have in the Guidelines about cue'd things…

musicEnfanthen:
See also #668

bwbohl:
2022-01-28 ODD meeting: @rettinghaus if you could please be so kind as to add some documentation to the guidelines. Apparently there isn't anything about `@cue` so far, some general example might be a first start, even when limiting it to layer-cueing.

rettinghaus:
Added a section about cue notes in the Guidelines.

source/examples/cmn/cmn-cue_layer.txt
```diff
@@ -0,0 +1,50 @@
...
+          <note dur="4" oct="2" pname="f" />
+        </layer>
+      </staff>
+    </measure>
+    <measure n="1">
+      <staff n="1">
+        <layer n="1">
+          <mRest />
+        </layer>
+        <layer n="2" cue="true">
```
ahankinson, line 28: I'm wondering if we should recommend that `cue` should also use `sameas` and point to the cued layer, rather than encoding the layer twice. Thoughts?

rettinghaus: Sorry, I don't understand.

kepper: I've run into this recently. I would say it should be optional, but not necessarily recommended. The cue notes / layer is still an entity in itself, even though it has a semantic connection to some other notes / layer and is derived from there. It could have incorrect notes, though. In that case, one may perhaps prefer a `@corresp` over `@sameas`?

ahankinson: The notes in a cue layer correspond to notes in another layer, correct? They're not played, but they're there to indicate that you should listen to those notes for your own entrance.

So should we use some of the available mechanisms (sameas, corresp, etc.) to indicate that this link exists, rather than to encode them as completely separate notes?

ahankinson: In the case of a cued layer, I would expect that a full layer would be cued. Otherwise, we would just mark four notes as cues.

rettinghaus: Yes, we can point to the original layer, if we have a full score encoded. If we have only a single part, where should it point at?

ahankinson: Your addition looks great! Thanks.

ahankinson (approved):

ahankinson:
Probably should have one more reviewer before merging. Since @craigsapp reviewed it there have been some changes.

musicEnfanthen:
Would it be helpful to include an example of the proposed linking mechanism for cued layers with `corresp` in chapter [10.1.1](https://music-encoding.org/guidelines/dev/content/analysisharm.html#analysisDescribingRelationships) of the guidelines and link there from the description here?

source/docs/04-cmn.xml
```diff
@@ -1104,10 +1104,33 @@
...
                      <figure>
                         <head/>
-                        <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/cmn/cmn-sample158.txt" parse="text"/></egXML>
+                        <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/cmn/cmn-ossia_layer.txt" parse="text"/></egXML>
                      </figure>
                   </p>
                </div>
+               <div xml:id="cmnCue" type="div3">
+                  <head>Cue</head>
+                  <p>Cue notes are smaller notes that usually occur in an orchestral or vocal part and are not to be played or sung by the corresponding musician, but by other instruments/singers. Cue notes are used for orientation, <abbr>i.e.</abbr>, to follow the music during longer pauses and to find the correct re-entry point more easily. In MEI the <att>cue</att> attribute is available to indicate such notes.</p>
```
bwbohl: instruments/singers --> instruments or singers?

bwbohl (commented):
only a minor comment

many thanks @rettinghaus

bwbohl (approved):

bwbohl:
@kepper could you revise your review please

source/docs/04-cmn.xml
```diff
@@ -1104,10 +1104,33 @@
...
+                     <figure>
+                        <head>Cue notes in the flute part of Brahms’ <hi rend="italic">Ein deutsches Requiem</hi>.</head>
+                        <graphic url="../images/modules/cmn/Brahms-flutePart.png"/>
+                     </figure>
+                     <figure>
+                        <head/>
+                        <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve" valid="true"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/cmn/cmn-cue_layer.txt" parse="text"/></egXML>
+                     </figure>
+                  </p>
+                  <p>If the voice from which the cue notes originate is also encoded, they should refer to their sounding counterpart with the <att>corresp</att> attribute.</p>
```
musicEnfanthen, line 1131: Would it be possible to include another encoding example here for this linking mechanism with `corresp`. This would be very helpful for users consulting this passage.

rettinghaus: As you suggested this should probably go into [chapter 13](https://music-encoding.org/guidelines/dev/content/analysisharm.html#analysisDescribingRelationships) where nothing can be found about `@corresp`at all. Opened #915.

musicEnfanthen: `@corresp` itself and all other attributes of `att.linking` are momentarily explained in chapter 10 (Analytical markup). If that's a good place for them, we should discuss over there in #915, right. 

For the passage here, I am not asking to explain `corresp` in general (what would be too much out of scope of this PR), just to add a minimal encoding example how you would target the "sounding counterpart" from the cued layer. (My initial thought was to link to the corresp section from here, but since there is nothing really helpful for the use case here, it seems easier for now to include a small example here which exemplifies that usage of cued layers.) What do you think?

craigsapp:
Here is another example that could use this feature that I have come across:

Screen Shot 2022-02-21 at 1 34 30 PM

From Stravinsky: Perséphone (Melodrama in Three Scenes for Solo Tenor, Choir and Orchestra)

The second staff (violin 1) has two layers, with the first one being cue-sized showing the sounding pitches (also in square brackets).   The performance fingerings are in the second layer.

kepper (approved):
:)

kepper:
by instruction of our technical co-chairs during ODD meeting today – many thanks ;-)

PR #862, opened 2021-10-15
Fix canonicalization

bwbohl:
The canonicalised mei-source produced by our build process has several issues that are being fixed by this PR as follow:

* Fix: update schema reference URI to match actual local path
* Fix: add missing attributes on TEI element
* Fix: remove empty `@xml:lang` on xicluded elements
* Fix: remove surplus xml-lang prcessing-instructions from xincluded files
* Fix: remove `@xml:base` on xincluded files as those would render all relative targets unavailable


bwbohl (commented):
Apparently my branch has some commits (blank.yml) tht should not be merged....
Will fix before merge!

Issue #865, opened 2021-10-27
Incompleteness of note head encoding


rettinghaus:
There are several things problematic about the current situation with assigning note heads: 

1. The description of `@head.shape` states that this attribute should be "Used to override the head shape normally used for the given duration."  
That leads to two questions: What's the difference to `@dur` and `@dur.ges` (which implicitly gives the note head), and what should happen with a stem (e. g. if one encodes a quarter note with `@head.shape="whole"`)?

2. `data.HEADSHAPE.list` is incomplete in terms of duration. While the former `data.HEADSTYLE` included the value "dblwhole", this was now move to `data.NOTEHEADMODIFIER.list`, although it is a note head for itself and not just "enclosing fences" that are added to a note head. Furthermore there are two styes for double whole notes, the modern oval note head with vertical lines and the more traditional square head (resembling a brevis note). Currently there seems to be no way to switch between them. There are cases where both forms are used at the same time: 
![dblwholes](https://user-images.githubusercontent.com/7693447/139076805-c4b7029d-e8b2-4154-9557-0e79f276de6c.png)

3. (Visual) Durations for non-standard note heads are not easy or simply impossible to encode. 
Take a look at triangular note heads (example from Gould, p. 13): 
￼![triangular heads](https://user-images.githubusercontent.com/7693447/139072126-44a9f502-6ba3-4e21-81b7-6d938e7e8a39.jpeg)
To encode the eighth notes you have to use `<note @head.shape="isotriangle" @head.fill="solid" dur="8" />`, which is a bit annoying but possible.  
When it comes to slash noteheads, it is not possible to distinguish quarter, half and whole notes encoding wise.
See https://github.com/rism-digital/verovio/issues/1866 and https://www.w3.org/2021/03/smufl14/tables/slash-noteheads.html.  
Furthermore, having values for different note head _styles_ doesn't comply with the description of `@head.shape` (see above). 

4. Switching to different note head shapes is difficult, because you have to encode it explicitly on every single note. It would be good to have a way to switch shapes for multiple notes at once (as it is basically a _notationType_).  
An overview of several note head styles is given in the [LilyPond Notation Reference]:(https://lilypond.org/doc/v2.22/Documentation/notation/note-head-styles.html)
![image](https://user-images.githubusercontent.com/7693447/139086704-fc1e0c56-5eb9-4baf-9309-61688c55457d.png)

This relates to #202 and #400.

bwbohl:
[...]

PR #867, opened 2021-10-28
Move dblwhole to head.shape

rettinghaus:
This moves `dblwhole` back to `head.shape` and extends `data.NOTEHEADMODIFIER.pat` to allow solmization syllables. 

Compare the SMuFL ranges
https://w3c.github.io/smufl/latest/tables/note-name-noteheads.html
https://w3c.github.io/smufl/latest/tables/note-name-noteheads-supplement.html

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -3077,12 +3077,22 @@
     <desc>Captures text rendered in the center of the notehead.</desc>
     <content>
       <rng:choice>
-        <rng:data type="string">
-          <rng:param name="pattern">centertext\((A|B|C|D|E|F|G)(f|♭|n|♮|s|♯)?\)</rng:param>
-        </rng:data>
-        <rng:data type="string">
-          <rng:param name="pattern">centertext\(H(s|♯)?\)</rng:param>
-        </rng:data>
+        <rng:choice>
+          <rng:data type="string">
+            <rng:param name="pattern">centertext\((A|B|C|D|E|F|G)(f|♭|n|♮|s|♯)?\)</rng:param>
+          </rng:data>
+          <rng:data type="string">
+            <rng:param name="pattern">centertext\(H(s|♯)?\)</rng:param>
+          </rng:data>
+        </rng:choice>
+        <rng:choice>
+          <rng:data type="string">
+            <rng:param name="pattern">centertext\((Do|Re|Mi|Fa|So|La|Ti|Si)\)</rng:param>
+          </rng:data>
+          <rng:data type="string">
+            <rng:param name="pattern">centertext\((Di|Ri|Ra|Fi|Se|Li|Le|Te)\)</rng:param>
+          </rng:data>
+        </rng:choice>
       </rng:choice>
     </content>
   </macroSpec>
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -1474,6 +1474,9 @@
         <valItem ident="whole">
           <desc>Unfilled, rotated oval (Unicode 1D15D).</desc>
         </valItem>
+        <valItem ident="dblwhole">
+          <desc>Unfilled, rotated oval with vertical lines on either side (Unicode 1D15C).</desc>
+        </valItem>
         <valItem ident="backslash">
           <desc>Unfilled backslash (~ reflection of Unicode 1D10D).</desc>
         </valItem>
@@ -3066,9 +3069,6 @@
         </valItem>
         <valItem ident="circle">
           <desc>Enclosing circle.</desc>
-        </valItem>
-        <valItem ident="dblwhole">
-          <desc>Enclosing "fences".</desc>
         </valItem>
       </valList>
     </content>


bwbohl (approved):

bwbohl:
@music-encoding/maintainers please review

ahankinson:
Are there any issues describing these changes? Particularly the extension to the note heads?

rettinghaus:
Allowing the solmization syllables as text in the notehead seemed to be missing, as pitch names were already allowed. 
Also this allows some semantics over using just SMuFL code points for these noteheads (which can also get quite messy regarding different durations).

ahankinson:
I have so many questions -- not particularly on this change, but on this section of the spec in general:

1. What does `.pat` stand for? (Helpful for understanding what this is trying to do...)
2. What is the `centertext` bit at the beginning of the pattern? Is this a literal string, or is it some sort of formatting?
3. What types of values is this pattern checking: `(H(s|♯)`?

source/modules/MEI.xml
```diff
@@ -3067,22 +3070,29 @@
...
+          <rng:data type="string">
+            <rng:param name="pattern">centertext\((A|B|C|D|E|F|G)(f|♭|n|♮|s|♯)?\)</rng:param>
+          </rng:data>
+          <rng:data type="string">
+            <rng:param name="pattern">centertext\(H(s|♯)?\)</rng:param>
+          </rng:data>
+        </rng:choice>
+        <rng:choice>
+          <rng:data type="string">
+            <rng:param name="pattern">centertext\((Do|Re|Mi|Fa|So|La|Ti|Si)\)</rng:param>
```
ahankinson, line 3090: Should this also allow “Ut”?

rettinghaus: As far as I know it is not used in modern solmization (and haven't really been used since the 17th century). As written above I simply followed the corresponding SMuFL ranges.

pe-ro:
> I have so many questions -- not particularly on this change, but on this section of the spec in general:
> 
>     1. What does `.pat` stand for? (Helpful for understanding what this is trying to do...)

data.NOTEHEADMODIFIER references 2 other datatypes: data.NOTEHEADMODIFER.list, which allows a closed list of values, and data.NOTEHEADMODIFIER.pat (where 'pat' is shorthand for 'pattern'), which allows values matching regex patterns.  The objective is to permit either a value from the closed list or a value that matches one of the patterns defined in data.NOTEHEADMODIFER.pat.

>     2. What is the `centertext` bit at the beginning of the pattern? Is this a literal string, or is it some sort of formatting?

It's a literal string. The value is expected to look like "centertext()", where the text that will appear in the notehead is inside the parentheses.  I'm not completely convinced it's necessary because the description of the datatype says "Captures text rendered in the center of the notehead."  Also, "centered-ness" is implied in the definition/description of data.NOTHEADMODIFIER.list.  It seems better to me to leave it to a renderer to decide whether the modifying symbol or text will actually be centered.

>     3. What types of values is this pattern checking: `(H(s|♯)`?

This allows "H", which is the designation for B natural in German-speaking regions.

source/modules/MEI.xml
pe-ro: I think "Ut" should be allowed for use in pre-modern notation.

pe-ro:
The solmization patterns should allow "Ut", in my opinion.

musicEnfanthen:
@ahankinson @pe-ro What's the status of this? Does this PR need more discussion, or is it safe to start a review?

pe-ro:
It's fine with me to provide a double whole value in head.shape.  However, I don't recommend removing the value "dblwhole" from data.NOTEHEADMODIFIER.list.  I would rename it to "fences" instead.  The idea is to separate the head shape from the presence of the fences, thereby allowing any shape to have fences or not.

kepper (changes requested):
during 2022 dev meeting in Berlin, @musicEnfanthen, @fujinaga, @pe-ro and @kepper came to the conclusion that this should have be done slightly differently. The new "dbwlwhole" value for `data.HEADSHAPE.list`would not be necessary, if instead we kept the "dblwhole" value for `data.NOTEHEADMODIFIER.list` and renamed it to "fences" instead. This means that you'd still have to use two separate attributes to get a whole note head with fences, but it would avoid a "shortcut" in the encoding, and would also avoid collision of having a "pre-built" fences combined with explicit "fences" as modifiers. The `.pat`stuff seems ok.

musicEnfanthen:
@rettinghaus Any chance to rework this according to the proposal above?

ahankinson:
Just because there is no reference to the original ticket (I think?) that motivated this change:

https://github.com/music-encoding/music-encoding/issues/865

(A further appeal to **please** reference issues in pull requests, otherwise we lose track of what we're trying to fix.)

musicEnfanthen:
Actually, there was a reference to #865 in the original PR description (see edited version). Somehow it got lost, maybe because the discussion in 865 developed into a more broader direction? But be it as it may, thanks for the reminder and re-reference.

ahankinson:
Could someone (@kepper?) provide an example of how you might encode a note with this plan?

musicEnfanthen:
Pinging @rettinghaus and @bwbohl to add some additional comments.

pe-ro:
MEI Development Meeting, 25/05/2023: Closing this pull request in favor of addressing #865 with a new pull request.

PR #870, opened 2021-11-03
Make dist target depend on clean

bwbohl:
This PR adds a dependency to the dist ant target to run the clean target before execution. Tis way it is assured that all previous build artefacts are being removed before running list.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -216,7 +216,7 @@
         </antcall>
     </target>
     
-    <target name="dist" depends="init"><!-- TODO does not work with depends="init" only works if init run before --> <!-- TODO if depends init works add to all other targets and introduce global initialized param so that init will only be reexecuted if init not run before -->
+    <target name="dist" depends="init, clean"><!-- TODO does not work with depends="init" only works if init run before --> <!-- TODO if depends init works add to all other targets and introduce global initialized param so that init will only be reexecuted if init not run before -->
         <classfileset refid="mei.classpath"/>
         <antcall target="build-customizations"/>
         <antcall target="build-compiled-odds"/>
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -48,6 +48,7 @@ jobs:
 
       ### PUBLISHING THE SCHEMA ###
       - name: Checkout SCHEMA_REPO into SCHEMA_DIR
+        if: github.repository_owner == 'music-encoding'
         uses: actions/checkout@v2
         with:
           # repository to check out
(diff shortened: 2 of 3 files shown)


build.xml
```diff
@@ -216,7 +216,7 @@
         </antcall>
     </target>
     
-    <target name="dist" depends="init"><!-- TODO does not work with depends="init" only works if init run before --> <!-- TODO if depends init works add to all other targets and introduce global initialized param so that init will only be reexecuted if init not run before -->
+    <target name="dist" depends="init, clean"><!-- TODO does not work with depends="init" only works if init run before --> <!-- TODO if depends init works add to all other targets and introduce global initialized param so that init will only be reexecuted if init not run before -->
```
ahankinson, line 219: Is the TODO still valid? I'm not sure what it's trying to say...

bwbohl: thx for spotting will double check

musicEnfanthen: @bwbohl Why  `clean` and not `reset` target (i.e., `clean` plus delete `/lib`) ? When `dist` depends on `init`, the `/lib` folder should be reinstantiated anyway, right? Pobably it would require to switch order of depends: `depends="reset,init"`

bwbohl: `reset` would be an option if you are developing the build process, but I deem it kind of annoying when the lib components are pulled again just when I (locally) do some schema development. But yeah that points me in another direction why not introduce a `build` or `build-all` target for those local development cases (or general building) and make `dist` a target that puts all those build artefacts in the right places in the dist folder; the primary function of `dist` then maybe should be our own build-and-deploy workflow?

bwbohl: concerning the TODO describing problems with `init`:
It's not working to just call `dist` at the moment (without calling `init` before) because the build-rng target fails without calling ant with`-lib` giving the path to Saxon. but if you give the path to where Saxon will be after init the call will fail due to missing file, although it is there hen build-rng is called, it wasn't when the build was started…

bwbohl: I have an idea o how to fix it though… ;-)

ahankinson: Do you want to fix it in this PR, or open a new one later?

ahankinson:
@bwbohl any updates on this?

rettinghaus (approved):

ahankinson:
@bwbohl could you resolve the conflicts on this and I can then merge? Thanks.

ahankinson:
Also, please resolve whether this is still a `draft` or if it's ready.

bwbohl:
still a draft, fix at hand from other project, will implement asap

rettinghaus:
@bwbohl What's the status of this?

PR #875, opened 2021-11-05
Fix encoding problems in TEI

rettinghaus:
This PR brings a pile of minor improvements to the Guidelines and the schema. 

1. add missing `sourceDesc` to all TEI files in the docs
2. fix an encoding error in the `<att>` element
3. unifies **i.e.** and **e.g.** to American style enclosing it in `abbr`


source/modules/MEI.critapp.xml
```diff
@@ -63,9 +63,9 @@
     <remarks>
       <p>The alternatives provided in <gi scheme="MEI">lem</gi> and/or <gi scheme="MEI">rdg</gi>
         sub-elements may be thought of as exclusive or as parallel. The <att>type</att> attribute
-        may contain any convenient descriptive word, describing the extent of the variation (e.g.
-        note, phrase, measure, etc.), its text-critical significance (e.g. significant, accidental,
-        unclear), or the nature of the variation or the principles required to understand it (e.g.
+        may contain any convenient descriptive word, describing the extent of the variation (e.g.,
```
ahankinson: I think they were missed here?

source/modules/MEI.edittrans.xml
```diff
@@ -617,7 +617,7 @@
     <remarks>
       <p>The <att>character</att> attribute describes characteristics of the hand, particularly
         those related to the quality of the writing, e.g., 'shaky', 'thick', regular'. A description
-        of the tint or type of ink, e.g. 'brown' or the writing medium, e.g. 'pencil', may be placed
+        of the tint or type of ink, e.g., 'brown' or the writing medium, e.g., 'pencil', may be placed
```
ahankinson: ... and here?

source/modules/MEI.edittrans.xml
```diff
@@ -962,7 +962,7 @@
         causative agent. When the presumed loss of text arises from action (partial deletion, etc.)
         assignable to an identifiable hand, the <att>hand</att> attribute signifies the hand
         responsible for the action. The <att>reason</att> attribute indicates why the text has to be
-        supplied, e.g. 'overbinding', 'faded ink', 'lost folio', 'omitted in original', etc. The
+        supplied, e.g., 'overbinding', 'faded ink', 'lost folio', 'omitted in original', etc. The
```
ahankinson: ... and here? Maybe something went wrong in an automated transform?

ahankinson (changes requested):
Surrounding the `e.g.,` with `<abbr>` seems to have stopped halfway through?

source/modules/MEI.critapp.xml
rettinghaus: Sorry, I should have been more clear. I added `abbr` tags in the docs for the Guidelines only, not in the schema files itself. I wasn't sure if this would be needed or even a good idea. If you say so, I'll add them there, too.

ahankinson: I don't see why not. The only reason to not do it is if the TEI markup isn't supported in this context, but I don't think that's the case.

rettinghaus: Done.

ahankinson (approved):

bwbohl:
Would you be so kind as to file an issue that the sourceDesc should be filled with some real content? Or maybe we can come up with a sensible proposal in this PR?

ahankinson:
@bwbohl I think it's pretty normal to have empty required tags, if there's nothing obvious that goes in them. AFAIK, `sourceDesc` is the TEI's manuscript description section; not sure why it's required, but I think the fact that it's empty is OK?

rettinghaus:
Yes, it is common practice to leave this empty. However I opened #877 for a later discussion.

PR #881, opened 2022-01-12
Clarify that @oct in <note> changes between B and C

alastair:
The main documentation for `<note>` says that pname and oct conform to Acoustical Society of America representation, but didn't say that oct changes between B and C, not G-A. The tutorial at https://music-encoding.org/tutorials/101-quickstart  also links to this Wikipedia page, so I added it here for more detail and added a sentence for clarification of oct.

I'm not sure how to build these docs to test them - I copied the `<ref>` syntax from another source file, I think it's correct.


rettinghaus (approved):

bwbohl (approved):

bwbohl:
thx @alastair

PR #884, opened 2022-01-12
Specify naming conventions of Scientific Pitch Notation

### LINKED ISSUE #883: Minor clarification of pitch specification in guidelines
@alastair has improved the description of pitch and octave by adding a link to Wikipedia to explain ASA pitch specification and an explicit pointer that the octave changes between B and C (and not between G and A). 

These are helpful, but we could be more explicit by clarifying the octave specification here rather than relying solely on the link. I'd suggest this has three parts (the first of which is now there): octave changes between B-C, C4(?) is middle C, and so B3 is the note one semitone below.

musicEnfanthen:
This PR further clarifies the naming conventions of Scientific Pitch Notation in the CMN chapter of the guidelines (section `Notes, Chords and Rests in CMN`).

Fixes #883.


rettinghaus (approved):

DILewis (changes requested):
I think this is great (it may be a little over detailed now, but that's a better place to be). Could we remove 'albeit', though? If something is absolutely needed here to replace it, 'in MEI', 'here' or 'but' are all preferable, I think.

musicEnfanthen:
Thanks for the feedback. 

Probably the main difficulty of the entire passage is that SPN uses capital letters to indicate pitch names (and the passage here refers to pitches mostly in capital letters, too), but the `pname` attribute in MEI uses lower case. I guess this could be another source of confusion when someone tries to enter upper case pitch name values?

So maybe (changes indicated in bold formatting):

The default values for `pname` and `oct` conform to the Scientific Pitch Notation (SPN) (also known as American Standard Pitch Notation) **in a slightly modified way**; that is, the letters A–G, **but** in lower case, indicate the musical note name of a pitch, and the numbers 0–9 indicate the octave range to which a note belongs. The value for <att>oct</att> changes between B and ...

Would that work?

DILewis:
How about something like:

"The default values for `@pname` and `@oct` use the Scientific Pitch Notation (SPN) (also known as American Standard Pitch Notation), where the letters A-G indicate the musical note name of a pitch, and the numbers 0–9 indicate the octave range to which it belongs. `@pname` differs only by using lower case values for pitches (a-g instead of A-G). The value for oct changes between B and ..."

source/docs/04-cmn.xml
```diff
@@ -256,13 +256,13 @@
...
+                        <p>The default values for <att>pname</att> and <att>oct</att> conform to the <ref target="https://en.wikipedia.org/wiki/Scientific_pitch_notation">Scientific Pitch Notation (SPN)</ref> (also known as American Standard Pitch Notation); that is, the letters A–G, albeit in lower case, indicate the musical note name of a pitch, and the numbers 0–9 indicate the octave range to which a note belongs. The value for <att>oct</att> changes between B and C, that is, octave ranges go from C, D, …, G, A, to B. For example, the so-called <hi rend="italic">middle C</hi> or <hi rend="italic">c'</hi> (the C in the middle, i.e. fourth C key from left, on a standard 88-key piano keyboard) is represented on the first ledger line in G clef notation and labelled as <hi rend="italic">C4</hi> in the naming convention of SPN, the note one semitone below would be labelled <hi rend="italic">B3</hi>, and <hi rend="italic">A4</hi> would refer to the first A above <hi rend="italic">C4</hi>.</p>
```
```suggestion
                        <p>The default values for <att>pname</att> and <att>oct</att> conform to the <ref target="https://en.wikipedia.org/wiki/Scientific_pitch_notation">Scientific Pitch Notation (SPN)</ref> (also known as American Standard Pitch Notation); that is, the letters A–G indicate the musical note name of a pitch, and the numbers 0–9 indicate the octave range to which a note belongs. <att>pname</att> values differ from this convention only by using lower case values for pitches (a–g instead of A-G). The value for <att>oct</att> changes between B and C, that is, octave ranges go from C, D, …, G, A, to B. For example, the so-called <hi rend="italic">middle C</hi> or <hi rend="italic">c'</hi> (the C in the middle, i.e. fourth C key from left, on a standard 88-key piano keyboard) is represented on the first ledger line in G clef notation and labelled as <hi rend="italic">C4</hi> in the naming convention of SPN, the note one semitone below would be labelled <hi rend="italic">B3</hi>, and <hi rend="italic">A4</hi> would refer to the first A above <hi rend="italic">C4</hi>.</p>
```

DILewis (approved):

source/docs/04-cmn.xml
```diff
@@ -256,13 +256,13 @@
                            <item>octave (using @oct)</item>
                            <item>duration (using @dur)</item>
                         </list>
-                        <p>A typical note, in this case a quarter note C4, is therefore encoded like so: <figure>
+                        <p>A single note, in this case a quarter note C4, is therefore encoded like so: <figure>
```
ahankinson:
```suggestion
                        <p>A single note, in this case a quarter note C4, is therefore encoded as: <figure>
```

source/docs/04-cmn.xml
```diff
@@ -256,13 +256,13 @@
...
+                        <p>The default values for <att>pname</att> and <att>oct</att> conform to the <ref target="https://en.wikipedia.org/wiki/Scientific_pitch_notation">Scientific Pitch Notation (SPN)</ref> (also known as American Standard Pitch Notation); that is, the letters A–G indicate the musical note name of a pitch, and the numbers 0–9 indicate the octave range to which a note belongs. <att>pname</att> values differ from this convention only by using lower case values for pitches (a–g instead of A-G). The value for <att>oct</att> changes between B and C, that is, octave ranges go from C, D, …, G, A, to B. For example, the so-called <hi rend="italic">middle C</hi> or <hi rend="italic">c'</hi> (the C in the middle, i.e. fourth C key from left, on a standard 88-key piano keyboard) is represented on the first ledger line in G clef notation and labelled as <hi rend="italic">C4</hi> in the naming convention of SPN, the note one semitone below would be labelled <hi rend="italic">B3</hi>, and <hi rend="italic">A4</hi> would refer to the first A above <hi rend="italic">C4</hi>.</p>
```
ahankinson:
```suggestion
                        <p>The default values for <att>pname</att> and <att>oct</att> conform to the <ref target="https://en.wikipedia.org/wiki/Scientific_pitch_notation">Scientific Pitch Notation (SPN)</ref> (also known as American Standard Pitch Notation); that is, the letters A–G indicate the musical note name of a pitch, and the numbers 0–9 indicate the octave range to which a note belongs. <att>pname</att> values differ from this convention only by using lower case values for pitches (a–g instead of A-G). The value for <att>oct</att> changes between B and C, that is, octave ranges go from C, D, …, G, A, to B. For example, <hi rend="italic">middle C</hi> or <hi rend="italic">c'</hi> (the C in the middle, i.e. fourth C key from left, on a standard 88-key piano keyboard) is represented on the first ledger line in G clef notation and labelled as <hi rend="italic">C4</hi>, in the naming convention of SPN. The note one semitone below would be labelled <hi rend="italic">B3</hi>, and <hi rend="italic">A4</hi> would refer to the first A above <hi rend="italic">C4</hi>.</p>
```

ahankinson: The diff doesn't really help with the suggestion. What I suggested is removing "so-called" (it's common enough practice to call it middle C that we don't need to caveat it), and splitting the "For example" sentence as it was starting to run on.  So:

"...first ledger line in G clef notation and labelled as C4 in the naming convention..."

becomes

"...first ledger line in G clef notation and labelled as C4. In the naming convention..."

source/docs/04-cmn.xml
musicEnfanthen: Thank you.

source/docs/04-cmn.xml
musicEnfanthen: Thank you for this observation, too. I initially used "so-called" because there are (non-piano-guided) views that see other C's than C4 in the middle of their instrumental ambitus and call them "middle C". But probably you are right and these cases are less common.

Agree with splitting the long example sentence. If you don't mind, I would split it after "in the naming convention of SPN", since the main point of the first part of this sentence is to clarify that middle C is expressed as C4 in SPN ;)

For the diff, I was wondering about the missing highlighting colors already, maybe something is wrong inline with my last commit (the three dots?)?

musicEnfanthen: I checked once more and there seems to be everything OK with the formatting, but a kind of length maximum for Github GUI's diff highlighting mechanism.

ahankinson: I've updated the sentence break.

musicEnfanthen: Many thanks!

bwbohl (approved):

PR #891, opened 2022-01-24
Add quotes to customization path example

annplaksin:
Just a quick thing: I realized using the `-Dcustomization.path="bla"` requires quotes around the value. Maybe the build instructions are easier to follow if this is added to the example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ Nevertheless it is possible to build any customization locally in your working c
 
    * Build a specific customization's RNG schema:
      ```bash
-     ant -lib lib/saxon/saxon9he.jar -Dcustomization.path=[PATH/TO/YOUR/CUSTOMIZATION] build-rng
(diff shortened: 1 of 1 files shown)


bwbohl:
does not hold true for macOS, I'll test it on Windows

annplaksin:
@bwbohl Okay, I was wondering why the quotes weren't there already. I tried it on Windows and only `-Dcustomization.path="customizations/mei-Mensural.xml"` worked. Every other variation like `...\\...`, `...\...` or `.../...` failed.

I'm checking it on my WSL now. It seems on that Ubuntu WSL 2, it works with and without quotes...

bwbohl:
ok, I can approve that it makes a differnece on my windows cmd, curiuous whether this still holds true after #888

bwbohl (approved):

PR #892, opened 2022-01-24
Added sforzando/sforzato `sf` to `data.ARTICULATION`

### LINKED ISSUE #608: sforzato as articulation
Typical instantaneous dynamics instructions define a loudness that applies to all notes at and after the position of the instruction until another instruction specifies a different loudness level. This is different with sforzato. It (often if not usually) applies to only the notes at but not after its position, similar to an accent. Basically, sf is the "maximum accent". I suggest adding `<valItem ident="sf">` to data.ARTICULATION.

axelberndt:
This pull request adds `sf` to the value list of `data.ARTICULATION` as discussed in issue #608.

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -216,7 +216,10 @@
           <desc>Soft accent, see SMuFL Articulation supplement (U+ED40–U+ED4F).</desc>
         </valItem>
         <valItem ident="sf">
-          <desc>Sforzando, sforzato.</desc>
+          <desc>Sforzando.</desc>
+        </valItem>
+        <valItem ident="sfz">
+          <desc>Sforzato.</desc>
         </valItem>
         <valItem ident="stacc">
           <desc>Staccato (Unicode 1D17C).</desc>
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -215,6 +215,9 @@
         <valItem ident="acc-soft">
           <desc>Soft accent, see SMuFL Articulation supplement (U+ED40–U+ED4F).</desc>
         </valItem>
+        <valItem ident="sf">
+          <desc>Sforzando, sforzato.</desc>
+        </valItem>
         <valItem ident="stacc">
           <desc>Staccato (Unicode 1D17C).</desc>
         </valItem>


ahankinson:
I believe `sforzando` and `sforzato` are two different symbols? `sf` and `sfz`?

axelberndt:
Yes, the visual rendition may differ. But I found no clear logical distinction in the literature. Please correct me if I am wrong.

ahankinson:
If we take it as an indicator which SMuFL glyph to use, then there are differences: 

https://smufl-browser.edirom.de/index.html?glyphname=dynamicSforzando&glyphname=dynamicSforzando1&glyphname=dynamicSforzato

(I'm actually not sure I agree with that first one; I would call that 'subito', but SMuFL calls it "Sforzando".) 

"Visual" and "logical" are not always clearly separable. If a composer wrote an "sfz" and MEI captures it as "sf", and Verovio only renders it as `sf`, then this is lossy encoding. We don't know what the composer's "logical" meaning was -- it could be based on their own compositional tradition, or it could be because they understand it as a different effect.

axelberndt:
In issue #608 we had this discussion going into the opposite direction, i.e. encoding it all as `acc` and just adapting the glyph. Now, as sforzando/sforzato is logically clearly different to an accent it makes sense to have it as a separate item in the value list, hence this pull request. 

However, there seems to be no logical difference between `sf` and `sfz`. Thus it makes sense to have only one representation for it in the value list and visual adjustment in the encoding by @glyph.num if required.

The alternative would be this:
````xml
<valItem ident="sf">
  <desc>Sforzando.</desc>
</valItem>
<valItem ident="sfz">
  <desc>Sforzato.</desc>
</valItem>
````
I can live with both. But according to the discussion in issue #608 the alternative is less preferable.

**Comment by ahankinson** [comment]
Hm, I read #608 and now just re-read it. I can't see where this was determined, though.

I do see a comment on "If we allow this, we should allow all of these as well." But I think that's a slightly different issue. In this case, you're saying

```xml
<desc>Sforzando, sforzato.</desc>
```

which, I think, is incorrect, since they can clearly have different symbols. Whether or not they have different **interpretations** is left to the interpreter.

So I think I would prefer your alternative; If people want to use this as a floodgate to justify adding `sffffffffffzzzzzffffzzss`, then we can cross that bridge when we come to it, but it won't be in this PR. :)

axelberndt:
Alright, I close this PR and start a new one with the alternative.

ahankinson:
You don't need to close this! Just add a commit to this branch... 

Closing it and re-opening a new PR makes it difficult for other reviewers to see the history of changes and the discussion that led to the current state. 

Edit: I see that you did add it; this is fine! You don't need to open a new PR.

axelberndt:
Done. Sorry.

lpugin:
This PR was discussed further during our last meeting on Friday 2022-01-28. It does remain unclear why having sforzando/sforzato is really necessary to be added as an articulation because these can currently be already be encoded with `<dynam>sf</dynam>`. Regarding the attachment to a particular note, this is perfectly possible using `@startid`. Regarding the argument of the semantic, having them in a `<dynam>` is no less semantic than what is currently provided for all dynamics encoded in MEI, including the most common ones such as `forte` or `piano`. 

By definition, they do apply to the onset of a note. However, the general category of dynamic seems appropriate for them even though it is true that the difference between an articulation and a dynamic can be difficult to make. This situation can be encountered elsewhere in MEI, for example between dynamics and hairpins, or between directives and other control events (dynamics, laissez vibrer, etc.). These are general ambiguities in music notation we need to live with. As it stands, it does not look like adding more values to articulations solves much. Quite on the contrary, it seems that would introduce more confusion.

axelberndt:
Thanks Laurent for your reply. Well, to put it pointedly, sforzato is not a dynamics instruction. It is an articulation, just as MEI does not handle accents as dynamics but articulation. `@startid` does not fix the problem. The discussion in issue #608 provides more details on this.

Regarding ambiguity and confusion, at the moment MEI has simply a wrong representation of sforzato. I understand that, for backward compatibility reasons, it would not be removed from the dynamics. But for correctness it should certainly be added to articulation.

lpugin:
You are very assertive. Everything would be so much easier if we could say this is wrong or this is correct in a simple way. I certainly do not agree that we can say that sforzato is not a dynamic instruction - which does not mean it does not also belong the the articulation realm. 

You seem to imply that sforzato should be removed from dynamics in the next major release of MEI where backward compatibility can be broken. How would you encode something like this, then?
image
Or are you suggesting we should have articulations as control events? I think that would be necessary to encode such cases, but that is a whole different story.

axelberndt:
Sorry if that came through as too strong an assumption. As I said, I expressed it pointedly. I do not suggest to remove sforzato from dynamics. You have a good example for that case. Can you share the source? I am interested to see what follows.

In this PR I suggest to add it also to articulation because that is how it is quite frequently used. ... [doing a quick web search] ... [Here](https://www.musicreadingsavant.com/5-music-articulation-symbols-you-need-to-know/) is an example of an instructional text. And [this](https://fthmb.tqn.com/71JE-D5AsZNd42GutQ3Og25CNOg=/3365x2641/filters:fill(auto,1)/the-hut-on-fowls-legs-baba-yaga-score-for-9-movement-of-pictures-at-exhibition-suite-for-piano-by-modest-mussorgsky-1839-1881-549582181-57c883063df78c71b641ca0f.jpg) is another sheet music example where it is used this way.

What is the difference? As a dynamics instruction it says, play everything with this loudness until another instruction is met. As an articulation it says, play this note with sforzato (not the subsequent ones).

Anyway, this discussion has alreaady taken place in the issue before preparing this PR. If there is so much concern against it, feel free to close this PR. I may have rushed ahead with it too swiftly.

ahankinson:
> `@startid` does not fix the problem.

I just looked again through #608 and could not see any mention of why `@startid` couldn't be used. Could you please clarify?

craigsapp:
> Regarding ambiguity and confusion, at the moment MEI has simply a wrong representation of sforzato. I understand that, for backward compatibility reasons, it would not be removed from the dynamics. But for correctness it should certainly be added to articulation.

Is this an ontological issue, or does this cause you data processing problems, since articulations are children of `<notes>` while dynamics are more distant relatives of notes?

> Well, to put it pointedly, sforzato is not a dynamics instruction. It is an articulation, just as MEI does not handle accents as dynamics but articulation. 

It is an accent, but it has a graphical syntax like dynamics.    It cannot be incorporated correctly into the stacking order for automatic layout if it is classified as an articulation, so to implement such an accent in verovio, it would have to be converted internally from an accent into a dynamic.

As discussed before (I believe), Humdrum treats sforzato as articulations, using the letter `z` to indicate them:

```tsv
**kern	**dynam
2cc'~z>	pp
*-	*-
!!!RDF**kern: > = above
```

Screen Shot 2022-01-31 at 6 24 12 AM

Here is example graphical behavior for sf that is different from other accents:

Screen Shot 2022-01-31 at 6 33 11 AM

https://github.com/craigsapp/beethoven-piano-sonatas/blob/master/reference-edition/sonata01-1.pdf
https://verovio.humdrum.org/?file=beethoven/sonatas/sonata01-1.krn#mh73

In Humdrum, these sf are attached to the notes, where they are placed before the note rather than centered on a note, and there is an extra layout parameter `rj` meaning "right justification" to place it on the note with right justification:

```
**kern	**kern	**dynam
*	*^	*
*clefF4	*clefG2	*clefG2	*
*k[b-e-a-d-]	*k[b-e-a-d-]	*k[b-e-a-d-]	*
*f:	*f:	*f:	*
*MM240	*MM240	*MM240	*
=73	=73	=73	=73
(4FF-	2r	8d-L	.
.	.	8B-	.
2EE-)	.	8d-	.
.	.	8B-J	.
!	!LO:DY:rj	!	!
.	2gz	8d-L	.
.	.	8B-	.
!LO:DY:rj	!	!	!
[4E-z	.	8d-	.
.	.	8B-J	.
!!LO:LB:g=z
=	=	=	=
*	*v	*v	*
*-	*-	*-
!!!RDF**kern: > = above
!!!RDF**kern: < = below
```

Screen Shot 2022-01-31 at 6 39 23 AM

Here is how I would encode @lpugin's example, which has one `sf` symbol shared across multiple staves (and in theory multiple layers on a staff):

Screen Shot 2022-01-31 at 6 56 10 AM

```tsv
**kern	**kern	**dynam
*clefF4	*clefF4	*
*k[b-]	*k[b-]	*
8BBB-'^^zy 8BB-'^^zy	8D'^^zy 8G'^^zy	sf
=	=	=
*-	*-	*-
```
Screen Shot 2022-01-31 at 6 59 23 AM


For readability I encode it as a dynamic, but  I should also add `zy` to each note.  The `z` indicate that the note should be performed with `sf`, and the `y` means that this is an interpretation (i.e., the note-level `sf` is invisible).

Perhaps more properly, I could make one chord with `z` not marked with `y` to indicate that one chord causes a visual marker:

Screen Shot 2022-01-31 at 7 03 47 AM

NB: In MEI, `dynam@plist` is best to be used to indicate multiple notes the dynamics accent applies to.

Humdrum's focus is on analysis where it makes most sense to associate with the note, while MEI's focus is on notation where it is more convenient to group with dynamics since it has the same graphical syntax as dynamics.  Since there is no conceptual overlap between regular dynamics and articulation dynamics, there is no loss of information when encoding in the current MEI style.

craigsapp:
> > `@startid` does not fix the problem.
> I just looked again through #608 and could not see any mention of why `@startid` couldn't be used. Could you please clarify?


Both `@tstamp` and `@startid` can be used to place dynamics:

Screen Shot 2022-01-31 at 7 19 05 AM

If `@tstamp` is used, then this should mean that it applies to all notes (or chords) on the same staff that have the same `@tstamp`.  If there are multiple notes (or chords) at the same time but only one of them is accented, then `@startid` should be used.   If the dynamic is attached to one staff, but applies to notes on another staff, then `@plist` should be used.  When `@plist` is used, then it should override the inferred links generated by `@tstamp` or `@startid` (as for @lpugin's example above).

<details>
<summary> Click to view MEI data for above example </summary>

```xml
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0.0-dev">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2022-01-31T07:16:02" version="3.9.0-dev-9f581c8">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="mstffdh">
    <score xml:id="sdqdweg">
     <scoreDef xml:id="skujnlx">
      <staffGrp xml:id="sbdqlz9">
       <staffDef xml:id="sqfuu1c" n="1" lines="5">
        <clef xml:id="cgfolzm" shape="G" line="2" />
        <meterSig xml:id="metersig-L2F1" count="3" unit="4" />
       </staffDef>
      </staffGrp>
     </scoreDef>
      <measure xml:id="measure-L1">
       <staff xml:id="sqgtl65" n="1">
         <note xml:id="note-L4F1" dur="4" oct="4" pname="c" accid.ges="n" />
         <note xml:id="note-L6F1" dur="4" oct="4" pname="d" accid.ges="n" />
         <note xml:id="note-L7F1" dur="4" oct="4" pname="e" accid.ges="n" />
        </layer>
       </staff>
       <dynam xml:id="dynam-L6F1" staff="1" tstamp="2.000000">sf</dynam>
       <dir xml:id="dir-L5F1" place="above" staff="1" tstamp="2.000000">
        <rend xml:id="r3uj4sc" fontstyle="normal">tstamp</rend>
       </dir>
      </measure>
      <measure xml:id="measure-L8" n="2">
       <staff xml:id="staff-L8F1N1" n="1">
        <layer xml:id="layer-L8F1N1" n="1">
         <note xml:id="note-L9F1" dur="4" oct="4" pname="c" accid.ges="n" />
         <note xml:id="note-L11F1" dur="4" oct="4" pname="d" accid.ges="n" />
         <note xml:id="note-L12F1" dur="4" oct="4" pname="e" accid.ges="n" />
        </layer>
       </staff>
       <dynam xml:id="dynam-L11F1" staff="1" startid="note-L11F1">sf</dynam>
       <dir xml:id="dir-L10F1" place="above" staff="1" tstamp="2.000000">
        <rend xml:id="ruvsse3" fontstyle="normal">startid</rend>
       </dir>
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

</details>

axelberndt:
@ahankinson: I do not see how `@startid` indicates that the sforzato is applied only to one note and the previous dynamics instruction keeps its validity for all notes that follow.

@craigsapp: I regard it as an ontological issue. Of cource, it also causes processing problems. When users ecode a sforzato articulation, it is undesired to play all music at full loudness from that point on.

My understanding of MEI is not that of a purely graphical music representation.

ahankinson:
If you have `@startid` with no `@endid`, then the target is a single note. (you can also have startid/endid be the same ID) It's an encoding problem (or an explicit decision) if you have another dynamic that overlaps that.

axelberndt:
Hm, I am not aware of a case where users encode overlapping dynamics. And it seems a rather complicated way to achieve the behaviour of articulation. What is the problem with adding it to articulation? On the logical side there is no ambiguity. One is the dynamics sforzato, the other is the articulation sforzato. They are semantically distinct.

ahankinson:
I can't think of any specific ones offhand, but I can certainly imagine a `sfz` on a specific note coming in the middle of a longer crescendo or decrescendo. 

@craigsapp indicates that one of the problems with adding it to articulations is that it doesn't "fit" well with the stacking setup for the existing articulations, which tend to be more symbol-like than text-like. 

> One is the dynamics sforzato, the other is the articulation sforzato. They are semantically distinct.

I'm not aware of any existing distinction between those two things, and I'm not sure anybody has ever thought about it in these ways, since they haven't been forced to make this distinction in such clear-cut ways.

As @craigsapp has shown, it is completely possible to get exactly what I think you want with:

```
<dynam xml:id="dynam-L11F1" staff="1" startid="note-L11F1">sf</dynam>
```

Is there a reason, beyond the name that it's being given, that this isn't sufficient?

axelberndt:
MEI is not designed around graphical rendering issues, I think. Having problems with the stacking of graphical symbols should not be a reason against it.

Your code snippet links the instruction to a certain note, where it is meant to start. But how does it indicate whether this is a dynamics instruction or an articulation? Simply put, where does it end and what happens after it?

lpugin:
I think that one issue with the PR is that it sets a controlled list for sforzato-like values. As pointed out already, it does not make much sense to have `sf` without having `sff`. But then why not `smfz` ?
https://www.henle.de/blog/en/2015/09/28/unconventional-dynamic-markings/

axelberndt:
We can't have them all, so we take none? Hm. Why?
No, really, if you have so much trouble with it, close this PR.

craigsapp:
> Your code snippet links the instruction to a certain note, where it is meant to start. But how does it indicate whether this is a dynamics instruction or an articulation? Simply put, where does it end and what happens after it?

I see your point, but the current convention is to use `@startid` for single attachment points in the score and add `@endid` when there is a second attachment point (i.e., there is more of a graphical focus rather than a semantic focus).   Verovio currently does not realize dynamics when rendering to MIDI, but when it does, this issue (instantaneous dynamics versus continuous dynamics) will be important, but also easy to solve since any dynamic containing the characters `r`, `s` or `z` are instantaneous dynamics (apply only to the current note(s)), and dynamics without those letters are continuous (apply to all successive notes until the next dynamic).  But there is a far worse problem: Piano music can have separate continuous dynamics streams at the same time, such as different ones on separate staves, or even in separate layers on the same staff.  This is not mentioning cases where there are cross-staff layers (does the home staff control the dynamics state, or the display staff's dynamic state?).

 `sf` cannot have a duration (and therefore `@endid` implicitly has the same value as `@startid`) since there is no concept of this in CMN otherwise.  If there were, it would behave similar to articulations, where _simile_ is used to continue articulations on notes (currently the best semantic encoding would be to add articulations to each note that has an articulation implied by the _simile_, but make that articulation not visible in notation rendering).


If you want to be semantically correct, `@endid` could be added, pointing to the same id:

```xml
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0.0-dev">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2022-01-31T08:02:52" version="3.9.0-dev-9f581c8">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="mxf5uz7">
    <score xml:id="sc7sc81">
     <scoreDef xml:id="sx92z71" midi.bpm="400.000000">
      <staffGrp xml:id="skkz22v">
       <staffDef xml:id="sy6v95m" n="1" lines="5">
        <clef xml:id="c5ldqwk" shape="G" line="2" />
       </staffDef>
      </staffGrp>
     </scoreDef>
      <measure xml:id="measure-L1">
       <staff xml:id="s3iufpy" n="1">
         <note xml:id="note-L2F1" dur="1" oct="4" pname="c" accid.ges="n" />
        </layer>
       </staff>
       <dynam xml:id="dynam-L2F1" staff="1"  startid="note-L2F1" endid="note-L2F1">sf</dynam>
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

`sf` can only apply to a single note and cannot have a duration, so adding `@endid` seems redundant.  Verovio does not care if `@endid` is given in this case, since both with and without will give the same graphical result (so this particular point is more about MEI than verovio).

-----------------------------------------------

> On the logical side there is no ambiguity. One is the dynamics sforzato, the other is the articulation sforzato. They are semantically distinct.

You are assigning a limiting definition to `<dynam>` that is not intended.  Here is the description of dynam from the documentation:

https://music-encoding.org/guidelines/v4/elements/dynam.html

> (dynamic) – Indication of the volume of a note, phrase, or section of music.

`sf` falls under "volume of a note".


--------------------------------

It is possible to informally clarify the semantic meaning of accent dynamics by using `dynam@type="artic"`:

```xml
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0.0-dev">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2022-01-31T08:02:52" version="3.9.0-dev-9f581c8">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="mxf5uz7">
    <score xml:id="sc7sc81">
     <scoreDef xml:id="sx92z71" midi.bpm="400.000000">
      <staffGrp xml:id="skkz22v">
       <staffDef xml:id="sy6v95m" n="1" lines="5">
        <clef xml:id="c5ldqwk" shape="G" line="2" />
       </staffDef>
      </staffGrp>
     </scoreDef>
      <measure xml:id="measure-L1">
       <staff xml:id="s3iufpy" n="1">
         <note xml:id="note-L2F1" dur="1" oct="4" pname="c" accid.ges="n" />
        </layer>
       </staff>
       <dynam xml:id="dynam-L2F1" staff="1" type="artic" tstamp="1.000000">sf</dynam>
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

craigsapp:
> We can't have them all, so we take none? Hm. Why? No, really, if you have so much trouble with it, close this PR.

In my last post, I point out that the current description of `<dynam>` is more encompassing that your interpretation.  You want to create two subcategories of dynamics (instantaneous and continuous), and then limit `<dynam>` to the continuous set, and move the instantaneous ones to `<artic>`.

I do not see a semantic advantage of allowing duplicate placement of instantaneous dynamics in both `<artic>` and `<dynam>` since there is no problem differentiating between continuous and instantaneous dynamics when both are inside of `<dynam>`.  Your argument is related more to semiotics rather than semantics.  Semiotics being about how to represent a symbol (or more specifically how meaning is derived from symbols), and semantics being about the actual meaning of a symbol.   In both the current behavior of MEI and your proposed one, the semantics would be the same (and if they are not, then it should be fixed).

I am sure @pe-ro will say to do it both ways 😜

axelberndt:
I went already through the whole process of expressive MIDI rendering during the development of meico and the MPM export. I stumbled over this case which is the reason why I initiated the issue in the first place.

You distinguish between instantanious and continuous dynamics instructions. The former creates stepwise dynamics change, the latter a continous transition. None represents a dynamics peak (up and down again). Likewise you cannot express an accent via dynamics.

So, to my understanding, MEI misses a feature. I did not regard adding it as a limitation. In fact, the addition is backward compatible and does not clash with anything in the dynamics section.

Anyway, your message came through. I close this.

craigsapp:
> You have a good example for that case. Can you share the source? I am interested to see what follows.

This use is fairly common in piano music.  Here is another example from Beethoven's first piano sonata that is ambiguous:

Screen Shot 2022-01-31 at 7 36 16 AM

https://github.com/craigsapp/beethoven-piano-sonatas/blob/master/reference-edition/sonata01-1.pdf

Does the `sf` apply to all notes, or to just one note?  If one note then it must be the D.

Here is another example near the end of the same movement that matches better to @lpugin's example:

Screen Shot 2022-01-31 at 7 38 57 AM

In this case all notes on both staves should be played `sf` (and the sf is applied to the notes that follow it visually, i.e., the second and fourth chords).


--------------------

It is dangerous for @lpugin to bring up this example to justify categorization of `sf` et al. as purely `<dynam>`, because the horizontal marcato accent mark is also used in the same manner:

https://github.com/craigsapp/chopin-mazurkas/blob/master/reference-edition/mazurka50-2.pdf

Screen Shot 2022-01-31 at 9 22 48 AM

The hairpin-looking things on the last quarter note of each measure is an accent mark shared across all notes on both staves at that point (functioning similar to `sf`).  

This was discussed c. 2005 at the Detmold meeting, but I have never seen a solution or further discussion about them? (unless I am missing something)

axelberndt:
I would encode these accents as articulations that are childern of the `<measure>` and either linked to all notes via `@plist` or simply placed at a certain `@tstamp` where it is implicitly applied to all notes at this position. Same with your second sforzato example. In the first example it is unclear if the sforzato should go on.

axelberndt:
Ah, sorry, my fault. `<artic>` cannot be child of measure. It might be placed in a `<layer>` instead.

craigsapp:
> I would encode these accents as articulations that are childern of the <measure> and either linked to all notes via `@plist` or simply placed at a certain `@tstamp` where it is implicitly applied to all notes at this position. 

That is a theoretical encoding that is not currently allowed, since artic cannot be a child of measure:

https://music-encoding.org/guidelines/v4/elements/artic.html

Screen Shot 2022-01-31 at 9 52 09 AM

(I just saw your message about the same thing that you posted as I was writing this).

Encoding in such a way is good, but this is only valid for the horizontal marcato and not for other articulations.  Ideally this type of accent would be promoted to an element (`<hairpinAccent>`?) which would have similar graphical language as `<dynam>`.

> Same with your second sforzato example. In the first example it is unclear if the sforzato should go on.

It is definite that the sforzato applies to a single chord (listen to any recording of this music).  Cite an example where a sforzato should go on.

axelberndt:
I found an instructional text for continuous sforzato [here](https://www.musicians-place.de/glossar-musikalische-grundbegriffe/57-sforzato.html), unfortunately only in German. I did understand Laurent's post in the same way.

rettinghaus:
Although it is said plenty already, I just wanted to point out that we also have the `@plist` attribute on `dynam`. So you actually can list all "active participants" and thus can be very precise to which notes the text-based articulation applies.

craigsapp:
> I found an instructional text for continuous sforzato here, unfortunately only in German. I did understand Laurent's post in the same way.

I can translate it into English no problem with my excellent Google translating skills:


> Occasionally one comes across the sforzato written as an abbreviation. The entire phrase is emphasized, i.e. played louder and more accentuated than the other notes. The sforzato instruction should then be overridden by another instruction, for example a dynamic instruction such as mezzopiano.

Screen Shot 2022-01-31 at 11 15 40 AM

That is a good start, but I would still like to see a real-world example (by a fairly well known composer printed by a reputable publishing house).

This notation is ambiguous, and should be avoided.  The strange placement of the `mp` helps to identify that it could be a continuous `sf`.  Adding a dashed line or using _simile_ would disambiguate such notation:

https://dictionary.onmusic.org/terms/3161-simile#:~:text=A%20directive%20to%20perform%20the,articulation%20symbol%20(see%20example).


Here is an example for Beethoven piano sonata 18 simile (but not using that word, which is more common):

https://github.com/craigsapp/beethoven-piano-sonatas/blob/master/reference-edition/sonata18-2.pdf

Screen Shot 2022-01-31 at 11 26 00 AM


Beethoven sonata 16 movement 2:

https://github.com/craigsapp/beethoven-piano-sonatas/blob/master/reference-edition/sonata16-2.pdf

Screen Shot 2022-01-31 at 11 36 13 AM

The word "staccato" here is implying the staccatissimos from the previous measure in the other hand (but no staccato on the first note where the text starts).



Here is a strange case that matches your example, but probably a mistake in the typesettings where staccatos on the second beamed group are likely missing:

Mozart K 576, movement 2 near end:

https://github.com/craigsapp/mozart-piano-sonatas/blob/master/reference-edition/sonata17-2.pdf

Screen Shot 2022-01-31 at 11 29 45 AM

axelberndt:
Yes, admittedly I doubt the reliance of the source I gave.

I just had a conversation with a composer/orchestrator/music theory lecturer here at the university of music. He clearly stated "sforzato _is_ an articulation." Which brings us back to the outset of this whole discussion. Why isn't it an articulation in MEI in the first place?! "Because it looks like a dynamics instruction" and "it is hard to render graphically" are no valid answers.

Anyway, I think it became clear that you don't want it to be an articulation in MEI. I accept this.

kepper:
I hesitate to join this party, as I see good arguments on both sides, and I doubt that my addition will make things any easier. In general, I sympathize with the idea of calling sforzato an articulation. The question that comes with this is which consequences such a statement will trigger. Apparently, just adding two or three values to a closed list of articulations – for which we probably all agree that it has room for improvement, to put it mildly – isn't sufficient. One major argument that hasn't been spelled out (unless I've missed it) is that `<artic>` isn't transcriptional: It has no child content, but uses the said `@artic` attribute with its closed list of values. This means that there is no way to have a controlled list on one side, but also the the full freedom to encode whatever composers / engravers have left us with – the Henle link from @lpugin is quite to the spot here. I believe it is necessary to support that kind of things in MEI, and I'm sure my Beethoven colleagues could easily bring up an example where LvB changed a "sf" statement to an "sfz" with an additional penstroke. Attributes are not capable of doing this. A potential solution could be to allow content inside `<artic>`, but that will be quite a can of worms. The other solution, apparently preferred by many here, is to relax a bit on the term "dynamics": (Most) articulation statements will eventually effect the loudness of a note. In that wider sense, they actually _are_ dynamics, and `<dynam>` would be appropriate. Yes, `<artic>` is more specific, but maybe it is already too specific – which could very well be a modelling issue in MEI. What it comes down to: The problem behind all this is much bigger than the (tempting!) simple solution can deal with, and it surely requires very careful re-consideration of many other existing models. That's not a reason to not think about this, but it may be an argument to not rush into this…

craigsapp:
Here is discussion from Behind Bars (by Elaine Gould), page 115:

Screen Shot 2022-01-31 at 1 09 00 PM

She is saying that `sf` and `sfz` are instantaneous, but `rf`, `rfz` and the textual abbreviation _rinf._ may also be used for continuous application.

-------------------------------------

> Anyway, I think it became clear that you don't want it to be an articulation in MEI. I accept this.

Sforzatos are articulations in the musical sense and I doubt that anyone here is claiming otherwise (see [straw man argument](https://en.wikipedia.org/wiki/Straw_man)), `<dynam>` and `<artic>` are not currently defined in the way that you want (hence there is no `sf` articulation).  I leave it up to you Germans to argue about fine ontological distinctions, which I am claiming are semiotic (semantic sugar) rather than semantic itself.  

In other words, I see no loss of semantics with the current system since instantaneous dynamics can be made explicit by providing  either `@endid` or `@pname`.   If they are continuous, then `@endid` should not be given.

I have pointed out that the description of `<dynam>`: https://music-encoding.org/guidelines/v4/elements/dynam.html

> (dynamic) – Indication of the volume of a note, phrase, or section of music.

Includes "volume of a note", which is an instantaneous dynamics marking.

----------------------------------

There is also the related (but reverse) problem of semantic encoding of _simile_.  How should this be done, since `<artic>` is a child of notes and not of measure?   This is more problematic semantically.   

Here is an example:

Screen Shot 2022-01-31 at 1 34 28 PM

```xml
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0.0-dev">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2022-01-31T13:34:08" version="3.9.0-dev-9f581c8">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="mgs9l6w">
    <score xml:id="sihbkm1">
     <scoreDef xml:id="s5o7426">
      <staffGrp xml:id="stvqxcv">
       <staffDef xml:id="sl7fx4m" n="1" lines="5">
        <clef xml:id="cirdo33" shape="G" line="2" />
        <meterSig xml:id="metersig-L2F1" count="4" unit="4" />
       </staffDef>
      </staffGrp>
     </scoreDef>
      <measure xml:id="measure-L1" n="1">
       <staff xml:id="smaymm8" n="1">
         <note xml:id="note-L4F1" dur="4" oct="4" pname="c" accid.ges="n">
          <artic xml:id="artic-L4F1" artic="stacc" />
         </note>
         <note xml:id="note-L5F1" dur="4" oct="4" pname="d" accid.ges="n">
          <artic xml:id="artic-L5F1" artic="stacc" />
         </note>
         <note xml:id="note-L6F1" dur="4" oct="4" pname="e" accid.ges="n">
          <artic xml:id="artic-L6F1" artic="stacc" />
         </note>
         <note xml:id="note-L7F1" dur="4" oct="4" pname="f" accid.ges="n">
          <artic xml:id="artic-L7F1" artic="stacc" />
         </note>
        </layer>
       </staff>
      </measure>
      <measure xml:id="measure-L8">
       <staff xml:id="staff-L8F1N1" n="1">
        <layer xml:id="layer-L8F1N1" n="1">
         <note xml:id="note-L10F1" dur="4" oct="4" pname="c" accid.ges="n">
          <artic xml:id="artic-L10F1" artic.ges="stacc" />
         </note>
         <note xml:id="note-L11F1" dur="4" oct="4" pname="d" accid.ges="n">
          <artic xml:id="artic-L11F1" artic.ges="stacc" />
         </note>
         <note xml:id="note-L12F1" dur="4" oct="4" pname="e" accid.ges="n">
          <artic xml:id="artic-L12F1" artic.ges="stacc" />
         </note>
         <note xml:id="note-L13F1" dur="4" oct="4" pname="f" accid.ges="n">
          <artic xml:id="artic-L13F1" artic.ges="stacc" />
         </note>
        </layer>
       </staff>
       <dir xml:id="dir-L9F1" place="below" staff="1" tstamp="1.000000">simile</dir>
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

In this case the staccatos are gestural, but the _simile_ text has no relation to the encoded articulations (if you want to remove the _simile_ text, you would also have to convert all of the following `@artic.ges` into `@artic`).

ahankinson:
Thanks all! Although the PR is closed now, I quite enjoyed following the discussion. Despite the change ultimately not getting accepted, I think some things were clarified. Thanks to @axelberndt for bringing this issue up, and to everyone else for providing review and feedback.

rettinghaus:
Revisiting this discussion again I wonder if someone could add a sentence or two about this to the Guidelines? It would definitely be a valued contribution.

PR #893, opened 2022-01-26
always specify full path

rettinghaus:
Couldn't build, because files were not found. This fixes it (on my machine).


rettinghaus:
@bwbohl pinging you, because I cannot assign anything.

bwbohl (approved):

Issue #82, opened 2015-05-28
remove fingerprint?


ahankinson:
_From [kep...@edirom.de](https://code.google.com/u/113787537940489926112/) on June 28, 2012 11:01:44_

<fingerprint> is hardly usable, as one has to remove the value before fingerprinting the file (otherwise you'll never receive the same value as you've stored there). Do we really need fingerprint, or can it go away safely?

_Original issue: http://code.google.com/p/music-encoding/issues/detail?id=82_

ahankinson:
_From [andrew.hankinson](https://code.google.com/u/andrew.hankinson/) on July 21, 2012 14:39:00_

For an audio file or image file, a <fingerprint> element would be very useful. Johannes is right -- fingerprinting the containing file is pretty much useless, but using it to describe an audio file would allow MEI to store references to exact recordings, even if the checksum differs on the digital file.

ahankinson:
_From [pd...@virginia.edu](https://code.google.com/u/103686026181985548448/) on July 21, 2012 14:59:15_

I'm confused (as usual).  I don't understand what you mean by the phrase "even if the checksum differs on the digital file".  If the checksum differed, then it wouldn't be the same file/recording.  Sorry, paint me a picture.  :-)

ahankinson:
_From [andrew.hankinson](https://code.google.com/u/andrew.hankinson/) on July 21, 2012 15:06:44_

A checksum (MD5, SHA) is used to give a way of verifying that the bits of two files are exactly the same. For audio, however, a WAV and an MP3 can be used to contain the same audio but have completely different checksums. (If you want to get really specific, we've also found that two different CD drives will rip WAV files from CDs with different checksums).

So the only way of referring to the same recording without taking into account the exact file is to refer to the fingerprint, since that refers to the perceptual features, not the digital.

ahankinson:
_From [pd...@virginia.edu](https://code.google.com/u/103686026181985548448/) on August 01, 2012 15:09:18_

I'm upgrading this issue to "bug" status.  I'll remove fingerprint for now and we can decide where/how to put it back later.

**Labels:** -Type-Other -Priority-Medium -Milestone-Release2013 Type-Bug Priority-High Milestone-ReleaseNext

ahankinson:
_From [pd...@virginia.edu](https://code.google.com/u/103686026181985548448/) on August 02, 2012 10:59:04_

This issue was closed by revision r640 .

**Status:** Fixed

PR #894, opened 2022-01-26
remove deprecated fingerprint

### LINKED ISSUE #885: Removing <fingerprint>
Proposal of @music-encoding/ig-metadata 
Short summary: We noticed the existence of the element `<fingerprint>` when updating the mei guidelines (chapter 3 Metadata). Our call to the community about usage, need, examples and description for the guidelines remained without result.
Two facts:
- This element is not described anywhere in the Guidelines (only the element definiton exists).
- comment in the encoding-tool [mei2010to2012.xsl](https://github.com/music-encoding/encoding-tools/blob/cb6bb19a75b983617f56cb4aad4eaaf9cb8558a5/mei2010To2012/mei2010To2012.xsl#L216): "use of mei:fingerprint deprecated"

The metadata ig will not include the element in the guidelines (chapter 3) and proposes to remove it.

This topic was discussed in the [Issue on fingerprint](https://github.com/music-encoding/metadata-ig/issues/13) at the metadata-ig repo.

rettinghaus:
This PR brings the long overdue removal of `fingerprint` that is deprecated since ten years. 
Closes #885

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -1195,21 +1195,6 @@
         Archival Description (EAD) standards.</p>
     </remarks>
   </elementSpec>
-  <elementSpec ident="fingerprint" module="MEI.header">
-    <desc>Contains a string that uniquely identifies an item, such as those constructed by combining
-      groups of characters transcribed from specified pages of a printed item or a file's
-      checksum.</desc>
-    <classes>
-      <memberOf key="att.common"/>
-      <memberOf key="att.authorized"/>
-      <memberOf key="att.bibl"/>
-      <memberOf key="model.msInline"/>
-      <memberOf key="model.physDescPart"/>
-    </classes>
-    <content>
-      <rng:text/>
-    </content>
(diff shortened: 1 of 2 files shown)


musicEnfanthen:
Seems to be safe to remove after thorough discussions and checks in #885 and music-encoding/metadata-ig#13.

Fun fact: As far as I can see, the element was already removed in 2012 (d8b523b508d0f12c8fd5be930059e7c4b27b28fe and 075a99729824da990cc4692731bf296e96706adc) and removal confirmed in 2015 (#82).

ahankinson:
Actually, the discussion in #82 was from 2012, and just migrated over when we moved from Google Code. :-D That was on May 28, 2015, which is why it has that timestamp.

musicEnfanthen:
Ah, I see. Thanks for the context @ahankinson !

bwbohl (approved):

musicEnfanthen:
Thank you @rettinghaus.

PR #895, opened 2022-01-27
Update build process to Saxon 10

### LINKED ISSUE #783: TEI Stylesheets and Saxon 10
I noticed that the TEI Stylesheets have switched to Saxon 10 for their build. We should confirm that this will or won't cause problems for our build process.

See: https://github.com/TEIC/Stylesheets/pull/498

rettinghaus:
This PR updates the ant process to use Saxon 10 for building schema and Guidelines. 
closes #783

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -50,7 +50,7 @@ jobs:
         # For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-ant
         run: |
           ant -noinput -buildfile build.xml init
-          ant -noinput -buildfile build.xml -lib lib/saxon/saxon9he.jar -Dgithub.sha=${{ github.sha }}
+          ant -noinput -buildfile build.xml -lib lib/saxon/saxon-he-10.5.jar -Dgithub.sha=${{ github.sha }}
 
       ### PUBLISHING THE SCHEMA ###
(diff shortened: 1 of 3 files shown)


musicEnfanthen:
Thank you, @rettinghaus . Could you please separate the changes to the header structure and indentation in the readme from this PR as it is not connected to the changes to the build process itself? I know it's tempting but we probably should separate different concerns in the PR's as much as possible for better readability and findability of changes.

rettinghaus:
@musicEnfanthen Done. Opened #898 in addition.

musicEnfanthen:
Many thanks, @rettinghaus. Does it build correctly on your machine(s) with the new configuration? 

As @ahankinson pointed out in #783, we should make sure that there aren't any side effects and that it is tested and confirmed on different devices / OS by different users.

rettinghaus:
@musicEnfanthen Tested it successfully on macOS (Apple Silicon) and on Linux (Ubuntu 18.04). Works as expected.

ahankinson:
The GitHub Actions should be updated to match.

https://github.com/music-encoding/music-encoding/blob/develop/.github/workflows/deploy.yml#L53

ahankinson:
I built it on my machine (macOS Intel) and it seems to work fine.

ahankinson (approved):

ahankinson:
Needs one more approving reviewer, who can also do the merge.

PR #897, opened 2022-01-27
Remove font name in favor of font family

### LINKED ISSUE #716: deprecate fontname

rettinghaus:
This removes `@lyric.name`, `@text.name` and `@fontname` with the respective `data.FONTNAME` from the schema. 
closes #716

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -1711,12 +1711,6 @@
           <rng:ref name="data.FONTFAMILY"/>
         </datatype>
       </attDef>
-      <attDef ident="lyric.name" usage="opt">
-        <desc>Sets the font name default value for lyrics.</desc>
-        <datatype>
-          <rng:ref name="data.FONTNAME"/>
-        </datatype>
-      </attDef>
       <attDef ident="lyric.size" usage="opt">
         <desc>Sets the default font size value for lyrics.</desc>
         <datatype>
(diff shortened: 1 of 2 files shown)


lpugin:
I added a comment to the issue https://github.com/music-encoding/music-encoding/issues/716 I think it would be wise to clarify how to refer to SMuFL in general before removing it. See also this open discussion https://github.com/music-encoding/music-encoding/pull/704

bwbohl:
#717 duplicates one change from this PR

bwbohl:
2022-02-24 ODD meeting: We need to discuss this in connection with #704 and come up with a general concept for fonts and fallbacks

bwbohl:
converted to draft as discussion has to take place first in order to come up with a sound concept

musicEnfanthen:
Since #716 was closed as `wontfix` during MEI Developer meeting 2022, can we close this draft PR, too? @rettinghaus ?

rettinghaus:
Absolutely

musicEnfanthen:
Thank you

Issue #878, opened 2021-11-10
Clarify remarks for elements that are modelled after other standards


doerners:
Some elements in MEI are modelled after other standards, e.g. [exhibHist](https://music-encoding.org/guidelines/dev/elements/exhibHist.html), however, the remarks do not seem to specify the exact element(s) that were used as templates for modelling and neither give the exact version of the respective standard used. These information would be helpful in order to provide sufficient descriptions for the guidelines and enable a better understanding of the elements for anyone not involved in the initial modelling process.

ahankinson:
I had a look, and I can't find any reference to `exhibHist` (or indeed `exhibitHist`, as it seemed to be called in earlier forms of MEI). It may be possible that this was available in a local customization of EAD (from the days when EAD was in SGML and you could define your own tags...)

bwbohl:
just spotted this after replying to #879, thanks!

bwbohl:
Would it be a good starting point to put together a list of those elements? To some degree the knowledge about this might, of course, be considered 'forbidden lore' (it's hard to find something, that isn't there ;-). @pe-ro, as you might be the best resource and our only chance for achieving this, may I assign this ticket to you?

kepper:
I already had a start on generating functional links into the guidelines for such elements – we usually have a sentence like "This element is modelled on an element in the Encoded Archival Description (EAD) standard." in the `remarks` (the one above is taken from the aforementioned `exhibHist` element. I don't think we will be able to get correct links for all of those automatically, but we should at least be able to get a list of elements to which this applies…

doerners:
I took the liberty to browse through the MEI modules and compile a list of the affected elements – it's 175 in total btw. 

I listed the respective module, the element name within MEI, the element name within the other standard according to the guidelines, the element name of the element I think might have been the template if the guidelines didn't specify that and a link to the element description in the respective standard (considering the current version of the standard). 

For some elements I couldn't easily find any equivalent in the designated other standard(s). This might be due to another version used during the modelling process. This affects e.g. the author element in EAD, probably because EAD took the route to go more generic and thus introduced the agent element at some point.  It might be good if someone with more profound knowledge of the initial modelling process could have a look at the list and provide some more insight.

Here it is as csv file:
[MEI-elements-from-other-standards.csv](https://github.com/music-encoding/music-encoding/files/7958219/MEI-elements-from-other-standards.csv)

<details>
<summary>Here it is as a markdown table</summary>

[...]

PR #899, opened 2022-01-28
Improve remarks

rettinghaus:
This is a first attempt to unify remarks that mention other standards and add generic references to TEI and EAD websites (see #878).

### DIFF
--- a/source/docs/02-shared.xml
+++ b/source/docs/02-shared.xml
@@ -305,7 +305,7 @@
                         <specDesc key="att.nNumberLike" atts="n"/>
                      </specList>
                   </p>
-                  <p>The <gi scheme="MEI">pb</gi> element can be used to mark page beginnings. When transcribing an existing document the <att>n</att> attribute should be used to record the page number displayed in the source. It need not be an integer, <abbr>e.g.</abbr>, 'iv', or 'p17-3'. The logical page number can be calculated by counting previous <gi scheme="MEI">pb</gi> ancestor elements. When used in a score context, a page beginning implies an accompanying system beginning. This element is modelled on an element in the Text Encoding Initiative (TEI) standard.</p>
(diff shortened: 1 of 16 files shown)


ahankinson:
There are two "ll"s in modelled.

rettinghaus:
Not in American English: https://www.oxfordlearnersdictionaries.com/definition/english/model_2

ahankinson:
*sigh* In all other English-speaking countries, there are two Ls. One L just looks *wrong*. 

Next thing you know we'll be changing everything to Farenheit. :-D

bwbohl:
Some comments:

* I think the question on ”modelled“ vs ”model“ is something that is tied to a general decision on language usage in the guidelines and might be a question for our editors in chief of the guidelines @kepper @pe-ro ?
* unfying the sentences referencing external standards in terms of order of HTML, EAD, TEI etc. seems fine to me also because it's more or less just a standard phrase.
* concerning tag-usage here (`tei:ref`) I'm wondering how this fits with #900 and also here I think some general tagging guidelines for the guidelines should be developed before we implement more such changes

but anyway thanks for bringing this up and starting others to think about those things ;-)

kepper (commented):
First: I'm certainly not opposed to this, even though I'm not approving. However, I am still undecided wether we should provide links to the very elements that stood model, or just the respective standard. I see the argument that we don't have literal copies, but I still believe that in an academic context, it's good practice to give references even when not quoting. Of course this means we need to make it clear that the connection to bespoke ancestor elements might be rather loose. This would require another reconsideration of this sentence…

bwbohl:
Although I personally prefer `modelled` over `modeled` I guess the guidelines tend to be us-american english, but I leave this to our native speakers and @music-encoding/guidelines-editors-in-chief to decide.

One question though concerning the `<ref>` tagging shouldn't we go a step more semantic in the sense of #913 and use `<name>` with `@ref`for tagging organisations/institutions e.g.:

``` xml
[…] is modeled on an element of the <name ref="https://www.tei-c.org">Text Encoding Initiative (TEI)</name>.
```

and later add `<ref>` tags if (not when ;-) we explicitly link to the specific element definitions?

rettinghaus:
Closing this and separating into different pull requests.

PR #900, opened 2022-01-28
Added abbreviation tags to MARC

rettinghaus:
Links may be added after discussion in #878.

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -2315,7 +2315,7 @@
         the recording medium, <abbr>i.e.</abbr>, 2-track monophonic recordings. In this example, the soundChan
         element should record the fact that there is a single output channel, while the <gi
         scheme="MEI">trackConfig</gi> element should capture the existence of two physical tracks.
-        This element is analogous to MARC field 344 subfield g.</p>
+        This element is analogous to <abbr>MARC</abbr> field 344 subfield g.</p>
     </remarks>
   </elementSpec>
   <elementSpec ident="source" module="MEI.header">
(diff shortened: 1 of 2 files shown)


bwbohl:
I see the sense in tagging MARC in order to make it possible to e.g. index it but would there be a semantically stronger option than `abbr`?  What about `term`?

bwbohl:
I thinks it's worthwhile to consider deeper and semantically richer tagging of the guidelines but maybe we should come up with a clear concept what we want to achieve with it rather earlier than later as TEI offers such a rich tags for so many purposes…

doerners:
Since MARC is short for **MA**chine **R**eadable **C**ataloguing I'd argue that technically `<abbr>` isn't wrong nor imprecise here. Considering the examples given for [`<term>`](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-term.html) within the TEI guidelines I'd say it doesn't really apply entirely to the MARC example here.

I like the idea of having a glossary as some of the TEI examples refer to. We might think about building a glossary for MEI, however, my intuition would be to focus more on the musical terms that are not abbreviations but of technical nature.

bwbohl:
ok, let's go with the abbreviation for a term ;-)

bwbohl:
Strategy then should be to markup similar abbreviations and likewise terms if they appear, then use a script to collect them all and then enhance the encoding with some key. an then enable automatic compilation of indices across the guidelines

bwbohl (dismissed):

kepper (approved):
sounds wonderful, thanks!

bwbohl:
just to link it, this goes hand in hand with #913

bwbohl (approved):

PR #902, opened 2022-01-29
Refine MIDI names description and link

rettinghaus:
As the allowed values for MIDI instrument names in MEI all have underscores and no parentheses they are not simply "taken from" but "based on" the official names list. 
Also this updates the link.

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -2746,9 +2746,9 @@
       </valList>
     </content>
     <remarks>
-      <p>Instrument names are taken from the list at <ref
-        target="https://www.midi.org/specifications/item/gm-level-1-sound-set"
-        >https://www.midi.org/specifications/item/gm-level-1-sound-set</ref>. </p>
+      <p>Instrument names are based on the official list in the <ref
+        target="https://midi.org/specifications/midi1-specifications/general-midi-specifications"
+        >General MIDI Specifications</ref>.</p>
       <p>MEI uses 0-based program numbers.</p>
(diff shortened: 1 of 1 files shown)


ahankinson:
It's really too bad that we can't find a way to link to the specific page that contains the sound set listing, instead of a login page for the MIDI specification.

rettinghaus:
I just wanted to complain too. But I found (thanks to Wikipedia) an official page: https://www.midi.org/specifications-old/item/gm-level-1-sound-set
I'll change this.

ahankinson:
I saw that too, but was a bit put off by the "specifications-old" part of the URL. I'm not sure how long it will be maintained...

craigsapp:
They used to charge $50 for a paper copy of the specs...

rettinghaus:
As the URL has changed before … Still I think it is better to refer to an _official_ page than something else.
Nowadays the specs are free, you "just" have to register.

kepper:
What if we do https://web.archive.org/web/20211204235352/https://www.midi.org/specifications-old/item/gm-level-1-sound-set instead?

rettinghaus:
Relying on the [Web Archive](https://web.archive.org/) isn't much better in my opinion.

ahankinson:
It is what Wikipedia does to prevent stale links...

kepper (approved):
the requested change for links to archive.org needs to be implemented for all links in the Guidelines (see #918), so this can go in like all the other links first.

bwbohl:
2022-02-24 ODD meeting: many thanks to everyone and bringing up the great idea in #918

PR #909, opened 2022-02-03
Fix SMuFL glyphs in guidelines

### LINKED ISSUE #908: Images from SMuFL browser don't work
Images in the Guidelines from the SMuFL-browser don't work, because they're pre-fixed with `../` making it a wrong relative URL.
See: https://music-encoding.org/guidelines/dev/content/cmn.html#cmnFermata

Should we switch to a SMuFL text font in general?

rettinghaus:
This PR fixes the wrong prefixing of URLs and brings a special treatment for SMuFL glyphs. 
Also this fixes the HTML lang attribute which was giving a validation error.
closes #908

### DIFF
--- a/utils/guidelines_xslt/generateGuidelines.xsl
+++ b/utils/guidelines_xslt/generateGuidelines.xsl
@@ -2052,7 +2052,7 @@
     <xsl:template name="getSinglePage">
         <xsl:param name="contents" as="node()*"/>
         
-        <html xml:lang="en">
+        <html lang="en">
             <head>
                 <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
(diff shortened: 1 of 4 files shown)


kepper (commented):
This doesn't address the huge fontsize of SMuFL glyphs yet, right? I would suggest to have another, more specific `when` that checks for a link to the SMuFL browser (so it needs to go in first), which adds another `@class` to such images. We can then have a CSS rule for that (probably something like `max-height: 1em;`). Obviously, this will need to look at the existing `@class` as well…

rettinghaus:
@kepper Please try it, the `class="smufl"` is specially for this purpose, and I haven't seen another rend on graphic in the guidelines.

bwbohl (approved):

PR #911, opened 2022-02-15
Remodel exhibHist et. al. after provenance

### LINKED ISSUE #879: Clarify <exhibHist> vs <eventList>
The element description of [`<exhibHist>`](https://music-encoding.org/guidelines/dev/elements/exhibHist.html) states that it encodes:

> A record of public exhibitions, including dates, venues, etc.

The element is only allowed within `<physDesc>`, although I would not consider an exhibition of an item an inherent physical property of a resource. Additionally, the element does not allow to record several exhibition events via structured descriptions, however, it is allowed to repeat the element itself.

My questions are: 

- What is the actual purpose of the element?  – Maybe I misunderstand why it is modelled in this specific way. This might be, because it is not clear on which element from the EAD standard it has been modelled (see [issue 878](#878)).
- Could the information be encoded better with the `<event>` element and do we need `<exhibHist>` at all, i.e. should we drop it?

doerners:
This PR remodels `<exhibHist>`, `<treatHist>` and `<treatSched>`after `<provenance>` as proposed by the Metadata IG in #879. It also adds `<history>` to `<physDesc>`. Closes #879

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -1119,11 +1119,29 @@
     <classes>
       <memberOf key="att.common"/>
       <memberOf key="att.bibl"/>
+      <memberOf key="att.datable"/>
       <memberOf key="att.lang"/>
-      <memberOf key="model.physDescPart"/>
     </classes>
     <content>
-      <rng:ref name="macro.struc-unstrucContent"/>
+      <rng:choice>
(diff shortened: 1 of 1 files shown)


kepper (approved):
happy to follow this compromise – it should come to an end ;)

Issue #913, opened 2022-02-16
Better tagging in Guidelines


rettinghaus:
As discussed in #900 and #910 we should use more TEI elements to achieve better, semantically richer Guidelines. 

**Proposal**: 

- use `abbr` for all used abbreviations
- use `term` for all musical/technical terms (like "glissando") 
- add `@rend` to the former to add quotation marks automatically instead of hardcoding them
- use `name` for names of composers and works with `@ref` to point to additional information
- use `val` for all named attribute values in textual content

(This list could be extended.)

musicEnfanthen:
Pinging @kepper since these questions (and their intended use-case) should be coordinated with the Editorial Board of the Guidelines.

kepper:
In general, I think this is very helpful. It may take some time to implement this throughout the Guidelines, but I'm happy to accept the challenge (fully aware that I haven't found much time in the last couple of months…). One question, though: Shouldn't we just rely on CSS (`:before` and `:after`) instead of an additional `@rend`? Seems to be presentational only, and could help to keep the Guidelines tidy…

bwbohl:
+1 for the CSS solution instead of `@rend`

lpugin:
Since #938 has been merge I am closing it. This would require a significant work on the guidelines and the need of it is not very clear.

PR #914, opened 2022-02-16
utils: extend templates for tei elements in guidelines XSLT

rettinghaus:
in preparation for #913 this extends the template for `tei:term` and add one for `tei:name`.

### DIFF
--- a/utils/guidelines_xslt/odd2html/guidelines.xsl
+++ b/utils/guidelines_xslt/odd2html/guidelines.xsl
@@ -400,11 +400,56 @@
     
     <xd:doc>
         <xd:desc>
-            <xd:p>term</xd:p>
+            <xd:p>Names</xd:p>
+        </xd:desc>
+    </xd:doc>
+    <xsl:template match="tei:name" mode="guidelines">
+        <xsl:choose>
+            <xsl:when test="not(@ref)">
+                <xsl:apply-templates select="node()" mode="#current" />
+            </xsl:when>
+            <xsl:otherwise>
+                <a class="link_ref link_external" href="{@ref}">
+                    <xsl:apply-templates select="node()" mode="#current" />
+                </a>
+            </xsl:otherwise>
+        </xsl:choose>
+    </xsl:template>
+    
+    <xd:doc>
+        <xd:desc>
+            <xd:p>Terms</xd:p>
         </xd:desc>
     </xd:doc>
     <xsl:template match="tei:term" mode="guidelines">
-        <span class="term"><xsl:apply-templates select="node()" mode="#current"/></span>
+        <span class="term">
+            <xsl:choose>
+                <xsl:when test="not(@ref)">
+                    <xsl:apply-templates select="node()" mode="#current" />
+                </xsl:when>
+                <xsl:when test="starts-with(@ref,'#')">
+                    <xsl:variable name="chapter.id" select="substring-after(@ref,'#')" as="xs:string" />
+                    <xsl:variable name="tocInfo" select="$all.chapters/descendant-or-self::chapter[@xml:id = $chapter.id]" as="node()?" />
+                    <xsl:choose>
+                        <xsl:when test="exists($tocInfo)">
+                            <a class="link_ref chapterLink" title="{$tocInfo/@number || ' ' || $tocInfo/@head}" href="#{$chapter.id}">
+                                <xsl:apply-templates select="node()" mode="#current" />
+                            </a>
+                        </xsl:when>
+                        <xsl:otherwise>
+                            <span class="wrong_ref" data-target="{$chapter.id}">
+                                <xsl:apply-templates select="node()" mode="#current" />
+                            </span>
+                        </xsl:otherwise>
+                    </xsl:choose>
+                </xsl:when>
+                <xsl:otherwise>
+                    <a class="link_ref link_external" href="{@ref}">
+                        <xsl:apply-templates select="node()" mode="#current" />
+                    </a>
+                </xsl:otherwise>
+            </xsl:choose>
+        </span>
     </xsl:template>
     
     <xd:doc>
--- a/utils/guidelines_xslt/odd2html/guidelines.xsl
+++ b/utils/guidelines_xslt/odd2html/guidelines.xsl
@@ -265,7 +265,7 @@
     <xsl:template match="tei:ref" mode="guidelines">
         <xsl:choose>
             <xsl:when test="starts-with(@target,'#')">
-                <xsl:variable name="chapter.id" select="substring-after(@target,'#')" as="xs:string"/>
+                <xsl:variable name="chapter.id" select="substring(@target,2)" as="xs:string"/>
                 <xsl:variable name="tocInfo" select="$all.chapters/descendant-or-self::chapter[@xml:id = $chapter.id]" as="node()?"/>
                 <xsl:choose>
                     <xsl:when test="exists($tocInfo)">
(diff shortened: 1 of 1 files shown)


utils/guidelines_xslt/odd2html/guidelines.xsl
```diff
@@ -265,7 +265,7 @@
     <xsl:template match="tei:ref" mode="guidelines">
         <xsl:choose>
             <xsl:when test="starts-with(@target,'#')">
-                <xsl:variable name="chapter.id" select="replace(@target,'#','')" as="xs:string"/>
+                <xsl:variable name="chapter.id" select="substring-after(@target,'#')" as="xs:string"/>
```
kepper: In general, the `replace` is more expensive, but also more robust (it still works if there is no hash). However, as the line above ensures that there is a hash to start with, it could be made even faster by just doing a `substring(@target,2)`, making it completely unnecessary to query the content…

kepper (dismissed):
I haven't tested the changes to guidelines.xsl, but have full trust in @rettinghaus ;-) many thanks!

utils/guidelines_xslt/odd2html/guidelines.xsl
rettinghaus: You're right, thanks for this!

bwbohl (dismissed):

bwbohl:
@kepper could you please make your review an approving one?

utils/guidelines_xslt/odd2html/guidelines.xsl
```diff
@@ -400,11 +400,56 @@
...
+                        <xsl:when test="exists($tocInfo)">
+                            <a class="link_ref chapterLink" title="{$tocInfo/@number || ' ' || $tocInfo/@head}" href="#{$chapter.id}">
+                                <xsl:apply-templates select="node()" mode="#current" />
+                            </a>
+                        </xsl:when>
+                        <xsl:otherwise>
+                            <span class="wrong_ref" data-target="{$chapter.id}">
+                                <xsl:apply-templates select="node()" mode="#current" />
+                            </span>
+                        </xsl:otherwise>
```
kepper, line 443: I have a question on this

kepper: it basically says that every internal link (starting with a hash, according to line +430) needs to point to a chapter, or it is labelled a `wrong_ref`. I haven't looked up if there are any such cases, but it feels a bit too opinionated. Probably, when pointing to elements _as musical concepts_ (like _tempo_), we would still encourage to use `<gi>` instead, but I wonder if we should already label every other link target as wrong. I would probably prefer a more neutral `unspecified_ref` or some such…

kepper (commented):
I'm not opposed to this, I just wonder if the CSS class of `wrong_ref` is a bit too strict (see my comments). Happy to follow if others see no problem in this, though.

utils/guidelines_xslt/odd2html/guidelines.xsl
rettinghaus: @kepper Actually this `wrong_ref ` wasn't my idea, but was simply copied over. I see your point, but this is not really something that is connected to the PR and should be addressed in another PR.

kepper: Thanks for the clarification. Then let's go ahead!

kepper (dismissed):
thanks to @rettinghaus' clarifications, this can move on. Further clarifications can come in later…

bwbohl:
This is connected to #913

musicEnfanthen:
When finished with the new build workflow (see #959), we will have to check this PR for any possible merge conflicts.

utils/guidelines_xslt/odd2html/htmlFile.xsl
```diff
@@ -248,7 +248,7 @@
                             </div>
                         </div>
                         
-                        <script type="text/javascript">
+                        <script>
```
kepper: This is a separate issue – please remove and commit independent from the other change

kepper (dismissed):
please separate the two changes of this into two PRs. The javascript one seems rather minor, though, so that probably isn't necessary or might even cause problems on outdated browsers

utils/guidelines_xslt/odd2html/guidelines.xsl
```diff
@@ -400,11 +400,56 @@
...
+            <xsl:when test="not(@ref)">
+                <xsl:apply-templates select="node()" mode="#current" />
+            </xsl:when>
+            <xsl:otherwise>
+                <a class="link_ref link_external" href="{@ref}">
+                    <xsl:apply-templates select="node()" mode="#current" />
+                </a>
+            </xsl:otherwise>
+        </xsl:choose>
+    </xsl:template>
```
kepper, line 417: This will affect the names of editors of the docs only. That content is contained in the headers of our guideline chapter files, which is not directly rendered to the web version of the Guidelines, and also not included in the PDF version. Instead, a special page is generated, which also deduplicates that content. In essence, this code has currently no effect.

utils/guidelines_xslt/odd2html/guidelines.xsl
```diff
@@ -400,11 +400,56 @@
...
+                                <xsl:apply-templates select="node()" mode="#current" />
+                            </span>
+                        </xsl:otherwise>
+                    </xsl:choose>
+                </xsl:when>
+                <xsl:otherwise>
+                    <a class="link_ref link_external" href="{@ref}">
+                        <xsl:apply-templates select="node()" mode="#current" />
+                    </a>
+                </xsl:otherwise>
```
kepper, line 450: There are no `<term>` elements with a `@ref` attribute in all our sources, so these lines have very little consequences _at the moment_ either. We could of course add that for future use, but it will not do anything now.

musicEnfanthen:
Given the comments by @kepper above, do you agree to push this down the road to another release than v5 @rettinghaus?

rettinghaus:
@musicEnfanthen sure. I think I was a preparation for something I planned and never came around to actually do.

musicEnfanthen:
Ok, thank you. Moved to v6

musicEnfanthen:
@rettinghaus Can you please double-check if this is still needed? The suggested changes seem to have no effect on anything in the repo, since the addressed tei elements are not used here?

utils/guidelines_xslt/odd2html/guidelines.xsl
```diff
@@ -265,7 +265,7 @@
     <xsl:template match="tei:ref" mode="guidelines">
         <xsl:choose>
             <xsl:when test="starts-with(@target,'#')">
-                <xsl:variable name="chapter.id" select="replace(@target,'#','')" as="xs:string"/>
+                <xsl:variable name="chapter.id" select="substring(@target,2)" as="xs:string"/>
```
bwbohl, line 268: this is much safer than replacing any `#` alternatively a substring-after could be used

bwbohl:
There are 13 occurrences of `<term` and 450 of `<name` in the source folder

utils/guidelines_xslt/odd2html/guidelines.xsl
rettinghaus: I think with the test in the preceding line this statement is clearer than `substring-after`.

bwbohl (dismissed):

rettinghaus:
@musicEnfanthen I was definitely planning to tag at least all persons in the guidelines.

bwbohl (approved):

musicEnfanthen:
Tested locally. Works both in HTML and PDF version of Guidelines

musicEnfanthen (approved):

PR #925, opened 2022-03-11
Extend value list for att.curvature@curvedir

### LINKED ISSUE #912: Extend curve direction for s-shaped slurs
The available values of `@curvature.curvedir` are insufficient when it comes to encoding s-shaped slurs. It is impossible to distinguish between the following shapes:
MixedShapes
For s-shaped cross staff slurs the shape can be deduced following Gould p. 321:
Cross-staff
However, for s-shaped slurs within a single staff the choice of shape must be encoded.

One suggestion would be to replace `curvature_CURVEDIR_mixed` by `curvature_CURVEDIR_mixed_above_below` and `curvature_CURVEDIR_mixed_below_above`. I am not happy with these long names, so feel free to propose better names.

lpugin:
* Add values for mixed slurs starting specifically above or below
* As discussed in #912


rettinghaus:
Thanks. The only other place we use a dot as separator in attribute values is in `data.NOTATIONTYPE` specifying a subtype. All other cases use a dash. Maybe use `above-below` and `below-above`?

musicEnfanthen:
Thank you, @lpugin . Looks good to me, too. 

I am wondering if we could add something to the documentation about curve directions? Maybe in [2.4.3.4. Curve Shape](https://music-encoding.org/guidelines/v4/content/shared.html#usersymbolsCurveShape)? Maybe at least name the possible curve directions at the end of the following sentence there?

```However, if @curvedir is present, the curve must respect the curvature direction specified there (`above`, `below`; for s-shaped slurs: `mixed`, `above-below`, `below-above`).```

File: source/modules/MEI.shared.xml [hunk 1]
```diff
@@ -757,7 +757,13 @@
             <desc>Downward curve.</desc>
           </valItem>
           <valItem ident="mixed">
-            <desc>A "meandering" curve, both above and below the items it pertains to.</desc>
+            <desc>A "meandering" curve, both above and below the items it pertains to, starting un-determinately above or below.</desc>
```
musicEnfanthen: I am wondering if "un-determinately" could give the false impression that it could not be determined if a curve starts above or below. Would "starting either above or below without specific indication." work?

Maybe a native speaker can clarify?

musicEnfanthen (dismissed):

musicEnfanthen:
Thank you, @lpugin for the addition. Looks good to me, one small final question from my side about the wording for the "mixed description, see above.

source/modules/MEI.shared.xml
lpugin: It works for me

pe-ro:
As a nominal native speaker, I think

<desc>A "meandering" curve, both above and below the items it pertains to.</desc>

is sufficient.  😊

musicEnfanthen (approved):
Thank you!

musicEnfanthen:
Pinging @bwbohl for second review.

kepper:
We discussed this properly at the 2022 dev workshop in Berlin, together with @th-we, @lpugin, @pe-ro, @rettinghaus, @musicEnfanthen and others. We came up with an updated idea which seems to go further than this (sneak peak ahead). It will be introduced in a separate PR later on…

IMG_6207

musicEnfanthen:
Discussion at MEI Developer Workshop Berlin lead to an alternative approach for modelling s-slurs. Will be introduced in another PR. Closing this PR.

rettinghaus:
Where is the other PR? :)

PR #927, opened 2022-03-14
Fix data.MIDINAMES

### LINKED ISSUE #812: MIDINAMES error in MEI 4/5 RelaxNGs
Most values from the [data.MIDINAMES](https://music-encoding.org/guidelines/v4/data-types/data.midinames.html) range have lost their underscores in the resulting [rng](https://music-encoding.org/schema/4.0.0/mei-all.rng). Looks like the `altIdent` values slipped into there.
Same applies for [development version](https://music-encoding.org/schema/dev/mei-all.rng).

rettinghaus:
This resolves the problematic `altIdent` in `data.MIDINAMES` and fixes a wrong value in this range, as *Bag pipe* should become "Bag_pipe" instead of "Bagpipe".

This is also a bugfix for MEI 4.
closes #812

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -2092,656 +2092,529 @@
     <content>
       <valList type="closed">
         <valItem ident="Acoustic_Grand_Piano">
-          <altIdent>Acoustic Grand Piano</altIdent>
-          <desc>Program #0.</desc>
+          <desc>Acoustic Grand Piano, MIDI Program Change #1.</desc>
         </valItem>
         <valItem ident="Bright_Acoustic_Piano">
-          <altIdent>Bright Acoustic Piano</altIdent>
-          <desc>Program #1.</desc>
+          <desc>Bright Acoustic Piano, MIDI Program Change #2.</desc>
         </valItem>
         <valItem ident="Electric_Grand_Piano">
-          <altIdent>Electric Grand Piano</altIdent>
-          <desc>Program #2.</desc>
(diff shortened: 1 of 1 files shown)


ahankinson:
Is there a reason it's now starting at 1, not 0?

MIDI program changes are 0-indexed, so this should probably be reflected in MEI.

See: https://www.recordingblogs.com/wiki/midi-program-change-message

rettinghaus:
This reflects the *MIDI Program Change* given in the [original specifications](https://www.midi.org/specifications-old/item/gm-level-1-sound-set). Otherwise we need to change the given *Key*s for percussions accordingly. 

The remarks state:
> MEI uses 0-based program numbers.

If that leads to confusion we should drop this, because the description in `data.MIDIVALUE` should be clear in this regard.

ahankinson:
About this particular change: The original says `Program #0`, while the proposed change says `MIDI Program Change #1`. I think both are correct -- it's the first program change in the list, but "Acoustic Grand Piano" **is** program 0. (message value `0x00`).

Given that we're specifically trying to fix an issue with it rendering to RNG correctly, and not actually trying to change the MIDI semantics, I would suggest keeping the original wording and not introducing this change here.

ahankinson (approved):

musicEnfanthen:
@rettinghaus thanks for the PR, and thanks for the approval @ahankinson. Will take this PR to the next ODD meeting.

musicEnfanthen (approved):
Thanks a lot. Looks good.

bwbohl:
thanks!

PR #932, opened 2022-03-15
BREAKING(modules): Change macro name of data.OTHERSTAFF

rettinghaus:
`data.OTHERSTAFF` is used for `@beam.with` and `@stem.with`. The latter one is used to combine voices in part writing, i.e., merge stems from different layers into one. So the description of this macro ("on another staff") is not correct. 
But even for `@beam.with` it is not very precise, because this could happen within one staff, or point to a staff that has multiple layers. 

This PR proposes to change the name from `data.OTHERSTAFF` to `data.NEIGHBORINGLAYER`, because it actually points to the next layer above or below the current one (even if this is "on another staff"). Descriptions are updated accordingly. 

This will not(!) change the values or anything else.


bwbohl:
Thanks for bringing this up @rettinghaus! I think being precise in the naming of things is always a good idea. But I somehow see a much bigger problem in the imprecision of the values `above` and `below`, in both cases staff and layer): what are they referring to, graphical position? logical position in the XML? What if there are multiple layers on the referred staff? What if the notes to be stemmed with are on the the above staff but not on the lowest (graphical? or logical?) layer – after all I think there is no definite rule on how to stack layers in the XML according to their graphical position, or is there?

image

bwbohl:
So the question concerning your proposed changes would be: what are ”adjacent layers”?

source/modules/MEI.xml
```diff
@@ -3112,16 +3112,16 @@
       </rng:data>
     </content>
   </macroSpec>
-  <macroSpec ident="data.OTHERSTAFF" module="MEI" type="dt">
-    <desc>For musical material designated to appear on another staff, the location of the staff
-      relative to the current one; <abbr>i.e.</abbr>, the staff above or the staff below.</desc>
+  <macroSpec ident="data.NEIGHBORINGLAYER" module="MEI" type="dt">
+    <desc>For musical material designated to appear on an adjacent layer or staff, the location of the layer
```
ahankinson: Perhaps "an adjacent layer on a different staff"?

ahankinson:
Could you provide an example of where you might use "beam with" or "stem with" on neighbouring layers on the same staff?

source/modules/MEI.xml
kepper: this would always require to cross the staff, which isn't always necessary. Therefore: no ;-)

kepper (dismissed):
After discussion at 2022 Berlin dev meeting, @musicEnfanthen, @rettinghaus, @fujinaga, @pe-ro and @kepper agree to accept this.

kepper:
The current proposal doesn't easily support connecting layers 1 and 3 plus 2 and 4 of four instruments spread across two staves. However, we didn't do that in the past, so we can skip it once more ;-)

kepper:
IMG_6208

kepper (approved):
still good, now with languages

musicEnfanthen (approved):

musicEnfanthen:
Thanks everyone! Merged at MEI Developer meeting 2022

Issue #489, opened 2017-12-03
Provide better description of manuscript material


pe-ro:
The attached ODD file describes a new MEI.msdesc module, as well as a few changes to existing MEI features, to better accommodate the capture of more physical details of manuscript material.  

The most significant change is in the handling of components of a resource; i.e., in the elimination of `<itemList>` and changes to the content model of `<componentGrp>` to allow it to replace `<itemList>` and handle the special relationships between manuscript components.  For example, manuscripts sometimes contain fragments of other, unrelated material.  Conversely, part of the manuscript being described may be located elsewhere.

[msDesc_PROPOSAL.odd.txt](https://github.com/music-encoding/music-encoding/files/1524945/msDesc_PROPOSAL.odd.txt)

pe-ro:
Here's the latest version of the proposal to enhance the description of manuscripts.

[msDesc_PROPOSAL.odd.txt](https://github.com/music-encoding/music-encoding/files/1569381/msDesc_PROPOSAL.odd.txt)

Highlights:

1. Backward-compatible changes:

- New administrative elements
- New descriptive elements that may appear within physDesc and/or physMedium
- New elements that may appear inline when MEI.msdesc is activated
- Make dimensions and extent available inline with or without msDesc activation
- Make dimensions member of att.evidence
- Make fingerprint and watermark available inline when the msDesc module is activated
- Restrict use of some inline elements to the metadata header
- Modify physMedium to accommodate more detailed description (for manuscripts and other materials)
- Add an attribute class for subordinate component type that provides a "short cut" for expressing the relationship between the component and its host item
- Add an attribute class for record type
- Make source a member of att.componentType and att.recordType
- Allow creation within source
- Revise description and remarks for contents element
- Make creation, provenance, relation, and resp members of att.datable
- Extend content model of provenance
- Revise constraint on relation within a bibliographic description

2. Backward-incompatible changes:

- Move componentGrp to MEI.header module, rename it to componentList, and modify its explanatory remarks
- Modify physLoc to group historical info within <history>
- Make provenance a child of history; modify its description and comments regarding history to make clear the distinction between it and the creation element

This proposal borrows heavily from TEI, but incorporates the borrowings into the existing MEI framework.

Comments and questions eagerly sought.  😀

axgeertinger:
Hi Perry
[...]

PR #935, opened 2022-03-23
Removal of explicit

rettinghaus:
The Metadata IG agreed to deprecate `explicit`, since this element regards mainly textual content and thus use cases for it are very uncommon/rare in the musicological context. Furthermore `explicit` isn't modeled the way `incipit` is, which works very well in MEI as it has been thought through thoroughly. If someone really needs/wants to encode the last measures etc. of a piece for identification purposes in the header the Metadata IG sees this as a case for a customization. 

Please refer to music-encoding/metadata-ig/issues/14 for more context.

### DIFF
--- a/source/modules/MEI.msDesc.xml
+++ b/source/modules/MEI.msDesc.xml
@@ -261,23 +261,6 @@
       <p>This element is modelled on an element in the Text Encoding Initiative (TEI) standard.</p>
     </remarks>
   </elementSpec>
-  <elementSpec ident="explicit" module="MEI.msDesc">
-    <desc>Contains the explicit of a manuscript item; that is, the closing words of the text proper,
-      exclusive of any rubric or colophon which might follow it.</desc>
-    <classes>
-      <memberOf key="att.common"/>
-      <memberOf key="att.bibl"/>
-      <memberOf key="att.facsimile"/>
-      <memberOf key="att.lang"/>
-      <memberOf key="model.physDescPart"/>
-    </classes>
-    <content>
-      <rng:ref name="macro.struc-unstrucContent"/>
-    </content>
-    <remarks>
-      <p>This element is modelled on an element in the Text Encoding Initiative (TEI) standard.</p>
-    </remarks>
-  </elementSpec>
   <elementSpec ident="foliation" module="MEI.msDesc">
     <desc>Describes the numbering system or systems used to count the leaves or pages in a
       codex.</desc>


ahankinson:
I'm -1 on removing `explicit`.

1) Having it in doesn't hurt anything, and gives people more options for doing catalogues and encodings, **particularly** where they aren't served by any other format. (i.e., MEI is unique in the notation formats for being able to encode explicits.)
2) We could fix `explicit` to function more like `incipit`; removing it isn't the only option to deal with this.

In some communities, encoding the final pitch/chord or cadential formulas are very useful pieces of information, and this is what an explicit could capture.

doerners:
The Metadata IG did have a longer discussion about that so this proposal or the resulting PR were not made lightly. 

> Having it in doesn't hurt anything, and gives people more options for doing catalogues and encodings, particularly where they aren't served by any other format. (i.e., MEI is unique in the notation formats for being able to encode explicits.)

As @rettinghaus pointed out and explained in his [comment](https://github.com/music-encoding/metadata-ig/issues/14#issuecomment-1065212330) the main problem with `explicit` in MEI is that it in fact doesn't offer the ability to encode *musical* explicits. So if anyone would want to encode the final pitch/chord or cadential formulas they couldn't do it. At least not with `explicit` as-is. On the other hand, if someone would want to encode only text/textual content why not use TEI for that, as this is in fact a text-centered format? Or leave that for a customization if that seems more appropriate?

> We could fix explicit to function more like incipit; removing it isn't the only option to deal with this.

The IG thought about and discussed that the functionalities `incipit` offers should not only be limited to the beginning of a work/piece, because there are use cases where one might want to encode excerpts from a work/piece other than the beginning, like explicits or maybe even a musical theme that appears neither at the beginning nor the end but maybe in some middle part. This is why deprecating `explicit` is from our point of view only the first step. The idea is to remodel `incipit` into `excerpt` (because we see that both `incipit` and `explicit` can be classified as types of excerpts) and to indicate via a `@form` attribute where the excerpt stemmed from. We documented this (although I admit very briefly) in our [issue](https://github.com/music-encoding/metadata-ig/issues/14) as well, however, implementing that is out of the scope for this PR.

ahankinson:
Thanks @doerners -- if I gave the impression that I thought the IG made this suggestion flippantly, I apologize. I was only doing my part as a member of the larger MEI community to review the suggestions made by the IG in light of its impacts specification. I did take your suggestion seriously, but do see some significant problems with the proposal. I hope the length of this response will further indicate that I consider it to be made in earnest.

That `explicit` does not follow the same model as `incipit` is very useful to point out. But the model of `explicit` is not fixed -- we can change it to be identical to `incipit` and so to encode musical content. Since we can fix it to encode musical content gives less of a reason to remove it entirely because it doesn't.

I'm less thrilled about changing it to excerpt. An incipit and explicit are a form of identification to differentiate them from other compositions with the same title, or indeed with no title. Incipits and explicits are a "thing" -- RISM has over 1.8 million incipits, and they are specifically **not** excerpts. (The difference being that you might expect an excerpt to have all voices present, while an incipit typically only has the top voice, or the first voice that sounds).

MEI is also the only format that I am aware of that lets you specify `incipit` and `explicit`, so in that sense there is a unique feature available here that is not available elsewhere. Whether or not anyone uses it *now* is not necessarily a reason to remove it. Having it in the spec, and indeed fixing it to make it more like `incipit`, does not harm anything and may, indeed, be useful to someone later.

doerners:
@ahankinson apologies if my choice of words came across harshly, I merely wanted to point out that indeed the description of this PR doesn't include/reflect the entirety of our discussion and reasoning but that we did i.a. talk about the aspects you mentioned, however, our conclusion was what is expressed by this PR.

To me – and from what I grasped from the IG's discussion I think it's fair to say that this is not only my single opinion – it seems like an awful lot of modelling effort to be made when following your suggestion. That is because:

[`<incip>`](https://music-encoding.org/guidelines/dev/elements/incip.html) does include `<incipText>` and `<incipCode>` as child elements and is member of the [model.incipLike](https://music-encoding.org/guidelines/dev/model-classes/model.incipLike.html) model class (in fact it is the only element in this model class). When we want to remodel `<explicit>` like `<incip>` it wouldn't work to simply move it to [model.incipLike](https://music-encoding.org/guidelines/dev/model-classes/model.incipLike.html)  as @rettinghaus [pointed out](https://github.com/music-encoding/metadata-ig/issues/14#issuecomment-1065212330). So we could introduce a new model class `model.explicitLike` and two new elements `<explicitText>` and `<explicitCode>` to achieve the same functionality as `<incip>`. If we, however, want to offer the same functionality (additionally) for any excerpt that is not necessarily on incipit or an explicit we could introduce the new elements `<excerpt>`, `<excerptText>` and `<excerptCode>` as well as the model class `model.excerptLike`. 

To me this seems quite redundant and I'm not convinced that it is best practice or efficient to have three different model classes that each contain only one element but are somewhat semantically connected. 

If we follow the idea of the IG I [mentioned](https://github.com/music-encoding/music-encoding/pull/935#issuecomment-1076335205) and remodel `<incip>` to `<excerpt>` with an additional `@form` attribute, I imagine that one could choose to designate if the encoding refers to an incipit or an explicit etc. with only one element, similar to:

```XML
<!-- incipit encoded here -->
<excerpt form="incipit">
<score>
...
</score>
</excerpt>
<!-- explicit encoded here -->
<excerpt form="explicit">
<score>
...
</score>
</excerpt>
<excerpt form="something other than incipit or explicit">
<score>
...
</score>
</excerpt>
```

Your point that 
> (The difference being that you might expect an excerpt to have all voices present, while an incipit typically only has the top voice, or the first voice that sounds).

to me would than seem to be a matter of thorough documentation as to what the users have to expect from the element or what its intention is, respectively.

> MEI is also the only format that I am aware of that lets you specify `incipit` and `explicit`

Actually TEI offers both [`incipit`](https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-incipit.html) and [`explicit`](https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-explicit.html), however, there they are both member of the [model.msQuoteLike](https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-model.msQuoteLike.html) model class, i.e. the modelling approach is different to MEI. I'm not really sure if it would be the best approach to copy that approach.

ahankinson:
Actually, following TEI and making a `quoteLike` model would be a great example to follow for both incipit and explicits! Thank you for pointing that out. We could also follow this example and remodel [`rubric`](https://music-encoding.org/guidelines/dev/elements/rubric.html) to follow it, since they are closely related.

Adding a  `quoteLike` model would also solve the problem of having a model class that is too narrowly defined for incipits.

Not all incipits are notation, and not all notation is encoded in MEI. `incipCode` is actually quite useful for projects like RISM, where we have millions of incipits encoded in Plaine and Easie. Using a different notation format other than MEI is useful there. (not to mention it supports parsons code and `**kern`). 

And `incipText` can be useful where you want to record the text of a piece without necessarily encoding the music.

rettinghaus:
Sorry, I don't understand this now: do you mean having `incipCode` and `incipText` is good, but we don't need these possibilities for `explicit`?

ahankinson:
Nope -- I'm saying having something similar to those would also be useful for `<explicit>`.

doerners:
And what about `excerpt`?

ahankinson:
According to your proposal above, we would lose the ability to encode incipits in non-MEI formats, and to capture the text incipits (not music) of scores. Unless I'm mistaken?

rettinghaus:
I think @doerners suggested out `excerptText` and `excerptCode`.

doerners:
Yes, `<excerptText>` and `<excerptCode>` are meant to be included...the dummy encoding was just supposed to illustrate how the use of the `@form` attribute could look like....at least in my imagination

ahankinson:
And what benefits would that bring us over having dedicated incipit and explicit encoding methods? (Noting, particularly, that you can constrain where an element appears using the schema, but you can't constrain where an element with a given attribute value appears, other than with Schematron. So in this model we would have to allow `<excerpt form="incipit">` anywhere we allow `<excerpt>`.)

But listen -- we could go back and forth for days. I am still very much -1 to this *specific* change, but would like to leave the floor open for others to chime in.

(there are a lot more things I could probably say about `excerpt` but I'm trying to contain my comments to the change in this PR, which is specifically deleting explicit -- there are no changes here that would replace it with anything else, so on the face of it this is a loss of functionality and a breaking change to the schema.)

doerners:
The benefit, in my opinion, would be that we could encode the different aspects with three instead of nine elements. Furthermore I think it shouldn't be a problem to constrain the use with Schematron as MEI already uses Schematron in other cases, so why not here?
`<incip>` is allowed in `<expression>`, `<physDesc>` and `<work>`, in which case would it be problematic to allow `<excerpt>` in these exact places? From the top of my head I can't name any.

musicEnfanthen:
@ahankinson If I read you correctly, the removal of the explicit element without a proper encoding alternative is something to be concerned about. Would it help if the Metadata IG (that has gone through a long decision finding process about this element) would prepare another PR to implement such an alternative? (Until then, this PR would remain in draft mode?)

ahankinson:
@musicEnfanthen yes, my specific objection at the moment is that this PR removes functionality without replacing it with something equivalent.

However, I believe there are larger issues with the direction of this change.

The reason `<incip>` and `<explicit>` do not have the same model is because they come from two different lineages. `<incip>` has, as far as I can tell, been in MEI since the beginning, primarily due to the significance of the incipit projects like RISM (and the work at CCARH with these incipits, which had a direct influence on the development of MEI). `<explicit>` was introduced in 2018 as part of an effort to improve manuscript descriptions in MEI, in response to community requests and to possibly align better with the (relatively new) msDesc functions of TEI. (See Issue #489 and commit cd39884a73284f97e860b52c39bca793a0307407). That it hasn't been widely adopted may be an artifact of its rather late introduction to the spec.

That `<explicit>` does not currently allow encoding of musical material is, I believe, an oversight that should be rectified. It should be aligned with `<incip>`, since they serve similar functions and, as we've seen, would be following the example set by the TEI. The `<incip>` element has been fairly widely used. The Carl Nielsen Catalogue of Works uses this in almost every entry; there are some examples encoded in the Delius catalogue as well. These are just two that I know of off the top of my head. This would suggest that this is a good model to follow.

The alternative solution (`excerpt`) has problems, since it suggests removing something that works and is widely adopted (`<incip>`) because of something that doesn't work (`<explicit>`) and replacing it with a third, new thing (`<excerpt>`), simply because implementing this would somehow be more "optimal". As I suggested before, if it's a question of the internal implementation not having the right names, we can adopt the TEI strategy and consolidate the modelling to something more generic and equivalent to their `quoteLike` model to align the internal implementation with the public-facing schema.

(Edit: Also, a claimed benefit is that "we could encode the different aspects with three instead of nine elements." While this is true, we'll still have nine *variants* of those elements (due to the available values of `@form`). When to use elements and attributes is something more like an art, than a science, but regardless of which is used, we retain the same overall necessary descriptive complexity to cover the semantics of the desired encoding.)

I am also a bit taken aback by the approach the IG has taken with this change. The initial indications (https://github.com/music-encoding/metadata-ig/issues/14) were that they were to discuss it for the purposes of improving the documentation for it. After a single meeting, they then announce a decision to completely remove it and have filed this PR. This was despite earlier indications that "... we didn't have an idea how to transfer this to music. But maybe there is a specialist for medieval music who can [help] us." On that ticket I provided several links to help them with the improvements to the documentation. As far as I can tell, no effort was made to communicate this change outside of the IG prior to this proposal, nor their findings from any consultations they may have had.

doerners:
> this PR removes functionality without replacing it with something equivalent.

This is not entirely true, because you wouldn't loose functionality when removing `<explicit>` (as it is currently modelled). Without `<explicit>` you could encode its content like:

```XML
<physDesc>
  <p class="explicit">your explicit text</p>
</physDesc>
```

Or even more simple like:

```XML
<physDesc>
  <p>Explicit: your explicit text</p>
</physDesc>
```

If one only cares for human-readability this would be fine I guess, it is in fact valid MEI. I'm not saying it's pretty, but that is not the point here.

> `<explicit>` was introduced in 2018 as part of an effort to improve manuscript descriptions in MEI [...]. That it hasn't been widely adopted may be an artifact of its rather late introduction to the spec.

To me it seems that `<explicit>` was strictly copied from TEI without much thought about its implication for the musical context. This is foremost indicated by its description, since it is more or less exactly the same as in TEI. Since there also seem to be no (other) issues or PRs discussing its shortcomings, this could be an indicator that it hasn't been used by the community at all (for whatever reason). Thus, the IG discussed that for rare cases where `<explicit>` is indeed needed, a customization seems to be an appropriate approach.

> That <explicit> does not currently allow encoding of musical material is, I believe, an oversight that should be rectified.

Yes this is true. On the other hand, considering **only** `<explicit>` neglects cases where one wants to encode an excerpt from a work that can't be classified as `<explicit>` or incipit.

> The alternative solution (excerpt) has problems, since it suggests removing something that works and is widely adopted (`<incip>`) because of something that doesn't work (`<explicit>`) and replacing it with a third, new thing (`<excerpt>`), simply because implementing this would somehow be more "optimal". As I suggested before, if it's a question of the internal implementation not having the right names, we can adopt the TEI strategy and consolidate the modelling to something more generic and equivalent to their quoteLike model to align the internal implementation with the public-facing schema.

The proposal of remodelling `<incip>` (which doesn't have the "right" name either) to `<excerpt>` didn't merely came from the impression that it would be more "optimal". It was an earnest concern to the IG how users could encode cases that can't be classified as either `<incip>` or `<explicit>`. The IG doesn't know of any existing functionality in MEI (that is comparable to the one of `<incip>`), but maybe we are mistaken, so please let us know. This perceived lack of functionality is why the IG concluded that a more generic element would solve the problem without convoluting the schema with several new elements. 

Furthermore, my impression is that MEI has no intention to seem like a mere customization of TEI (substantiated by this [comment](https://github.com/music-encoding/music-encoding/issues/878#issuecomment-1024618688)), I feel it is therefore a little contradictory simply copying elements from TEI and adopting all of TEI's approaches. Don't get me wrong, I'm not saying we have to invent the wheel all over again but I think we definitely should discuss the implications for MEI and its use cases and this maybe means that it is reasonable to differ from TEI now and then.

> When to use elements and attributes is something more like an art, than a science, but regardless of which is used, we retain the same overall necessary descriptive complexity to cover the semantics of the desired encoding.

I don't understand the point you are making here. It was never up for discussion that we would need proper documentation.

> I am also a bit taken aback by the approach the IG has taken with this change. The initial indications (https://github.com/music-encoding/metadata-ig/issues/14) were that they were to discuss it for the purposes of improving the documentation for it. After a single meeting, they then announce a decision to completely remove it and have filed this PR. This was despite earlier indications that "... we didn't have an idea how to transfer this to music. But maybe there is a specialist for medieval music who can [help] us." On that ticket I provided several links to help them with the improvements to the documentation. As far as I can tell, no effort was made to communicate this change outside of the IG, nor their findings from any consultations they may have had.

tbh, to me this reads like a very harsh critique to the IG and their work. The issue was opened on January 13th, so it has been open for over two month and we had 3 meetings in between. Furthermore, everyone is free to attend the meetings to discuss issues and the IG repo is public so basically anyone can comment there as well. And even though we opened a PR this doesn't mean that there can't be further discussion here with the wider community. If I read your comment correctly, your critique is that we didn't take enough time and consideration to come to our conclusion and didn't properly communicate with the community. So I would like to ask, what is your concrete suggestion here?

Should the community come up with a guideline or rule how long things have to be discussed in order to open PRs and how the decision making process has to be communicated? (I couldn't find anything regarding this in the By-Laws)

And yes you provided links. Those links show the application of `<explicit>` with TEI. If we just want to copy what TEI does, fine, but I would argue that the links aren't entirely helpful to understand the element in the specific MEI context or the considerations for including it as-is. This makes proper documentation hard, no matter which approach is taken.

ahankinson:
@doerners thank you for your response. I'm not going to engage directly, because I don't think it will be particularly helpful at this point, until others have had a chance to speak.

I'll just point out that I am just one vote (hence the "-1") on the larger technical team. If others are on board with your approach, then I won't stand in the way.

musicEnfanthen:
Thank you both very much @doerners and @ahankinson for sharing your inputs, ideas and concerns. I think that – beside different perspectives concerning the expected outcome – this discussion is still very constructive and will help to improve the subject under consideration. In fact, I see perfect agreement in your arguments that explicit, as it is now, needs to be changed. The question that arises is whether this change should manifest in an improvement to the documentation, a deprecation of `<explicit>`, the removal of the element, or a rebuild of `incip` and `explicit` as `excerpt`. Looking at the initial issue in the metadata repo, it seems that all these different options were taken in consideration at one point or another. 

I agree with @ahankinson that it is somewhat difficult to track the decision finding process regarding these different options from the description in the original issue or this PR for non-metadata-ig members: starting as "discuss explicit" to improve its documentation, going to a proposal to "deprecate" the element ([comment](https://github.com/music-encoding/metadata-ig/issues/14#issuecomment-1070984921)), which finally resulted in this PR that removes the element. I also agree with @ahankinson  that the removal without alternative or the remodelling as `excerpt` would have far-reaching implications that need to be considered thoroughly, also from members of the @music-encoding/technical-team that are not members of the @music-encoding/ig-metadata . 

On the other hand, I totally see and understand the frustration of the @music-encoding/ig-metadata (and eventually other IG's) resulting from the lack of guidance on what an appropriate workflow should look like so that an idea developed in an IG and deemed important to that IG's use cases might someday find its way into the schema. And I see it as the responsibility of the @music-encoding/technical-team-co-chairs to come up with such guidance in order to give a little more security of action and communication here. So that is something that @bwbohl and I will definitely work on, in coordination with the @music-encoding/mei-board  and also the @music-encoding/technical-team and IG members. I would propose to dedicate a good part of our next ODD meeting to that topic, and to start/continue discussion about that crucial subject in another issue/GH discussions/any of our SLACK channels (but outside of this PR here).

pe-ro:
I'm no longer a member of the technical team, but for what it's worth, I'm a -1 here too.  Sort of.  More below.

It seems to me that this particular tempest is based on a too literal definition of "incipit".  Granted, the current definition in the guidelines/schema, "The opening music and/or words of a musical or textual work.", doesn't currently spell it out, but the term "incipit" usually understood in thematic catalogs may be applied to excerpts as well as the literal opening music/text.  For example, introductory material is usually omitted.  Also, multiple incipits may be provided to identify a work, obviously not all of which are literally the "opening" words or music.  Think of the "vamp" at the start of Mozart's 40th, the piano introduction of most songs, or the entrances of each voice in a polyphonic piece.  Using this broader definition of "incipit" obviates the need for an `<excerpt>` element.

True, the content model for `<explicit>` was borrowed directly from TEI.  But that was done so that discussions like this one would be triggered.  I'm in favor of using the content model of `<incipit>` in `<explicit>` or another model that makes it possible/easy to encode musical or textual explicits -- perhaps a model that combines the current model of `<incipit>` with the current model of `<explicit>`.

Regarding the workflow for getting suggestions/recommendations from an IG into the schema, I can only say that in my opinion 1) hard-and-fast rules will only lead to more/other frustrations as each instance has its own maturation schedule and process, and 2) that an abundance of patience is necessary for collaborative endeavors.  In my experience, the most important thing is to assume good intentions from all parties involved in the discussion and to not expect that everything will always go your way even when yours is the most fan-damn-tastic idea in the history of mankind.  Remember Nikola Tesla. 😄

lpugin:
I agree that having a better definition for `<explicit>` would be highly desirable. Before dropping it, I would suggest to take into account a couple of additional points. Maybe they have been discussed by the IG already, and I apologize for going through them again if that is the case. I hope they will still be useful.

We should not forget that MEI documents can also be text (only) ones. For example, it would not be unrealistic to have a project wiling to build an MEI corpus out of which some of the documents are just text. One can argue they should mix MEI and TEI documents, but there is no reason to force them to do so. In such cases, having a purely textual `<explicit>` element might make sense, in the same way one might have a purely textual `<incipit>`.

Another point that is also in favour of a revision / extension of `<explicit>` instead of dropping as suggested by @pe-ro is that fact that music can sometimes be equally well (if not better?) described by the end of a piece rather than its beginning. I am thinking in particular about the question of mode identification and the role of the final here. Using `<explicit>` might not be the traditional way of encoding excerpt of a document, but raising the role of the final might be a perfectly valid argument for willing to do things differently.

bwbohl:
it took me a while to know how to proceed here…
Today @musicEnfanthen and I had a chat about this. 

First: This issue brought to our attention that there is no general procedure for deprecating things in MEI, but we think there should be one (how exactly that would work should be discussed in another place).
With that in the back of our heads, our proposed compromise for this 'explicit' issue here would be a multi-step process:

1. We think the current content model of `<explicit>` seems to be a bug or oversight; fixing this should be the priority. So we propose to align the content model of `<explicit>` with that of `<incip>` to get these two in sync.
  (Side note: maybe it hasn't been used just because of that odd behavior)
2. Meanwhile, if the need arises for encoding other 'excerpts', a proposal for a corresponding encoding model could be fleshed out. – maybe even with merging the three cases in a general concept for encoding 'typed excerpts' (incipit, explicit, …). Then a fleshed-out proposal on how to replace explicit/incipit or any other middle excerpts should be drafted and elaborated and discussed with the wider community before implementation.

Does that seem plausible and acceptable to all those involved here so far?

ahankinson:
Thanks @bwbohl -- Just a few notes.

1. There is no `<incipit>` -- only `<incip>`. I think this is because of their different histories in MEI.
2. In principle no problem with aligning the content models, but I wouldn't want to see `<incip>` or `<explicit>` removed and replaced with `<excerpt type="incipit">` (or something similar). In that sense, I think we should follow the example of the TEI with their [`model.msQuoteLike`](https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-model.msQuoteLike.html) class and keep the individual elements available, and possibly expand the content model for our own `<rubric>` element.

rettinghaus:
I’m -1 on remodeling `explicit`.
Please let me recap why the IG believes removing this element is the best way to go with the least impact on the schema. Trying to align `explicit` with `incip` would lead to too many duplications, since you would have to match `incipText` and `incipCode` to the parent `explict` element. Although one could try to remodel `incipCode` into `musicCode` (from TEI), this would be a _real_ breaking change. Eventually a generalized replacement (e.g. `excerpt`) for both would be desirable, but we do not see this for the upcoming version of MEI. 
The IG didn’t took the decision light hearted, and it was in fact discussed over several weeks. Every IG meeting is public and everyone is invited to join them. The IG is made up of librarians, information scientists, and musicologists, most of whom have worked on work catalogues themselves. Given the fact that `explicit` currently is poorly modeled (and yes, it hurts), and that a musical ending of a piece is barely used in the context of work catalogues for identification purposes, we decided unanimously that just removing it would pretty much solve the problem. I am saddened that the IG is denied competence in such matters. Some people are beginning to wonder why you need an IG at all. Throwing constantly red herrings into the discussion on a pull request (e.g. the key word of doom: „breaking change“) isn’t very helpful either to say the least.

I'm frankly disturbed that what the IG clearly sees as the worst possible course of action here is now supposed to be the best. (BTW is the name of the element `incip` not `incipit`, so to bring it in line wouldn’t `explicit` have to be renamed into `explic`?)

Interestingly, we had a fairly similar discussion about the removal of fingerprint, where it was declared indispensable until we found out that it was originally removed more than ten years ago, whereupon the voices in the wilderness suddenly fell silent. 

I would be really interested in seeing three real-world examples from a project that actually used `explicit` in MEI.

ahankinson:
Because I feel a certain sense of responsibility for this response, I will try to address it as best I can. 

> I am saddened that the IG is denied competence in such matters.

At no point was any member of the Metadata's IG's competence questioned. I think that needs to be stated clearly, outright and up-front to avoid any misunderstanding. This proposal came from a certain viewpoint, and that viewpoint was always understood to come from a place of professional competence. I, nor any other participant in this discussion, ever once called into question the competence of any person making this proposal. To suggest otherwise is disingenuous.

What I think is misunderstood, however, is that the IG does not have *authority* over any part of the schema. Authority in this community is, principally, the elected MEI Board, and in day-to-day operations, the Technical Team. When the IG proposes a change, that change is subject to critique and review by the Technical Team, who have their own competencies (including metadata, but who are not part of the metadata IG). When setting up the IG mechanisms there was never any precedent, nor expectation, that the IGs would have singular authority over any part of the schema, and whose proposals would be "rubber stamped" and accepted simply because it comes from an IG rather than an individual. 

Additionally, this statement worries me:

> we decided unanimously that just removing it would pretty much solve the problem.

No person, or group of people, "decides unanimously" to make a change. Again, the mechanisms for change here are through the Technical Team (whose membership is wide open).

My sense is that the IG felt it had sufficient expertise within its own ranks to make a decision, and then were surprised when this viewpoint was not unanimously accepted outside of the IG. This says to me, principally, that the change was not consulted widely enough before it was "decided". When it finally did make it out of the IG it was presented to the community as a "fait accompli", and there was surprise, and indeed some idignation, that this decision would be questioned. 

If we were to re-run this whole scenario, I would hope the process would go something like this:

1. **Metadata IG**: Hey, we've noticed that there's a problem with `<explicit>`. We've looked at it and noted that it has X and Y problems. We've filed an issue on the schema repo describing these problems.
2. **Technical Team / MEI Community**: Thank you for noticing this! It is true that there are problems here. We should make sure (X project | Y community) knows about this. (Discussions happen). OK, we have decided to (_remodel explicit | remove the element | do something else_). Would the IG be willing to make a pull request?
3. **Metadata IG**: Sure! Here is the PR that implements the changes that were decided. 
4. **Technical Team**: Great! Thanks. 

This is in contrast to this process:

1. **Metadata IG**: We're going to discuss `<explicit>`. (Discussions happen). OK, we've decided to remove it. Here's the PR.
2. **Technical Team / MEI Community**: Uhm. Right. Uh. Why? What's wrong with it? What options did you consider? Can we examine this decision or propose alternate solutions? Who did you talk to? 
3. **Metadata IG**: No, we discussed all of those extensively, and decided that this is the best option.
4. **Technical Team**: But we don't agree with that!

> I'm frankly disturbed that what the IG clearly sees as the worst possible course of action here is now supposed to be the best. 

I'm likewise disturbed by the fact that you, or the IG, seems to think its decisions need to be unilaterally approved and are not open to critique or discussion, improvement or alteration. All Interest Group decisions, by any IG, are subject to the Technical Teams advice and review. To be "disturbed" by alternate proposals and compromise solutions by the larger community is a highly problematic attitude, particularly coming from a member of the MEI Board.

riedde:
> If we were to re-run this whole scenario, I would hope the process would go something like this:

I would also like to have done it in another way. I also do not want to argue and stay out of it. However, just some background information about the workflow during the last few years (from my view):

1. **Metadata IG:** Discussing Issues. Agreeing on a proposal. Bringing "wishes" to the ODD team.
2. **ODD-Meetings**: If you like to have something like this, please prepare a PR, and _then_ we can discuss it.
3. **Metadata IG:** Some members starting to learn ODD. Opening PRs for discussion.

Besides: To make everything much more transparent we brought all our discussions and issues (most of them are also discussions right now) to our IG-repo. We'll try to adjust this by using discussions for all kinds of ideas and questions and using issues only for working on 'real issues' or 'noticed problems' (from our perspective) in the future.

What I think we need is a basic standardized workflow of "how to contribute" because we all have different ideas about how it works or should work, and to avoid further frustration. This workflow should, no, must be defined by a schema authority. As I know the technical board ( @bwbohl and @musicEnfanthen ) are already working on that issue.

KristinaRichts:
Perhaps we could bring the discussion back into an objective mode. Nobody in the IG thinks that our decisions have to be accepted just like that. I am sorry if this impression has been created. We are all well aware of the steps it takes to initiate change.

> My sense is that the IG felt it had sufficient expertise within its own ranks to make a decision, and then were surprised when this viewpoint was not unanimously accepted outside of the IG. This says to me, principally, that the change was not consulted widely enough before it was “decided”.

That is not fair. What do you think interest groups are for, if not to pool expertise on a subject and to contribute their knowledge? Just to express that you like metadata? I think we all have enough other things to do.
As one of the convenors of the IG, I’m a little frustrated right now to read all this. The Interest Group has done so well in the last year. There really is a will to bring expertise and help improve the format. Sorry to say this, but this discussion is not conducive to this workflow and mixes up different things. There has been a feeling in the group for some time that metadata concerns have to take a bit of a back seat to other things that always seem too important.
Margrethe and I have always encouraged contributing things in git, not to be deterred. Klaus is just standing up for the whole group here. Maybe we should talk about the workflow and the role of IGs in general separately, I’m with Dennis on that. The discussion about `explicit`, `incip` and `excerpt` should perhaps be taken a step back again. Thank you.

musicEnfanthen:
Thank you very much @KristinaRichts and @riedde for sharing your feelings and perspective on that issue. And with issue I do not mean the technical content of this PR ("Removal of explicit"), but the path this discussion has taken towards a perceived friction between the interests of the Metadata IG and those of the Technical Team which, in my opinion, should not be in contradiction.

First of all let me say that @bwbohl and I take this issue very seriously! When feelings get hurt, we all should be very careful and prudent, and try to avoid further escalations and find ways to restore mindful communication and trust.

Saying this, I am convinced that all participants of the discussion of this PR are acting on good will, that is with the intention to help and support each other. And I am also convinced that no one involved here has ever intended to deny or diminish the work, competency, or expertise of other individual members of the community or an Interest group as a whole. 

Unfortunately, however, this is one of the perceptions/hurt feelings that emerged from the discussion above, and I feel very sorry about this. As part of the Technical Co-Chair Team I also feel responsible for this and would like to apologize for this impression to all members of the Metadata IG. You are doing an amazing job and your contributions to the development of MEI are highly appreciated, valued and always welcome! And so are you, as individuals, colleagues and members of this community. Please do never assume otherwise.

I totally see and understand the frustration of the IG, having several open tickets and PR's over the last months that seem to be delayed, questioned or rejected. We will need to identify reasons for that and solutions to improve the situation. 

As @ahankinson pointed out above, one of the responsibilities of the Technical Team is that of a supervision, in the sense of advice and support like a second eye in proofreading or reviewing. This supervision, in my view, should not be understood by any party as "the final obstacle in a hurdle race" preventing the ideas of the PR creators from running, but as a springboard to let them fly. There should be no competition between people who prepare a PR and those who review it, but ideally it is a symbiotic process which helps to improve the issue at hand. That this is not always possible without discussion, dissent or objection, and even sometimes leads to final rejection of a good idea, is without saying, but finding a "best possible compromise" is probably not possible without searching, checking alternatives, or listening to contradictory perspectives and concerns. For the thorough and responsible work they have done in this respect over the past years, I would like to thank all the members of the Technical Team. Your critical voices are always challenging, and can be quite annoying and frustrating, but I believe them to be crucial and vital to this community. We should just make sure that our communication and feedback is perceived as constructive and doesn't leave people with bad feelings, especially since purely written communication has its own pitfalls and tends to be more prone to misunderstandings.

Concerning the need for a "basic standardized workflow" for contributions that was raised multiple times above: I would not like to see the Technical Co-Chairs dictate such a workflow. Rather, I would suggest having a joint meeting (or multiple) with interested people, members of the IGs, of the Technical Team and the Board to collect needs and requirements from all parties. The [contributing guidelines](https://github.com/music-encoding/music-encoding/blob/develop/CONTRIBUTING.md) we have already could serve here as a point of departure from where discussion, dissent or obejction could finally lead to a "best possible compromise" as well.

[On monday, we will take this topic to our next Board meeting, too, to discuss further approaches. Meanwhile, I invite everyone to stay in touch, to continue talking to each other, and to address any issues of discomfort as soon as they arise.]

PR #938, opened 2022-03-24
Add and handle val tags

rettinghaus:
This PR tries to address one part of #913 by unifying the handling of attribute values in the Guidelines. 
Before values were treated very differently in different locations, sometimes they were in italics, sometimes in single or double quotes, sometimes in plain text. I tried to find every occurrence, wrap it into `tei:val` tags and added a corresponding css rule, so that every value should be shown in single quotes (as it is done in the TEI Guidelines). Furthermore I created a new `guidelines.css` to collect specific rules there.

Additionally this brings some typographically correct possessive apostrophes.


bwbohl:
@rettinghaus could you please resolve conflicts and then ping me for review?

rettinghaus:
@bwbohl merge conflicts have been resolved.

bwbohl (dismissed):

source/docs/08-lyricsperfdir.xml
```diff
@@ -126,17 +126,17 @@
...
                <p>The use of <gi scheme="MEI">syl</gi> described in this section is common to CMN and other notation systems, such as mensural notation. Other uses specific to certain types of notation and repertoires are addressed in other chapters. See for example <ptr target="#neumes"/>.</p>
             </div>
             <div xml:id="lyricsInEvents" type="div2">
                <head>Vocally Performed Text Encoded Within Notes</head>
                <p>Each lyric syllable can be encoded directly within an associated note, either by using the <att>syl</att> attribute on <gi scheme="MEI">note</gi> or the <gi scheme="MEI">verse</gi> element.</p>
                <p>Using the <att>syl</att> attribute on notes is the simplest way of encoding vocally performed text and is recommended only for simple situations or for those encodings which do not focus on vocally performed text.</p>
-               <p>The following example from Handel's <hi rend="italic">Messiah</hi> (HWV 56) shows the use of <att>syl</att>:</p>
+               <p>The following example from Händel’s <hi rend="italic">Messiah</hi> (HWV 56) shows the use of <att>syl</att>:</p>
```
musicEnfanthen: Wondering why this is changed to the German version of the name? (Here and in lines 139 and 153) 

Also: How is it connected to the PR title and description?

source/docs/03-metadata.xml
```diff
@@ -342,8 +342,7 @@
...
-                     <p>A virtual unit (vu) in MEI is a measure of distance. It is determined by half the distance between adjacent staff lines where the interline space is measured from the middle of a staff line. Unless otherwise specified, the MEI virtual unit is set as the default unit.</p>
+                     <p>The <att>unit</att> attribute is restricted to certain values: <val>byte</val> (Byte), <val>char</val> (Character), <val>cm</val> (Centimeter), <val>deg</val> (Degree), <val>in</val> (Inch), <val>issue</val> (Serial issue), <val>ft</val> (Foot), <val>m</val> (Meter), <val>mm</val> (Millimeter), <val>page</val> (Page), <val>pc</val> (Pica), <val>pt</val> (Point), <val>px</val> (Pixel), <val>rad</val> (Radian), <val>record</val> (Record), <val>vol</val> (Serial volume), and <val>vu</val> (MEI virtual unit).</p>
```
musicEnfanthen, line 345: What happened to the explanation of the virtual unit in former line 346 here? Was it removed by accident?

source/docs/05-mensural.xml
```diff
@@ -184,7 +184,7 @@
...
...he other note levels (<hi rend="italic">brevis</hi>-<hi rend="italic">semibrevis</hi> and <hi rend="italic">semibrevis</hi>-<hi rend="italic">minima</hi>). This means that the <hi rend="italic">breves</hi> in <hi rend="italic">Ars antiqua</hi> do not have a "perfecta" / "imperfecta" quality, and this is why there is no <att>dur.quality</att> attribute for the <hi rend="italic">breves</hi> in the previous example. However, the <hi rend="italic">brevis</hi> can have a ternary division (indicated by <hi rend="italic">tempus=”3”</hi>), being divided either into three (equal) <hi rend="italic">minor semibreves</hi> or into a <hi rend="italic">minor</hi>-<hi rend="italic">maior</hi> pair of <hi rend="italic">semibreves</hi>. The encoding also allows for the possibility of encoding a binary division of the breve in <hi rend="italic">Ars antiqua</hi> notations: the indication <hi rend="italic">tempus=”2”</hi> indicates the breve is divided into two equal <hi rend="italic">semibreves</hi>. This is why in this example with <hi rend="italic">tempus="3"</hi>, the <hi rend="italic">semibreves</hi> do have a <att>dur.quality</att> attribute (with values <val>minor</val> or <val>maior</val>).</p>
```
musicEnfanthen, line 187: Would it make sense to use val-tags for `"perfecta" / "imperfecta"` here, too?

source/docs/05-mensural.xml
```diff
@@ -281,7 +281,7 @@
...
                      </specList>
                   </p>
-                  <p>The signs for the Italian <hi rend="italic">divisiones</hi> can be encoded in the <att>sign</att> or <att>mensur.sign</att> attribute (to be used with <att>mensur</att> or <att>staffDef</att> respectively), with the values: "t" for ternaria, "q" for quaternaria, "si" and "i" for senaria imperfecta, "sp" and "p" for senaria perfecta, "oc" for octonaria, "n" for novenaria, and "d" for duodenaria. And the additional values for senaria gallica, "sg" and "g", and senaria ytalica, "sy" and "y".</p>
+                  <p>The signs for the Italian <hi rend="italic">divisiones</hi> can be encoded in the <att>sign</att> or <att>mensur.sign</att> attribute (to be used with <att>mensur</att> or <att>staffDef</att> respectively), with the values: <val>t</val> for ternaria, <val>q</val> for quaternaria, <val>si</val> and <val>i</val> for senaria imperfecta, <val>sp</val> and <val>p</val> for senaria perfecta, <val>oc</val> for octonaria, <val>n</val> for novenaria, and <val>d</val> for duodenaria. And the additional values for senaria gallica, <val>sg</val> and "g", and senaria ytalica, <val>sy</val> and "y".</p>
```
musicEnfanthen: "g" and "y" at the end should be also embraced in val tags?

source/docs/11-scholarlyediting.xml
```diff
@@ -270,10 +270,10 @@
...
+                     <p>Like any other ‘controlEvent’ (see <ptr target="#eventsControlevents"/>), <gi scheme="MEI">cpMark</gi> is placed in the score using the <att>staff</att> and <att>tstamp</att> attributes. The end point of the mark itself, when necessary, may be indicated using the <att>tstamp2</att> attribute. The source material, which is intended to be inserted in the space indicated by the copy mark, can be identified by the attributes <att>origin.tstamp</att>, <att>origin.tstamp2</att>, <att>origin.staff</att> and <att>origin.layer</att>. While <att>origin.tstamp</att> provides the relative distance from the beginning of the "gap", <att>origin.tstamp</att> is relative to the position identified by <att>origin.tstamp</att>. However, <att>origin.tstamp</att> defaults to the same value as <att>tstamp2</att> and should only be provided when necessary. When neither <att>origin.staff</att> nor <att>origin.tstamp</att> are provided, they take the same values as the cpMark’s <att>staff</att> and <att>tstamp</att> attributes; that is, they indicate a strict ‘vertical’ or ‘horizontal’ copy.</p>
```
musicEnfanthen, line 273: Multiple single or double quotation marks in this paragraph that still need to be replaced with val tag?

source/modules/MEI.edittrans.xml
```diff
@@ -616,8 +616,8 @@
     </attList>
     <remarks>
       <p>The <att>character</att> attribute describes characteristics of the hand, particularly
-        those related to the quality of the writing, <abbr>e.g.</abbr>, 'shaky', 'thick', regular'. A description
-        of the tint or type of ink, <abbr>e.g.</abbr>, 'brown' or the writing medium, <abbr>e.g.</abbr>, 'pencil', may be placed
+        those related to the quality of the writing, <abbr>e.g.</abbr>, <val>shaky</val>>, <val>thick</val>, <val>regular</val>. A description
```
musicEnfanthen:
```suggestion
        those related to the quality of the writing, <abbr>e.g.</abbr>, <val>shaky</val>, <val>thick</val>, <val>regular</val>. A description
```

source/modules/MEI.analytical.xml
```diff
@@ -228,7 +228,7 @@
     <attList>
       <attDef ident="key.accid" usage="opt">
         <desc>Contains an accidental for the tonic key, if one is required, <abbr>e.g.</abbr>, if key.pname
```
musicEnfanthen: Wouldn't `key.pname` and `key.accid` qualify to be marked up with att tag?

source/modules/MEI.midi.xml
```diff
@@ -17,7 +17,7 @@
         </datatype>
       </attDef>
       <attDef ident="midi.duty" usage="opt">
-        <desc>Specifies the 'on' part of the duty cycle as a percentage of a note's duration.</desc>
+        <desc>Specifies the 'on' part of the duty cycle as a percentage of a note’s duration.</desc>
```
musicEnfanthen, line 20: Is 'on' a value here?

source/modules/MEI.visual.xml
```diff
@@ -1655,8 +1655,8 @@
           <rng:ref name="data.BOOLEAN"/>
         </datatype>
         <remarks>
-          <p>This attribute is ignored when the bar.method attribute's value is "mensur" or
-            "takt".</p>
+          <p>This attribute is ignored when the bar.method attribute’s value is <val>mensur</val>
```
musicEnfanthen: bar.method with att?

source/modules/MEI.xml
```diff
@@ -478,9 +478,7 @@
     </content>
   </macroSpec>
   <macroSpec ident="data.CERTAINTY" module="MEI" type="dt">
-    <desc>Values for certainty attribute. Certainty may be expressed by one of the values 'high',
-      'medium', or 'low'. The value 'unknown' should be used in cases where the encoder does not
-      wish to assert an opinion.</desc>
+    <desc xml:lang="en">Values for certainty attribute.</desc>
```
musicEnfanthen: Why did the further description go to remarks?

source/web/css/guidelines.css
```diff
@@ -0,0 +1,22 @@
+/* layout for MEI Guidelines */
+p, div.p {}
```
musicEnfanthen: empty rule?

source/web/css/guidelines.css
```diff
@@ -0,0 +1,22 @@
...
+
+span.att {
+    font-weight: bolder;
```
musicEnfanthen: The former font-weight in mei-screen was 700, which would correspond to "bold" (with default normal being 400, bolder would only be 500). Change intentional?

utils/guidelines_xslt/odd2html/htmlFile.xsl
```diff
@@ -61,6 +61,8 @@
                             href="{$cssPath}css/mei-website.css" />
                         <link rel="stylesheet" media="screen" type="text/css"
                             href="{$cssPath}css/mei-screen.css" />
+                        <link rel="stylesheet" media="screen" type="text/css"
+                            href="{$cssPath}css/guidelines.css" />
```
musicEnfanthen: What's the difference of this new `guidelines.css` file to `mei-screen.css`? Would it make sense to have those declarations there, too?

musicEnfanthen (changes requested):

utils/guidelines_xslt/odd2html/htmlFile.xsl
musicEnfanthen: Also, do we need to handle those val and att tags in the printed guidelines (`mei-print.css`) as well (cc @kepper?)?

source/docs/03-metadata.xml
bwbohl: good catch

musicEnfanthen:
Thank you @rettinghaus. Please find some comments and open questions above.

source/docs/08-lyricsperfdir.xml
rettinghaus: Something I stumbled upon so I decided to give him his original name. But both variants are valid and it doesn't belong to this PR, so I'll change it back.

source/docs/03-metadata.xml
rettinghaus: Somehow got lost by merging.

source/docs/11-scholarlyediting.xml
rettinghaus: No.

source/modules/MEI.visual.xml
rettinghaus: How is it connected to the PR title and description? But happy to add it.

source/docs/05-mensural.xml
rettinghaus: I don't think so.

source/modules/MEI.xml
rettinghaus: This one was copied over directly from TEI, see https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-teidata.certainty.html 
Maybe it would be good to also copy "defines the range of attribute values expressing a degree of certainty."?

source/web/css/guidelines.css
rettinghaus: This CSS is a bare bone file to be populated later. This line here is a placeholder indicating that the other rules are not overwritten for now.

source/web/css/guidelines.css
rettinghaus: I found the attributes to be too thick, so this is a personal choice. For comparison: TEI highlights them only by adding a `@`.

utils/guidelines_xslt/odd2html/htmlFile.xsl
rettinghaus: My idea was to separate rules to show what is specific for the guidelines. I'd leave `mei-print.css` to @kepper, as I don't know anything about this and how to test it.

rettinghaus:
@musicEnfanthen thanks for the thorough review.

utils/guidelines_xslt/odd2html/htmlFile.xsl
musicEnfanthen: I understand your intention, but is the function of mei-screen.css not the same as your intended function for guidelines.css (both having now rules specific for the guidelines)? But maybe I do not fully understand what mei-screen.css does exactly.

source/docs/05-mensural.xml
musicEnfanthen: Aren't these "values", too, in the sense you applied the val tag to other double quoted values?

source/docs/11-scholarlyediting.xml
musicEnfanthen: So what is the intended difference then between "gap" in double-quotation and 'vertical'/'horizontal' in single quotation, and shouldn't we stick to one or the other formatting, if there is no difference?

source/modules/MEI.xml
musicEnfanthen: It took me a while to understand that the description in TEI is only in the remarks, so you copied it over here from desc to remarks to follow the original modelling in TEI, right? If so, would that need to be checked with other datatypes, attributes, etc., as well, and how is it connected to the title and description of this PR here? (Sorry to be annoying about the latter, but it makes things easier to review and follow if there are less "side effects" not connected to the actual PR.)

source/web/css/guidelines.css
musicEnfanthen: I see your point, but my CSS analysis tool would throw me a "major code smell" for empty blocks: https://rules.sonarsource.com/css/RSPEC-4658

So maybe remove it now and include it as soon as it is really used. Or, if it needs to be kept, I would suggest to add a comment inside the empty declaration that this is intentionally left empty until implementation. Otherwise, my future self will always stumble upon this empty rule and bug you with questions about why it's there ;)

source/web/css/guidelines.css
musicEnfanthen: I have no strict opinion on the boldness of attributes in the guidelines, so fine with me. Just wanted to make sure it is not a typo. As stated above, for such changes not primarily related to the title and desc of the PR, it would help in the future to either avoid them at all (and put them in a separate PR) or at least indicate such changes in the description, to give others a hint about what to expect.

source/web/css/guidelines.css
rettinghaus: removed `guidelines.css`

source/web/css/guidelines.css
rettinghaus: moved rule back to `mei-screen.css`

source/docs/11-scholarlyediting.xml
rettinghaus: Agreed, but this is not connected to this PR.

source/modules/MEI.xml
rettinghaus: It felt easier to just copy the whole block, but if this is awkward, I'll revert this.

musicEnfanthen: Yes, please. I think it is better to have a separate PR for this change, where the content of desc could also be updated according to your proposal.

utils/guidelines_xslt/odd2html/htmlFile.xsl
musicEnfanthen: I guess these two lines are now obsolete, too, since the rules were re-integrated with mei-screen.css

rettinghaus: Exactly! Thanks.

source/modules/MEI.midi.xml
rettinghaus: No.

musicEnfanthen (dismissed):

musicEnfanthen:
Thank you @rettinghaus for addressing the comments. Looks good to me.

source/modules/MEI.cmn.xml
```diff
@@ -2383,7 +2383,7 @@
         that is, for repetition of the same pitch. In the case of irrational durations, such as such
         as two quarter notes in the time of five 8th notes in a measure of 5/8 time, decimal values
         may be used in the <att>dur.ges</att> attribute. For example, the <att>dur.ges</att>
-        attribute would take the value "2.5" if the <att>midi.div</att> attribute's value was "1".
+        attribute would take the value <val>2.5</val> if the <att>midi.div</att> attribute’s value was <val>1</val>.
         The <att>num</att> and <att>numbase</att> attributes may be used for explicit labelling of a
         tuplet, such as, '3' with an 8th-note triplet, '3:2' over a quarter-note triplet, etc. The
```
musicEnfanthen, line 2388: Sorry, just another observation: If the `2.5` and `1` above are embraced by `val` tags, `3` and `3:2` should be tagged here, too, shouldn't they?

source/docs/10-analysisharm.xml
```diff
@@ -297,7 +297,7 @@
                         <specDesc key="att.meterConformance.bar" atts="metcon"/>
                      </specList>
                   </p>
-                  <p>When used on <gi scheme="MEI">staff</gi> or <gi scheme="MEI">layer</gi> elements, the <att>metcon</att> attribute can be used to indicate if the duration of the contained events is equal to ("c"), less than ("i"), or greater than ("o") that predicted by the time signature. When used on the <gi scheme="MEI">measure</gi> element, <att>metcon</att> takes a boolean value, where "true" indicates conformance by all staff and layer descendants of the measure.</p>
+                  <p>When used on <gi scheme="MEI">staff</gi> or <gi scheme="MEI">layer</gi> elements, the <att>metcon</att> attribute can be used to indicate if the duration of the contained events is equal to ("c"), less than ("i"), or greater than ("o") that predicted by the time signature. When used on the <gi scheme="MEI">measure</gi> element, <att>metcon</att> takes a boolean value, where <val>true</val> indicates conformance by all staff and layer descendants of the measure.</p>
```
musicEnfanthen: "c", "i", and "o" with val tags? Like "true"

rettinghaus: Indeed. Thanks for spotting this one.

source/modules/MEI.cmn.xml
rettinghaus: `3:2` is not an allowed value in these attributes, so I think both, number and ratio, are supposed to show what "meaning" can be expressed with the two attributes.

musicEnfanthen (approved):

musicEnfanthen:
ODD meeting, May 25, 2022: Thank you @rettinghaus . @kepper will have a final look and check about consequences for the XLST workflows.

musicEnfanthen:
Ping @kepper. Would you be able to have a look until the next ODD meeting?

bwbohl (approved):

musicEnfanthen:
Merged at ODD meeting (July 2022). @kepper If you think that there is some change needed afterwards, please let us know.

musicEnfanthen:
Thank you @rettinghaus for the improvements.

PR #947, opened 2022-04-22
fix data.COLOR

rettinghaus:
This is a small fix to correct the description after the update from #599.

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -1022,7 +1022,7 @@
         <rng:data type="token">
           <rng:param name="pattern">#[0-9A-Fa-f]{6,6}</rng:param>
         </rng:data>
-        <!-- AARRGGBB values -->
+        <!-- #RRGGBBAA values -->
         <rng:data type="token">
           <rng:param name="pattern">#[0-9A-Fa-f]{8,8}</rng:param>
         </rng:data>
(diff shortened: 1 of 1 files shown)


ahankinson:
I think these are all taken from the available CSS notation formats (https://developer.mozilla.org/en-US/docs/Web/CSS/color_value) so do we need to list them all individually? Can we point to the CSS specification for all available formats? (If that's the case then we need to also make sure our tools support `rgba()` and `hsl()` functions... but that's a slightly different story.

There's also differing guidance on whether it's RRGGBBAA or AARRGGBB. 

https://www.digitalocean.com/community/tutorials/css-hex-code-colors-alpha-values

https://caniuse.com/css-rrggbbaa

From what I can tell it's supposed to be RRGGBBAA for CSS and browsers.

rettinghaus:
Changed `AARRGGBB` to `RRGGBBAA`. However, it would be indeed good to know if this is _only a fix_ or a breaking change.

craigsapp:
> From what I can tell it's supposed to be RRGGBBAA for CSS and browsers.

Yes it is only RGBA and never ARGB (so it is a fix rather than a backwards compatibility problem).

ahankinson:
I did find reference to some platforms using Alpha channel first (Android) but I think this is probably the exception and not the rule.

craigsapp:
They must have colorful webpages 😉

rettinghaus:
Just to add: MEI supports a subset from the [CSS Color Module Level 4](https://www.w3.org/TR/css-color-4):

> The color-functions are [<rgb()>](https://www.w3.org/TR/css-color-4/#funcdef-rgb), [<rgba()>](https://www.w3.org/TR/css-color-4/#funcdef-rgba), [<hsl()>](https://www.w3.org/TR/css-color-4/#funcdef-hsl), [<hsla()>](https://www.w3.org/TR/css-color-4/#funcdef-hsla), [<hwb()>](https://www.w3.org/TR/css-color-4/#funcdef-hwb), [<lab()>](https://www.w3.org/TR/css-color-4/#funcdef-lab), [<lch()>](https://www.w3.org/TR/css-color-4/#funcdef-lch), [<oklab()>](https://www.w3.org/TR/css-color-4/#funcdef-oklab), [<oklch()>](https://www.w3.org/TR/css-color-4/#funcdef-oklch), and [<color()>](https://www.w3.org/TR/css-color-4/#funcdef-color).

We adopted all color names from CSS 4 draft, but `data.COLORVALUES` wasn't touched.

ahankinson (approved):

bwbohl (approved):

musicEnfanthen:
Thank you everyone!

PR #948, opened 2022-04-22
schema: develop placement clarification

kepper:
This PR is a continuation of the discussions at #853. 
These commits are the first step to clarify the use of place attributes. Here's a summary of the problems we tried to solve: 

- In MEI.neumes, the `@place` attribute is individually added to `<hispanTick>`, `<episema>`, and `<signifLet>`. They all use the `data.EVENTREL` datatype. 
- This datatype is somewhat similar to `data.STAFFREL`, including its child data types. There are slightly different values, though. 
- In the CMN part, there are two different `@place` attributes, as defined by the `att.placementRelStaff` and `att.placementRelEvent` attribute classes. 
- Right now, the only way a user can notice the difference between those is by the description of the attribute(s) shown in some editors. Without those descriptions, a user cannot tell the difference between staff-related and event-related placements. 

We wanted to keep all attributes as they are, so that instances don't get invalid. Here's a summary of what we did:

- `data.STAFFREL` has been merged with `data.EVENTREL` and renamed to `data.POSITION`.
- `data.POSITION` now references `data.POSITION.cmn` and `data.POSITION.neumes`, which are located in the proper modules and take care of allowing only the intended values, depending on activated modules. 
- `att.placementRelStaff` and `att.placementRelEvent` are merged into `att.placement`.

This leaves the question whether a user should be able to specify the anchor to which `@place` refers. In the current setup (before the changes in here), this was not possible, but could be infered from the elements that bear the `@place` attribute. The same could be achieved by proper documentation. If, however, more fine-grained control is wanted, a new attribute seems like a reasonable option. Potential names are `@place.relativeTo` with values like `event`, `staff`, or maybe even `notehead` and `stem`. 

Before proceeding, we would like to discuss this proposal to make sure it doesn't have any unwanted side effects. As soon as the specs are worked out, documentation will be adjusted.

### DIFF
--- a/source/modules/MEI.cmn.xml
+++ b/source/modules/MEI.cmn.xml
@@ -56,6 +56,56 @@
       </valList>
     </content>
   </macroSpec>
+  <macroSpec ident="data.POSITION.cmn" module="MEI.cmn" type="dt">
+    <desc>Location of musical material from the CMN repertoire.</desc>
+    <content>
+      <rng:choice>
+        <rng:ref name="data.POSITION.cmn.basic"/>
+        <rng:ref name="data.POSITION.cmn.extended"/>
+      </rng:choice>
+    </content>
+    <constraintSpec ident="between_requires_adjacent_staves" scheme="isoschematron">
+      <constraint>
+        <sch:rule context="mei:*[@place eq 'between']">
+          <sch:assert test="count(tokenize(normalize-space(string(@staff)), '\s+')) = 2">The @staff
+            attribute must contain 2 numerically-adjacent integer values.</sch:assert>
+          <sch:let name="tokenizedStaff" value="tokenize(normalize-space(string(@staff)), '\s+')"/>
+          <sch:let name="maxValue"
+            value="max((number($tokenizedStaff[1]), number($tokenizedStaff[2])))"/>
+          <sch:let name="minValue"
+            value="min((number($tokenizedStaff[1]), number($tokenizedStaff[2])))"/>
+          <sch:assert test="$maxValue - $minValue = 1">Staves <sch:value-of select="$minValue"/> and
+            <sch:value-of select="$maxValue"/> are not adjacent.</sch:assert>
+        </sch:rule>
+      </constraint>
+    </constraintSpec>
+  </macroSpec>
+  <macroSpec ident="data.POSITION.cmn.basic" module="MEI" type="dt">
+    <desc>Location of symbol relative to a staff.</desc>
+    <content>
+      <valList type="closed">
+        <valItem ident="above">
+          <desc>Above the staff.</desc>
+        </valItem>
+        <valItem ident="below">
+          <desc>Below the staff.</desc>
+        </valItem>
+      </valList>
+    </content>
+  </macroSpec>
+  <macroSpec ident="data.POSITION.cmn.extended" module="MEI" type="dt">
+    <desc>Location of symbol relative to a staff.</desc>
+    <content>
+      <valList type="closed">
+        <valItem ident="between">
+          <desc>Between staves.</desc>
+        </valItem>
+        <valItem ident="within">
+          <desc>Within/on the staff.</desc>
+        </valItem>
+      </valList>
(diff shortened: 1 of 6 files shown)


source/modules/MEI.cmn.xml
```diff
@@ -876,17 +926,16 @@
     </attList>
   </classSpec>
   <classSpec ident="att.numberPlacement" module="MEI.cmn" type="atts">
-    <desc>Attributes that record the placement and visibility of numbers that accompany a bowed
-      tremolo or tuplet.</desc>
+    <desc>Attributes that record the placement and visibility of numbers that accompany a notational object.</desc>
     <attList>
       <attDef ident="num.place" usage="opt">
-        <desc>States where the tuplet number will be placed in relation to the note heads.</desc>
+        <desc>Indicates where the number is written in relation to its reference object.</desc>
```
ahankinson, line 932: The data type, `data.POSITION.cmn.basic` says "`<desc>Location of symbol relative to a staff.</desc>`" but this says "Indicates where the number is written in relation to its reference object."

If the data type specifies that it's in relation to the staff, then I don't see how the instance of the data type can be more generic ('...in relation to its reference object...') 

So the data type should probably be made more generic in its description, and then the instance (`num.place` here) should try to be a bit more specific than just "reference object", if possible.

source/modules/MEI.cmn.xml
```diff
@@ -56,6 +56,56 @@
...
+          <desc>Above the staff.</desc>
+        </valItem>
+        <valItem ident="below">
+          <desc>Below the staff.</desc>
+        </valItem>
+      </valList>
+    </content>
+  </macroSpec>
+  <macroSpec ident="data.POSITION.cmn.extended" module="MEI" type="dt">
+    <desc>Location of symbol relative to a staff.</desc>
```
ahankinson, line 97: Location of a symbol relative to one or two staves? `between` needs two staves to make sense, no?

source/modules/MEI.cmn.xml
kepper: Thanks for catching this, @ahankinson. Actually, I didn't pay much attention to the `<desc>`s yet. Will look for better descriptions…

source/modules/MEI.cmn.xml
kepper: see my comment above – not much attentation paid to this yet – thanks for catching, and of course you're right ;-)

musicEnfanthen:
ODD meeting, May 25, 2022: Pushed to the next ODD meeting. @kepper and @rettinghaus will have another look to adjust the descriptions.

rettinghaus:
The more I think about that, the more I believe we need more than just "above" and "below", e.g. we don't have the possibility yet to say an `<ornam>` is left to an associated event:
ornaments left

Same with `accid@place`. currently we cannot encode explicitly the "normal" position of the accidental.

source/modules/MEI.cmn.xml
bwbohl: This is reflected in the constraints.

bwbohl:
@kepper could you please pick up here and see how we get it improved for MEI 6, thanks!

source/modules/MEI.cmn.xml
```diff
@@ -56,6 +56,56 @@
...
+          <sch:let name="minValue"
+            value="min((number($tokenizedStaff[1]), number($tokenizedStaff[2])))"/>
+          <sch:assert test="$maxValue - $minValue = 1">Staves <sch:value-of select="$minValue"/> and
+            <sch:value-of select="$maxValue"/> are not adjacent.</sch:assert>
+        </sch:rule>
+      </constraint>
+    </constraintSpec>
+  </macroSpec>
+  <macroSpec ident="data.POSITION.cmn.basic" module="MEI" type="dt">
+    <desc>Location of symbol relative to a staff.</desc>
```
bwbohl, line 84:
```suggestion
    <desc>Location of symbol relative to its reference object.</desc>
```

PR #949, opened 2022-04-23
remove remaining remnants

rettinghaus:


kepper (approved):
thanks for spotting those – one regex too few ;-)

kepper:
thanks!

PR #951, opened 2022-04-26
Develop custos

### LINKED ISSUE #944: Custos after staff
Sometimes a custom is not printed/written (completely) on the lines of the staff.  
image

Is there a possibility to encode this? 
@music-encoding/ig-mensural Would it make sense and be sufficient if we add the `@onstaff` attribute here, or is there a finer method needed, that allows saying something like "80% off the staff"? 

(Example taken from [Petrucci's _Frottole libro 3_](https://www.digitale-sammlungen.de/de/view/bsb00082309))

rettinghaus:
This PR adds the `@onstaff` attribute to `custos`. 
Based on [this comment](https://github.com/music-encoding/music-encoding/issues/944#issuecomment-1102526832) it introduces a new `@custos.onstaff` for `scoreDef` and `staffDef` as a global styling option.

Closes #944.

### DIFF
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -375,6 +375,18 @@
       </attDef>
     </attList>
   </classSpec>
+  <classSpec ident="att.custodes.vis" module="MEI.visual" type="atts">
+    <desc>Used by staffDef and scoreDef to provide default rendition for custodes.</desc>
+      <attList>
+        <attDef ident="custos.onstaff" usage="opt">
+           <desc>Indicates the default placement of a custos on the page. 
+             A value of <val>true</val> means before the end of the staff and <val>false</val> after the end of the staff.</desc>
+           <datatype>
+              <rng:ref name="data.BOOLEAN"/>
+           </datatype>
+        </attDef>
+     </attList>
+   </classSpec>
   <classSpec ident="att.clefGrp.vis" module="MEI.visual" type="atts">
     <desc>Visual domain attributes.</desc>
   </classSpec>
@@ -1472,6 +1484,7 @@
     <classes>
       <memberOf key="att.barring"/>
       <memberOf key="att.cleffing.vis"/>
+      <memberOf key="att.custodes.vis"/>
       <memberOf key="att.distances"/>
       <memberOf key="att.endings"/>
       <memberOf key="att.keySigDefault.vis"/>
@@ -1587,6 +1600,7 @@
     <classes>
       <memberOf key="att.barring"/>
       <memberOf key="att.cleffing.vis"/>
+      <memberOf key="att.custodes.vis"/>
       <memberOf key="att.distances"/>
       <memberOf key="att.keySigDefault.vis"/>
       <memberOf key="att.lyricStyle"/>
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -409,6 +409,7 @@
       <memberOf key="att.altSym"/>
       <memberOf key="att.color"/>
       <memberOf key="att.extSym"/>
+      <memberOf key="att.placementOnStaff"/>       
       <memberOf key="att.staffLoc"/>
       <memberOf key="att.typography"/>
     </classes>


lpugin:
Looking at the Pettrucci's print linked in the issue it seems that it is just a printing artefact and that the positioning of the custos simply depends on the space available at the end  of the staff. 

Would these be considered `onstaff` or not?

image

image

image

(I personally cannot really see the purpose of encoding this and I am not very convinced by the pertinence of this change.)

rettinghaus:
That was my original question: do we need a finer method to say "how much on or off the staff" a custos is? Taking it from the discussion I got the impression, that `@onstaff` is _good enough_.

ahankinson:
I think @lpugin is questioning whether this is actually intentional (as in, there is some musical reason why it does not appear on the staff) rather than an artefact of the printing process. If the former, then we should try to encode it. If the latter, then there's not really much need to support this in the encoding.

Are there any other sources where this sort of custos appears? Or is this the only one? Looking at the original source, I would tend to agree that the printers were... *ahem* liberal ... with their positioning of the custos symbols. I don't think we need to support it if we're simply visually encoding sloppiness by a printer.

rettinghaus:
Here are some more random examples:





Verovio usually puts them "onstaff":

craigsapp:
How would you encode the accidental placements:

When there is space they are show as in modern style:

Screen Shot 2022-04-26 at 8 28 19 AM

If there is no space, the are placed below (or above) the note the apply to:

Screen Shot 2022-04-26 at 8 28 04 AM

And they are repeated for each note (because natural signs had not yet been invented 😉 ?)

Screen Shot 2022-04-26 at 8 27 32 AM

`accid@place="below"` seems like a good encoding for this (It is not yet available in verovio).

ahankinson:
Looking at the examples, I'm inclined to think that this more a practical page layout decision than any musical stylistic decision. Where the custos is "off the staff" it's pretty much always because they ran out of staff lines area (probably because they were pre-drawn -- the print examples show multiple-impression prints, and the MSS would have been pre-drawn as well.). Given that this isn't necessarily a consideration in a digital rendering environment, I would be inclined to agree with @lpugin that this is more of a printing artifact than any "conscious" musical layout decision.

ahankinson:
Looking at the examples, I'm inclined to think that this more a practical page layout decision than any musical stylistic decision. Where the custos is "off the staff" it's pretty much always because they ran out of staff lines area (probably because they were pre-drawn -- the print examples show multiple-impression prints, and the MSS would have been pre-drawn as well.). Given that this isn't necessarily a consideration in a digital rendering environment, I would be inclined to agree with @lpugin that this is more of a printing artifact than any "conscious" musical layout decision.

rettinghaus:
Saying that these are artifacts that we shouldn't be allowed to encode?

lpugin:
> Saying that these are artifacts that we shouldn't be allowed to encode?

That is not the point. The question is about the purpose of the encoding. Are you planning to do some statistical analysis of on-staff vs not-on-staff custos? Or do you want to do look-alike transcriptions? I think `@onstaff` `true|false` would work for the former, but so much for the latter. One reason is that the custos are not precisely placed on the staff or not. I know you pointed that yourself. Where I am not convinced is more about the general direction we would take here. Were do you stop on that path? For example, why would you not want to encode that the text is overlapping the notes on this example ?

image

If you want to do look-alike transcriptions, then I think you should better look at using and extending the page-based customization we have for Verovio. That works already as it is to have the custos placed exactly as there are on the source.

ahankinson:
"Allowed" is a strong term. The question is who is the consumer of this type of data. If you want to describe particular features of an original source for humans to read, then you can use prose descriptions. If, however, you want to get specific about exact placement of symbols, then (I think) there needs to be a **musical** use case for supporting that. 

MEI isn't a page layout format, and it isn't a graphical encoding format (like SVG). In this particular case, we can see that the custos is only "off the staff" because of limitations of the medium on which this music was printed. There are no musical semantics being changed because it is on or off the staff. 

If we start trying to describe every graphical nuance, but in a music encoding format, then there is actually no end to what we can describe. It's like making a semantic encoding format to try and encode a painting...

```
<bowl>
    <pear color="green" />
    <grape type="cluster" num="12" stem="upper-left" />
    <apple rend="core-only" rotate="90" />
</bowl>
```

In other words: Trying to accurately reconstruct a painting from an XML encoding of it would be an extremely challenging task. Trying to make a consistent standard where multiple people can semantically encode different paintings and they all come out exactly as the original is actually impossible. 

Music notation is a graphical format, yes, but we need to be clear about where our boundaries are between what we should encode, and what we shouldn't, because we can't encode everything in the visual domain. To me, that line is drawn where there is a **musical** use case for encoding something. (And I recognize that the counter argument will be "well, that's not consistent, what about MEI's ability to encode [insert something else here]..." My goal in stating this is not to make global statements about what has been done in the past, it's about not making a bad situation worse in the future.)

rettinghaus:
I'm confused. Where is the musical semantics in saying a fermata is above or below the system? Following that I'm wondering where the purpose is in encoding a clef, when the encoded pitches of the notes are sufficient. So MEI is just a bad MIDI? (I am exaggerating excessively!) No, I think there were different practices to print/write custodes, and if one goal of MEI is to encode "what you see" this is indeed needed. 
A possible scenario for this is to take a bunch of mensural encodings and compare the sources if this is actually a printer specific thing (for what I get the impression it is) or not, does it change over time etc. etc. 
One could possibly encode this information in the header, but simple prose is not helpful in the described scenarios. We see them in the musical context, so this encoding belongs to the MEI body imho.

ahankinson:
> No, I think there were different practices to print/write custodes, and if one goal of MEI is to encode "what you see" this is indeed needed.

I don't think one goal of MEI is to "encode what you see." In fact, I think if you want to encode what you see, you draw it, or you point to a picture. Otherwise we end up with a format that does the digital equivalent of diplomatic transcriptions from the 1960s: Poor mechanical approximations of the "original appearance" of a source, which (IMO) harmed more than helped scholarship. If you want to show the original, we have the technology to show the original. But don't pretend that encoding every visual nuance of every score, regardless of its musical importance, is being "truer" to the original text somehow. 

I think we just discovered the uncanny valley of music encoding... :-D

> A possible scenario for this is to take a bunch of mensural encodings and compare the sources if this is actually a printer specific thing (for what I get the impression it is) or not, does it change over time etc. etc.

Yes, that is a valid use case. It is also not necessary to have this encoded on every custos -- you can add it once as a classification term or some other descriptive property and do your analysis on that. 

> One could possibly encode this information in the header, but simple prose is not helpful in the described scenarios. We see them in the musical context, so this encoding belongs to the MEI body imho.

As Johannes mentioned [in this comment](https://github.com/music-encoding/metadata-ig/issues/15), making an artificial break between "metadata" and "music" is not always helpful. And the header elements does not include just prose -- it allows you to classify your documents in any way you like, in machine readable form.

lpugin:
I must say that I am still not sure how you are going to evaluate if this is a printer specific thing when it is going to be highly subjective to decide if the custos should be encoded as on the staff or not for most cases. 

Anyway, this is without discussion a quite minor need - at least much less than having clefs, and one concern for me is to see it being an attribute in MEI.visual for `<scoreDef>` and `<staffDef>` when these two elements are already quite overloaded with attributes in my opinion.

kepper:
We had a conversation at the 2022 dev meeting in Berlin. Everyone in the room (@fujinaga, @rettinghaus, @musicEnfanthen, @pe-ro, @kepper) agrees that we will close this. Argument: Basically, new observations should be encoded using `@type`. If that shows sufficient benefit, it can be promoted to the use of `@class`, using some other external (classification), which will show the support of a wider community. Eventually, things can be promoted to dedicated (optional) elements in MEI. Some cases are so obvious that they can start higher up the ladder, but we agreed that for the current solution, it will be sufficient to stick with `@type` and observe it further on…

Issue #959, opened 2022-05-10
Roadmap to PDF Guidelines


musicEnfanthen:
Steps needed:

XSLTs
- [x] prepare XSLTs to handle PDF Guideline creation
- [x] integrate XSLTs into build workflow
- [x]  adjust to different build environments
- [x] merge pdf-guidelines branch to dev

DOCKER
- [x] prepare docker build workflows for the following steps:
    - [x] render live MEI examples to SVG (https://github.com/Edirom/staticVerovioConverter)
    - [x] generate complete PDF with [princexml](https://www.princexml.com/)
    - [x] integrate everything in docker-mei container [https://github.com/music-encoding/docker-mei](https://github.com/music-encoding/docker-mei)

- [x] documentation of build workflow with docker (README)

GH ACTIONS
- [x] include docker workflows in [deploy action](https://github.com/music-encoding/music-encoding/blob/develop/.github/workflows/deploy.yml)
  - [x] [build with ANT]
  - [x] check path structure of local and CI builds (#960)
  - [x] render live MEI examples to SVG 
  - [x] generate PDF  

The build process should be the same for dev build and tagged releases and work with both. Additional features for a release should be

- [ ] automatically creating a release on GitHub
- [ ] automatically adding release assets (Guidelines HTML & PDF , Schemata)

POTENTIAL ANT IMPROVEMENTS

- [ ] https://github.com/music-encoding/music-encoding/compare/pdf-guidelines-test...improve-gh-param-handling?expand=1
- [x] https://github.com/TEIC/Stylesheets/issues/544
- [ ] https://github.com/bwbohl/music-encoding/tree/feature-build-on-pr
- [ ] https://github.com/bwbohl/music-encoding/tree/feature_cleanBeforBuild

POTENTIAL STYLING IMPROVEMENTS

- [ ] https://github.com/bwbohl/music-encoding/tree/fix/classHeadingPadding

POTENTIAL OTHER IMPROVEMENTS

- [ ] https://github.com/music-encoding/music-encoding/pull/806

ALSO CHECK:

- [ ] #834

lpugin:
@kepper don't you want to have the script for converting the MEI within this repository? I am sure that https://github.com/Edirom/staticVerovioConverter works but then things (e.g., options for the conversions) are spread over different repository. It seems to me that a Python script call from the GH action would do the job - we use that a lot.

kepper:
[...]

PR #961, opened 2022-05-17
fix file paths for build process

musicEnfanthen:
This PR fixes the wrong file paths in the PDF guidelines build process. Tested for print version. CSS assets for online version need to be moved in correct folder in another step.

Adresses #959


kepper:
shall I check if it runs locally? It does look reasonable, but actually testing it requires more time than I have right now – can get back to it later no problem, but is it really necessary?

bwbohl (approved):

kepper (approved):
thanks!!!

PR #974, opened 2022-06-14
Create att.tuning class

### LINKED ISSUE #973: Create att.tuning
The tuning attributes (`@tune.pname`, `@tune.Hz`, `@tune.temper`) should be grouped into a separate `att.tuning` class. 

Maybe `@tune.accid` could be added, to allow semitones as reference.

Currently they are allowed only on `scoreDef`, but should probably also be allowed on `staffDef` and `layerDef`.

rettinghaus:
This PR moves tuning attributes to `att.tuning` and adds a `tune.accid` attribute to allow e.g. b flat as reference pitch. 
`scoreDef`, `staffDef`, and `layerDef` are all made part of this new class.
closes #973


lpugin:
Could you elaborate a bit more on the purpose of `tune.accid`? Maybe with a few examples? That would be great.

source/modules/MEI.shared.xml
```diff
@@ -3469,6 +3469,38 @@
...
+      </attDef>
+      <attDef ident="tune.temper" usage="opt">
+        <desc>Provides an indication of the tuning system, <val>just</val>, for example.</desc>
+        <datatype>
+          <rng:ref name="data.TEMPERAMENT"/>
+        </datatype>
+      </attDef>
+    </attList>
+    <remarks>
+      <p>If tune.pname is not specified, a4 can be assumed as the reference pitch.</p>
```
ahankinson: Is this supposed to say "A 440"? Not sure if "a4" means pname + oct?

rettinghaus: It's referring to scientific pitch notation. "A440" would mean pitch + tuning.

ahankinson: Sure; but does the "4" mean the octave then? If so, how can we capture the octave in the tuning attributes? As far as I can tell there is no way of specifying that the tuning is, for example, `c5` (to make up an example).

rettinghaus:
`@tune.accid` is meant to work like `@key.accid`. If a piece/movement is in C sharp major you would encode `key.pname="c" key.accid="s" key.mode="major"`. So if you want to a assign a tune frequency not to _a_ but to _b flat_ you would need `tune.pname="b" tune.accid="f"`.

source/modules/MEI.shared.xml
rettinghaus: Yes. So either we add another `tune.oct` attribute, or we restrict tunings to the "fourth" octave.

ahankinson: Do you have some real-world examples of where this might be used, or what experiences you've had that prompted the change? I'm struggling to think of some, so having some examples would help.

Specific questions:

 - In this case, I'm not sure what `tune.pname` means; My assumption is that it's necessary to have a fixed Hz reference point for the note name, e.g., A=440; A=466, etc. So just having `tune.pname="A"` seems to omit half of the necessary information?
 - Are there cases where the reference pitch would be anything other than A? I've never heard A=440 referred to as Bb=466... And in that specific case they would be equivalent, so it would be unnecessarily complicated. Can't you always reference a `pname` to A?
 - And, in the case that the reference pitch *is* something different, wouldn't you also need the temperament as well? A=466 or 440 doesn't tell you the temperament, so you need to include that to know the distance to the next note -- A=466 with meantone temperament means something, but what would `tune.pname=Bb`, `tune.temper=meantone` mean? I would guess you would need the fixed `tune.Hz` for this -- `tune.pname=Bb  tune.Hz="???" tune.temper="meantone"`

craigsapp:
> but what would tune.pname=Bb, tune.temper=meantone mean?

For Pythagorean temperament or a meantone temperament, this sort of parameter is used to specify the location of the [wolf fifth](https://en.wikipedia.org/wiki/Wolf_interval), where it would be opposite to the `tune.pname` fifth.

Here are diagrams from https://www.hpschd.nu/index.html?nav/nav-4.html&t/welcome.html&https://www.hpschd.nu/tech/tmp/mean.html for meantone :

Screen Shot 2022-06-16 at 7 50 01 AM

This configuration could be thought of as quarter-tone meantone on D, with the wolf fifth E-flat/G-sharp is placed opposite to D/A. (Lines within the circle of fifths indicates pure major thirds).

In this configuration:

Screen Shot 2022-06-16 at 7 49 51 AM

The wolf fifth is placed at A-flat/C-sharp which is opposite to G/D, and therefore it could be considered to be quarter-tone meantone centered on G rather than D.

In your system, the first example would instead be referred to as being on "A" (as that is the references), and the second example would be on "D".   But notice that this definition is arbitrary to a standard position, which you assign to "A", but the standard position is not clearly defined.


---------------------

Note that "[meantone](https://en.wikipedia.org/wiki/Meantone_temperament)" is a set of temperaments, but typically means quarter-comma meantone, where each perfect fifth is flattened by 1/4th of a pythagorean comma (the difference between C and B-double-sharp in a pythagorean tuning).  1/4-comma meantone produces pure major thirds, while 1/3-comma meantone (the next most common meantone) produces pure minor thirds.  Pythagorean temperament is also meantone in the sense that it is "zero-comma meantone", and Equal temperament is meantone where the comma adjustment is 1/12 of a comma.

-----------------------

What `tune.pname` would mean for temperaments other than pythagorean and meantone is not very clear.  Well-temperaments are irregular and there is no wolf fifth, so considering "D" the home pitch of the default position of the tuning makes no sense, and rather a transposition interval would probably be more useful.

--------------------

> `@tune.accid` is meant to work like `@key.accid`. If a piece/movement is in C sharp major you would encode key.pname="c" key.accid="s" key.mode="major". So if you want to assign a tune frequency not to a but to b flat you would need tune.pname="b" tune.accid="f".

This way of thinking is more like MIDI synthesizers than of historical temperament use.   Harpsichords were not tuned differently for each  musical key, but rather a temperament was chosen that allowed a range of keys to be played within it.   This is especially true for well-temperaments: they had irregularities that were fixed to specific pitches, and not relative to the key. I.e., "well-tempered" means there is one temperament in which all keys can be played in (so they are unlikely to be transposed from their standard position).

rettinghaus:
Aha, so `@tune.pname` is _not_ the pitch to assign a frequency (`@tune.Hz`) to, but indeed for specifying the center tone for a temperament. Then I guess we don't need the `@tune.accid` at all, but should be more clear in the description of `@tune.pname`.

And `@tune.Hz` always means the tuning for a4, correct?

ahankinson:
Not that it matters, but I'm still wondering what a value other than `A` would mean for `@tune.pname`? Is that actually a thing?

musicEnfanthen (dismissed):
Clarifying the meaning and usefulness of `tune.pname` (as tuning reference pitch name or central tone of a tuning system) should be part of another PR. This PR looks fine in moving the tuning attributes into a separate class.

bwbohl (dismissed):
thanks for remodelling the tune attributes for better reuse in other places of MEI files

source/modules/MEI.shared.xml
musicEnfanthen: ODD meeting (July 2022): We agreed that the meaning and description of tune.pname is misleading and needs clarification. This should be done in another PR, since the PR here does not change any existing attributes, but just moves tuning related things into a separate class. 

@ahankinson Are you ok with resolving this conversation?

musicEnfanthen (dismissed):

kepper (approved):
discussed during MEI dev meeting Berlin 2022, approved by Ich, Klaus, Stefan, Perry, jo

musicEnfanthen (approved):

source/modules/MEI.shared.xml
kepper: outdated

musicEnfanthen:
Thanks everyone! Merged at MEI Developer Workshop 2022 in Berlin

PR #976, opened 2022-06-15
Develop style

rettinghaus:
This PR is an attempt to bring the style of the source files a bit more in line with TEI. 

- extracts resolved abbreviations into `gloss` elements with an additional `versionDate` attribute
- adds `xml:lang` attributes to `desc`, `gloss`, and `remarks`

Further adjustments to the XSLTs are needed.

### DIFF
--- a/source/modules/MEI.analytical.xml
+++ b/source/modules/MEI.analytical.xml
@@ -5,143 +5,143 @@
   xmlns:xi="http://www.w3.org/2001/XInclude" xmlns:rng="http://relaxng.org/ns/structure/1.0"
   xmlns:sch="http://purl.oclc.org/dsdl/schematron" xml:id="module.MEI.analytical">
   <moduleSpec ident="MEI.analytical">
-    <desc>Analytical component declarations.</desc>
+    <desc xml:lang="en">Analytical component declarations.</desc>
   </moduleSpec>
   <classSpec ident="att.accid.anl" module="MEI.analytical" type="atts">
-    <desc>Analytical domain attributes.</desc>
+    <desc xml:lang="en">Analytical domain attributes.</desc>
   </classSpec>
   <classSpec ident="att.ambNote.anl" module="MEI.analytical" type="atts">
-    <desc>Analytical domain attributes.</desc>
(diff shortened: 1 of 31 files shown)


ahankinson:
@rettinghaus are you still working on this issue?

rettinghaus:
Thanks for the reminder! Obviously I need to resolve those conflicts. And I'd like to have some feedback on the changes in the XSLTs from @kepper. 
Or we split things into multiple PRs and keep this just for the changes in the sources.

rettinghaus:
Conflicts resolved.

kepper (approved):
discussed during MEI dev meeting Berlin 2022, approved by Ich, Klaus, Stefan, Perry, jo

musicEnfanthen (approved):

musicEnfanthen:
Thanks everyone! Merged at MEI Developer Workshop 2022 in Berlin

ahankinson:
Sorry, I had a question about this one: Given that `xml:lang` applies to all elements and their children, is it enough to just set it once on all the elements on the root element, and then only add/change it for those paragraphs where it is not in English? 

https://www.oxygenxml.com/dita/1.3/specs/archSpec/base/xmllang.html

"The `@xml:lang` attribute applies to all attributes and content of the element where it is specified, unless it is overridden with `@xml:lang` on another element within that content."

rettinghaus:
Well you could do that, but then it applies also to every specified element name and attribute value. But values like "adlib" and "sprech" are not English. Do we want to specify the language of every value? Also I think that it is not possible to 'reset' the language.

ahankinson:
But not every term within the paragraphs are English either -- It's music; half the terms are probably Italian! 😄 I think the main point is that the guidelines are written in English. 

It is possible to "reset":

```
<?xml version="1.0" encoding="utf-8" ?>
<doc xml:lang="en">
 <list title="Titre en français" xml:lang="fr">
  <p>Texte en français.</p>
  <p xml:lang="fr-ca">Texte en québécquois.</p>
  <p xml:lang="en">Second text in English.</p>
 </list>
 <p>Text in English.</p>
</doc>
```

From: https://www.opentag.com/xfaq_lang.htm

PR #991, opened 2022-09-16
pureODD: Rename macroSpec to dataSpec

rettinghaus:
This PR changes the outdated `macroSpec` into `dataSpec` for data type specifications.


musicEnfanthen:
@rettinghaus Thank you for this PR. Can we find a more specific PR title, please, like "Rename macroSpec to dataSpec"? And could you provide some short background information why/when macroSpec was outdated?

ahankinson:
It looks like it was deprecated with the move to "Pure ODD". 

Documentation here: https://tei-c.org/Vault/P5/3.4.0/doc/tei-p5-doc/en/html/ref-macroSpec.html

Discussion here: https://github.com/TEIC/TEI/issues/1737

Other discussion on the TEI mailing list here: https://lists.tei-c.org/pipermail/tei-council/2018/024993.html

musicEnfanthen:
Thanks. Do I read it correctly from the sources above, that the `type` attribute was completely deprecated and removed? So what do we do with `macroSpec[@type = 'pe']`

Also, there are some more XSLTs that need to be considered for `macroSpec[@type = 'dt']`:
- https://github.com/music-encoding/music-encoding/blob/1e2ee013d822d7c7b99ae22b83fc5c854d789055/utils/guidelines_xslt/odd2html/specs/moduleSpecs.xsl#L119
- https://github.com/music-encoding/music-encoding/blob/c3f27ed65f07cc27cebb01910959912ecec7be0f/utils/guidelines_xslt/odd2html/globalVariables.xsl#L157
- https://github.com/music-encoding/music-encoding/blob/c3f27ed65f07cc27cebb01910959912ecec7be0f/utils/guidelines_xslt/odd2html/specs/dataTypeSpecs.xsl#L18

ahankinson:
Yes, I think the `@type` parameter should probably be removed from all the `macroSpec` elements.

ahankinson:
@rettinghaus @musicEnfanthen Can we prioritize merging this? I'm noticing that not having this change in means that there are some newer PRs that still have the old `macroSpec` so we're going to get out-of-sync.

musicEnfanthen:
Pinging @lpugin @kepper and @pe-ro who are still in the meeting room

musicEnfanthen:
But maybe it is easier to go with other PRs first (with the old macroSpec), and in the end @rettinghaus would need to adjust this draft here anyway?

ahankinson:
What else is left to be done? I think @rettinghaus fixed the things you brought up, and also removed the `@type` from `macroSpec`

kepper:
I am very strong on pushing this down the road. Changing this markup will break guidelines generation, and I don't want to rush this. This doesn't mean it's a bad idea in general, but it's a bad idea for now. Nothing should wait for it. Sorry…

rettinghaus:
Schemas build fine, Guidelines seems to work. Needs some more checks.

bwbohl:
I'm all in for pureODD, but please let us tackle this after we come up with a stable MEI 5 release.

musicEnfanthen:
Maybe we could meanwhile prepare a list that contains all the necessary steps to move to pure odd (in MEI 6)? That could help to provide a bigger picture of the expected changes and workload.

rettinghaus:
closing due to too many conflicts

PR #992, opened 2022-09-22
Fix/934

### LINKED ISSUE #934: edit/update schematron rule in `<history>`
Since `<exhibHist>` et.al. are now modelled after `<provenance>` (#911), they are children of `<history>`.  [`<history>`](https://music-encoding.org/guidelines/dev/elements/history.html) has defined a Schematron rule to prevent the use of `<acquisition>` and `<provenance>` at the `work` and `expression` levels. Also `<exhibHist>` et.al. should not be allowed on these levels, however the Schematron rule is not updated yet.

Personally, I think we should first discuss whether `<exhibHist>` et.al. should be allowed at the `manifestation` level in general or whether it would not make more sense to allow them at this level only in the case of encoding a `manifestation singleton`. However, since there is a general lack of information/descriptions for `manifestation singleton` (#906), I am not sure what such a rule might look like or whether there are considerations against such a restriction.

bwbohl:
As `<exhibHist>`, `<treatHist>` and `<treatSched>` have been remodelled after `<provenance>` in [#911](https://github.com/music-encoding/music-encoding/issues/911), the corresponding Schematron rule to report an error in work and expression contexts had to be updated. Moreover, in [#934](https://github.com/music-encoding/music-encoding/issues/934), we agreed also to apply this rule to the manifestation context if the manifestation is not a singleton.
This PR applies the necessary changes.


rettinghaus (dismissed):
Looks good to me, thanks!

source/modules/MEI.header.xml
```diff
@@ -1417,9 +1417,9 @@
     </content>
     <constraintSpec ident="history_restriction" scheme="isoschematron">
       <constraint>
-        <sch:rule context="mei:history[parent::mei:work or parent::mei:expression]">
-          <sch:assert test="not(mei:acquisition or mei:provenance)">The acquisition and provenance
-            elements are not permitted in the work or expression context.</sch:assert>
+        <sch:rule context="mei:history[parent::mei:work or parent::mei:expression or parent::mei:manifestation[not(@singleton='true')]]">
+          <sch:assert test="not(mei:acquisition or mei:provenance or mei:exhibHist or mei:treatHist or mei:treatSched)">The acquisition, provenance and exhibHist
+            elements are not permitted in the work context or expression context, if expression is not singleton.</sch:assert>
```
doerners:
```suggestion
          <sch:assert test="not(mei:acquisition or mei:provenance or mei:exhibHist or mei:treatHist or mei:treatSched)">The elements acquisition, provenance, exhibHist, treatHist and treatSched are not permitted at the work or expression level and are only permitted at the manifestation level, if the manifestation is a manifestation singleton.</sch:assert>
```

doerners (changes requested):
The schematron rule looks find to me, thanks @bwbohl! However, I noticed two things I think should be changed within the description:

1. Not all elements affected by the schematron rule are listed within the description. I think it would be good to list all of them.
2. My impression was that expression and manifestation were mixed up in the second half of the description – if I'm not totally mistaken there is no singleton expression.

The wording I provided is just a suggestion and I'm not set that it has to stay like this :)

bwbohl:
thanks for the valuable contribution @doerners ! Merged it, If you would be so kind as to update your review ;-)

doerners (approved):
Thanks @bwbohl looks great!

rettinghaus (approved):
Even better!

Issue #968, opened 2022-05-24
Request for making several elements valid in mei-neume dev-5.0.0 schema


yinanazhou:
As shown in the figure below, the underlined elements are not valid in the mei-neume dev-5.0.0 schema. The details and examples are as follows. All the examples are from Einsiedeln (CH-E_611).

![mei-Neumes   dev copy](https://user-images.githubusercontent.com/74484576/169932914-70b01afa-afa1-47a3-b293-ed637cea5ed8.png)

1. `<accid>` is not valid both inside and outside a syllable, but it happens both inside and outside.

inside:
<img width="970" alt="018 flat inside 1 - folio" src="https://user-images.githubusercontent.com/74484576/169933659-512e74d9-fb95-432d-8236-2bd1036f163e.png">
<img width="856" alt="018r flat inside 1 - mei" src="https://user-images.githubusercontent.com/74484576/169934080-349709e2-39d9-417b-a2ed-a651a4df5d85.png">

<img width="960" alt="018r flat inside 2 - folio" src="https://user-images.githubusercontent.com/74484576/169933974-26541b54-9168-4d1d-978d-d40c1180975a.png">
<img width="883" alt="018r flat inside 2 - mei" src="https://user-images.githubusercontent.com/74484576/169934107-cc07609c-8cf0-41c2-8f37-8ab017374e46.png">

outside:
<img width="1010" alt="260r flat outside - folio" src="https://user-images.githubusercontent.com/74484576/169933998-7ca3c1b2-6bbd-4c73-ab48-8c9a08c0d0e6.png">
<img width="865" alt="260r flat outside - mei" src="https://user-images.githubusercontent.com/74484576/169934137-dec55c06-0d07-4059-be67-0875e1284cca.png">

<img width="850" alt="261v flat outside - folio" src="https://user-images.githubusercontent.com/74484576/169934029-7ef5bc32-faa6-4893-be64-4ecf1f9f0742.png">
<img width="833" alt="261v flat outside - mei" src="https://user-images.githubusercontent.com/74484576/169934157-2c75eaa8-30d0-43f1-9d02-fbaa55046e41.png">

2. `<divLine>` is only valid inside a syllable, but it is mostly outside.

<img width="960" alt="044r divisio outside - folio" src="https://user-images.githubusercontent.com/74484576/169934987-0c12a4a7-d0fa-474a-b312-eeac6d420388.png">
[...]

PR #1000, opened 2022-10-02
Allow divLine inside and outside syllable

rettinghaus:
Another piece to the puzzle #968

### DIFF
--- a/customizations/mei-Neumes.xml
+++ b/customizations/mei-Neumes.xml
@@ -23,11 +23,11 @@
   <teiHeader>
     <fileDesc>
       <titleStmt>
-        <title>Music Encoding Initiative Guidelines: <title type="sub">Neumes
-          Customization</title></title>
+        <title>Music Encoding Initiative Guidelines: <title type="sub">Neumes Customization</title></title>
         <respStmt>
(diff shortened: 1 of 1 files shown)


kepper:
trying to work out the difference between this and #1019, I tried to validate according to the requirements specified in https://github.com/music-encoding/music-encoding/issues/968#issuecomment-1146583421. Compared to that, this solution doesn't support `<accid>` neither inside nor outside of `<syllable>` (which means inside `<layer>`). I prefer the simplicity of this solution, but it doesn't go the whole way, and so I think we should go with #1019 instead of this one for now, insisting on the proposal in the comment over there: 

> we encourage the IG to schedule a meeting to jointly clean up this part of MEI

bwbohl:
@martha-thomae and @fujinaga in the light of #1019 having been merged, is there anything left here or can this be closed?

fujinaga:
> @martha-thomae and @fujinaga in the light of #1019 having been merged, is there anything left here or can this be closed?

I think it can be closed.

musicEnfanthen:
@rettinghaus Also ok with closing this?

musicEnfanthen:
Thank you @rettinghaus

PR #1002, opened 2022-10-02
Unify spelling

rettinghaus:
This PR supersedes #899 by just Americanize spelling of `modeled`

### DIFF
--- a/source/docs/02-shared.xml
+++ b/source/docs/02-shared.xml
@@ -305,7 +305,7 @@
                         <specDesc key="att.nNumberLike" atts="n"/>
                      </specList>
                   </p>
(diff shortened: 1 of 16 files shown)


doerners (approved):
Looks good to me, thanks @rettinghaus

kepper (commented):
I think we shouldn't necessarily get into such language discussions. Personally, I have a certain preference for `modelled` to look more nicely, but I see the benefit of using one lingo consistently, which probably should be american english. But, following arguments of my co: @music-encoding/guidelines-editors-in-chief, it really doesn't matter a whole lot. I'd like to hear @ahankinson on that again…

ahankinson:
I want the bikeshed to be yellow.

(seriously, though, the consistency is fine. I still think the Americans are wrong, but that ship sailed into Boston harbour a long time ago.)

kepper:
@pe-ro and I suggest then to use different wording here: 

> This model of this element is based on the respective element of the TEI, [link here].

That way, we can just go around the discussion. Essentially, we all seem to agree that bikeshed colo[u]rs don't matter a whole lot… May I ask @doerners to implement accordingly?

ahankinson:
I think this is the wrong PR @kepper :)

musicEnfanthen:
Superseded by #1027

PR #1003, opened 2022-10-02
Add value pedline

### LINKED ISSUE #644: pedal line vs. pedstar
In `<pedal>` we can use `form="pedstar"` or `form="line"` to encode the first and the last case of the following example. 
grafik

To encode the middle one (which is quite common in modern notation) I recommend to add the value `"mixed"` to the value list.

rettinghaus:
closes #644

Maybe it would be better to extract these values into a new datatype?

### DIFF
--- a/source/modules/MEI.cmn.xml
+++ b/source/modules/MEI.cmn.xml
@@ -991,25 +991,9 @@
       rendition.</desc>
     <attList>
       <attDef ident="pedal.style" usage="opt">
-        <desc xml:lang="en">Determines whether piano pedal marks should be rendered as lines or as terms.</desc>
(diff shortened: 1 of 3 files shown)


ahankinson:
Since it's used in two different places, I think a new datatype would make sense. 

I don't think the added comma is necessary. There are only two components to the "and" so the serial comma doesn't apply.

ahankinson:
Can this wait until after #991? Then we don't have to change the `macroSpec`/`dataSpec` twice...

lpugin (approved):

kepper (approved):
looks good, thanks

Issue #969, opened 2022-05-25
Make barline with serpent available


rettinghaus:
Currently it not easy/possible to encode a `segno` serpent. 

![](https://smufl-browser.edirom.de/E04B.png)

How about adding a `dblsegno` value to `data.BARRENDITION`?

musicEnfanthen:
Thank you @rettinghaus .

Short question: Is the segno serpent always used with a double barline? There seems to be another standalone version in SMuFL: https://w3c.github.io/smufl/latest/tables/repeats.html

musicEnfanthen:
Just for the record, some discussion about that sign in the Steinberg Forum (with links to lilypond approach): https://forums.steinberg.net/t/serpent-segno/127943

lpugin:
We can add `segno` and `dblsegno` to match SMuFL.

rettinghaus:
Solved with #1004

Issue #787, opened 2021-04-15
Thick double bar lines


rettinghaus:
Currently it is not possible to encode a double bar line consisting of two thick lines which can be found in several older prints.
![image](https://user-images.githubusercontent.com/7693447/114868730-88f1dd80-9df6-11eb-8f88-fb84fac37bdf.png)
Example from: _Leroy, Eugène: Méthode de Trompette_

Could we add another value to `data.BARRENDITION`?

kepper:
sure. quite apparently, it's necessary :-)

craigsapp:
A variant of this style would be for repeats:

https://github.com/craigsapp/haydn-piano-sonatas/blob/master/reference-edition/sonata12-2.pdf

<img width="275" alt="Screen Shot 2021-04-15 at 12 15 34 PM" src="https://user-images.githubusercontent.com/3487289/114925807-69ea5000-9de4-11eb-9b3f-80b98dedcac5.png">

rettinghaus:
Also in older prints repeat bar lines usually have four dots (on each side) instead of two. 
![image](https://user-images.githubusercontent.com/7693447/140763866-aef0b14c-72f1-4f3b-a44e-bd6fbf2d591e.png)
(Beethoven's Bagatelles Op. 119)

rettinghaus:
Obviously there are so many variations that I don't see the point in trying to catch every possibility in a hard-coded list. 
Would there be a way to encode these notation styles in general in MEI?

kepper:
Styling needs to be addressed in the future anyway. Semantically, this is still a double barline, so this can be encoded in some way already. We will come back to this later in a different context.

Issue #520, opened 2018-05-12
coda and segno encoding


craigsapp:
How are coda and segno encoded?  They should use `<dir>` as mentioned in the guidelines:

http://music-encoding.org/docs/MEI_Guidelines_v3.0.0.pdf

But there is not an example, which would better than me using telepathy to figure it out :-)

Example to fill in:

<img width="680" alt="screen shot 2018-05-11 at 10 45 17 pm" src="https://user-images.githubusercontent.com/3487289/39954046-cc06f48c-556c-11e8-8435-a28233baea45.png">

MEI data without coda or segno:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="4.0.0">
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title />
            </titleStmt>
            <pubStmt />
        </fileDesc>
        <encodingDesc>
            <appInfo>
                <application isodate="2018-05-11T22:41:13" version="2.0.0-dev-509c169-dirty">
                    <name>Verovio</name>
                    <p>Transcoded from Humdrum</p>
                </application>
            </appInfo>
        </encodingDesc>
        <workDesc>
            <work>
                <titleStmt>
                    <title />
                </titleStmt>
            </work>
        </workDesc>
    </meiHead>
    <music>
        <body>
            <mdiv xml:id="mdiv-0000000775059617">
                <score xml:id="score-0000001883574307">
                    <scoreDef xml:id="scoredef-0000000355467483" midi.bpm="400">
                        <staffGrp xml:id="staffgrp-0000001812563648">
                            <staffDef xml:id="staffdef-0000001748516128" clef.shape="G" clef.line="2" key.sig="0" meter.count="4" meter.unit="4" n="1" lines="5">
                                <label xml:id="label-0000001930042499" />
                            </staffDef>
                        </staffGrp>
                    </scoreDef>
                    <section xml:id="section-0000000562905122">
                        <measure xml:id="measure-L5" n="1">
                            <staff xml:id="staff-L5F1N1" n="1">
[...]

PR #1004, opened 2022-10-02
Develop barlines

lpugin:
This PR add data types to bar lines
* segno and dblsegno (#969)
* heavy and dblheavy (#787)

It also add the corresponding SMuFL codes whenever appropriate

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -364,13 +364,13 @@
     <content>
       <valList type="closed">
         <valItem ident="dashed">
-          <desc xml:lang="en">Dashed line (Unicode 1D104).</desc>
+          <desc xml:lang="en">Dashed line (SMuFL E036 and Unicode 1D104).</desc>
         </valItem>
         <valItem ident="dotted">
-          <desc xml:lang="en">Dotted line.</desc>
+          <desc xml:lang="en">Dotted line (SMuFL E037).</desc>
         </valItem>
         <valItem ident="dbl">
-          <desc xml:lang="en">(Unicode 1D101).</desc>
+          <desc xml:lang="en">Double barline (SMuFL E031 and Unicode 1D101).</desc>
         </valItem>
         <valItem ident="dbldashed">
           <desc xml:lang="en">Double dashed line.</desc>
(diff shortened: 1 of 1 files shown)


source/modules/MEI.xml
```diff
@@ -364,37 +364,49 @@
...
         <valItem ident="rptboth">
-          <desc xml:lang="en">Repeat start and end.</desc>
+          <desc xml:lang="en">Repeat start and end (SMuFL E042).</desc>
         </valItem>
         <valItem ident="rptend">
-          <desc xml:lang="en">Repeat end (Unicode 1D107).</desc>
+          <desc xml:lang="en">Repeat end (SMuFL E041 and Unicode 1D107).</desc>
+        </valItem>
+        <valItem ident="segno">
+          <desc xml:lang="en">Segno serpent (SMuFL E04A).</desc>
```
rettinghaus: I think we only want U+E04B as double barline with segno glyph.

lpugin: When discussing it it seemed appropriate also to include the segno serpent that @musicEnfanthen pointed us to.

kepper: but the E04B is not included now, just E04A (see https://www.smufl.org/version/latest/range/repeats/)? The last comment ("_also_") seems to contradict the code…

rettinghaus: E04A is just the serpent symbol without a barline, so why should we encode this as barline? 
I think the approach of SMuFL for this is to have it available for combination with barlines drawn with primitives, so e.g., in Verovio we would use E04A to put it on top of the drawn barlines. But for referencing the look in the schema we should only use E04B.

rettinghaus: @kepper and E04B _is_ in there with "dblsegno".

ahankinson: From what I can tell, the serpent segno with and without double lines is used as an alternative to the barline. So it would make sense to have both. (This doesn't actually preclude *also* having serpent segnos as symbols on their own if they *don't* appear as a barline, too.)

rettinghaus: No, the one without lines is not an alternative to the barline. Though it can appear between notes on a staff you would not encode this as a barline. This should be encoded as a <segno> (see #520), which may appear on or above a staff.

lpugin: I think it is good to have `segno` just in case someone  encounters a single barline with a segno. I removed the references to SMuFL because that can be misleading.

kepper (approved):
looks good to me

ahankinson (approved):

ahankinson:
I'll wait for @musicEnfanthen to review and merge, since his review was requested, but it looks good to me.

rettinghaus (approved):

musicEnfanthen (approved):
You've got four yes'es 😉

PR #1006, opened 2022-10-02
Add `pedline` value to att.pedal.vis@form

### LINKED ISSUE #644: pedal line vs. pedstar

lpugin:
The value is for pedal represented with "Ped." and a line as requested in #644. Closes #644.


rettinghaus:
Duplicates  #1003

Issue #572, opened 2019-01-06
Add <pgFootLast>


pe-ro:
Currently, MEI can deal with a page header and footer on the first page, and on the following pages.  However, sometimes a page footer on the last page is unique; that is, different from those on the preceding pages.  Adding `<pgFootLast>` would make it possible to represent this situation.

For example:
```
<scoreDef>
  <pgHead>Top of first page</pgHead>
  <pgHead2>Top of subsequent pages</pgHead2>
  <pgFoot>Foot of first page</pgFoot>
  <pgFoot2>Foot of subsequent pages</pgFoot2>
  <pgFootLast>Foot of last page</pgFootLast>
</scoreDef>
```
could be used for a score with a unique header on the first and last pages and repeating headers and footers on the intervening pages.  In this case, `<pgFoot2>` should be replaced by `<pgFootLast>` on the last page.

The following example captures unique header on page 1 and unique footer on the last page:
```
<scoreDef>
  <pgHead>Top of first page</pgHead>
  <pgFootLast>Foot of last page</pgFootLast>
</scoreDef>
```
`<pgHeadLast>` could also be added, but I haven't yet seen an real-life examples of this.

ahankinson:
`<pgFoot type="last">`? Or something along those lines?

pe-ro:
Nope.  That assigns special meaning to a particular value of `@type`, which is a no-no.

What's your objection to `<pgFootLast>`?

ahankinson:
The "something along those lines" meant "maybe not type, but something else?"

I tend to think of elements as "things" and attributes as "and this what makes this thing different than the others." It's still a page footer -- it's just a specific type of page footer.

pe-ro:
That works -- sometimes.

But at what point does a "thing of a certain type" become a new thing?  Horse-drawn carts and jet airplanes are both types of conveyance, but are vastly different in practical terms.  In this case, a pattern has already been established -- we have `<pgHead>` for a header on the first page, `<pgHead2>` for headers on subsequent pages, `<pgFoot>` for footers on the first page, and `<pgFoot2>` for footers on pages other than the first.  In other words, different elements for different "things".  Adding a new `<pgFootLast>` seems appropriate to me.  This relieves processors from the burden of dealing with a mix of controlled and uncontrolled values of `@type`.

[...]

PR #1008, opened 2022-10-02
Develop header footer

lpugin:
Following the discussion in #572, the PR removes `pgHead2` and `pgFoot2`. They are replaced with a `@func` attribute on `pgHead` and `pgFoot`. The possible values are:
* all
* first
* last
* alt1
* alt2

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -7547,13 +7547,7 @@
         <rng:ref name="pgHead"/>
       </rng:optional>
       <rng:optional>
-        <rng:ref name="pgHead2"/>
-      </rng:optional>
-      <rng:optional>
         <rng:ref name="pgFoot"/>
-      </rng:optional>
-      <rng:optional>
-        <rng:ref name="pgFoot2"/>
       </rng:optional>
       <rng:optional>
         <rng:ref name="instrGrp"/>
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -6897,7 +6897,7 @@
       <attDef ident="func" usage="opt">
         <desc xml:lang="en">Records the function (i.e., placement) of the page footer.</desc>
         <datatype>
-          <rng:ref name="data.PGFORMFUNC"/>
+          <rng:ref name="data.PGFUNC"/>
         </datatype>
       </attDef>
     </attList>
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -6902,51 +6902,13 @@
       </attDef>
     </attList>
     <remarks xml:lang="en">
-      <p>This element is used to capture the textual data that often appears on the first page of
+      <p>This element is used to capture the textual data that often appears in
         printed music. It may also be used for similarly formatted material in manuscripts. When
         used within <gi scheme="MEI">pb</gi>, it records a temporary suspension of the pattern of
         page footers established by the use of <gi scheme="MEI">pgFoot</gi> within a previous <gi
         scheme="MEI">scoreDef</gi>. Auto-generated page numbers may be indicated with a processing
-        instruction. The pgHead, pgHead2, pgFoot, and pgFoot2 elements should <hi rend="bold"
+        instruction. The pgHead and pgFoot elements should <hi rend="bold"
         >*not*</hi> be used to encode textual notes/annotations.</p>
-    </remarks>
-  </elementSpec>
-  <elementSpec ident="pgFoot2" module="MEI.shared">
-    <gloss versionDate="2022-05-18" xml:lang="en">page footer 2</gloss>
-    <desc xml:lang="en">A running footer on the pages following the first.</desc>
-    <classes>
-      <memberOf key="att.common"/>
-      <memberOf key="att.facsimile"/>
-      <memberOf key="att.lang"/>
-    </classes>
-    <content>
-      <rng:zeroOrMore>
-        <rng:choice>
-          <rng:text/>
-          <rng:ref name="model.textComponentLike"/>
-          <rng:ref name="model.textPhraseLike.limited"/>
-          <rng:ref name="model.editLike"/>
-          <rng:ref name="model.transcriptionLike"/>
-          <rng:ref name="model.appLike"/>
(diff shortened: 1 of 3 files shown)


source/modules/MEI.xml
```diff
@@ -3071,6 +3071,31 @@
       </rng:data>
     </content>
   </macroSpec>
+  <macroSpec ident="data.PGFORMFUNC" module="MEI" type="dt">
```
kepper: Could you please explain why this is named `data.PGFORMFUNC`? Specifically, why have the "FORM" part in there? If it's necessary, it should also be part of the `<desc>`, I guess?!

kepper: having spoken to you, this is about the function of the page form. The wording seems a bit poor, and is under revision right now…

kepper (dismissed):
thanks for the change

source/modules/MEI.xml
lpugin: I renamed the data type to PGFUNC

musicEnfanthen (commented):
Thanks for the fix. Do we have any place in the Guidelines (text) where the old attributes are mentioned except the listing of attributes in 02-shared.xml?

Like here: https://github.com/music-encoding/music-encoding/blob/304e9f2dda61ad3fd3fb3c442ab94bc8c2d98b00/source/modules/MEI.shared.xml#L7550 and https://github.com/music-encoding/music-encoding/blob/304e9f2dda61ad3fd3fb3c442ab94bc8c2d98b00/source/modules/MEI.shared.xml#L7556

source/modules/MEI.shared.xml
```diff
@@ -6973,55 +6939,23 @@
           <rng:ref name="data.HORIZONTALALIGNMENT"/>
         </datatype>
       </attDef>
+      <attDef ident="func" usage="opt">
+        <desc xml:lang="en">Records the function (i.e., placement) of the page header.</desc>
+        <datatype>
+          <rng:ref name="data.PGFORMFUNC"/>
```
musicEnfanthen: Correct? Or data.PGFUNC?

source/modules/MEI.xml
```diff
@@ -3071,6 +3071,31 @@
...
+        <valItem ident="alt1">
+          <desc xml:lang="en">The first of an alternating pattern of headers or footers.</desc>
+        </valItem>
+        <valItem ident="alt2">
+          <desc xml:lang="en">The second of an alternating pattern of headers or footers.</desc>
+        </valItem>
+      </valList>
+    </content>
+    <remarks xml:lang="en">
+      <p>An alternating pattern with "alt1" and "alt2" starts from the first page. However, if header of footer with a func="first" is also defined, it will shift the pattern by one page. A header or footer with func="last" will interupt the pattern accordingly.</p>
```
```suggestion
      <p>An alternating pattern with "alt1" and "alt2" starts from the first page. However, if header or footer with a func="first" is also defined, it will shift the pattern by one page. A header or footer with func="last" will interupt the pattern accordingly.</p>
```

source/modules/MEI.xml
```diff
@@ -3071,6 +3071,31 @@
       </rng:data>
     </content>
   </macroSpec>
+  <macroSpec ident="data.PGFUNC" module="MEI" type="dt">
+    <desc xml:lang="en">Page header and footer function; a value that defines the function (i.e., the placement) of the header or the footer.</desc>
+    <content>
+      <valList type="closed">
+        <valItem ident="all">
+          <desc xml:lang="en">Header or footer for all the pages, including first and last unless provided.</desc>
```
```suggestion
          <desc xml:lang="en">Header or footer for all pages, including the first and last page, unless provided.</desc>
```

musicEnfanthen: "unless provided" is unclear to me. In the sense of "unless declared otherwise"? Meaning that an additional declaration of "first" or "last" overrides the value for "all" in the first and last page?

musicEnfanthen (changes requested):

source/modules/MEI.shared.xml
lpugin: We went back and forth with different ideas. `PGFUNC` might seem clearer but can also be misleading because it is not about defining the function of the page.

pe-ro (changes requested):
I would remove the unnecessary word "accordingly" from the description.

source/modules/MEI.xml
lpugin: `Header or footer for all pages, including the first and the last page, unless a page header or footer for the first or the last page is provided.`
Does it work for you?

musicEnfanthen: Yes, much clearer for me, thanks!

source/modules/MEI.shared.xml
lpugin: Sorry, I am now understanding the question. I will fix it.

musicEnfanthen: Sorry, should have been clearer in my question that the newly added datatype PGFUNC was addressed as PGFORMFUNC here. Thanks for the fix.

musicEnfanthen (approved):

source/modules/MEI.xml
```diff
@@ -3071,6 +3071,31 @@
...
+        <valItem ident="alt1">
+          <desc xml:lang="en">The first of an alternating pattern of headers or footers.</desc>
+        </valItem>
+        <valItem ident="alt2">
+          <desc xml:lang="en">The second of an alternating pattern of headers or footers.</desc>
+        </valItem>
+      </valList>
+    </content>
+    <remarks xml:lang="en">
+      <p>An alternating pattern with "alt1" and "alt2" starts from the first page. However, if header or footer with a func="first" is also defined, it will shift the pattern by one page. A header or footer with func="last" will interupt the pattern accordingly.</p>
```
bwbohl: requested change by @pe-ro 
```suggestion
      <p>An alternating pattern with "alt1" and "alt2" starts from the first page. However, if header or footer with a func="first" is also defined, it will shift the pattern by one page. A header or footer with func="last" will interupt the pattern.</p>
```

pe-ro (approved):
Looks good.

musicEnfanthen:
Thanks for the reviews.

@lpugin That one is still good to go also from your side?

PR #1010, opened 2022-10-02
improve consistency for keysig naming

### LINKED ISSUE #525: Harmonize key.sig et al. attributes on scoreDef, staffDef, key and keySig

kepper:
AH, Edit for clarity: This PR refers to #525

### DIFF
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -779,11 +779,16 @@
     <desc xml:lang="en">Visual domain attributes.</desc>
     <classes>
       <memberOf key="att.color"/>
-      <memberOf key="att.visibility"/>
     </classes>
     <attList>
       <attDef ident="sig.showchange" usage="opt">
         <desc xml:lang="en">Determines whether cautionary accidentals should be displayed at a key change.</desc>
+        <datatype>
+          <rng:ref name="data.BOOLEAN"/>
+        </datatype>
+      </attDef>
+      <attDef ident="sig.show" usage="opt">
+        <desc xml:lang="en">Determines whether key signature is visible.</desc>
         <datatype>
           <rng:ref name="data.BOOLEAN"/>
         </datatype>
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -1328,7 +1328,7 @@
     <desc xml:lang="en">Used by staffDef and scoreDef to provide default values for attributes in the logical
       domain that are related to key signatures.</desc>
     <attList>
-      <attDef ident="key.sig" usage="opt">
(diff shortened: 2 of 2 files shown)


ahankinson:
Is there an issue or a conversation or something justifying this change?

source/modules/MEI.visual.xml
```diff
@@ -788,6 +787,12 @@
           <rng:ref name="data.BOOLEAN"/>
         </datatype>
       </attDef>
+      <attDef ident="sig.show" usage="opt">
```
ahankinson, line 790: I made a couple comments before here, but I'm not sure where they went... so apologies if you are getting spammed!

Looking at this, I understand that `sig.show` is supposed to correspond with `keysig.show`. But how does `@sig.show` align with `@visible` which is already available on `<keySig>`?

ahankinson: Ah, after seeing the issue mentioned in the commit message -- I'm not sure I would agree with `@sig.show`... we tend to use the word "visible", `@visible` in MEI, and AFAIK this is the only place where we're using "show". 

So perhaps, instead, it should be `keysig.visible` instead of `keysig.show`?

rettinghaus: `show` is referring to the naturals appearing in key changes. So you can say that the key change shows cancelling naturals or actually doesn't show them.

Compare MusicXML's [`cancel element`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/cancel/).

ahankinson:
Update: I have edited the description to link the PR to the issue since I could never find it when looking at @kepper s original message.

@lpugin has updated the issue with a new proposal based on further discussion.

musicEnfanthen:
@ahankinson @lpugin @kepper This is superseded by #1170, isn't it?

lpugin:
Yes, I think it can be closed

PR #1011, opened 2022-10-02
guidelines: improve documentation for cue vs. small notes

### LINKED ISSUE #668: `@cue` needs clarification

### LINKED ISSUE #926: Differentiate between cue and cue-sized notes
Issue https://github.com/rism-digital/verovio/issues/2625 highlights that there are two semantically different types of notes that currently are lumped together with `note@cue="true"`.   There should be an encoding method to differentiate between these two types of visually similar note types:

* cue notes, which are small notes that should not be played by a performer.
* cue-sized notes which should be played by the performer.

One practical difference is that cue notes should have cue-size lyrics, while cue-size notes should have regular-sized lyrics.

I propose adding the attribute `<note@size>` to handle the second case.   The main units would be in percentages of the default size (probably dependent on if the note is regular, cue or grace).   For example `<note@size="80%">` would make the note 80% of its normal size.

There should be two additional symbolic sizes allowed in `<note@size>`: `cue` and `grace`.  These would in turn be controllable by some `staffDef` or `scoreDef` attribute/element that allows specifying the  size scaling for true cue and grace notes that could be shared with cue-like and grace-like notes that instead use `note@size` to control their size.

In the future, there will probably be a need for a separate size control `<note@head.size>` .  This controls the notehead size only, without adjusting the size (length) of the stem.  So consideration for how `<note@size>` and `<note@head.size>` interact should be made sooner rather than later.

--------------

And alternate possibility is to mirror how `note@grace` works: In this case there are three categories: `acc`, `unacc` and `unknown`.  For `note@cue`, there could be `cue` and `cue-like`.

kepper:
fixes #926. Instead of a separate `@size`, it is encouraged to use the existing `@fontsize`. Added a sentence in the Guidelines. Co-authored-by @riedde

### DIFF
--- a/source/docs/04-cmn.xml
+++ b/source/docs/04-cmn.xml
@@ -1129,7 +1129,9 @@
                      </figure>
                   </p>
                   <p>If the voice from which the cue notes originate is also encoded, they should refer to their sounding counterpart with the <att>corresp</att> attribute.</p>
-                  <p>Cue notes must not be confused with other small notes such as grace notes or <ref target="https://en.wikipedia.org/wiki/Fioritura">fiorituras</ref>.</p>
+                  <p>Cue notes must not be confused with other small notes such as grace notes or <ref target="https://en.wikipedia.org/wiki/Fioritura">fiorituras</ref>. 
+                     The visual size of a note can be controlled with <att>fontsize</att>, which is independent of the information carried by <att>cue</att> (though cue notes are usually 
+                     written in smaller font size).</p>
                </div>
                <div xml:id="cmnDir" type="div3">
                   <head>Directives and Rehearsal marks</head>


ahankinson:
Is it a problem that `<beam>` and `<tuplet>` doesn't have `@fontsize` but has `@cue`?

lpugin:
It seems that there is more work needed here:
* It would be better to have the attribute describing size to be call `size` and not `fontsize`. That way it can also be applied to elements elsewhere (beam, tuplet, clef, etc.)
* This means moving it out of the att.typography class into a more generic att.sizeable class
* Probably other attributes on staffDef should be refactored: `music.size` and `scale` should be replaced by a single attribute `size`

bwbohl:
this pr should also address #668

Issue #578, opened 2019-03-12
Empty beam element with sameas attribute is not valid


lpugin:
Validation fails with empty `<beam>` elements using `@sameas`
```xml
<staff xml:id="staff_45387" n="4">
	<layer xml:id="layer_45393a" n="1">
		<note xml:id="note_45399a" pname="b" oct="3" dur="8" accid.ges="f"/>
		<beam xml:id="beam_45405a">
			<note xml:id="note_45411a" pname="c" oct="4" dur="8" accid="s"/>
			<note xml:id="note_45417a" pname="d" oct="4" dur="8" />
		</beam>
		<beam xml:id="beam_45426a">
			<note xml:id="note_45432a" pname="a" oct="3" dur="8" />
			<note xml:id="note_45438a" pname="b" oct="3" dur="8" accid.ges="f"/>
		</beam>
		<beam xml:id="beam_45447a">
			<note xml:id="note_45453a" pname="e" oct="3" dur="8" accid="n"/>
			<note xml:id="note_45459a" pname="f" oct="3" dur="8" />
		</beam>
		<note xml:id="note_45468a" pname="d" oct="3" dur="8" artic="spicc"/>
	</layer>
	<layer xml:id="layer_45393b" n="2">
		<note xml:id="note_45399b" pname="b" oct="3" dur="8" accid.ges="f"/>
		<!-- beams with @sames -->
		<beam xml:id="beam_45426a" sameas="#beam_45405a"/>
                <beam xml:id="beam_45426b" sameas="#beam_45426a"/>
		<beam xml:id="beam_45447b" sameas="#beam_45447a"/>
		<note xml:id="note_45468b" pname="d" oct="2" dur="8" artic="spicc"/>
	</layer>
</staff>
```
![image](https://user-images.githubusercontent.com/689412/54200106-d2aac400-44ca-11e9-8fe3-0ce84807fd51.png)

craigsapp:
The main problem is a duplicated ID:

<img width="1073" alt="Screen Shot 2019-03-12 at 10 23 53 AM" src="https://user-images.githubusercontent.com/3487289/54207559-bd28a080-44b0-11e9-9fff-124dd2193a53.png">

The id for the first sameas beam should be  `beam_45426b`.

lpugin:
Here is a complete and shorter example:
```xml
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #1012, opened 2022-10-02
Relax schematron on beams

kepper:


rettinghaus (approved):
Makes sense

lpugin (dismissed):

source/modules/MEI.cmn.xml
```diff
@@ -1346,10 +1346,10 @@
     </content>
     <constraintSpec ident="When_not_copyof_beam_content" scheme="isoschematron">
       <constraint>
-        <sch:rule context="mei:beam[not(@copyof)]">
+        <sch:rule context="mei:beam[not(@copyof) and not(@sameas)]">
```
musicEnfanthen: is it "AND" or "OR" ?

musicEnfanthen: SInce the text says "or"

musicEnfanthen: true AND true = true; 
true AND false = false; (rule does not apply when one attribute is not there but the other is)
true AND true = false;

ture OR true = true;
true OR false = true (rule applies when only one attribute is not there)
false OR false = false

Or do I mess up and and and and and and or?

ahankinson: Wow. You win the gold star for that last sentence. Perfect. ⭐

pe-ro: Good catch.  I think it's "or".

musicEnfanthen (changes requested):
~~Can be merged after fixing AND to OR~~ EDIT: See comments above

source/modules/MEI.cmn.xml
bwbohl: No, it definitely has to be `and` because what we want to express in words is:
If a beam has neither `@copyof` nor `@sameas` it has to have at least two descendants.
With `or` the expression would evaluate true if just one of them was missing.

Or am I misinterpreting the intended model?

musicEnfanthen: Maybe... or maybe I was confused by the `or` in the textual description: `A beam without a copyof or sameas attribute...`

So as it stands (with `and`), if it has no `@copyof` AND no `@sameas` attribute, the rule is applied.
If it has a `@copyof` but no `@sameas` attribute, the rule is NOT applied.
If it has a `@sameas` but no `@copyof` attribute, the rule is NOT applied.

Is that correct and the intended outcome? Then maybe clarify the textual description:
`A beam that contains neither a copyof nor sameas attribute must have ....`

musicEnfanthen:
Fixes #578

source/modules/MEI.cmn.xml
rettinghaus:
```suggestion
        <sch:rule context="mei:beam[not(@copyof or @sameas)]">
```

rettinghaus: I think this was meant.

bwbohl: Please be aware that this XPath pattern is setting the context for the following `sch:assert`. This is to say only nodes matching this pattern will be tested against the XPath expression in `sch:assert`, being: 
```xpath 
count(descendant::*[local-name()='note' or local-name()='rest' or local-name()='chord' or local-name()='space']) &gt; 1
```
We only have to test for the presence of at least two child nodes if both attributes (`@copyof` and `@sameas`) are missing.

The following table illustrates the results of the two different XPath patterns in `sch:rule/@context`.

|    | mei:beam[not(@copyof) _and_ not(@sameas)]  | mei:beam[not(@copyof) _or_ not(@sameas)]  | mei:beam[not(@copyof or @sameas)] |
|:--------|:-------|:-------|:-------|
| `<beam/>` |  match   | match   | match |
| `<beam copyof=""/>`   | no match   | match  | no match |
| `<beam sameas=""/>`   | no match  | match   | no match |
| `<beam copyof="" sameas=""/>`   | no match   |  no match   | no match |

If this table holds `match` in a cell that means, the element will be tested, i.e. child `note`, `rest`, `chord`, and `space` will be counted. An error will be reported if this number is less than `2`.

The big difference in the XPath patterns is the two rows in the middle. While _AND_ would not count their child elements, _OR_ will, because one of the attributes is missing and thus the pattern matches those `beam`s.

rettinghaus: Sorry. I looked only at the bit that was changed.

rettinghaus (dismissed):

source/modules/MEI.cmn.xml
```diff
@@ -1346,10 +1346,10 @@
...
           <sch:assert
             test="count(descendant::*[local-name()='note' or local-name()='rest' or               local-name()='chord' or local-name()='space']) &gt; 1"
-            >A beam without a copyof attribute must have at least 2 note, rest, chord, or space
+            >A beam without a copyof or sameas attribute must have at least 2 note, rest, chord, or space
```
musicEnfanthen:
```suggestion
            >A beam that contains neither a copyof nor sameas attribute must have at least 2 note, rest, chord, or space
```

source/modules/MEI.cmn.xml
bwbohl: Same here, didn't realise until now that you changed the parenthesis … Your proposed expression `mei:beam[not(@copyof or @sameas)]`delivers the same results as `mei:beam[not(@copyof) and not(@sameas)]`

musicEnfanthen (approved):

bwbohl (approved):

PR #1013, opened 2022-10-03
Use official rng for svg from w3c

rettinghaus:
tei-c.org is down, so building the schema fails because of the hard-coded url for `svg11.rng`. 
In this PR it is replaced with the "official" rng from w3c.

### DIFF
--- a/customizations/mei-CMN.xml
+++ b/customizations/mei-CMN.xml
@@ -92,7 +92,7 @@
         <moduleRef key="MEI.visual"/>
 
         <!-- Include SVG -->
-        <moduleRef url="https://www.tei-c.org/release/xml/tei/custom/schema/relaxng/svg11.rng"
+        <moduleRef url="../source/svg11.rng"
           prefix="svg_">
           <content>
             <rng:define name="mei_model.graphicLike" combine="choice">
--- a/customizations/mei-Mensural.xml
+++ b/customizations/mei-Mensural.xml
@@ -92,7 +92,7 @@
         <moduleRef key="MEI.visual"/>
 
         <!-- Include SVG -->
-        <moduleRef url="https://www.tei-c.org/release/xml/tei/custom/schema/relaxng/svg11.rng"
+        <moduleRef url="../source/svg11.rng"
           prefix="svg_">
           <content>
             <rng:define name="mei_model.graphicLike" combine="choice">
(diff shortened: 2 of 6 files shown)


ahankinson:
I'm very much +1 on this change. Have you verified this works? 

Because the TEI page is down I can't see the differences. Browsing around the TEI github it looks like they might have this one:

https://raw.githubusercontent.com/TEIC/Stylesheets/4ee458aedd5b6cb62d4882d1ae5ac2f0dfacbcb8/profiles/iso/schema/svg11.rng

The W3C one uses modules, while the TEI one seems to have things inline. Shouldn't be a problem but we should make sure it doesn't break anything.

rettinghaus:
Unfortunately the build process doesn't resolve the linked modules correctly, resulting in non-functioning schemas. I'm not sure what the best solution is here. Should we just store the svg rng in our own repo?

ahankinson:
Is it possible to use the DTD?

kepper:
I took the rng from https://web.archive.org/web/20220320030131/https://tei-c.org/release/xml/tei/custom/schema/relaxng/svg11.rng locally, which wouldn't break things. But yes, we should have it locally…

musicEnfanthen:
So this needs to be reworked, right?

rettinghaus:
@kepper where should it go? Into `customizations` or `source`?

ahankinson:
I think we should just host it on music-encoding.org, and not in the core schema. Or are there technical reasons to host it in the schema?

kepper:
I would keep it locally in the core schema repo (preferably in source), so that you have everything in place to build the schemata from there, without external dependencies. Next time, it might be a Paderborn server that goes down and stops us from building, so having it in the same place seems safest.

ahankinson:
Isn't the MEI website hosted on GitHub? No Paderborn servers involved. :-D

Although it might be worthwhile considering the case for building the schema offline?

rettinghaus:
Added the RNG file to the source directory.

ahankinson (approved):
I'm approving this, because I think it's the most reasonable and expedient fix. 

However, some small part of me feels like we're losing something by not appealing to an external authority, but are instead making a local copy. Totally understand the reasons, but it's not *really* how it's supposed to work...

bwbohl (approved):
completely agree with @ahankinson

ahankinson:
@bwbohl I think you can probably merge this one.

PR #1015, opened 2022-10-03
Add num.log and numbase.log to att.mensur.log

lpugin:
This PR add two attributes `num.log` and `numbase.log` to `mensur` that can be used to override the visual values given in `num` and `numbase`. This relates to the discussion in #686 and with @pe-ro, @kepper and @fujinaga

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -1872,6 +1872,22 @@
       <memberOf key="att.duration.ratio"/>
       <memberOf key="att.mensural.shared"/>
     </classes>
+    <attList>
+      <attDef ident="num.log" usage="opt">
+        <desc xml:lang="en">Along with numbase.log, describes logical duration as a ratio. num.log is the first value in the
+          ratio, while numbase.log is the second. The logical duration can be provided when it differs from the visual duration as given in num.</desc>
+        <datatype>
+          <rng:data type="positiveInteger"/>
+        </datatype>
+      </attDef>
+      <attDef ident="numbase.log" usage="opt">
+        <desc xml:lang="en">Along with num.log, describes logical duration as a ratio. num.log is the first value in the
+          ratio, while numbase.log is the second. The logical duration can be provided when it differs from the visual duration as given in numbase.</desc>
+        <datatype>
+          <rng:data type="positiveInteger"/>
+        </datatype>
+      </attDef>
+    </attList>
   </classSpec>
   <classSpec ident="att.meterConformance" module="MEI.shared" type="atts">
     <desc xml:lang="en">Attributes that provide information about a structure’s conformance to the prevailing


ahankinson:
Typically the logical attributes do not have the `.log` suffix. I think these would be the only ones in the entire spec.

lpugin:
Yes, we are aware of this. Elsewhere, we use `.ges`, but it is not exactly what it is about. MEI puts essentially the visual domain first. This is a typical case where one need to be able to encode a logical value when it differs from the visual one. Do `.log` seems to be the best option assuming that `.ges` is not appropriate here. Do you have another idea?

ahankinson:
I think the logical domain is the primary one, but also that you only need an attribute for a given domain where it deviates from what may otherwise be expected -- in other words, all domains co-exist at the same time, but it's only when they vary that you need a way of trying to address them specifically. But that's maybe for a longer conversation.

In this case, specifically, the `num` and `numbase` are the "controls" for the mensuration sign -- they're the ones that set the machine-readable and actionable values to compute the mensuration. This change would shift that function to `num.log` and `numbase.log`, which runs counter to just about every other attribute's behaviour. So if the visual didn't change (the norm in 99% of the cases) you would always need to supply `num.log` and `numbase.log`. 

If there has to be a specific attribute for the visual identity of a symbol, then we should make it `num.vis` and `numbase.vis`. This is also unique in the codebase, but it has the benefit of making the "logical" behaviour the default un-suffixed attribute, and varying it only if needed.

lpugin:
> I think the logical domain is the primary one, but also that you only need an attribute for a given domain where it deviates from what may otherwise be expected -- in other words, all domains co-exist at the same time, but it's only when they vary that you need a way of trying to address them specifically. But that's maybe for a longer conversation.

> If there has to be a specific attribute for the visual identity of a symbol, then we should make it num.vis and numbase.vis. This is also unique in the codebase, but it has the benefit of making the "logical" behaviour the default un-suffixed attribute, and varying it only if needed.

I think we agree that visual and logical domains are very closely linked and there there is no need to make a distinction between the two in most of the cases - and we do not. In the very few cases where we need this, I personally think that MEI would put the visual first, essentially because of its background. But as you said, this is a long discussion.

lpugin:
> If there has to be a specific attribute for the visual identity of a symbol, then we should make it `num.vis` and `numbase.vis`.

I don't think we have any attribute ending with `.vis`, so these would be the only ones in the entire spec too.

ahankinson:
Yes, I acknowledged that, but also said that it has the benefit of "making the "logical" behaviour the default un-suffixed attribute, and varying it only if needed."

Since 99% of the time, `num` and `numbase` are all that is needed, and it's only the case where the visuals differ that we would then need `num.vis` and `numbase.vis`.

lpugin:
> Since 99% of the time, `num` and `numbase` are all that is needed, and it's only the case where the visuals differ that we would then need `num.vis` and `numbase.vis`.

Yes, the PR works in the same way, but just the opposite way. Since 99% of the time, `num` and `numbase` are all that is needed, and it's only the case where the logicals differ that we would then need `num.log` and `numbase.log`.

ahankinson:
There are lots of cases in MEI where a visual symbol is specified in a different attribute to display a symbol if the encoder wants a different visual appearance from the expected logical behaviour -- all the `glyph.*` attributes do this.

I don't know of any where the logical functioning shifts to a different attribute in the presence of a "visual" override.

kepper:
I can see arguments for both perspectives. In my understanding, the different domains are almost always entangled and overlap to some degree. So this comes down to what is considered the default (whatever that means), and what is the deviation from that. I think the IG should be in charge of deciding which is which (I certainly don't know enough about this repertoire to judge). Looking at the list provided in [#686](https://github.com/music-encoding/music-encoding/issues/686#issue-651120217), it seems that there is large interest in prioritizing the visual here. The (plausible) argument I see in the solution here is that the IG will get that approved, and get's an opportunity to override this visual encoding with logical in case it's not straightforward to do that. For `@accid` and `@accid.ges`, this seems to be roughly the same direction – in fact, one could argue that `@accid.ges` should really be `@accid.log` instead. But, as one can hear accidentals, but not information about meter, it was put in the gestural domain – which in this case seems to be closer entangled to the logical than the visual. Makes sense?!

ahankinson:
> I think the IG should be in charge of deciding which is which...

I don't really know how to progress this conversation (don't want to get into the weeds on a GH issue) but I pretty fundamentally disagree with this. I think we need to have a pretty clear understanding at the core level about this issue, and we shouldn't leave it to the IG to decide since it can have an impact across the entire spec.

lpugin:
OK, I'll mark it as a draft for now and we will make sure there is more discussion before it goes forward - if it does at all.

craigsapp:
How does this relate to the examples I presented in the November 2021 Mensural-IG?

http://bit.ly/mensuration-notes

Examples of mensuration signs along with schematics of their logical meanings (red numbers in the schematics indicate explicit information taking from the mensuration signs, while blue numbers indicate implicit information from the mensuration signs):



Screen Shot 2022-10-03 at 9 43 39 AM


This is a traditional mensuration sign which shifts the division levels up one durational level.   Normally the mensuration sign indicates the divisions of  breves to semibreves (tempus) and semibreves to minims (prolation).  "C3" instead indicates the divisions of longs into breves (modus or modus minor) and breves to semibreves (tempus).  In this case the "3" is not a proportion.

However there is the same visual mensuration sign:

Screen Shot 2022-10-03 at 9 44 07 AM

which can be interpreted differently as a 3:2 proportion (sesquialtera) in the C mensuration. 
A related case, where only the proportion is given (with "3" implying "3/2"):

Screen Shot 2022-10-03 at 9 53 47 AM

This mensuration sign means to apply sesquialtera to the tempus (breve to semibreve) level of the prevailing mensuration sign (such as Cut-C or C).  In this case the mensuration is treated as O, with the semibreves being played 3/2 faster than those in C (the perfect breve in quasi O mensuration having the same duration as the imperfect breve in C mensuration).

However, the sesquialtera proportion can instead alternatively be applied to the prolation (semibreve to minim) level instead of the tempus level:

Screen Shot 2022-10-03 at 10 01 04 AM

There are also "C2" and "2" mensurations which behave similar to the above cases.   And there are cases where the "3/2" or "2/3" proportion are fully spelled out, rather than giving one of those two numbers, such as "Cut-C 3/2" being equivalent to "Cut-C 3" or "3".

In other words, how are these four mensuration signs to be indicated in MEI without ambiguity in meaning?  When performing such music, "C3" and "3" ambiguities need to be resolved by the musician who has to figure out which of the two possible interpretations make sense for performance.

lpugin:
This PR does not address the problem of the tactus, which should be addressed separately. It focuses on the logical vs visual distinction. 

For example, if you have a `C3` that is **not** a proportion, you do:
```xml
<mensur tempus="2" sign="C" num="3" num.log="1" />
```

On the contrary, if the `C3` is a proportion (sesquialtera), then you do:
```xml
<mensur tempus="2" sign="C" num="3" numbase.log="2" />
```

This is not dissimilar to what MEI does with `accid` and `accid.ges`, and similarly encoders could always provide the `.log` values. (At least there is no Schematron rule that prevents this.)

craigsapp:
How would the two "3" interpretations be encoded?

craigsapp:
I am wondering about this one:

> On the contrary, if the C3 is a proportion (sesquialtera), then you do:

```xml
 <mensur tempus="2" sign="C" num="3" numbase.log="2" />
```

I am thinking you mean:

```xml
<mensur tempus="3" sign="C" num="3" numbase.log="2" />
```

Normally the tempus for "C" is `2`, but with sesquialtera it becomes "3" with an added `3/2` proportion to set the duration of the perfect breve to that of the  original `C`'s imperfect breve (the proportion meaning 3 [semibreves] in the time of 2 [semibreves]).

lpugin:
> How would the two "3" interpretations be encoded?

My understanding is that we need another attribute that indicates at which level the modification operates. I thought this could be treated separately, and that we could make a first step in MEI 5.0. I understand this was not a good idea and we better do this properly in MEI 6.0 because there is no reason to rush.

I am closing this PR because discussing fundamental questions in a PR is not appropriate.

craigsapp:
>  because there is no reason to rush.

Indeed: issues have to be aged to produce a fine vintage: https://github.com/music-encoding/music-encoding/issues/686

PR #1016, opened 2022-10-03
Add ornamentation accid gestural

### LINKED ISSUE #710: @upperaccid.ges/@loweraccid.ges
Here is an example of a turn, where the lower-note accidental is placed above the turn:

strange-turn-acc1

For visual fidelity, this accidental must be placed in `turn@accidupper`, but for gestural fidelity it should be placed in `turn@accidlower`; otherwise, rendering of this turn (such as in a MIDI conversion) will be incorrect.

Extra attributes should be added to `<turn>` to allow this case.   I think I have seen this in printed editions as well: the most common note to alter would be the lower auxiliary note, converting it into a minor second leading tone.   In this case the scribe leaves it to the performer to figure that out.

Additional attributes to distinguish between visual and gestural accidental would be useful for other corner cases.  I propose adding `@accidupper.ges` and `@accidlower.ges` which would act in a similar manner to `note@accid` and `note@accid.ges`.

So the second turn in the above example would be encoded as:

```xml
<turn staff="1" accidupper="s" accidupper.ges="n" accidlower.ges="s" form="upper" />
```

Adding `@accidupper.ges` and `@accidlower.ges` would also localize the accidentals of the auxiliary notes which would simplify conversion to MIDI, since the accidental states of the auxiliary notes would not need to be inferred from the key signature and previous accidental states of notes in the measure.  This is a non-trivial process, similar to inferring `note@accid.ges`.

This is related to issue https://github.com/humdrum-tools/verovio-humdrum-viewer/issues/425.

kepper:
fixes #710

### DIFF
--- a/source/modules/MEI.gestural.xml
+++ b/source/modules/MEI.gestural.xml
@@ -541,6 +541,23 @@
       <memberOf key="att.timestamp2.gestural"/>
     </classes>
   </classSpec>
+  <classSpec ident="att.ornamentAccid.ges" module="MEI.gestural" type="atts">
+    <desc xml:lang="en">Gestural accidentals associated with ornaments.</desc>
+    <attList>
+      <attDef ident="accidupper.ges" usage="opt">
+        <desc xml:lang="en">Records the sounding accidental associated with an upper neighboring note.</desc>
+        <datatype>
+          <rng:ref name="data.ACCIDENTAL.GESTURAL"/>
+        </datatype>
+      </attDef>
+      <attDef ident="accidlower.ges" usage="opt">
+        <desc xml:lang="en">Records the sounding accidental associated with a lower neighboring note.</desc>
+        <datatype>
+          <rng:ref name="data.ACCIDENTAL.GESTURAL"/>
+        </datatype>
+      </attDef>
+    </attList>
+  </classSpec>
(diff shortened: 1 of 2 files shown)


source/modules/MEI.gestural.xml
```diff
@@ -541,6 +541,23 @@
       <memberOf key="att.timestamp2.gestural"/>
     </classes>
   </classSpec>
+  <classSpec ident="att.ornamentAccid.ges" module="MEI.gestural" type="atts">
+    <desc xml:lang="en">Gestural accidentals associated with ornaments.</desc>
+    <attList>
+      <attDef ident="accidupper.ges" usage="opt">
```
ahankinson, line 547: Perhaps `upperaccid.ges` -- would align with `accid.ges`?

lpugin: Are you concerned about the naming? Since the plain attribute is `accidupper` then `accidupper.ges` is inline on that side. I think this is more important.
image

ahankinson: OK, sure.

ahankinson (approved):

musicEnfanthen (approved):

musicEnfanthen:
Thanks for the fix.

Issue #994, opened 2022-09-22
resolve literature references in specs


bwbohl:
Sometimes there are still literature references in the specs as xml comments in the form of, e.g., `<!-- (Read, p. 97-98). -->`

https://github.com/music-encoding/music-encoding/blob/cff88ae4e9d5b406be1358174a08d5a27c889d45/source/modules/MEI.cmn.xml#L1922

I'd rather we resolve them to proper XML encoding and and the cited resource in the metadata

PR #1018, opened 2022-10-03
Add documentation of manifestation singleton

### LINKED ISSUE #906: provide description and documentation for `manifestation/@singleton`
Neither the specs nor the guidelines seem to contain any information on `manifestation/@singleton`

riedde:
- add documentation of `manifestation/@singleton` at Specs and Guidelines
- adding `<availability>` to `<manifestation>` because it is available in `<item>` and it is needed in case of `@singleton="true"`
- restricting `<availability>` to this case

Closes #906 

Co-authored-by: @doerners 
Co-authored-by: @mss2221

### DIFF
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -1310,7 +1310,7 @@
                   </specList>
                </p>
                <p>The content of the <gi scheme="MEI">item</gi> is quite similar to <gi scheme="MEI">manifestation</gi> with a few omissions:</p>
-               <p>Please note that in case of handwritten sources the <att>singleton</att> attribute is available for the <gi scheme="MEI">manifestation</gi> element to indicate a unique physical object, precisely, manifestation singleton are sources <quote>produced as unique objects, with no siblings <hi rend="italic">intended</hi> in the course of their production.</quote> (<ref target="https://repository.ifla.org/handle/123456789/659">FRBRoo</ref>, p. 57). These cases result in <gi scheme="MEI">item</gi> and <gi scheme="MEI">manifestation</gi> being essentially the same.</p>
+               <p>Please note that in case of handwritten sources the <att>singleton</att> attribute is available for the <gi scheme="MEI">manifestation</gi> element to indicate a unique physical object, precisely, manifestation singleton are sources <quote>produced as unique objects, with no siblings intended in the course of their production.</quote> (<ref target="https://repository.ifla.org/handle/123456789/659">Definition of FRBRoo</ref>, p. 57). These cases result in <gi scheme="MEI">item</gi> and <gi scheme="MEI">manifestation</gi> being essentially the same.</p>
                <p>
                   <specList>
                      <specDesc key="head"/>
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -1310,7 +1310,7 @@
                   </specList>
                </p>
                <p>The content of the <gi scheme="MEI">item</gi> is quite similar to <gi scheme="MEI">manifestation</gi> with a few omissions:</p>
-               <p>Please note that in case of handwritten sources the <att>singleton</att> attribute (inspired by <ref target="https://repository.ifla.org/handle/123456789/659">FRBRoo</ref>) is available in <gi scheme="MEI">manifestation</gi> to indicate a unique physical object, because <gi scheme="MEI">item</gi> and <gi scheme="MEI">manifestation</gi> are essentially the same.</p>
+               <p>Please note that in case of handwritten sources the <att>singleton</att> attribute is available for the <gi scheme="MEI">manifestation</gi> element to indicate a unique physical object, precisely, manifestation singleton are sources <quote>produced as unique objects, with no siblings <hi rend="italic">intended</hi> in the course of their production.</quote> (<ref target="https://repository.ifla.org/handle/123456789/659">FRBRoo</ref>, p. 57). These cases result in <gi scheme="MEI">item</gi> and <gi scheme="MEI">manifestation</gi> being essentially the same.</p>
                <p>
                   <specList>
                      <specDesc key="head"/>
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -1310,6 +1310,7 @@
                   </specList>
                </p>
                <p>The content of the <gi scheme="MEI">item</gi> is quite similar to <gi scheme="MEI">manifestation</gi> with a few omissions:</p>
+               <p>Please note that in case of handwritten sources the <att>singleton</att> attribute (inspired by <ref target="https://repository.ifla.org/handle/123456789/659">FRBRoo</ref>) is available in <gi scheme="MEI">manifestation</gi> to indicate a unique physical object, because <gi scheme="MEI">item</gi> and <gi scheme="MEI">manifestation</gi> are essentially the same.</p>
                <p>
                   <specList>
                      <specDesc key="head"/>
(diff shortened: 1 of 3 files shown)


source/docs/03-metadata.xml
```diff
@@ -1308,6 +1310,7 @@
                   </specList>
                </p>
                <p>The content of the <gi scheme="MEI">item</gi> is quite similar to <gi scheme="MEI">manifestation</gi> with a few omissions:</p>
+               <p>Please note that in case of handwritten sources the <att>singleton</att> attribute (inspired by <ref target="https://repository.ifla.org/handle/123456789/659">FRBRoo</ref>) is available in <gi scheme="MEI">manifestation</gi> to indicate a unique physical object, because <gi scheme="MEI">item</gi> and <gi scheme="MEI">manifestation</gi> are essentially the same.</p>
```
ahankinson: I'm not sure this is technically true. 

There are cases where there is only one **surviving** print (unica). The item and the manifestation are not necessarily the same, it's just we don't have more than one copy of the item to point to.

There are also many cases where, for example, the composer has written annotations on a print. In that case it is both a print (many copies), but also a unique print due to its annotations. In this case the manifestation is not unique, but the item is.

doerners: Well, the documentation is not as precise as the scope note of Manifestation Singleton in [FRBRoo](https://www.ifla.org/wp-content/uploads/2019/05/assets/cataloguing/FRBRoo/frbroo_v_2.4.pdf), which is what inspired it in MEI, but - at least - I understand them to be conceptually the same. 
The scope note precisely says:
> This class comprises physical objects that each carry an instance of F2 Expression,
and that were produced as unique objects, with no siblings intended in the course of
their production. It should be noted that if all but one copy of a given publication
are destroyed, then that copy does not become an instance of F4 Manifestation
Singleton, because it was produced together with sibling copies, even though it now
happens to be unique. Examples of instances of F4 Manifestation Singleton include
manuscripts, preparatory sketches and the final clean draft sent by an author or a
composer to a publisher.

So I think both cases you described @ahankinson do not satisfy the "intent" criterion and are thus not examples of (true) manifestation singletons.

ahankinson: We have many cases in RISM where a library holds a very special copy of a print -- for example, one with composer handwritten annotations, or annotations from a famous conductor.

They often mis-catalogue these sources as "manuscript" since, to them, the manuscript component of these prints is what makes them worth cataloguing; not the print.

I say this only to suggest that the wording as it is, is ambiguous about this. The wording from the FRBRoo scope note is more precise, in that it specifically states **"produced as unique objects, with no siblings intended in the course of
their production."** That bit is important, I think, for understanding what a 'manifestation singleton' is. The second sentence is even clearer on the case of unica. 

I think they put these in the scope notes because it's precisely the type of confusion that would arise here. 

So I might suggest adding something similar to the MEI guidelines -- and probably also a link to the definition in FRBRoo, because that is helpful -- to ensure people understand the specific circumstances where it should be used.

Also -- I'm not sure what happens with multiple copies of the same manuscript... is it a singleton or not?

source/modules/MEI.frbr.xml
```diff
@@ -374,11 +374,21 @@
...
+      </constraint>
+    </constraintSpec>
     <attList>
       <attDef ident="singleton">
+        <desc xml:lang="en">Indicates the manifestation is a unique physical object.</desc>
         <datatype>
           <rng:ref name="data.BOOLEAN"/>
         </datatype>
+        <remarks xml:lang="en"><p>This attribute is inspired by the <ref target="https://repository.ifla.org/handle/123456789/659">FRBRoo</ref> concept of manifestation singleton.</p></remarks>
+        <remarks xml:lang="en"><p>Manifestation singleton encompasses: manuscripts, preperatory sketches and final clean drafts.</p></remarks>
```
ahankinson:
```suggestion
        <remarks xml:lang="en"><p>Manifestation singleton encompasses: manuscripts, preperatory sketches, and final clean drafts.</p></remarks>
```

source/docs/03-metadata.xml
```diff
@@ -1308,6 +1310,7 @@
                   </specList>
                </p>
                <p>The content of the <gi scheme="MEI">item</gi> is quite similar to <gi scheme="MEI">manifestation</gi> with a few omissions:</p>
+               <p>Please note that in case of handwritten sources the <att>singleton</att> attribute is available for the <gi scheme="MEI">manifestation</gi> element to indicate a unique physical object, precisely, manifestation singleton are sources <quote>produced as unique objects, with no siblings <hi rend="italic">intended</hi> in the course of their production.</quote> (<ref target="https://repository.ifla.org/handle/123456789/659">FRBRoo</ref>, p. 57). These cases result in <gi scheme="MEI">item</gi> and <gi scheme="MEI">manifestation</gi> being essentially the same.</p>
```
```suggestion
               <p>Please note that in case of handwritten sources the <att>singleton</att> attribute is available for the <gi scheme="MEI">manifestation</gi> element to indicate a unique physical object, precisely, manifestation singleton are sources <quote>produced as unique objects, with no siblings intended in the course of their production.</quote> (<ref target="https://repository.ifla.org/handle/123456789/659">Definition of FRBRoo</ref>, p. 57). These cases result in <gi scheme="MEI">item</gi> and <gi scheme="MEI">manifestation</gi> being essentially the same.</p>
```

ahankinson: The em was removed because it was not in the original statement, so not part of the direct quote. Also the full title of the PDF was provided. (Will we want to convert this to a full bibliographic reference, as per #994 ?)

doerners: Thanks @ahankinson. Yeah you're right that the italics are not part of the direct quote, however, I thought it might be good to add some emphasis because I read the "intended" as an important word in the sentence. Personally I like the idea of havin full bibliographic references but this might be better adressed in a different PR?

ahankinson: I think it doesn't lose anything without the emphasis. However, if you do want to include it you should add "[emphasis added]" to the quote to indicate that it wasn't in the original -- otherwise, you're actually misquoting since you're implying a meaning that may or may not be there in the original. See: https://writeanswers.royalroads.ca/faq/199102

A quick look suggests that doesn't seem to be any real standards for citation, so I guess we should add it later...

ahankinson (approved):

riedde:
Thank you @ahankinson for your review!

source/modules/MEI.header.xml
```diff
@@ -43,6 +43,9 @@
       <rng:optional>
         <rng:ref name="pubStmt"/>
       </rng:optional>
+      <rng:optional>
+        <rng:ref name="availability"/>
```
musicEnfanthen, line 47: You added `availability` to the `manifestation` element via `macro.bibldescPart`, right? However, in `item` availabilty is added directly (together with physDesc and physLoc). Does it make sense to use the same mechanism with `macro.bibldescPart` there too?

rettinghaus: @musicEnfanthen I think this would be a different thing. If you think that should be changed, maybe open another issue?

rettinghaus (approved):

source/modules/MEI.header.xml
riedde: I thought about this. If we would add this element directly, we would force an other order of the elements. The macro is only used by manifestation so I don't see any problems with this solution, but maybe I'm wrong.
My intention was to keep the order of the elements within manifestation and item analog against each other. Item has just some omittions.

musicEnfanthen: OK, so maybe that's soemthing to check for in a future review of this part of the specs ;)

musicEnfanthen (approved):

musicEnfanthen:
Thank you!

PR #1020, opened 2022-10-03
Develop metersig

### LINKED ISSUE #504: Clarify meterSig sym / form and numbers
See discussion in https://github.com/rism-ch/verovio/issues/800

Currently, the documentation says that `@sym` "Indicates the use of a meter symbol instead of a numeric meter signature". Howerver, in some cases displaying both can be desirable.

Also, `@form` deals with the visibility of the numbers, but not exclusively because the value "invis" indicates that the "Meter signature [is] not rendered" [entirely].

I would suggest to revise the description of `@sym` and remove the "instead of a numeric meter signature", and to add a value "symandnum" to `@form` for specifying that both the symbol and the numbers should be displayed. Then it remains up to the application to decide what to do if `@form` is not provided and both a symbol and some numbers are provided. I would expect it to display only the symbol but this is debatable and a Schematron rule could also be added.

lpugin:
Add sym+norm to data.METERFORM
* Second step for the implementation of https://github.com/music-encoding/music-encoding/issues/504#issuecomment-387104919
* A small difference with the proposal is the value, namely "sym+norm" rather than "norm+sym" because the common practice is to show the symbol before the numbers.

Add other as possible value in att.meterSigGrp.log@func
* This step of the proposal https://github.com/music-encoding/music-encoding/issues/504#issuecomment-387104919
* Instead of "unknown", the attribute value added is "other" because it seemed more appropriate

Closes #504

### DIFF
--- a/source/modules/MEI.cmn.xml
+++ b/source/modules/MEI.cmn.xml
@@ -1885,7 +1885,7 @@
   </elementSpec>
   <elementSpec ident="meterSigGrp" module="MEI.cmn">
     <gloss versionDate="2022-05-18" xml:lang="en">meter signature group</gloss>
-    <desc xml:lang="en">Used to capture alternating, interchanging, and mixed or any other non-standard meter signatures.</desc>
+    <desc xml:lang="en">Used to capture alternating, interchanging, mixed or other non-standard meter signatures.</desc>
     <classes>
       <memberOf key="att.common"/>
       <memberOf key="att.facsimile"/>
--- a/source/modules/MEI.cmn.xml
+++ b/source/modules/MEI.cmn.xml
@@ -797,6 +797,9 @@
             <desc xml:lang="en">Meter signatures with different unit values are used to express a complex metrical
               pattern that is not expressible using traditional means, such as 2/4+1/8.</desc>
           </valItem>
+          <valItem ident="other">
+            <desc xml:lang="en">Meter signatures with different unit values in an unspecified relationship.</desc>
+          </valItem>
         </valList>
       </attDef>
     </attList>
--- a/source/modules/MEI.cmn.xml
+++ b/source/modules/MEI.cmn.xml
@@ -1885,7 +1885,7 @@
   </elementSpec>
   <elementSpec ident="meterSigGrp" module="MEI.cmn">
     <gloss versionDate="2022-05-18" xml:lang="en">meter signature group</gloss>
-    <desc xml:lang="en">Used to capture alternating, interchanging, and mixed meter signatures.</desc>
+    <desc xml:lang="en">Used to capture alternating, interchanging, and mixed or any other non-standard meter signatures.</desc>
     <classes>
       <memberOf key="att.common"/>
       <memberOf key="att.facsimile"/>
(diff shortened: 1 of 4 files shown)


lpugin:
Anything problematic with this PR? Pinging @kepper and @musicEnfanthen

source/modules/MEI.xml
```diff
@@ -2094,6 +2094,9 @@
         <valItem ident="norm">
           <desc xml:lang="en">Meter signature rendered using traditional numeric values.</desc>
         </valItem>
+        <valItem ident="sym+norm">
+          <desc xml:lang="en">Show both the symbol and the traditional numeric values.</desc>
```
```suggestion
          <desc xml:lang="en">Meter signature rendered using both the symbol and the traditional numeric values.</desc>
```

source/modules/MEI.cmn.xml
```diff
@@ -797,6 +797,9 @@
             <desc xml:lang="en">Meter signatures with different unit values are used to express a complex metrical
               pattern that is not expressible using traditional means, such as 2/4+1/8.</desc>
           </valItem>
+          <valItem ident="other">
+            <desc xml:lang="en">Meter signatures with different unit values in an unspecified relationship.</desc>
```
musicEnfanthen: I am having issues to understand what that means without a look into the corresp. issue. Do you want the relationship explicitly to be "unspecified", or is it just a fallback for all cases not specified above?

musicEnfanthen (changes requested):

musicEnfanthen:
@lpugin No, just two small questions about clarification. Thanks for the reminder

musicEnfanthen:
Also, would we need to adjust the desc of meterSigGrp element? https://music-encoding.org/guidelines/dev/elements/meterSigGrp.html

 – Used to capture alternating, interchanging, **mixed or any other non-standard** meter signatures. [??]

source/modules/MEI.cmn.xml
lpugin: It is a fallback. Could you propose a better description

pe-ro (changes requested):
A modest change to the wording of the description for clarity:

Used to capture alternating, interchanging, mixed or other non-standard meter signatures.

pe-ro (dismissed):
Looks good.  Thanks.

source/modules/MEI.cmn.xml
lpugin: Sorry @musicEnfanthen, my comment was meant to be a question. Having looked at the issue, could you imagine a better description?

musicEnfanthen: No problem @lpugin. I see that it is quite tricky to express without using something like "above", etc. Maybe:

`Meter signatures with different unit values in a relationship not covered by alternating, interchanging or mixed values.` 

(But maybe @ahankinson or @pe-ro can help with a better wording?)

pe-ro: How about `Meter signatures in a relationship not covered by the values alternating, interchanging or mixed.`?

musicEnfanthen (approved):

musicEnfanthen:
@pe-ro May I ask you to re-review since Github dismisses a former review when new commits are added to a PR. Many thanks!

pe-ro (approved):

musicEnfanthen:
Thank you

PR #1021, opened 2022-10-03
Add `@opening.vertical` and `@angle.optimize` to att.hairpin.vis

### LINKED ISSUE #733: Angled hairpins
Currently it is not possible to encode (programmatically-determined) angled hairpins, because `@startvo` and `@endvo` by description record "a vertical adjustment of a feature's programmatically-determined" point. 

lily-fa2a919d

It would be good to add an attribute for encoding the rotation here, something like `@angle`?

lpugin:
Closes #733

### DIFF
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -686,7 +686,7 @@
         </datatype>
       </attDef>
       <attDef ident="angle.optimize" usage="opt">
-        <desc xml:lang="en">Indicates that the slope of the haipin can be adjusted to follow the content in order to optimize spacing.</desc>
+        <desc xml:lang="en">Indicates that the slope of the hairpin can be adjusted to follow the content in order to optimize spacing.</desc>
         <datatype>
           <rng:ref name="data.BOOLEAN"/>
         </datatype>
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -679,6 +679,18 @@
           <rng:ref name="data.BOOLEAN"/>
         </datatype>
       </attDef>
+      <attDef ident="opening.vertical" usage="opt">
+        <desc xml:lang="en">Indicates if the opening points must be aligned vertically, including when the hairpin is angled with @optimize.angle or through @endvo/@startvo adjustments.</desc>
+        <datatype>
+          <rng:ref name="data.BOOLEAN"/>
+        </datatype>
+      </attDef>
+      <attDef ident="angle.optimize" usage="opt">
+        <desc xml:lang="en">Indicates that the slope of the haipin can be adjusted to follow the content in order to optimize spacing.</desc>
+        <datatype>
+          <rng:ref name="data.BOOLEAN"/>
+        </datatype>
+      </attDef>
     </attList>
   </classSpec>
   <classSpec ident="att.halfmRpt.vis" module="MEI.visual" type="atts">
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -680,7 +680,7 @@
         </datatype>
       </attDef>
       <attDef ident="opening.vertical" usage="opt">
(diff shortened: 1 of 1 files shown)


source/modules/MEI.visual.xml
```diff
@@ -679,6 +679,18 @@
           <rng:ref name="data.BOOLEAN"/>
         </datatype>
       </attDef>
+      <attDef ident="opening.vertical" usage="opt">
+        <desc xml:lang="en">Indicates if the opening points must be aligned vertically, including when the hairpin is angled with @optimize.angle or through @endvo/@startvo adjustments.</desc>
```
ahankinson: I think what is meant is that the points should be connected with a line that is always 90° perpendicular to the horizontal plane, regardless of any angle or start/end adjustments. As it is, just saying that the points should be "aligned vertically" but doesn't say how they are to be aligned.

lpugin: Except that the points are not connected. So I would suggest
```
Indicates if the opening points must be aligned with a line that is always 90° perpendicular to the horizontal plane, regardless of any angle or start/end adjustments, including when the hairpin is angled with @optimize.angle or through @endvo/@startvo adjustments.
```
Better?

ahankinson: Sorry, in an earlier draft I had "an imaginary line" but that got missed in the editing.

ahankinson:
```
Indicates that the opening points must be aligned with an imaginary line that is always 90° perpendicular to the horizontal plane, regardless of any angle or start/end adjustments, including when the hairpin is angled with @optimize.angle or through @endvo/@startvo adjustments.
```

ahankinson: Just a couple tweaks

rettinghaus: `Indicates that the opening points are …` would make it clear that this is what the encoder has seen in the source _and_ the engraver would expect.

ahankinson (dismissed):

rettinghaus (dismissed):

source/modules/MEI.visual.xml
```diff
@@ -679,6 +679,18 @@
           <rng:ref name="data.BOOLEAN"/>
         </datatype>
       </attDef>
+      <attDef ident="opening.vertical" usage="opt">
+        <desc xml:lang="en">Indicates that the opening points are aligned with an imaginary line that is always 90° perpendicular to the horizontal plane, regardless of any angle or start/end adjustments, including when the hairpin is angled with @optimize.angle or through @endvo/@startvo adjustments.</desc>
```
```suggestion
        <desc xml:lang="en">Indicates that the opening points are aligned with an imaginary line that is always 90° perpendicular to the horizontal plane, regardless of any angle or start/end adjustments, including when the hairpin is angled with @angle.optimize or through @endvo/@startvo adjustments.</desc>
```

source/modules/MEI.visual.xml
```diff
@@ -679,6 +679,18 @@
...
+        <datatype>
+          <rng:ref name="data.BOOLEAN"/>
+        </datatype>
+      </attDef>
+      <attDef ident="angle.optimize" usage="opt">
+        <desc xml:lang="en">Indicates that the slope of the haipin can be adjusted to follow the content in order to optimize spacing.</desc>
```
```suggestion
        <desc xml:lang="en">Indicates that the slope of the hairpin can be adjusted to follow the content in order to optimize spacing.</desc>
```

source/modules/MEI.visual.xml
```diff
@@ -679,6 +679,18 @@
           <rng:ref name="data.BOOLEAN"/>
         </datatype>
       </attDef>
+      <attDef ident="opening.vertical" usage="opt">
```
musicEnfanthen, line 682: Would it make sense to move this below  "opening" attDef? In general, do we follow alphabetical order for attDefs?

musicEnfanthen (changes requested):

source/modules/MEI.visual.xml
lpugin: We can. I left it after `@angle.optimize` because they it relates to it. I am not sure we follow alphabetical order systematically.

musicEnfanthen: Ok, fine with that

musicEnfanthen (approved):

pe-ro (approved):
Looks good to me.

musicEnfanthen:
Thank you, everyone!

PR #1022, opened 2022-10-03
A expanMark element for encoding coda, segno and similar marks

lpugin:
This PR is a proposal for issue https://github.com/music-encoding/music-encoding/issues/520

Instead of having distinct element for each mark, it introduces a `repeatMark` element. The function of the mark has to be encoded in a `@func`. The possible values are:
```xml
<valItem ident="coda">
    <desc xml:lang="en">Coda (SMuFL E048 or Unicode 1D10C).</desc>
</valItem>
<valItem ident="segno">
    <desc xml:lang="en">Segno (SMuFL E047 or Unicode 1D10B).</desc>
</valItem>
<valItem ident="dalSegno">
    <desc xml:lang="en">Dal segno (SMuFL E045 or Unicode 1D109).</desc>
</valItem>
<valItem ident="daCapo">
    <desc xml:lang="en">Da capo (SMuFL E046 or Unicode 1D10A).</desc>
</valItem>
<valItem ident="fine">
    <desc xml:lang="en">Fine.</desc>
</valItem>
<valItem ident="other">
    <desc xml:lang="en">Unspecified symbol.</desc>
</valItem>
```

The element `repeatMark` also support `@glyph.num` or `@glyph.name` (through `att.extSym`), which makes it possible to choose a specific glyph (e.g., codaSquare in SMuFL).

In addition to this, textual content can be added. For example, for cases where the mark is "Repeat dal segno" instead of the common mark "D.S.".

If the element is empty, a renderer is expected to show a default sign (see SMuFL or Unicode references). When text is provided, then not default symbol would be rendered but only what is given in the text. Of course, the text content can itself contain a symbol. For example
```xml
<repeatMark func="coda">
  Coda sign is <symbol glyph.num="U+1234"/>
</repeatMark>
```

When textual content is provided, there is a Schematron rule that checks that no `@glyph.num` or `@glyph.name` is given on `<repeatMark>` because that would not make sense.


kepper (dismissed):
specs looking very good, especially good when someone adds Schematron 👍. As soon as there is something in the Guidelines, in can be moved from a draft to a PR ;-)

ahankinson:
I'll hold off approving until it's no longer a draft, but it looks good to me.

source/modules/MEI.cmn.xml
```diff
@@ -2262,6 +2262,53 @@
...
     </remarks>
   </elementSpec>
+  <elementSpec ident="repeatMark" module="MEI.cmn">
+    <gloss versionDate="2022-05-18" xml:lang="en">repetition mark</gloss>
+    <desc xml:lang="en">
+      An instruction expressed as a combination of text and symbols – segno and coda – typically above, 
+      below, or between staves, but not on the staff.</desc>
+    <classes>
+      <memberOf key="att.common"/>
+      <memberOf key="att.facsimile"/>
```
rettinghaus, line 2255: att.extSym is missing here

source/modules/MEI.cmn.xml
```diff
@@ -2262,6 +2262,53 @@
         model.textPhraseLike.limited.</p>
     </remarks>
   </elementSpec>
+  <elementSpec ident="repeatMark" module="MEI.cmn">
```
rettinghaus, line 2248: 'coda' and 'fine' are not really _repeat marks_ (even though they often appear in repetition contexts). Most of the presented values here mark jumps, so I'd prefer something like `jumpMark`.

source/modules/MEI.cmn.xml
```diff
@@ -2262,6 +2262,53 @@
...
+      below, or between staves, but not on the staff.</desc>
+    <classes>
+      <memberOf key="att.common"/>
+      <memberOf key="att.facsimile"/>
+      <memberOf key="att.lang"/>
+      <memberOf key="att.repeatMark.log"/>
+      <memberOf key="att.repeatMark.vis"/>
+      <memberOf key="att.repeatMark.ges"/>
+      <memberOf key="att.repeatMark.anl"/>
+      <memberOf key="model.controlEventLike"/>
```
rettinghaus, line 2261: Is `model.controlEventLike` sufficient? As said before some of these symbols may appear on a staff directly.

rettinghaus (changes requested):

source/modules/MEI.cmn.xml
lpugin: Yes, you got it right. The element represents marks within repetition patterns. It is a matter of preference.

source/modules/MEI.cmn.xml
lpugin: Yes
```xml
<repeatMark func="dalSegno"  place="within">
```

source/modules/MEI.cmn.xml
lpugin: It is in att.repeatMark.vis

rettinghaus: Oh yes, overlooked it.

rettinghaus (dismissed):

source/modules/MEI.cmn.xml
pe-ro: I think model.controlEventLike is sufficient.

source/modules/MEI.cmn.xml
pe-ro: `jumpMark` sounds too "jargon-y" to me, too "modern" perhaps.  These are "navigation marks" to me, so "navMark" is a possibility.  Still, I prefer "repeatMark".

lpugin: `navMark` sounds like `satNav` 🤣

pe-ro: `<sequenceMark>`
`<sequencingMark>`
`<seqMark>` (short form of either of the above)
`<execMark>` (short for execution mark)
`<expanMark>` (short for expansion mark, related to `<expansion>`)

I think I like the last one best because it indicates the function of the mark is analogous to the functionality of `<expansion>`.

lpugin: `performMark`?

pe-ro: I know `<expanMark>` is somewhat cryptic, but I prefer it over `<performMark>`.  `<performMark>` creates semantic confusion between "performance marks", like coda, segno, etc. and "performance directives"; i.e., `<dir>` elements. It's the same reason `<repeatMark>` isn't a very good name -- it can be confused with repeat marks on barlines.  At least, `<expanMark>` aligns semantically with `<expansion>`, the description of which (I hope) makes the concept clearer.

lpugin: This all makes sense. My only concern with `<expanMark>` is similar to yours with `<performMark>`, namely that it could create confusion, with `<expansion>`. In the end users will need to read the docs, and even if XML is self-documenting, we still want to avoid names becoming too long in order just because we want to make sure there is no possible confusion. So I am still fine with `<repeatMark>`. (I am fine with changing it too if we feel it is seriously wrong or somebody finds something much better.)

ahankinson:
I haven't seen much on how this would work with `<ending>` and the other ways of actually encoding repeats. Is this only for the symbol, or are there other semantics involved with this?

lpugin:
This PR focuses on the symbol and is based on the assumption that the semantics is to be handled with `<expansion>`. At least this is my understanding of it.

pe-ro:
@lpugin, has it right -- `<repeatMark>` is just about encoding the coda, segno, etc. signs/directives. The meaning (the performance order of the sections of notation marked by the symbol) should still be encoded using `<expansion>`.

This linkage between the sign and the semantics was my point about naming this element `<expanMark>`.  I think this makes it easier for the user to associate the two.

lpugin:
I got your point @pe-ro, and your are right that the relation between `<expanMark>` and `<expansion>` is actually a good think. So unless there is an objection to it - please speak up now - I will rename `<repeatMark>` to `<expanMark>`.

pe-ro:
👍

source/modules/MEI.shared.xml
```diff
@@ -2720,6 +2720,40 @@
...
+            <desc xml:lang="en">Dal segno (SMuFL E045 or Unicode 1D109).</desc>
+          </valItem>
+          <valItem ident="daCapo">
+            <desc xml:lang="en">Da capo (SMuFL E046 or Unicode 1D10A).</desc>
+          </valItem>
+          <valItem ident="fine">
+            <desc xml:lang="en">Fine.</desc>
+          </valItem>
+          <valItem ident="other">
+            <desc xml:lang="en">Unspecified symbol.</desc>
```
ahankinson: What is an 'other' mark? How would a user specify the actual symbol if they were to choose `other`?

ahankinson: If the user specifies `other` then I think it should be required to define the symbol with the `glyph.*` attributes.

rettinghaus:
I still prefer `rptMark` or `perfMark` (although the latter might cause confusion indeed). The problem with `expanMark` is that it will _not_ expand something. Instead it will mark a point visually where a repetition _should_ happen. So this is more similar to `"rptstart"` and `"rptend"` values from `data.BARRENDITION`.

ahankinson:
@rettinghaus given that this is only the symbol, and the actual expansion will need to be encoded elsewhere, it would make sense to me to align the symbol name with the function. 

In other words, I don't think simply encoding an `<expanMark>` (or whatever we call it) would be enough for a client to work out the expansion. There needs to be more encoding to make that happen.

Perhaps the `<expanMark>` needs an attribute to relate it to an expansion? That would make the relationship between the two quite clear.

source/modules/MEI.shared.xml
rettinghaus: Good point. Also the attribute name is `func`, not `form`. Are there more possible functions we don't know about yet? Then it would be better to name this `unkown` over `other`, because the symbol could be encoded whatsoever.

ahankinson: +1 to renaming it to `form` -- arguably, the function is similar, or the same, for some of these, but the form is different.

source/modules/MEI.shared.xml
```diff
@@ -2720,6 +2720,40 @@
...
+  <classSpec ident="att.repeatMark.log" module="MEI.shared" type="atts">
+    <desc xml:lang="en">Logical domain attributes.</desc>
+    <classes>
+      <memberOf key="att.controlEvent"/>
+      <memberOf key="att.duration.additive"/>
+      <memberOf key="att.startEndId"/>
+      <memberOf key="att.timestamp2.logical"/>
+    </classes>
+    <attList>
+      <attDef ident="func" usage="req">
```
ahankinson, line 2723:
```suggestion
      <attDef ident="form" usage="req">
```

source/modules/MEI.shared.xml
```diff
@@ -2720,6 +2720,40 @@
...
+        <desc xml:lang="en">Indicates the function of the mark.</desc>
```
ahankinson:
```suggestion
        <desc xml:lang="en">Indicates the form of the mark.</desc>
```

source/modules/MEI.cmn.xml
```diff
@@ -2262,6 +2262,53 @@
...
+    </constraintSpec>
+    <constraintSpec ident="repeatMark_with_glyph_has_to_be_empty" scheme="isoschematron">
+      <constraint>
+        <sch:rule context="mei:repeatMark[@glyph.num or @glyph.name]">
+          <sch:assert test="not(element()) and not(text())">When @glyph.name or @glyph.num is present, repeatMark must not have content.</sch:assert>
+        </sch:rule>
+      </constraint>
+    </constraintSpec>
+    <remarks xml:lang="en">
+      <p>When only @func is provided to describe the function of the repeat mark (i.e., without glyph information and no textual content), then a renderer
```
ahankinson:
```suggestion
      <p>When only @form is provided to describe the function of the repeat mark (i.e., without glyph information and no textual content), then a renderer
```

rettinghaus:
@ahankinson how should this work? Expansion links to sections and endings. 
`<expansion plist="#section1 #dalSegnoPart #dalSegnoPart" />`. How should `<expanMark func="fine" />` relate to this?

ahankinson:
As I understand it, the `<expanMark>` indicates the specific symbol that is used, so I would expect to see it used in much the same way as `<dir>` and allow placement of the symbol. So it would go inside `<measure>`, I think?

So given:

`<expansion xml:id="e1" plist="#section1 #dalSegnoPart #dalSegnoPart" />`

Then the various `<expanMark>`s might be (elsewhere in the encodings, whereever the symbol is needed):

```
<expanMark func="dalSegno" expansion="#e1" />
...
<expanMark func="dalSegno" expansion="#e1" />
...
<expanMark func="fine" expansion="#e1" />
```

(Hard to figure out the exact encoding without a visual, but I hope you'll get the meaning.)

source/modules/MEI.shared.xml
lpugin: I do not think we can bind the use of `other` with the definition of a glyph. The element can contain text because sometimes this is how a mark is described. The cases where `other` is used are probably the most likely not to be just a single symbol. For example, some thing like:
```xml
<expanMark func="other">Repeat this section as many times as needed until all the dancers have left the stage</expanMark>
```
I hope this also explains why `@func` seems preferable than `@form`.

ahankinson: In that instance "other" doesn't actually bring anything, though, and it would be better to omit it entirely.

That is, there's no real difference between your example and:

```xml
<expanMark>Repeat this section as many times as needed until all the dancers have left the stage</expanMark>
```

(and yes, I see that it's currently marked as "required" but that could be changed.)

lpugin:
> As I understand it, the `<expanMark> `indicates the specific symbol that is used, so I would expect to see it used in much the same way as `<dir>` and allow placement of the symbol. So it would go inside `<measure>`, I think?

Yes. `<expanMark>` is a control event. You can think of it as a more specific `<dir>`, similarly to `<tempo>` or `<ornam>`. (See #520). However, it is meant to be able to accommodate various cases - not just a specific symbol.

Examples (for the sake of readability without `@startid` or `@tstamp` and without `@staff`) :

### A generic symbol for a defined role

```xml
<expanMark func="coda"/>
```

### A specific symbol for a defined role

```xml
<expanMark func="coda" glyph.name="codaSquare"/>
```

### Some text for a define role

```xml
<expanMark func="dalSegno">
 Dal Segno
</expanMark>
```

### Some text with a symbol for a defined role

```xml
<expanMark func="coda">
  Coda sign is <symbol glyph.name="coda"/>
</expanMark>
```

### Some text for an undefined role

```xml
<expanMark func="other">Repeat this section as many times as needed until all the dancers have left the stage</expanMark>
```

### A specific symbol for an undefined role

(This would be a weird case in my opinion, but still possible)

```xml
<expanMark func="other" glyph.name="codaJapanese"/>
```

ahankinson:
@lpugin sounds reasonable, but given that it's not actually a *functional* mark (that is, the _function_ of the expansion is encoded elsewhere) I'm still not sure about `@func`. Your examples seem to illustrate that, since the use of "other" seems to indicate that the **form** of the symbol is what is changing.

That is, you have

```
<expanMark func="other" glyph.name="codaJapanese"/>
```

Where, if it was **functional** I would actually assume it would be:

```
<expanMark func="coda" glyph.name="codaJapanese"/>
```

But I still don't see the need for the "other" value in either form or function -- that would mean that there would actually be no way of deriving the symbol automatically, and you would either have to rely on textual content, or on other attributes, to have any sort of symbol to show anyway. That means, I think, that this possible value could reasonably be dropped. 

IOW, I don't see any difference between:

```
<expanMark func="other" glyph.name="codaJapanese"/>
```

and 

```
<expanMark glyph.name="codaJapanese"/>
```

What does `func="other"` bring to this encoding?

lpugin:
> Your examples seem to illustrate that, since the use of "other" seems to indicate that the form of the symbol is what is changing.

Then which one of the other given values would you pick for
```xml
<expanMark func="other">Repeat this section as many times as needed until all the dancers have left the stage</expanMark>
```
? 

I precisely expect `other` to be useful in the cases where the mark is given more a text directive but still clearly acts as a mark in the same way segno and coda function. You could argue that we can use `dir` for that but I think it would be better to let people use `expanMark` in such cases because it does bring some additional semantic - namely that the directive is an expansion / repetition mark.

> What does `func="other"` bring to this encoding?

As I said, this would be a weird case. But what it bring is that we can make `func` a required attribute. I do not think we can drop `other` without making the attribute optional.

What we can do it add a Schematron rule that forbids the use of `other` without textual content. I would also forbid `other` with only a glyph as above, again because I do not really see a use for it.

ahankinson:
Personally, I think `@func`/`@form` **should** be an optional attribute. The element is just the symbol, and has no bearing on the actual behaviour of the encoded notation.

Isn't:

`<expanMark func="coda">`

Just a shorthand for:

`<dir glyph.name="coda" glyph.auth="smufl">`

?

lpugin:
> Personally, I think `@func/@form` should be an optional attribute. The element is just the symbol, and has no bearing on the actual behaviour of the encoded notation.

The reasoning behind making `@func` required (and behind this PR in general) is to enforce semantics and to avoid things like
```xml
<expanMark glyph.name="coda"/>
```
In other words, we do not add distinct elements `<coda>`, `<segno>` or `<fine>` but compensate it with a required attribute to bring the information we would have with these. Additional we leave the door open to undefined marks with `func="other"`.

> Isn't:
`<expanMark func="coda">`
Just a shorthand for:
`<dir glyph.name="coda" glyph.auth="smufl">`
?

Yes, minus semantic. So if somebody is annoyed by having to have `func`, they can use `dir`. That is already possible and will still be.

ahankinson:
OK, after an offline conversation we've arrived at a proposal:

1. Change `@func` to `@markType`
2. Remove "other" from the possible values
3. Disallow, with schematron, an `<expanMark>` element that has neither `markType` attribute, nor textual content. Example below.
4. Move the attribute from required to recommended (`usage="rec"`)

So these would be allowed:

```xml
<expanMark markType="coda" />
<expanMark>Vamp until sunrise</expanMark>
<expanMark markType="coda" glyph.name="codaJapanese"/>
<expanMark markType="coda">Coda, but written out</expanMark>
```

But these wouldn't:

```xml
<expanMark />
<expanMark glyph.name="codaJapanese" />
<expanMark>
    <rend glyph.name="codaJapanese" /> (or any other symbol-only carrier element)
</expanMark>
```

So: if you have a symbol, you need to type it somehow. If you don't have a symbol, there needs to be text content.

Not sure about:

```
<expanMark>
    Go until you reach <rend glyph.name="codaJapanese" /> and then turn around.
</expanMark>
```

lpugin:
> Not sure about:
> ```xml
> <expanMark>
>     Go until you reach <rend glyph.name="codaJapanese" /> and then turn around.
> </expanMark>
> ```

Yes, I am not sure either. So maybe the best is simply to drop `other` from `markType` and still make it required. Then this would be encoded with a `dir`. I think this is actually better.

ahankinson:
If you drop "other" but also make it required, then what do you put for:

```xml
<expanMark markType="????">Vamp until sunrise</expanMark>
```

Given that this is **just the symbol** then I think for 

```xml
<expanMark>
    Go until you reach <rend glyph.name="codaJapanese" /> and then turn around.
</expanMark>
```

we allow this. It gets rendered as text, and the corresponding `<expansion>` is what takes care of the actual repeat expansion.

lpugin:
Looking at some more examples and in the literature, I think it is safe to say that coda, segno, dal segno, da coda and fine form a pretty well defined set. So it was probably not a good idea in the first place also to include `other` in the list for encoding something that is really a `dir`. If one day somebody can come with a good explanation and some examples, we can always revise the list. Hopefully the new insights will allow us to find a value that is better than `other`.

source/modules/MEI.gestural.xml
```diff
@@ -625,6 +625,13 @@
   <classSpec ident="att.reh.ges" module="MEI.gestural" type="atts">
     <desc xml:lang="en">Gestural domain attributes.</desc>
   </classSpec>
+  <classSpec ident="att.repeatMark.ges" module="MEI.gestural" type="atts">
+    <desc xml:lang="en">Gestural domain attributes.</desc>
+    <classes>
+      <memberOf key="att.duration.gestural"/>
+      <memberOf key="att.timestamp2.gestural"/>
```
musicEnfanthen:
```suggestion
      <memberOf key="att.timestamp2.ges"/>
```

source/modules/MEI.gestural.xml
```diff
@@ -625,6 +625,13 @@
   <classSpec ident="att.reh.ges" module="MEI.gestural" type="atts">
     <desc xml:lang="en">Gestural domain attributes.</desc>
   </classSpec>
+  <classSpec ident="att.repeatMark.ges" module="MEI.gestural" type="atts">
+    <desc xml:lang="en">Gestural domain attributes.</desc>
+    <classes>
+      <memberOf key="att.duration.gestural"/>
```
musicEnfanthen:
```suggestion
      <memberOf key="att.duration.ges"/>
```

source/modules/MEI.shared.xml
```diff
@@ -2711,6 +2711,40 @@
       </attDef>
     </attList>
   </classSpec>
+  <classSpec ident="att.repeatMark.log" module="MEI.shared" type="atts">
+    <desc xml:lang="en">Logical domain attributes.</desc>
+    <classes>
+      <memberOf key="att.controlEvent"/>
+      <memberOf key="att.duration.additive"/>
+      <memberOf key="att.startEndId"/>
+      <memberOf key="att.timestamp2.logical"/>
```
```suggestion
      <memberOf key="att.timestamp2.log"/>
```

rettinghaus (approved):

musicEnfanthen:
@ahankinson May I ask you to close the open conversations if obsolete?

ahankinson (approved):

musicEnfanthen (approved):

rettinghaus:
🥳 Thanks!

PR #1031, opened 2022-10-10
Add border and improve line breaks in classBox

### LINKED ISSUE #66: classBox missing borders
The `classBox` in data-types has no borders on the right and at the bottom. 
Is this intentional, or missing in CSS?

rettinghaus:
closes https://github.com/music-encoding/guidelines/issues/66

### DIFF
--- a/source/web/css/mei-screen.css
+++ b/source/web/css/mei-screen.css
@@ -181,7 +181,7 @@
 
 .specPage .classBox .classHeading {
 	background-color: #f5f5f5;
-	padding: 0 0 .1rem .5rem;
+	padding: 0 .5rem .1rem .5rem;
 	margin-left: -.5rem;
 	border-bottom: .5px solid #999999;
 }
@@ -191,6 +191,7 @@
 }
 
 .specPage .classBox .classHeading .classDesc {
+	display: inline;
 	padding-left: .5rem;
 	font-size: .7rem;
 	font-weight: 500;
--- a/source/web/css/mei-screen.css
+++ b/source/web/css/mei-screen.css
@@ -191,7 +191,6 @@
 }
 
 .specPage .classBox .classHeading .classDesc {
-	display: inline-block;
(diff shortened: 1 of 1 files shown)


source/web/css/mei-screen.css
```diff
@@ -192,6 +191,7 @@
 }
 
 .specPage .classBox .classHeading .classDesc {
+	display: inline-block;
```
bwbohl: I can see what this changes but I'd rather add o.5rem padding to `.specPage .classBox .classHeading` then `.classDesc` would always stay inline with `.classLabel` but no run into the right border when it is longish

bwbohl:
moreover what about sth lik

```css
.classContent .def:nth-of-type(2n) {
    background-color: #f5f5f5;
}
```

source/web/css/mei-screen.css
rettinghaus: This is just to keep `classDesc` better separated from `classLabel`.
It changes breaks from this:

image

to this:

image

rettinghaus:
Thanks @bwbohl, but I want to keep this PR small. There are still many things to improve.

source/web/css/mei-screen.css
bwbohl: This is true only for those `classDesc` that are too long to fit into the line.

rettinghaus: Ok, seems we cannot agree on this, so I removed it again.

bwbohl (approved):

musicEnfanthen:
@rettinghaus May I ask you for a short "before/after" screenshot, so it's easier to compare the result of the changes?

rettinghaus:
before 

image

after 

image

kepper (approved):
I left that open to save space, but I'm fine with this change :-) Thanks for caring about this, @rettinghaus 👍

PR #1033, opened 2022-10-11
Fix(ci): Update deploy action

rettinghaus:
This PR updates the used actions and combines the git steps.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,7 +7,7 @@ name: Deploy Schema and Guidelines
 
 on:
   push:
-    branches: [ develop ]
+    branches: [develop]
 
 env:
   # schema
@@ -44,7 +44,8 @@ jobs:
       - name: Set up JDK 1.8
         uses: actions/setup-java@v3
         with:
-          java-version: 1.8
+          distribution: "temurin"
+          java-version: "17"
 
       - name: Build with Ant
         # For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-ant
@@ -103,7 +104,7 @@ jobs:
         run: |
           rm -rf $GUIDELINES_DIR/dev 
           cp -r dist/guidelines/dev/web $GUIDELINES_DIR/
-          mv $GUIDELINES_DIR/web $GUIDELINES_DIR/dev 
+          mv $GUIDELINES_DIR/web $GUIDELINES_DIR/dev
 
       - name: Check git status and commit
         working-directory: ${{ env.GUIDELINES_DIR }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -41,7 +41,7 @@ jobs:
           # Default: false
           submodules: recursive
 
-      - name: Set up JDK 1.8
+      - name: Set up Java
         uses: actions/setup-java@v3
         with:
           distribution: "temurin"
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -42,9 +42,10 @@ jobs:
           submodules: recursive
 
       - name: Set up JDK 1.8
-        uses: actions/setup-java@v1
+        uses: actions/setup-java@v3
         with:
-          java-version: 1.8
+          distribution: "temurin"
+          java-version: "17"
```
musicEnfanthen, line 48: Is ant running with it?

.github/workflows/deploy.yml
```diff
@@ -42,9 +42,10 @@ jobs:
           submodules: recursive
 
       - name: Set up JDK 1.8
```
musicEnfanthen:
```suggestion
      - name: Set up Java

File: .github/workflows/deploy.yml [hunk 3]
```diff
@@ -69,23 +70,15 @@ jobs:
       - name: Copy built schema to SCHEMA_DIR
         run: cp -r dist/schemata/dev $SCHEMA_DIR/
 
-      - name: Check git status before commit
```
musicEnfanthen: I know that these three steps are somewhat verbose, but it helps to keep things separated (also for debugging). So would prefer not to combine them.

.github/workflows/deploy.yml
```diff
@@ -111,25 +104,17 @@ jobs:
         run: |
           rm -rf $GUIDELINES_DIR/dev 
           cp -r dist/guidelines/dev/web $GUIDELINES_DIR/
-          mv $GUIDELINES_DIR/web $GUIDELINES_DIR/dev 
+          mv $GUIDELINES_DIR/web $GUIDELINES_DIR/dev
 
-      - name: Check git status before commit
```
musicEnfanthen, line 116: Same as above

.github/workflows/deploy.yml
```diff
@@ -24,13 +24,13 @@ env:
 
 jobs:
   build:
-
+    name: Build and deploy Schema
```
musicEnfanthen:
```suggestion
    name: Build and deploy Schema and Guidelines
```

musicEnfanthen (changes requested):

.github/workflows/deploy.yml
rettinghaus: Yes, tested.

musicEnfanthen: Great

.github/workflows/deploy.yml
musicEnfanthen: Thank you

musicEnfanthen (approved):

bwbohl:
@rettinghaus what do you mean with 
> combines the git steps

can't really see a big change, although I think the small ones make a big difference ;-)

bwbohl:
Moreover I'd rather see this merged to [pdf-guidelines-test ](https://github.com/music-encoding/music-encoding/tree/pdf-guidelines-test) which holds the latest developments of the build process

musicEnfanthen:
@bwbohl This one updates the versions of different actions used right now (checkout, java setup, etc.). It should be fine to go for now in develop. The full script will be revised on pdf-guidelines-test for sure, but that is not primarily connected to this PR.

bwbohl (approved):

PR #1037, opened 2022-10-18
Fix(docs): Fix issue 120 (guidelines repo)

### LINKED ISSUE #120: 3.1.1.3: PhysDesc not allowed in fileDesc
Chapter 3.1.1.3. "Physical Description of the File" should be renamed and the example corrected accordingly (remove physDesc)

riedde:
This PR fixes music-encoding/guidelines#120

- Renaming guidelines section 3.4.1.2 to prevent misunderstanding
- update example (invalid encoding)

The example tried to offer a list of possible encodings, but the wrapping element is not allowed for this case. Trying to keep the intention, I just deleted the wrapper and added comments to clarify that the encoded examples are substitutable.

### DIFF
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -306,7 +306,7 @@
                         </figure>
                      </p>
                   </div>
-                  <div xml:id="headerPhysicalDescription" type="div4">
+                  <div xml:id="headerFileExtent" type="div4">
                      <head>Extent of the File</head>
                      <p>The third component of the fileDesc is a description of the physical qualities of the file. The <gi scheme="MEI">extent</gi> element is provided for this purpose.</p>
                      <p>
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -307,7 +307,7 @@
                      </p>
                   </div>
                   <div xml:id="headerPhysicalDescription" type="div4">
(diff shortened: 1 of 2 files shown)


riedde:
closes https://github.com/music-encoding/metadata-ig/issues/26

kepper (dismissed):
looks sensible to me

source/docs/03-metadata.xml
```diff
@@ -307,7 +307,7 @@
                      </p>
                   </div>
                   <div xml:id="headerPhysicalDescription" type="div4">
```
musicEnfanthen:
```suggestion
                  <div xml:id="headerFileExtent" type="div4">
```

musicEnfanthen (changes requested):

source/docs/03-metadata.xml
musicEnfanthen: Did not find any references to this id in the repo, but would be good to double-check

riedde: Same here.

musicEnfanthen (approved):

source/docs/03-metadata.xml
musicEnfanthen: Great, thanks

bwbohl (approved):

musicEnfanthen:
Thank you @riedde

PR #1039, opened 2022-10-19
Add gloss to some attributes

rettinghaus:
This PR adds glossaries to some attributes that contain their full name.

It also standardizes the spelling of "bar line" and brings the description of "@facs" in line with the TEI.

(Currently these `gloss` elements are just sitting in the specs and will not find their way to the Guidelines.)


bwbohl:
I love the efforts but would prefer to keep to one global change at a time.

1. I'm completely in on the `gloss` elements and would approve straight away if it was only those.
2. "bar line" vs "barline" I don't even know whether this actually is an issue. (@music-encoding/guidelines-editors native speakers, please help). To my understanding both uses are common. While "bar line" might be the etymologically older version, barline seems to be ok, also according to the standards of forming compounds in English. Of course, it should be consistently used in MEI and the corresponding element name `barLine` even indicates an open compound due to its camel case. Nevertheless, I think if we start down a language style path, we should first come up with some language style guidelines. So I'm a no at the moment.
3. As for `@facs`: like the rephrase but please in separate PR for the sake of generating release reports automatically.

rettinghaus:
Closing in favor of separated PRs #1045 #1046 #1047.

PR #1044, opened 2022-10-27
Add clarification for GG clef in data.CLEFSHAPE

th-we:
Sibmei so far output GG clefs as `<clef shape="GG" dis="8" dis.place="below"/>`, but in my understanding, `GG` is already transposing by itself without the `dis*` attributes. Before changing that in Sibmei, I'd like to make sure this is agreed upon and while I'm at it, document it for anybody else who wonders.

I carefully phrased "As this clef is usually not written with an "8" below" because should there be the odd and unlikely case of a `GG` clef with an 8 written below, then the `dis*` attributes would of course be appropriate.  But shall we open the can of worms if those attributes are visual or gestural?  And there are odd uses, like Carl Maria von Weber apparently writes them for flutes, oboes and some other instruments I wouldn't expect [here](https://www.loc.gov/resource/ihas.200154550.0/?sp=10&st=image):

grafik

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -541,7 +541,9 @@
           <desc xml:lang="en">G clef (Unicode 1D11E).</desc>
         </valItem>
         <valItem ident="GG">
-          <desc xml:lang="en">Double G clef.</desc>
+          <desc xml:lang="en">Double G clef. Sounds one octave lower than G clef. 
+            As this clef is usually not written with an "8" below the clef, do not combine with 
+            <att>dis</att>/<att>dis.place</att>/<att>clef.dis</att>/<att>clef.dis.place</att>.</desc>
         </valItem>
         <valItem ident="F">
           <desc xml:lang="en">F clef (Unicode 1D122).</desc>


ahankinson:
Can I suggest that these attributes are both visual and gestural? The domains of encoding are useful for thinking about different aspects of a symbol, but they shouldn’t be considered as always mutually exclusive from one another.

Personally I think Sibmei is correct here, and the GG should be accompanied by dis.place, simply because it centralizes the control of the octave shift in one attribute, rather than spreading the logic for computing this across two attributes, with an implicit octave shift.

bwbohl:
I'm not really sure about this, as `@dis.place` is of datatype `data.STAFFREL.basic` (above, below). Thus it actually is mere graphical. Of course, implicitly it's also gestural as the position above or below the clef (staff) indicates the direction of displacement. As on `staffDef` there are also the attributes `clef.shape`, `clef.dis`, and `clef.dis.shape`. I wonder how this would interact with `trans.diat` and `trans.semi`. To my understanding, the clef things are to be considered mere graphical while the trans would be gestural (or logical). So if you would want to encode e.g. a clef change I'd always prefer that in a `staffDef` and only use `clef` for e.g. cautionary clefs.

rettinghaus:
In Germany the double clef was often used to indicate that two voices share one staff. (I'm not sure though if there is a distinction in usage by county is possible.)

If shape is in the logical domain we should probably remove the value GG as there are different meanings possible.

ahankinson:
> If shape is in the logical domain we should probably remove the value GG as there are different meanings possible.

The `@shape` attribute is actually in several domains all at once. It sets the tessitura of the staff (logical) and describes the  appearance (visual). In some cases, like your example, it can also describe other features of the staff that may or may not belong to one or more of these domains.

The domains are useful methods for thinking about the different roles a musical symbol (in both appearance and function) takes, but I'm not sure it's that useful to try and break down whether one attribute is exclusively in one domain or another. They can be in more than one at the same time, and that's OK.

ahankinson:
> Thus it actually is mere graphical.

Except that it controls whether the octave is an octave UP or an octave DOWN. `@dis` sets the octave displacement, but it doesn't say which direction it goes; `@dis.place` does.

musicEnfanthen:
In general, I would agree that `<clef shape="GG"/>` resp. `<staffDef n="1" lines="5" clef.shape="GG" clef.line="2">...</staffDef>` should be sufficient to indicate the standard behaviour/meaning of a G-clef transposed one octave down. (Side note: Verovio handles it like that right now, any `dis.` attributes do not change anything for GG shaped clefs.) But now there is this:

> In Germany the double clef was often used to indicate that two voices share one staff. (I'm not sure though if there is a distinction in usage by county is possible.)

That seems to be actually the case with the "odd" Weber example given by @th-we above. The voices of Flutes, Oboes and Horns are each given on a single staff, but obviously notated in the correct octave position. So no octave displacement involved in that case.

So would we need someting to clarify that distinction?

th-we:
@rettinghaus Thanks for the interesting Weber explanation!

Shouldn't we clarify if `@dis`/`@dis.place` imply a number written above/below the clef? If they (always) do, then how do we differentiate between Weber style GG clefs and French style GG clefs? If they don't always imply a number, how do we specify when a number is written and when it's not?  Will we list clefs that can have numbers and clefs that can't?

Would adding `@pname`/`@oct` to `<clef>` be sensible, allowing to specify what's the pitch on the `@line`? `@oct='4'` would encode Weber-style, `@oct='3'`  would be French style (`@pname` not really needed here). As a side effect, that would make exotic clefs encodable. In the presence of `@oct`, I guess `@dis`/`@dis.place` would be reduced to their visual implications.

ahankinson:
`@oct` already exists on `<clef>`. Does that solve it?

ahankinson:
Discussed at the 23 Feb. ODD meeting. Noted that there are two possible meanings: "GG" ("German") meaning "two instruments" and "GG" ("French") which records an octave shift. In the first (German) meaning, there would need to be additional layers anyway, so the clef shape does not affect the notation.

In the latter (French) we would need to specify the octave the clef refers to. 

So it was suggested that `<clef shape="GG" />` would mean the German style, and `<clef shape="GG" oct="3" />` would be the French.

@th-we does that sound OK?

craigsapp:
This does not make great semantic sense for use with GG clef:

```xml
<clef shape="GG" oct="3" />
```

The staff does not define `@oct="3"` as a `@loc`, but it is rather defined by the clef.  So using it to place the clef is circular logic.

Here are the different clefs used to display middle C:

Screenshot 2023-02-23 at 9 51 40 AM


<details>
<summary> Click to view MEI data for above example </summary>

```xml
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0.0-dev">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2023-02-23T09:50:36" version="3.15.0-dev-2f1ce00-dirty">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="mcj7z1q">
    <score xml:id="smfzr79">
     <scoreDef xml:id="s1l2oiro" midi.bpm="400.000000">
      <staffGrp xml:id="s1qe5psk">
       <staffDef xml:id="staffdef-L1F1" n="1" lines="5">
        <clef xml:id="c9i25za" shape="G" line="2" />
       </staffDef>
      </staffGrp>
     </scoreDef>
      <measure xml:id="measure-L1">
       <staff xml:id="staff-L1F1" n="1">
         <note xml:id="note-L4F1" dur="1" oct="4" pname="c" accid.ges="n" />
        </layer>
       </staff>
      </measure>
      <scoreDef xml:id="s1qeiu7a">
       <staffGrp xml:id="s12wir7">
        <staffDef xml:id="ssptl3b" n="1">
         <clef xml:id="clef-L6F1" shape="G" line="2" dis="8" dis.place="below" />
        </staffDef>
       </staffGrp>
      </scoreDef>
      <measure xml:id="measure-L5">
       <staff xml:id="staff-L5F1N1" n="1">
        <layer xml:id="layer-L5F1N1" n="1">
         <note xml:id="note-L7F1" dur="1" oct="4" pname="c" accid.ges="n" />
        </layer>
       </staff>
       <dir xml:id="dir-L8F1" place="above" staff="1" tstamp="5.000000">
        <rend xml:id="r1j1ocah" fontstyle="normal">German</rend>
       </dir>
      </measure>
      <scoreDef xml:id="s14w5oyf">
       <staffGrp xml:id="s1w1lv4e">
        <staffDef xml:id="s1fqa3mj" n="1">
         <clef xml:id="clef-L10F1" shape="GG" line="2" />
        </staffDef>
       </staffGrp>
      </scoreDef>
      <measure xml:id="measure-L9">
       <staff xml:id="staff-L9F1N1" n="1">
        <layer xml:id="layer-L9F1N1" n="1">
         <note xml:id="note-L11F1" dur="1" oct="3" pname="c" accid.ges="n" />
        </layer>
       </staff>
       <dir xml:id="dir-L12F1" place="above" staff="1" tstamp="5.000000">
        <rend xml:id="r5o1dxu" fontstyle="normal">French</rend>
       </dir>
      </measure>
      <scoreDef xml:id="s1v7wims">
       <staffGrp xml:id="s1k0ny2c">
        <staffDef xml:id="svzfobm" n="1">
         <clef xml:id="clef-L14F1" shape="GG" line="2" dis="8" dis.place="below" />
        </staffDef>
       </staffGrp>
      </scoreDef>
      <measure xml:id="measure-L13">
       <staff xml:id="staff-L13F1N1" n="1">
        <layer xml:id="layer-L13F1N1" n="1">
         <note xml:id="note-L15F1" dur="1" oct="4" pname="c" accid.ges="n" />
        </layer>
       </staff>
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

</details>

When rendering to MIDI, verovio currently transposes `clef@shape="GG"` an octave lower (ignoring the `@dis="8"` attribute), which prevents the German usage.  So verovio will have to be fixed one way or the other in order to differentiate between a German or French usage of the `GG` clef.

One alternative would be to create `clef@shape="G G"` for the German usage (a doubled `G` clef) and French usage would be `clef@shape="GG"`, which is equivalent to the vocal tenor clef given in the second measure of the example.  

There are more generalized cases in mensural notation where there are two (or more clefs) and/or mensural signs given, which typically means repeat the music in the different clefs (or that there are two parts singing the same music in different clefs, which is somewhat like the German usage here).

Another possibility would be to use transposition for French style, which is similar to implicitly transposed instruments such as the piccolo, xylophone (glockenspiel/celeste at two octaves up) and double bass (one octave down) and most contra instruments (also one octave down in addition to their transposing interval):

```xml
     <scoreDef xml:id="sa0bopv">
      <staffGrp xml:id="s1jsous8">
       <staffDef xml:id="staffdef-L1F1" n="1" lines="5" trans.diat="-7" trans.semi="-12">
        <clef xml:id="clef-L2F1" shape="GG" line="2" />
       </staffDef>
      </staffGrp>
     </scoreDef>
```

ahankinson:
> The staff does not define `@oct="3"` as a `@loc`, but it is rather defined by the clef. So using it to place the clef is circular logic.

In fact, each note sets its own octave, so I would interpret the presence of `@oct` on `<clef shape="G">` as simply saying what octave the G line is in so that the notes can be rendered on the staff instead of with ledger lines. This is wholly appropriate information for the clef, IMO: since the shape sets the assumed pitch of the line, setting the octave of the line (if it differs from the norm) is complementary information. This is no different than specifying `@line` if the clef is on a different line than it would otherwise be assumed.

The absence of `@oct` means that it is in the "normal" octave for the clef, whatever that is. This is more accurate than using octave _displacement_, since the French "GG" is already assumed to have an octave shift. (the original complaint). 

> One alternative would be to create clef@shape="G G" for the German usage (a doubled G clef) and French usage would be clef@shape="GG", which is equivalent to the vocal tenor clef given in the second measure of the example.

A space in an attribute value means two attributes, like `@class` values in HTML elements; This would probably have to be `G_G` or `G-G` vs `GG`, since it is still one symbol.

Another solution which was brought up by @rettinghaus is to simply encode the "German" G clef as a normal one, and use the SMuFL symbol [`U+E055`](https://smufl-browser.edirom.de/index.html?class=clefsG) as a visual replacement. Also possible.

craigsapp:
> Another solution which was brought up by @rettinghaus is to simply encode the "German" G clef as a normal one, and use the SMuFL symbol [U+E055](https://smufl-browser.edirom.de/index.html?class=clefsG) as a visual replacement. Also possible.

That is an interesting idea.   Is the Weber usage common?   And what about cases where a C or F clef are doubled for the same reason (two parts sharing a staff with C or F clefs)?   If it is common, then that solution would be cumbersome, but otherwise, the meaning of the double G clef in that situation is the same as a single G clef, so it is primarily symbol substitution since there is no transposition involved.

> In fact, ...  assumed.

Maybe more to the point: Notes placed on the same line as the standard G clef are in the 4th octave.  So `clef@oct="4"` is clearly the default octave for G clefs, and `clef@oct="3"` (for C and G clefs) then clearly means notes are to be transposed down an octave when sounding from their written positions (and there is no circularity problem).

Currently in verovio the "GG" clef is automatically assuming to have `clef@oct="3"` (French system), and therefore the music transposes down an octave when sounding (MIDI).   The German system could use `clef@oct="4"` to indicate that the clef does not cause a transposition.  This is probably the best solution.   Does the German system occur in print or only in manuscript?  If only in manuscript then the French system should clearly be the default since I have seen it in print.

Here is another example using `clef@oct`:

Screenshot 2023-02-24 at 2 07 10 AM

<details>
<summary> Click to view MEI data for above example </summary>

```xml
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0.0-dev">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt />
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2023-02-24T02:05:16" version="3.15.0-dev-2f1ce00-dirty">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <workList>
   <work>
    <title />
   </work>
  </workList>
 </meiHead>
 <music>
  <body>
   <mdiv xml:id="m10s92yj">
    <score xml:id="s18dxpmx">
     <scoreDef xml:id="s1meeyyp" midi.bpm="400.000000">
      <staffGrp xml:id="sq5bvrg">
       <staffDef xml:id="staffdef-L1F1" n="1" lines="5">
        <clef xml:id="clef-L2F1" shape="G" line="2" />
       </staffDef>
      </staffGrp>
     </scoreDef>
      <measure xml:id="measure-L1">
       <staff xml:id="staff-L1F1" n="1">
         <note xml:id="note-L4F1" dur="1" oct="4" pname="c" accid.ges="n" />
        </layer>
       </staff>
       <dir xml:id="dir-L5F1" place="above" staff="1" tstamp="5.000000">
        <rend xml:id="r3rkucq" fontstyle="normal">German</rend>
       </dir>
       <dir xml:id="dir-L6F1" place="below" staff="1" tstamp="5.000000">
        <rend xml:id="raapzlu" fontstyle="normal">@oct="4"</rend>
       </dir>
      </measure>
      <scoreDef xml:id="s11qxeyd">
       <staffGrp xml:id="sntb9u2">
        <staffDef xml:id="szrir3j" n="1">
         <clef xml:id="clef-L8F1" shape="GG" line="2" oct="4" />
        </staffDef>
       </staffGrp>
      </scoreDef>
      <measure xml:id="measure-L7">
       <staff xml:id="staff-L7F1N1" n="1">
        <layer xml:id="layer-L7F1N1" n="1">
         <note xml:id="note-L9F1" dur="1" oct="4" pname="c" accid.ges="n" />
        </layer>
       </staff>
       <dir xml:id="dir-L10F1" place="above" staff="1" tstamp="5.000000">
        <rend xml:id="r179cz2j" fontstyle="normal">French</rend>
       </dir>
       <dir xml:id="dir-L11F1" place="below" staff="1" tstamp="5.000000">
        <rend xml:id="r4qbuef" fontstyle="normal">@oct="3"</rend>
       </dir>
      </measure>
      <scoreDef xml:id="sq3qr9y">
       <staffGrp xml:id="s1y2js9l">
        <staffDef xml:id="s1tvu0x4" n="1">
         <clef xml:id="clef-L13F1" shape="GG" line="2" oct="3" />
        </staffDef>
       </staffGrp>
      </scoreDef>
      <measure xml:id="measure-L12" right="end">
       <staff xml:id="staff-L12F1N1" n="1">
        <layer xml:id="layer-L12F1N1" n="1">
         <note xml:id="note-L14F1" dur="1" oct="4" pname="c" accid.ges="n" />
        </layer>
       </staff>
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

</details>

Verovio is not (yet) aware of the `clef@oct` system, so the second measure is being written an octave too high.   

Here is the German system:

```xml
         <clef shape="GG" line="2" oct="4" />
```

And here is the French system (probably should be the default for GG clefs):

```xml
         <clef shape="GG" line="2" oct="3" />
```

ahankinson:
>  then clearly means notes are to be transposed down an octave when sounding from their written positions

Except that `@oct` is on each note too, so you wouldn't actually transpose them by an octave when sounding just because of the clef -- that would be double transposition. The main difference would be whether a renderer could assume a specific octave for the "clef line" or not, and then draw the notes in the right place (on the staff, rather than with ledger lines). 

> probably should be the default for GG clefs

Assuming "French" `@oct="3"` is the default for GG clefs is fine, and then requiring `clef@oct="4"` for "German" clefs would be necessary.

> And what about cases where a C or F clef are doubled for the same reason (two parts sharing a staff with C or F clefs)

I didn't see any SMuFL symbols indicating that C or F clefs can be doubled for the same reason. Not saying that can't happen, just it doesn't seem to be in its repertoire.

craigsapp:
> Except that `@oct` is on each note too, so you wouldn't actually transpose them by an octave when sounding just because of the clef -- that would be double transposition. The main difference would be whether a renderer could assume a specific octave for the "clef line" or not, and then draw the notes in the right place (on the staff, rather than with ledger lines).

Yes, that is why I am getting confused since it is the opposite from the `@dis`/`@dis.place` system: in the dis system `note@oct` is the written pitch and the `clef@dis` indicates the octave transposition needed to get notes to sounding pitch.

In the `clef@oct` system, the `note@oct` is the sounding pitch and the `clef@oct` indicates the transposition needed to get to written pitch.

> The main difference would be whether a renderer could assume a specific octave for the "clef line" or not, and then draw the notes in the right place (on the staff, rather than with ledger lines).

I do not understand this sentence.  Are you meaning that the renderer's job is to automatically decide whether the French or German system is being used for "GG"?   That seems dangerous since there will be corner cases where the renderer's choice will be wrong, and this would also involve analysis of the pitches on the staff, which would be computationally complex and be a waste of time in real-time rendering situations.

N.B.: I do something similar when converting from Humdrum to MEI when there is no clef in the Humdrum data: I calculate the average pitch for the first 20 notes, and then assign treble clef if it is above a threshold (F3 I think) and bass clef otherwise.

craigsapp:
> I didn't see any SMuFL symbols indicating that C or F clefs can be doubled for the same reason. Not saying that can't happen, just it doesn't seem to be in its repertoire.

That is why it is interesting to know if Weber ever doubled the C or F clefs.   I would say that SMuFL is implicitly only aware of the French system, and this would be particularly so if the German system also doubled other clefs for cases where two parts share the same staff.

ahankinson:
> Are you meaning that the renderer's job is to automatically decide whether the French or German system is being used for "GG"?

No. 

I'm saying it's the encoder's job to specify whether, in the case of a G clef, the second line of the staff is in octave 3, or octave 4, since there are two possible interpretations of the meaning of the same shape.

[Renderers must provide implicit octave information for clefs anyway](https://github.com/rism-digital/verovio/blob/develop/src/clef.cpp#L69-L102). The fourth line for an "F" clef is octave 3. 

So I'm saying is that the `@oct` parameter would modify this assumption so that we could distinguish between two clefs that have the same shape but where the "implied" line on the staff is in a different octave.

craigsapp:
The function you quote: https://github.com/rism-digital/verovio/blob/develop/src/clef.cpp#L69-L102
returns the diatonic distance from the bottom line of a staff to middle C. Or in MEI terminlogy, it returns the `@loc` of middle C for the given clef.

For `G2` (treble) clef, it is `-2` since you must go down two steps to get to middle C (middle C is at `@loc="-2"`).  For `F4` (bass) clef it is `+10` diatonic steps to get from the bottom line to middle C (i.e., middle C is at `@loc="10"` for `F4` clefs).

Here is the code that should enable `clef@oct` in verovio:

```cpp
    int offset = 0;
    int defaultOct = 4; // C clef
    if (this->GetShape() == CLEFSHAPE_G) {
        defaultOct = 4;
        offset = -4;
    }
    else if (this->GetShape() == CLEFSHAPE_GG) {
        defaultOct = 3;
        offset = 3;
    }
    else if (this->GetShape() == CLEFSHAPE_F) {
        defaultOct = 3;
        offset = 4;
    }

    if ([HAS_CLEF_OCTAVE_ATTRIBUTE]) {
        int oct = this->GetOct();
        int octDifference = oct - defaultOct;
        offset -= octDifference * 7;
    }
```

I was initially complaining because `clef@oct` is an absolute value, but ultimately the calculation needs a relative octave displacement, which is provided by the `defaultOct` variable in the code above to calculate `octDifference`.

Example for

```xml
<clef shape="GG" line="2">
```

The default offset for this clef is `+5` since middle C is at `@loc="5"` in this clef (in the French interpretation currently being used by verovio).

For this clef:

```xml
<clef shape="GG" line="2" oct="4">
```

This code will be executed:

```cpp
    if ([HAS_CLEF_OCTAVE_ATTRIBUTE]) {
        int oct = this->GetOct();
        int octDifference = oct - defaultOct;
        offset -= octDifference * 7;
    }
```

`oct = 4`, `defaultOct = 3`, so `octDifference = +1`, and this will subtract `7` from the default offset of `+5`, resulting in an updated offset of `-2` (which matches the standard G2 clef's offset value).



> The fourth line for an "F" clef is octave 3.

More precisely the fourth line for an "F4" clef is in octave 3.   But more useful is that whichever line the F clef is on, that line is in octave 3.    So `defaultOct` is set to that 3 for calculating the offset when the shape of the clef is an `F`:

```cpp
    else if (this->GetShape() == CLEFSHAPE_F) {
        defaultOct = 3;
        offset = 4;
    }
```

If `clef@oct` is set, then probably `clef@dis` should be ignored in `Clef::GetClefLocOffset()`.   This is similar to `GG` clefs, where the function does not allow `clef@dis` to be used:

```cpp
    // ignore disPlace for gClef8vbOld
    if (this->GetShape() == CLEFSHAPE_GG) disPlace = 0;
```

lpugin:
Sorry for being late in the discussion but I agree with @craigsapp that it is weird to use `@oct` on clef only for `GG` when we use a different approach for octava clefs.

If it is agreed than `GG` clefs have by default an implicit displacement, why not simply adding a `@dis="0"` when it is a double instrument clef with no displacement ? Then it is more a documentation issue

ahankinson:
> it is weird to use @oct on clef only for GG when we use a different approach for octava clefs.

The difference between `GG` and octava clefs is that, for GG clefs, it's exactly the same symbol, but different octave _interpretations_ of the line the clef intersects, depending on the repertoire. That's not the same as octave displacement. 

In other words, to say that Weber used an octave displacement of "0" (or "8", depending on what you think of as "normal") on his GG clefs isn't correct -- but it would be correct to say that he thought of them as being in a different octave.

Like I said, I'm OK with the interpretation of the `GG` clef to be the "normal" French-style, which is correct in Verovio now. But I really think we need to use `@oct` for the second interpretation of the symbol, and *not* `@dis`.

ahankinson:
Since the PR is just modifying the documentation, here's my take on it that might clarify what I mean. I just added one sentence at the end.

```xml
<desc xml:lang="en">
Double G clef. Sounds one octave lower than G clef. As this clef is usually 
not written with an "8" below the clef, do not combine with <att>dis</att>/<att>dis.place</att>/<att>clef.dis</att>/<att>clef.dis.place</att>. 
For cases where double-G clefs indicate two instruments, and not octave displacement, 
use <att>oct</att> to indicate the sounding octave of the instruments.
</desc>

**Comment by lpugin** [comment]
> The difference between GG and octava clefs is that, for GG clefs, it's exactly the same symbol, but different octave interpretations of the line the clef intersects, depending on the repertoire. That's not the same as octave displacement.

Octava clefs are not the same symbols, and it is an MEI decision to use the same shape attribute value `G` when the symbol is actually not the same. One could therefore question the need of `@dis` and `@dis.place` when we could encode the same with a clef `@shape` such as `G-8` or `G+8`. 

What I am questioning is the use of a different approach for that particular case when `@dis` would work in my opinion. I am certainly not pretending to know what Weber thought and not claiming for this to be the truth, but only to be one way to encode it, to carry the information we need, and to be in line with other similar cases. This being said, using `@oct` works, and I am not opposed to it because it is a quite marginal issue.

**Comment by craigsapp** [comment]
> This being said, using `@oct` works, and I am not opposed to it because it is a quite marginal issue.

<details>
<summary> Click to view complaint about documentation of clef@oct </summary>

One thing that would be useful would be to enhance the `clef@oct` description in the element documentation:

https://music-encoding.org/guidelines/v4/elements/clef.html
Screenshot 2023-02-25 at 1 25 22 PM

Compare to the description for note:

https://music-encoding.org/guidelines/v4/elements/note.html
Screenshot 2023-02-25 at 1 26 32 PM

That description has no meaning in the context of clef.  Better would be something like: "Specifies which octave is represented by the clef's placement line.   For example for the treble clef (G clef on the second staff line) the default octave is 4, and for bass clef (F clef on the fourth staff line) the default octave is 3."

The list of chapters in the prose documentation is good:

Screenshot 2023-02-25 at 1 43 05 PM

But nowhere in that list of places where clefs are discussed can I find mention or examples of `clef@oct` use.

The Remarks section of the element pages might be a useful place:

Screenshot 2023-02-25 at 1 52 00 PM

But I presume this would not be an appropriate place to add examples (images of musical examples).   So perhaps some intermediate level documentation between the Guidelines and Element lists would be useful to discuss such things.

This would be similar to the SCORE documentation.   There is a "User Guide" which functions similar to MEI's "Guidelines" and tutorial, and there is a "Reference Guide" which is organized by element, usually having an introduction, then a detailed discussion about each attribute, and finally some deeper discussions about the element.

Here is the SCORE equivalent of the Element list (https://music-encoding.org/guidelines/v4/elements.html):

Screenshot 2023-02-25 at 2 04 27 PM

Each row is an element and each column is an attribute.   Here is the equivalent to https://music-encoding.org/guidelines/v4/elements/clef.html:

Screenshot 2023-02-25 at 2 06 43 PM

And here is the Reference Manual entry for Clefs, which gives a more verbose description of clefs, while also being systematic about the descriptions of the attributes:

Screenshot 2023-02-25 at 2 15 37 PM

Beginning of the table of contents, where each chapter represents one element:

Screenshot 2023-02-25 at 2 16 01 PM


Screenshot 2023-02-25 at 2 11 06 PM

Screenshot 2023-02-25 at 2 12 03 PM

Screenshot 2023-02-25 at 2 12 42 PM

In such theoretical MEI element-centric documentation for the `<clef>` element, attributes such as `@dis` and `@dis.place` would be discussed as well as `@oct`, giving examples of how these attributes affect the musical data.

</details>

It would be good for verovio to implement `clef@oct`, and it seems that the implementation should be easy (see post https://github.com/music-encoding/music-encoding/pull/1044#issuecomment-1444061142). 

`clef@oct` and `clef@dis`/`clef@dis.place` should not be mixed since they have similar functions but the `note@oct` values will be encoded differently in each system (the two systems cannot be mixed because `clef@oct` has notes encoded at sounding pitch, while `clef@dis` has notes encoded at written pitch).   The `clef@dis` method is an explicit notational system where the 8/15 symbols are shown in the graphical music notation.  The `clef@oct` method is an implicit way of doing the same thing, where there is no direct visual hint that transposition is happening (other than the instrument name).  There are also musical examples where music for tenor voices are displayed with a regular treble clef, but a vocal-tenor clef is implied (a treble clef with an 8 underneath it), so `clef@oct="3"` would be suitable for such situations.

Verovio now only uses the existing 8/15 octave modifiers that are provided in SMuFL (https://www.smufl.org/version/latest/range/clefs/), and does not allow cases where a number is otherwise requested with `@dis`/`@dis.place`, such as a C-clef with an `8` above the clef.   In such cases it would be useful to have `@oct` as a fall-back to indicate the correct vertical placement of notes on the staff.   (These sorts of cases will be very rare.)   In any case, `clef@dis` should have priority if `clef@oct` is also given.

Beyond the use for the French/German usage of the `GG` clef being discussed in this issue, I can see another useful application of `clef@oct` for implicitly octave-transposing instruments, including piccolo, xylophone, glockenspiel (sounding up two octaves), and double bass.   When displaying an orchestral score as a "sounding score", these octave-transposing parts retain their transpositions and are not actually transposed to their sounding pitches (since there would typically be way too many ledger lines that make the music almost unreadable).   It is better to encode these parts at sounding pitch, and then use `clef@oct` to transpose them implicitly.   Their display would be the same in both written and sounding scores, so avoiding using the transposing system.

Here is an example for a double bass:

One way of encoding a double bass score would be to have an octave transposition on the written score:

```xml
       <staffDef n="1" lines="5" trans.diat="-7" trans.semi="-12">
           <clef  shape="F" line="4" />
       </staffDef>
...
         <note dur="1" oct="4" pname="c" accid.ges="n" />
```

The note is written as if it were middle-C:

Screenshot 2023-02-25 at 2 36 36 PM

But when sounding (and rendered in a MIDI file), the written pitch is transposed down an octave.  

When converting to a sounding score, there is a problem: double basses are not transposed to sounding pitch like other non-octave transposing parts, such as B-flat clarinets.   The double-bass parts should retain their octave transpositions.  An algorithm to convert from written to sounding score would either have knowledge about which instruments to transpose or not, or would need to be told to not transpose any part that is purely an octave transposition.  Instruments such as a bass clarinet and  contrabass clarinet, etc.  would probably have an intermediate form: transpose the sub-octave amount, but leave the octave transposition.

Here is the same double bass note using the `clef@oct` method:

```xml
       <staffDef n="1" lines="5">
           <clef  shape="F" line="4" oct="2" />
       </staffDef>
...
         <note dur="1" oct="3" pname="c" accid.ges="n" />
```

In this case there is no transposition applied to the part because it is encoded at sounding pitch.   The `clef@oct="2"` means to write the notes an octave higher than their sounding pitches.   When converting between a written and sound score, the display of such a part would not change since there is no transposing information to apply.

ahankinson:
So far, the only real use case I've seen for `clef@oct` is to distinguish between two clefs, having **exactly the same shape**, but where the interpretation of their **octave position** is different.

> Octava clefs are not the same symbols...

Yes, I agree! I think you misunderstood me, though. I said "The difference between GG and octava clefs is that, **for GG clefs**, it's exactly the same symbol..." -- meaning there was no way, without an extra attribute, to distinguish between the "octave 3" and "octave 4" usages of the "GG" symbol shape. 

Also, if we accept that the "normal" case for `<clef shape="GG">` is that it implies `@oct="3"`, then nothing has to be done with Verovio. This is the default, "French" style.

The suggestion to use `@oct="4"` **would only apply** to the case mentioned by @rettinghaus :

> In Germany the double clef was often used to indicate that two voices share one staff.

This is to clarify that the "German" GG is not a "displacement" clef, but simply a different octave interpretation of the exact same shape -- two uses of the same shape for two different purposes.

lpugin:
I understand we need an attribute to disambiguate two usages of `GG`. As I said, using `@oct` works, and I was only suggesting to use `@dis` and `@displace` and not `@oct` because this is more inline with similar (but not identical) situations. I was also willing to avoid the need to discuss and decide when `@oct` should or can be used and which would be the allowed and not allowed combinations.

ahankinson:
I think the consensus is that this change is good, but that we should add an additional line about using `@oct` to indicate the use of the `GG` clef in a non-octave-displacing manner. @th-we are you able to do this?

ahankinson:
Superseded by #1124

PR #1045, opened 2022-10-30
Bring facs description in line with TEI

rettinghaus:
This PR brings the description of `@facs` in line with the TEI.
Also a link to the `surface` element has been added.


pe-ro (changes requested):
Descriptions typically begin with a capital letter, so "Points" instead of "points".

pe-ro (approved):
Looks good.  Thanks.

bwbohl (approved):

PR #1046, opened 2022-10-30
Fix(docs): Unify spelling of bar line

rettinghaus:
This PR unifies the spelling of bar line. 
Motivation: 

* up to now throughout the guidelines the spellings "barline" and "bar line" were mixed, so this improves consistency
* the element name is `barLine` (combined from "bar" + "line")
* every dictionary I looked it up prefers "bar line" over "barline":
    * [Merriam-Webster](https://www.merriam-webster.com/dictionary/bar%20line)
    * [Cambridge Dictionary](https://dictionary.cambridge.org/dictionary/english/bar-line)
    * [Collins](https://www.collinsdictionary.com/dictionary/english/bar-line)


bwbohl:
thx for making this a separate pull request. I checked and found two other occurrences of `barline` and two occurrences of `barLine` that ought to be fixed. Provided those in a pull request to your branch https://github.com/rettinghaus/music-encoding/pull/3 ;-)

bwbohl (changes requested):
https://github.com/rettinghaus/music-encoding/pull/3

musicEnfanthen:
Just to add some more music-related perspectives:

Elaine Gould uses "barline" in Behind Bars; 
Musescore switched from "bar line" to "barline" in 2015 [1]; 
SMuFL uses "barline" [2] as well as Humdrum documentation [3], music21 [4], and MusicXML [5];
Avid (the company behind Sibelius) and Finale both use "barline" [6][7].

On the other hand, Lilypond uses "bar line" in consistency with their "BarLine" objects [8]. 

In general, I have absolutely no strong feelings about the issue (as long as we do not have any editorial guidelines for the guidelines), but totally agree that the spelling should be at least consistent throughout the documentation. Looking at the usage with related projects above (including company products), there seems to be quite a preference for the compound version. Why not follow that spelling?


[1] https://musescore.org/en/node/77186
[2] http://www.smufl.org/version/1/range/barlines/
[3] https://www.humdrum.org/rep/barlines/index.html
[4] https://web.mit.edu/music21/doc/moduleReference/moduleBar.html
[5] https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/barline/
[6] https://www.avid.com/search#q=barline&t=KB&sort=relevancy
[7] https://usermanuals.finalemusic.com/FinaleWin/Content/Finale/Barlines.htm
[8] https://lilypond.org/doc/v2.21/Documentation/internals/barline

rettinghaus:
@musicEnfanthen Main reason is my second point: `barLine` is the element name not `barline`. 
In British English "barline" seems to be preferred, MEI prefers American English.

Is there the need for more references?
_Beyond MIDI_ uses "bar-line" here and there.

lpugin:
It seems that there is a general trend leaning toward the use of barline. Maybe we should change `barLine` to `barline` in MEI?

(Not directly related, but if we do so maybe we could take the opportunity of breaking compatibility with this change for also renaming `measure@left` and `measure@right` to `measure@barline.left` and `measure@barline.right` ?)

bwbohl (approved):

kepper (approved):
I don't believe that we'll ever achieve full consistency in the language of our guidelines. MEI is a community thing, and I believe it's a good thing that we see different people contributing. Given that, I don't care about this in particular, as the multitude we have seems perfectly comprehensible. Hoewever, I see the argument of improved consistency, and therefore support implementing this. I don't think that this will lead to ultimate consistency once and forever, though ;-)

rettinghaus:
Thanks!

PR #1048, opened 2022-10-31
Fix(MEI.edittrans): Replace special entities with proper links

rettinghaus:


musicEnfanthen:
Thank you @rettinghaus. There seems to be at least another place with special entities in https://github.com/music-encoding/music-encoding/blob/develop/source/modules/MEI.shared.xml#L3769

rettinghaus:
Yes, because this addressed in #1046.

rettinghaus:
Since #1046 is merged, this should be good to go now, @musicEnfanthen.

musicEnfanthen (approved):

musicEnfanthen:
@bwbohl @ahankinson Could one of you do a second review?

bwbohl (approved):

ahankinson (approved):
Nice catch @rettinghaus

PR #1050, opened 2022-10-31
Replace rng:empty with tei:empty

rettinghaus:
This is a small step towards pure odd. 
Not affecting any script we have in the repo and shouldn't affect external things.


bwbohl:
You're quite correct that we should make this move. NEvertheless I would like to see this in a general step to so coalled "pure ODD". Adding to Mileston MEI 6

rettinghaus:
Added to #991, so closing this PR.

bwbohl:
I don't see, how merging this into a #991 helps in the end… leaving this one open was fine from my point of view

PR #1029, opened 2022-10-05
Make tempo member of att.extender

rettinghaus:
Tempo changes are often expressed with extender lines:

image

However, this was yet missing in MEI.

PR #1055, opened 2022-12-05
BREAKING(MEI.shared): Move att.extender from tempo to att.tempo.vis

rettinghaus:
Instead of having `att.extender` directly in `tempo` this moves it to `att.tempo.vis` being more in line with `dir` and `dynam`. 

NB: The latter are also member of `att.verticalGroup`. Should `tempo` also be a member?

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -8059,7 +8059,6 @@
     <classes>
       <memberOf key="att.common"/>
       <memberOf key="att.bibl"/>
-      <memberOf key="att.extender"/>
       <memberOf key="att.facsimile"/>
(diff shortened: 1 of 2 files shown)


ahankinson:
Is there a problem this is solving? Or a future use for this? What is the motivation for this change?

rettinghaus:
As implied in the description. `tempo` is a special case of `dir`, so it should be structured in the same manner.

ahankinson (approved):

musicEnfanthen:
Just for reference: This is a follow-up to #1029.

Thanks @rettinghaus

musicEnfanthen (approved):

PR #1063, opened 2022-12-30
Add notationType to manifestation and extend value list

riedde:
Co-authored-by: @music-encoding/ig-metadata

At the last meeting of the metadata ig in 2022 we discussed the need of having a possibility to encode the notation type within the header. @KristinaRichts came up with some usecases and arguments for implementing this. During the meeting we figured out that the best place, from our view, is the manifestation element. To keep consistency we decided to propose using the existing `@notationType` with an extension instead of creating something new. The extension of the value list (datatype) at `@notationType` is necessary because the scope of the values isn't wide enough for describing the music and (that's the bigger issue I think) the `valList/@type` is closed!

So, the metadata-ig proposes to set the value list to semi and add values for an extension (list is provided by @irmlindcapelle). Also the proposed extension of the value list certainly does not contain all possible notation types. Nevertheless, this beginning should also be an invitation to complete the list over time.

This PR contains
- adding att.notationType to manifestation
- setting `valList/@type` to semi instead of close
- adding an extention for the datatype/valItems
- adding a notice in documentation (proposals for rewording are welcome!)

### DIFF
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -1284,7 +1284,7 @@
             </div>
             <div xml:id="msdesc" type="div2">
                <head>Encoding Sources in MEI</head>
-               <p>The <gi scheme="MEI">manifestation</gi> and <gi scheme="MEI">item</gi> elements allow detailed description of various types of sources, for instance, a printed text or manuscript, another computer file, an audio or video recording, or a combination of these. Both <gi scheme="MEI">manifestation</gi> and <gi scheme="MEI">item</gi> are part of the <ptr target="#FRBR"/> implementation in MEI. Please note: in MEI 3.0.0, the <gi scheme="MEI">source</gi> element was used to capture this type of information. The <gi scheme="MEI">manifestation</gi> element may contain the following elements:</p>
+               <p>The <gi scheme="MEI">manifestation</gi> and <gi scheme="MEI">item</gi> elements allow detailed description of various types of sources, for instance, a printed text or manuscript, another computer file, an audio or video recording, or a combination of these. Also the format of the notation can be classified within <gi scheme="MEI">manifestation</gi> using <att>notationType</att> and <att>notationSubtype</att>. Both <gi scheme="MEI">manifestation</gi> and <gi scheme="MEI">item</gi> are part of the <ptr target="#FRBR"/> implementation in MEI. Please note: in MEI 3.0.0, the <gi scheme="MEI">source</gi> element was used to capture this type of information. The <gi scheme="MEI">manifestation</gi> element may contain the following elements:</p>
                <p>
                   <specList>
                      <specDesc key="head"/>
--- a/source/modules/MEI.frbr.xml
+++ b/source/modules/MEI.frbr.xml
@@ -310,6 +310,7 @@
       <memberOf key="att.bibl"/>
       <memberOf key="att.componentType"/>
       <memberOf key="att.dataPointing"/>
+      <memberOf key="att.notationType"/>
       <memberOf key="att.pointing"/>
       <memberOf key="att.recordType"/>
       <memberOf key="att.targetEval"/>
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -2932,7 +2932,7 @@
   <macroSpec ident="data.NOTATIONTYPE" module="MEI" type="dt">
     <desc xml:lang="en">Notation type and subtype</desc>
     <content>
-      <valList type="closed">
+      <valList type="semi">
         <valItem ident="cmn">
           <desc xml:lang="en">Common Music Notation.</desc>
         </valItem>
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -2962,6 +2962,34 @@
       </valList>
     </content>
   </macroSpec>
+  <macroSpec ident="data.NOTATIONTYPE.extended" module="MEI" type="dt">
+    <desc xml:lang="en">Notation type extention</desc>
+    <content>
+      <valList type="semi">
+        <valItem ident="byzantine">
+          <desc xml:lang="en">Byzantine notation.</desc>
+        </valItem>
+        <valItem ident="choral">
+          <desc xml:lang="en">Choral/Square notation.</desc>
+        </valItem>
+        <valItem ident="graphic">
+          <desc xml:lang="en">Graphical notation.</desc>
+        </valItem>
+        <valItem ident="greek">
+          <desc xml:lang="en">Greek notation.</desc>
+        </valItem>
+        <valItem ident="modal">
+          <desc xml:lang="en">Modal notation.</desc>
+        </valItem>
+        <valItem ident="oldslavic">
+          <desc xml:lang="en">Old Slavic notation.</desc>
+        </valItem>
+        <valItem ident="ottoman">
+          <desc xml:lang="en">Ottoman notation.</desc>
+        </valItem>
+      </valList>
+    </content>
+  </macroSpec>
   <macroSpec ident="data.NOTEHEADMODIFIER" module="MEI" type="dt">
     <desc xml:lang="en">Captures any notehead "modifiers"; that is, symbols added to the notehead, such as
       slashes, lines, text, and enclosures, etc.</desc>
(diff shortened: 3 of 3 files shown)


ahankinson:
Typically this sort of use case would be handled with the `<classification>` element, which is already available in the header and on the manifestation. `@notationType` was added to the music body to allow renderers to differentiate between specific classes and subclasses of notation, e.g., white vs black mensural, or gothic vs. square neumes so that they might be displayed properly. So `@notationType` is more “data” than “metadata” (as problematic as that sounds…)

Was the possibility of using `<classification>` discussed? What were the problems with it?

riedde:
Thank you @ahankinson you for your comment! Yes, we discussed the use of `<classification>`, but I can't remember the arguments against this solution. I'll bring it up at the next meeting of @music-encoding/ig-metadata

riedde:
Report from IG-Meeting: Reusing `@notationType` was ment as a short cut (or easy alternative) to `<classification>` but should not exclude using it!

From me:

> @notationType was added to the music body to allow renderers to differentiate between specific classes and subclasses of notation

The idea was to allow recording and to "differentiate between specific classes and subclasses of notation" in the header. So from my view the idea and definition of `@notationType` does not exclude using it in this contexts. But I'm open to feedback!

After having a second eye on the attribute definition of `@notationType` I realized that it charactatrizes an element that contains notation. While `<manifestation>` doesn't contain notation (directly) I also feel a bit insecure with this proposal. 

However, the other thing proposed here was the extension of the value list for `@notationType` and to open the list (from `closed` to `semi`). Independent from what will happen to the first part of this PR I would like to keep the second part.

lpugin:
> However, the other thing proposed here was the extension of the value list for `@notationType` and to open the list (from closed to semi). Independent from what will happen to the first part of this PR I would like to keep the second part.

Whereas `@notationtype` should be an open or semi-open list was extensively discussed when this attribute was introduced. It was agreed to keep it close because its purpose is precisely to allow for a broad but univocal qualification of the notation type. Some flexibility was maintained by adding `@notationsubtype`, which is not constraint in that way and that can be used freely to describe the notation in a more refined way.

If you think there are some values missing in the list they should certainly be discussed, but I would be against opening the list because it seems to me that the flexibility you are looking for is available in other ways.

ahankinson:
Thanks @riedde !

> However, the other thing proposed here was the extension of the value list for @notationType and to open the list (from closed to semi). Independent from what will happen to the first part of this PR I would like to keep the second part.

Could you say a bit more about where the values from the list came from? For instance, where does the value "choral/square" come from? And the list itself is not exhaustive: I'm not sure how you would describe "common notation", for example, with this system? What is the difference between "Byzantine" and "Greek" notation?

Is this attempting to describe the system of notation independent of visuals, or a system of notation for a given repertoire? i.e., would "square" and "st-gall" notation the same because they are both "chant" notation, or are they different because they have a different visual appearance? 

Having a fixed set of values for notation types also does not permit the use of URIs or external vocabularies --  this is the biggest advantage of using the `<classification>` method. Like you said, this doesn't necessarily preclude the use of `<classification>` but it also means that MEI itself becomes its own "authority" on what notation types there are by providing this list, but in this method it is not done in a way that can be re-used by others -- we're not publishing our own URIs or classification scheme in a reusable way.

> @KristinaRichts came up with some usecases and arguments for implementing this.

Maybe these can be shared here somehow so we can review them?

ahankinson:
It also occurs to me that there are perhaps two different issues here: One is to define a set of notation types, and the other is to figure out how to put them into MEI. 

A shareable authority for notation types doesn't *really* exist, so that would be a pretty big accomplishment if we were able to define one. There is the [RDA Form of Musical Notation](http://www.rdaregistry.info/termList/MusNotation/) but it is pretty sparse, and a bit weird in what it describes. If a new vocabulary could be published as a separate shared vocabulary, perhaps led by the IG, that would be something that could be very useful!

The second is, given such a vocabulary, how do you use it in MEI? If we define a shareable vocabulary of notation types, then I think the `<classification>` method would still be most appropriate. We might think about a separate short-hand, like `<manifestation notationClass="http://some/uri/blah">` but I think, in this case, `@notationType` should probably be left on its own -- it's main purpose is for the use of renderers, which need a much more constrained set of values than all possible notation types.

Would you also want to have multiple notation types per manifestation?

riedde:
Thank you very much @lpugin and @ahankinson for the input, that helps me very much to open my view on this! I'll try to answer all your comments.

First of all: I totally agree that we are talking about different issues here. And I also see that `@notationType` is not suitable for what we (@music-encoding/ig-metadata) had intended.

> @lpugin If you think there are some values missing in the list they should certainly be discussed, but I would be against opening the list because it seems to me that the flexibility you are looking for is available in other ways.
> @ahankinson `@notationType` [...] -- it's main purpose is for the use of renderers, which need a much more constrained set of values than all possible notation types.

Yes, concerning opening the list I can absolutely follow @lpugin's argument. Maybe this all came up because the attribute definition of `@notationType` is very general and so the question came up, why less notation types are present in its value list. What about adding a remark at the definition of `@notationType`  that clarifies that what @ahankinson commented on this? Because when this is clear the value list is exhaustive from my view (until other notation types can be rendered).

> @ahankinson Could you say a bit more about where the values from the list came from?

The list has been made by @irmlindcapelle and I think it came from the DFG-Viewer project context. They had catalog entries for, e.g., byzantine music that uses a different notation system. Maybe @irmlindcapelle or @KristinaRichts can comment on this more precicely

>  And the list itself is not exhaustive

I have not claimed that either ;) I know that it's not, but it could be the beginning of a set of classifications. 

>  I'm not sure how you would describe "common notation", for example, with this system?

Sorry, I didn't get it. There is still the value `cmn`. If not, it's a bug and I failed on writing the ODD!

> What is the difference between "Byzantine" and "Greek" notation?

I'm absolutely not an expert for that, but a quick look on some facsimiles made me thinking that it is a system with a special set of characters for music notation (as cmn, mensural or neumes are). And as I get it it's not just a different visual representation but a different, yeah, system.

> Is this attempting to describe the system of notation independent of visuals, or a system of notation for a given repertoire? i.e., would "square" and "st-gall" notation the same because they are both "chant" notation, or are they different because they have a different visual appearance?

That's a really good point and has to be clarified. As I got it was meant as a classification of systems of notation and not just an other visual representation. However, "chant" for example is a special form of music/notation, but should actually always be clarified using a `subtype`. This for example is a case where a multi-level classification could be necessary...

> Would you also want to have multiple notation types per manifestation?

I wouldn't say that this is common, but can certainly happen...

All in all, I think @ahankinson pointed it already out:
> The second is, given such a vocabulary, how do you use it in MEI? If we define a shareable vocabulary of notation types, then I think the `<classification>` method would still be most appropriate. We might think about a separate short-hand, like `<manifestation notationClass="http://some/uri/blah">` but I think, in this case, `@notationType` should probably be left on its own [...].

ahankinson:
> Sorry, I didn't get it. There is still the value cmn. If not, it's a bug and I failed on writing the ODD!

Oh! My mistake. I didn't see the hidden bits in GitHub, so I wasn't aware of the "regular" values that were still there. Apologies.

This highlights an artificial difference between "normal" and "extended" notations, though -- one that is very Western-centric. I would think we would try to put all notations on equal footing in any "standard" that we try to put forward.

> I wouldn't say that this is common, but can certainly happen...

I would say this is very common -- tablature and staff notation often co-exist. Mixed cmn + mensural notations are also quite common, particularly in sources during the "transitional" periods. For non-Western notations, usually it's CMN with additional characters for microtones, or parallel transcriptions.

lpugin:
Maybe one more point related to the background of `@notationtype`. It was introduced for cases where it is not really possible to make the difference of the visual appearance of different notations when considering the music MEI encoding. For example, if you have
```
<ligature>
   <note dur="brevis" oct="4" pname="c" />
   <note dur="brevis" oct="3" pname="a" />
</ligature>
```
Is the visual representation encoded or intended this?
image
or that?
image

This is the purpose of `mensural.white` and `mensural.black`. 

In other words, the list contains values for which different notation types cannot be (or can hardly be) distinguished otherwise. I think this is a fairly atypical purpose. If you want to have additional values in there is is probably be helpful if you can make a similar case.

riedde:
Thank you for the constructive discussion!

PR #1066, opened 2023-01-26
Move lg to text module

### LINKED ISSUE #1051: Revise line and line group
Currently `lg` and `l` mix the visual appearance of a group of text lines with the semantic structure of verses. 
I'd like to propose to take the original descriptions from TEI to make it clear that these are elements to encode lines of verses and you still need formatting elements like `lb`.

rettinghaus:
This PR brings the descriptions of `lg` and `l` in line with TEI. Also it moves `lg` into the `MEI.text` module.
Closes #1051

### DIFF
--- a/source/modules/MEI.text.xml
+++ b/source/modules/MEI.text.xml
@@ -185,8 +185,8 @@
     </remarks>
   </elementSpec>
   <elementSpec ident="l" module="MEI.text">
-    <gloss versionDate="2022-05-18" xml:lang="en">line of text</gloss>
-    <desc xml:lang="en">Contains a single line of text within a line group.</desc>
+    <gloss versionDate="2023-01-26" xml:lang="en">verse line</gloss>
+    <desc versionDate="2023-01-26" xml:lang="en">Contains a single, possibly incomplete, line of verse.</desc>
     <classes>
       <memberOf key="att.common"/>
       <memberOf key="att.facsimile"/>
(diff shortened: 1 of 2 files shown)


source/modules/MEI.text.xml
```diff
@@ -223,6 +223,41 @@
       <p>The model of this element is based on the <ref target="https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-l.html">l</ref> element of the Text Encoding Initiative (TEI).</p>
     </remarks>
   </elementSpec>
+  <elementSpec ident="lg" module="MEI.shared">
```
ahankinson:
```suggestion
  <elementSpec ident="lg" module="MEI.text">
```

source/modules/MEI.text.xml
```diff
@@ -185,8 +185,8 @@
     </remarks>
   </elementSpec>
   <elementSpec ident="l" module="MEI.text">
-    <gloss versionDate="2022-05-18" xml:lang="en">line of text</gloss>
-    <desc xml:lang="en">Contains a single line of text within a line group.</desc>
+    <gloss versionDate="2023-01-26" xml:lang="en">verse line</gloss>
+    <desc versionDate="2023-01-26" xml:lang="en">Contains a single, possibly incomplete, line of verse.</desc>
```
ahankinson, line 189: I'm OK with the internal move of the element to the text module, but am hesitating to accept this PR because you're also changing the user-facing description, and therefore the semantics of the element itself at the same time. Is there a specific issue you are trying to address here?

source/modules/MEI.text.xml
```diff
@@ -223,6 +223,41 @@
       <p>The model of this element is based on the <ref target="https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-l.html">l</ref> element of the Text Encoding Initiative (TEI).</p>
     </remarks>
   </elementSpec>
+  <elementSpec ident="lg" module="MEI.shared">
+    <gloss versionDate="2022-05-18" xml:lang="en">line group</gloss>
+    <desc versionDate="2023-01-26" xml:lang="en">Contains one or more verse lines functioning as a formal unit, e.g. a stanza, refrain,
```
ahankinson, line 228: Also here -- why are you changing the description of the element?

ahankinson (changes requested):

ahankinson:
Fundamentally I think we need to diverge from TEI here. As TEI is a markup language, they are concerned mostly with paragraphs and, in their case, they deal with the concept of "lines" and "line groups" in verse markup.

MEI, in dealing primarily with music, has much less of a "markup" concept, and where text is used it's typically outside of a prose setting. Re-framing it as "verse" will confuse our users because of the musical concept of "verses" and "verse groups", while the original description made it clear that any line / line group of text can be marked up with these elements.

If you think otherwise it's a valid issue for the community, but it shouldn't be combined in a "mechanical" PR that re-organizes the internal module structure, since it's changing the definition and therefore the semantics of the elements.

source/modules/MEI.text.xml
rettinghaus: thanks

bwbohl:
2023-02-23 ODD meeting:
The change seems plausible, although we might want to revisit element-module-association globally at some point. @ahankinson's caveat to intentionally diverge from the TEI definition to accommodate other "non-lyrical" text lines is generally agreed upon.
Moreover we will have to fix the guidelines accordingly, by moving the element descriptions from the Shared description to text.

ahankinson:
From MEI Developer meeting: Line and Line Group were in `shared` because they were intended to be shared across modules and customizations. Furthermore, the changes to the definitions of the elements to align with TEI were rejected.

Given a similar sentiment in the ODD meeting, we will close this PR. Please re-open if there are any objections.

PR #1074, opened 2023-02-06
BREAKING(MEI.header): Remodel acquisition after provenance

### LINKED ISSUE #1072: Model of `<acquisition>`
Recently I noticed that the model of `<acquisition>` is different from all other "history-like" elements now that #911 has been merged. In particular, `<acquisition>` is now the only element in this series that does not allow `<eventList>` as a child element, even though its model is based on the TEI [acquisition](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-acquisition.html) element which allows a TEI `<listEvent>` element. 

Is there an important reason for this that I am missing? 
Otherwise, I would suggest to just remodel `<acquisition>` after provenence.

doerners:
This PR changes the model of `<acquisition>` to be in line with the other 'history-like' elements and the respective model in TEI.
fixes #1072

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -380,7 +380,25 @@
       <memberOf key="att.lang"/>
     </classes>
     <content>
-      <rng:ref name="macro.struc-unstrucContent"/>
+      <rng:choice>
+        <rng:group>
+          <rng:zeroOrMore>
+            <rng:ref name="model.headLike"/>
+          </rng:zeroOrMore>
+          <rng:zeroOrMore>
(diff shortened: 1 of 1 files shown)


ahankinson (approved):

ahankinson:
Looks good to me

bwbohl (approved):

musicEnfanthen:
Thanks everyone!

bwbohl:
Maybe we could create a macro for all history elements and reference it from within their content elements. This way we would guarantee that their model always stays identical.

PR #1080, opened 2023-02-14
Fix(docs): Remove confusing paragraph about attributes

### LINKED ISSUE #1079: Change parsing instructions in Guidelines for note attributes
Referencing discussion #1071 , this statement in [Section 4.2.5.1.1](https://music-encoding.org/guidelines/dev/content/cmn.html#cmnNotesBasic) seems to be causing some confusion about how certain attributes are to be parsed and inherited.

"Because these attributes may not be required in all situations (such as `@dur` for the notes of a chord), processing software should anticipate retrieving the information that would have been provided by missing attributes from a preceding `<note>` or `<chord>` parent in the same `<layer>`. Only information from `@pname`, `@oct` and `@dur` attributes can be gathered in this fashion. No other attributes can be treated this way."

I propose that we simply delete this paragraph altogether, since it's pretty clear that inheriting these attributes from previous (peer) notes is highly problematic. 

The case for inheriting `@dur` on `<chord>` is made later in [Section 4.2.5.2.1](https://music-encoding.org/guidelines/dev/content/cmn.html#cmnChordsBasic) so I don't think we need to replace this.

ahankinson:
Removes the paragraph about attributes being inherited.

Fixes #1079
Refs #1071


rettinghaus (approved):
Thanks!

musicEnfanthen (approved):

bwbohl (approved):

PR #1100, opened 2023-04-28
Add ellipse as text enclosing

rettinghaus:
This PR makes the distinction between circle and ellipse as text enclosing explicit. 

Before it was not possible to encode differences in layout like the following: 

image

image

This becomes evidently if you try to reproduce this with a renderer, and for longer texts you only have the option "circle".

image

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -3576,10 +3576,13 @@
           <desc xml:lang="en">Enclosed in box.</desc>
         </valItem>
         <valItem ident="circle">
-          <desc xml:lang="en">Enclosed in ellipse/circle.</desc>
+          <desc xml:lang="en">Enclosed in circle.</desc>
         </valItem>
         <valItem ident="dbox">
           <desc xml:lang="en">Enclosed in diamond.</desc>
+        </valItem>
+        <valItem ident="ellipse">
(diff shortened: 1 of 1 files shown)


musicEnfanthen:
Thank you @rettinghaus . Is there an open issue for that PR? If not, may I ask you to create it and describe the underlying issue very briefly?

ahankinson:
Conversation at MEI Developers meeting (24.05.2023): The distinction between "circle" and "ellipse" are not necessarily clear from a rendering perspective. We don't make the distinction between "rectangle" and "square", for example, in the same list. We also don't define the type of triangle used for the "triangle" attribute.

@pe-ro suggests that the intention of this value is more "_en_circle" rather than specifically rendering a perfect circle, and that the choice of drawing a circle vs. an ellipse should be a choice left to the rendering engine to avoid collisions and other considerations.

So we're a -1 on accepting this PR.

pe-ro:
MEI Development Meeting 25/05/2023 -- the consensus is to reject this request based on our understanding of current practice and the desire to limit the number of shape classes -- square-like, circle-like, triangle-like.  Given specific use cases, the issue may be reopened at a later date.

craigsapp:
See example need for differentiating "circle" and "circle-like": https://github.com/rism-digital/verovio/pull/3454#issuecomment-1596840198

Issue #579, opened 2019-03-14
What happened to content model of `<source>`?


raffazizzi:
I was looking at [`<source>`](https://github.com/music-encoding/music-encoding/blob/develop/source/modules/MEI.header.xml#L2320) in v4 and noticed that it cannot contain `<identifier>` and `<titleStmt>` any longer. Is this intentional?

[The examples on the Guidelines](https://music-encoding.org/guidelines/v4/content/header.html#headerSourceDescription) still show those elements as children of `<source>`

kepper:
Hi Raff, yes, this is intentional. In v4, the double use of `<source>` for a) describing the material used to encode the file at hand and b) a frbr:manifestation has been removed. Now, `<source>` serves only the first purpose, while the latter is dealt with by a new `<manifestation>` element. `<source>` allows bibliographic content (`<head>`, `<bibl>`, `<biblStruct>`,…), while `<manifestation>` is closer to the  model of `<source>` in v3. 
However, I absolutely agree that this isn't clear at all. The problem is that our documentation isn't anywhere close to matching the state of v4. This affects both the Guidelines and the examples. Right now, Benni and I (as this year's Tech Chairs) are trying to get a better picture of the situation and come up with a plan to resolve this asap. Until the situation is resolved, the release of v4 is still an ongoing process, and we have to make it much clearer that this is still work in progress…

bwbohl:
The discussion leading towards this descision (in part) took l´place in:
* https://github.com/music-encoding/music-encoding/issues/478
* https://github.com/music-encoding/music-encoding/issues/61
* https://github.com/music-encoding/music-encoding/issues/546

ahankinson:
Can this be closed now?

bwbohl:
closing as now referenced from guidelines issue

Issue #546, opened 2018-09-23
Clarification of FRBR levels / source


kepper:
On various occasions, members of the MEI Community have expressed confusion that `<workDesc>` is a first-level child of `<meiHead>`, while `<sourceDesc>` is nested into `<fileDesc>`. This has historical reasons: Before the introduction of FRBR into MEI, `<source>` was used to describe both the material used in the preparation of the file, and sources (prints, manuscripts, …) for which there is an editorial interest of some kind. Sometimes, these two fell together, and sometimes they didn't. When implementing the FRBR model, these two use cases were not differentiated clearly. As such a change to the MEI model would clearly break backwards compatibility, and at the same time affect a great number of MEI instances, it seems good to not implement it in the upcoming MEI 4.0.0, but instead in a later revision of MEI. At the same time, MEI 4.0.0 could include a deprecation warning, giving users and developers sufficient room for discussing a potential solution and testing its consequences on their use of MEI.

pe-ro:
Here's what I have so far --

```
<!-- Clarification of FRBR levels / source, issue #546 -->
<!-- Replace the former content of sourceDesc to allow bibliographic citations -->
<elementSpec ident="sourceDesc" module="MEI.header" mode="replace">
  <desc>(source description) – A container for the descriptions of the source(s) used in the
    creation of the electronic file.</desc>
  <classes>
    <memberOf key="att.common"/>
  </classes>
  <content>
    <rng:zeroOrMore>
      <rng:ref name="model.headLike"/>
    </rng:zeroOrMore>
    <rng:choice>
      <rng:ref name="bornDigital"/>
      <rng:oneOrMore>
        <rng:ref name="model.biblLike"/>
        <!--<rng:choice>
                <rng:ref name="bibl"/>
                <rng:ref name="biblStruct"/>
              </rng:choice>-->
      </rng:oneOrMore>
    </rng:choice>
  </content>
</elementSpec>
<!-- Add an element to explicitly capture the fact that the MEI file has no bibliographic sources -->
<elementSpec ident="bornDigital" module="MEI.header" mode="add">
  <desc>(born digital) – Used to explicitly indicate that the MEI file relies on no other
    sources.</desc>
  <classes>
    <memberOf key="att.common"/>
    <memberOf key="att.bibl"/>
  </classes>
  <content>
    <rng:empty/>
  </content>
</elementSpec>
<!-- Add element to capture structured bibliographic citations -->
[...]

Issue #478, opened 2017-11-09
Suggested revision of <sourceDesc>


BruxDDay:
Preface:

As the MEI header has become more commonly adapted to the creation of catalogs or metadata focused applications separate from files of MEI music notation encoding we see a problem with a dual interpretation of the element `sourceDesc`

The two roles can be summarized as follows:

1. The original intent of `sourceDesc` was to identify materials used to created digital music notation. This usage does not require a physical description of a specific document (FRBR Item), rather only a bibliographic identification. With this intent in mind, it makes sense to locate the element as part of the `fileDesc` element

2. The evolving use of `sourceDesc` is not linked to music notation necessarily, but mostly a physical description of a specific source/document. It may involve a very detailed physical and (and possibly other aspects) description. This use of `sourceDesc` would apply to a collection catalog. Maybe it would also apply to thematic catalog but the descriptive elements would be more selectively applied.

Suggested solutions:

1. Change the name of `sourceDesc` in the `fileDesc` to `sourceIdentifer`

2. Create a new level in the MEI Header titled `sourceDesc` independent of both `fileDesc` and `workDesc.` In this new level develop and refine elements that could be used to describe a physical document (independent of the abstract work described in `workDesc`.)

3. Place the new `sourceDesc` category after `workDesc` in the MEI hierarchy. In this sense the new`sourceDesc` will correspond with the "Item" level of a FRBR description. Manuscript sources will be unique by definition, and detailed descriptions of specific print sources would also be unique and correspond with "Item" in the FRBR hierarchy. This new use of `sourceDesc` might require some rethinking of `@analog` to describe FRBR relationships.

4. We further suggest that `revisionDesc` should be relocated as a sub-element of `fileDesc` because it is intended to describe revisions to MEI code, not musical revisions in a specific document. It would also be good to rename the current `revisionDesc` to `codeRevisionDesc` in order to avoid any confusion with a future use of `revisionDesc` in the new category of `sourceDesc`

bwbohl:
I've tried to tackle this huge issue a couple of times now… but still don't manage to address everything. Something that might be connected to this is #61 (allow `<bibl>` in `<source>` as an alternative to the structured elements)

BruxDDay:
[...]

PR #1122, opened 2023-05-23
schema: allow physDesc on Source

### LINKED ISSUE #1078: Encoding a physical description outside the FRBR module not possible
Various refactoring steps since MEI 3.0 have made it not possible to use `<physDesc>` outside the FRBR module. This is problematic for cases where the metadata is about a single source and where providing some metadata about the physical description is desirable. In MEI 3.0, it was possible to do
```xml
<source>
   <titleStmt>
      <title>Soprani</title>
   </titleStmt>
   <physDesc>
      <!-- ... -->
   </physDesc>
</source>
```
and I think this is still desirable. 

Furthermore, `<physDesc>` is part of the `header` module, and it seems awkward that it cannot be used without `FRBR`. See a related discussion [here](https://github.com/music-encoding/music-encoding/issues/803#issuecomment-850473806)

I would suggest to investigate a possible fix by allowing (back) `<physDesc>` within `<source>` but I have not evaluated the consequences of this. Any thoughts?

ahankinson:
Also removed a reference to a 'manifestation' since this module is independent of the FRBR module.

Built locally and looks like it works OK.

Fixes #1078

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -2500,6 +2500,9 @@
           <rng:ref name="model.biblLike"/>
         </rng:choice>
       </rng:zeroOrMore>
+      <rng:optional>
+        <rng:ref name="physDesc"/>
+      </rng:optional>
     </content>
     <constraintSpec ident="check_source_target" scheme="schematron">
       <constraint>
@@ -2516,9 +2519,7 @@
     </constraintSpec>
     <remarks xml:lang="en">
       <p>This element contains, or references via its <att>target</att> attribute, a description of
-        a source used in the creation of the electronic file. For description of a physical
-        embodiment of an expression of a work use the <gi scheme="MEI">manifestation</gi>
-        element.</p>
+        a source used in the creation of the electronic file.</p>
       <p>The <att>data</att> attribute may be used to reference one or more musical features found
         in the content of this particular source.</p>
     </remarks>
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -96,6 +96,7 @@
                <head>Introduction</head>
                <p>Metadata means "data about data", <abbr>i.e.</abbr>, information about various aspects of an encoding at hand. There are many different types of metadata, which MEI tries to order according to their respective scope or perspective, as described in <ptr target="#headerstructure"/>. MEI’s approach to metadata is heavily influenced by other existing standards and models, such as TEI, MARC, and FRBR. It attempts to reflect both current library practice and common scholarly methods, for example in the field of source descriptions (see chapter <ptr target="#msdesc"/>).</p>
                <p>This chapter thus addresses the description of an encoded item so that the musical text, as well as its sources, encoding, and revisions are all thoroughly documented. Such documentation is necessary for scholars using the texts, for software processing them, and for catalogers in libraries and archives. Together these descriptions and declarations provide an electronic analog to the title page attached to a printed work. They also constitute an equivalent for the content of the code books or introductory manuals customarily accompanying electronic data sets.</p>
+               <p>For large encoding projects, the Functional Requirements for Bibliographic Records (<abbr>FRBR</abbr>) capabilities of MEI provides most flexibility and descriptive capabilities. However, MEI also provides simpler, non-FRBR methods for source description for those cases where the FRBR capbilities are not needed. These are further discussed in <ptr target="nonfrbrsourcedescription" />./></p>
             </div>
             <div xml:id="headerstructure" type="div2">
                <head>Structure of the MEI Header</head>
@@ -115,6 +116,18 @@
                   <item>This concept is covered in section <ptr target="#headerusecases"/>.</item>
                   <item>A <hi rend="italic">revision history</hi>, tagged <gi scheme="MEI">revisionDesc</gi>, which allows the encoder to provide a history of changes made during the development of the electronic text. The revision history is important for version control and for resolving questions about the history of a file. The MEI elements used to encode the revision description are described in section <ptr target="#headerRevisionDescription"/>.</item>
                </list>
+               <div xml:id="nonfrbrsourcedescription" type="div3">
+                  <head>Non-FRBR Source Description</head>
+                  <p>MEI allows two methodologies to be used when doing metadata capture and source description. The <abbr>FRBR</abbr> methodology, described later in this chapter, gives encoders a mechanism for building detailed descriptions of bibliographic entities. Most bibliographic-based projects should favor the FRBR method.</p>
+                  <p>However, there are cases where detailed bibliographic analysis at the level required by the FRBR module is not possible or desirable. These may include situations such as retro-conversion of non-FRBR catalogues, or capturing the output of notation encoding software where only minimal source description capabilities are available. In these cases, encoders may capture metadata in the MEI header without using FRBR encoding entities and instead choose a simplified methodology for source description.</p>
+                  <p>The <gi scheme="MEI">sourceDesc</gi> element as a child of the <gi scheme="MEI">fileDesc</gi> element may be used to capture simple bibliographic descriptions. An example is given below.</p>
+                  <p>
+                     <figure>
+                        <head/>
+                        <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/header/header-sample080.txt" parse="text"/></egXML>
+                     </figure>
+                  </p>
+               </div>
             </div>
             <div xml:id="commonconcepts" type="div2">
                <head>Common Metadata Concepts</head>
--- a/source/examples/header/header-sample080.txt
+++ b/source/examples/header/header-sample080.txt
@@ -1,0 +1,28 @@
+<fileDesc>
+    <titleStmt>
+        <title>La clé du caveau</title>
+    </titleStmt>
+    <pubStmt>
+        <publisher>Capelle &amp; Renand</publisher>
+        <pubPlace>Paris</pubPlace>
+    </pubStmt>
+    <sourceDesc>
+        <source>
+            <physDesc>
+                <dimensions>
+                    <width unit="cm">9.5</width>
+                    <height unit="cm">14</height>
+                </dimensions>
+                <extent unit="page">376p</extent>
+            </physDesc>
+            <physLoc>
+                <identifier>XA Li 1811/1</identifier>
+                <repository>
+                    <corpName>Universität Bern, Institut für Musikwissenschaft</corpName>
+                    <geogName>Bern</geogName>
+                    <country>Switzerland</country>
+                </repository>
+            </physLoc>
+        </source>
+    </sourceDesc>
+</fileDesc>


kepper:
What I would like to see here, in addition to the change in the specs, is some more documentation, maybe along the lines of https://music-encoding.org/guidelines/dev/content/metadata.html#headerRelatedItemVsFRBR. We should ideally give example(s) on how `<source>` and `<manifestation> / <item>` should be combined, and what the purpose of both is. https://music-encoding.org/guidelines/dev/content/metadata.html#msdesc is just a stub, and work on that is underway, but it seems necessary to have that clarification somewhere already if we want to merge this now (which makes perfect sense – it should be possible to give details about a source without having to use full FRBR…)

ahankinson:
@kepper In addition to the "Related Item vs FRBR" section, the [introduction to the FRBR module](https://music-encoding.org/guidelines/dev/content/metadata.html#FRBR) contains the following:

"MEI header information may refer to different levels of description of the encoded work: Some information may apply the work in all its various forms and realizations, e.g., the name of its composer. Other information may describe a certain version of the work, or a source such as the printed first edition, or only a single copy of that source. Core MEI limits the header information to two such levels of description: work and source, respectively.

However, when the FRBR module is available more detailed descriptions are possible. ..."

So I didn't add anything more since it seems we already cover this. Personally, I would like to stay away from any prescriptive descriptions of how these things can be combined, since I prefer to see it as an either / or (use FRBR, or don't) rather than trying to mix them in the same encoding. For keeping them separate, sections 3.3 and 3.4 seem to already cover most of the ground (particularly §3.4.1.6.)

source/modules/MEI.header.xml
```diff
@@ -2516,9 +2519,7 @@
     </constraintSpec>
     <remarks xml:lang="en">
       <p>This element contains, or references via its <att>target</att> attribute, a description of
-        a source used in the creation of the electronic file. For description of a physical
```
riedde, line 2519: I'm not sure if reducing documentation here will help with comprehensibility but I guess it is proposed because the element definition defines the source element as a bibliographic description, so this sentence seems a bit weird here. So, it makes sense, absolutely! Thanks!

riedde (dismissed):
I dropped just one thought in a comment. Thanks for your work!

source/modules/MEI.header.xml
ahankinson: 👍

doerners (dismissed):

rettinghaus:
I'm not ok with this change. Not allowing physDesc in source came with a good reason, changing this back causes more confusion. I still don't see why this should be important now.

ahankinson:
>  Not allowing physDesc in source came with a good reason

Do you remember what the reason was? As far as we can tell it was dropped with the addition of the FRBR module, but not sure if that was done intentionally. As a result it means we cannot do non-FRBR source description, which seems like a bug.

ahankinson:
It was possible, for example, in MEI 3, but was removed from MEI 4. https://music-encoding.org/guidelines/v3/elements/source.html

rettinghaus:
Yes, the idea was to have full FRBR support and not to confuse manifestation with source. 
I'm a bit confused now that @riedde approved this, because I think we came in the metadata IG to another conclusion before.

ahankinson:
There is still full FRBR support. However, without this change there is no longer non-FRBR support. That is the issue. Sometimes we need to do source description without performing a full analysis of whether something is describing it at the manifestation or the item level -- we just want to describe the source that we transcribed it from.

If you want FRBR support, it still exists and this does not change it.

rettinghaus:
Who is we and why want you do it this way? I still think it's a feature, not a bug. FRBR was implemented this way to not allow non-FRBR.

ahankinson:
The issue is described in #1078. 

The background is that we wanted to encode incipits in MEI, with some minimal RISM data in the MEI header, so that we could support basic metadata input in Verovio. (https://github.com/rism-digital/verovio/pull/3278).

The FRBR hierarchy can be useful, but sometimes it's like trying to drink water from a firehose, and you just want to capture a few attributes about a single source.

lpugin (dismissed):
Looks good to me.

rettinghaus:
Perfect reason for a customization. The rest is silence.

musicEnfanthen:
Maybe it could help to add some explanations/remarks to `source` to clarify the intended (non-FRBR) usage and point to manifestation for FRBR usage (and maybe even vice versa)? To make sure "not to confuse manifestation with source" as @rettinghaus pointed out.

kepper:
Not sure it helps, but here's another personal take on this. From my perspective, FRBR can be very helpful to describe the available sources / dissemination / known situation of a work. However, whenever this concept of a work is challenged for any reason (be it practical cataloguing issues as in Andrew's situation, be it for a very detailed look at small chunks of sketch material as in my case), the strong assumptions that FRBR makes start to stand in the way – they're just not as helpful anymore as they promise to be at first sight. However, this doesn't mean they cannot be used to describe available sources anymore. This requires a level of investigation that may exceed the available resources, though. According to my understanding (which seems to be covered by the docs), `<sourceDesc>` is meant to capture the material that was considered in the preparation of the encoding at hand. That material may or may not be a manifestation singleton / item encoded elsewhere. In our project, we encode such material in both places, as those seem to be different statements. Obviously, we're not providing two `<physDesc>`s in that case, but have a very light encoding on one end, which is then linked to the other. I believe it is perfectly sensible to not take away the possibility to specify what your encoding is based on, even when you don't find the time to deal with all the intricacies of FRBR – which can be really painful in some situations. I therefore think this is a really good and necessary improvement, which will lead to better encodings.

kepper (dismissed):

riedde:
@rettinghaus yes, we discussed this in the @music-encoding/ig-metadata meeting and it was a very contrasting discussion. As far as I remember we arrived at a stalemate situation because we didn't find reasons to do that change (from our perspective). For me these reasons are given in the discussion above.

Yes I'm also not absolutely happy with a `<physDesc>` in `<source>`, but there is a documentation what `<source>` and `<manifestation>` and `<item>` are for. Further those are different use-cases and the `MEI.header` is independent from `MEI.frbr`.

The only thing that is a bit weird to me is that `<source>` is defined as _A bibliographic description of a source used in the creation of the electronic file._ And `<physDesc>` would allow to make a physical description inside of a bibliographic description. Therefore I wouldn't use it this way, but it might be helpful in a customization that does not contain `MEI.frbr` module.

Conclution: For me, the arguments outweigh the counterarguments

doerners:
Personally I'm +1 with @rettinghaus here. My biggest concern is that this change leads the way to use source as a direct substitute for manifestation. My understanding of the source element is, however, that it fulfils a different task than the manifestation element. I base this interpretation partially on the element definition of a similar element in EAD. However, my impression is that any counterargument or concern is not perceived as having the same weight as the pro arguments. Hence, the approval on this PR.

ahankinson:
> And <physDesc> would allow to make a physical description inside of a bibliographic description

Physical descriptions are part of bibliographic descriptions. Size and extent are very common features of bibliographic description; colour and other features less so, but are certainly not unheard of. A good chunk of the [3XX tag range](https://www.loc.gov/marc/bibliographic/bd3xx.html) and [5XX tag range](https://www.loc.gov/marc/bibliographic/bd5xx.html) in MARC21 deals with physical description.

riedde:
@ahankinson thanks! I didn't know that.

ahankinson:
Developer meeting, 25.05.2023: After a fairly extensive conversation, we decided that `<source>` and `<sourceDesc>` need a major rework.

The general consensus and motivation is that encoders should be able to describe a source without using FRBR structures (manifestation / item, specifically). However, after doing the example it was discovered that `<physLoc>` was also not available on `<source>` anymore. 

Looking back through earlier versions of MEI, it seems that most of the source description capabilities were removed in 4.0. The content models for MEI 1.8, MEI 2, and MEI 3 were consulted. In these previous versions it seems that there was confusion with FRBR elements being allowed within `<source>`, which seems strange.

Further investigations covered an investigation of `<bibl>` and `<biblStruct>`, but these have very weird content models as well. (e.g., `<pubStmt>` and `<imprint>` being two different ways of expressing a publication.)

It was agreed that a re-work of the `<source>` and `<sourceDesc>` elements was needed, and so we should convert this PR to a WIP draft while we consult more broadly with the community, particularly @music-encoding/ig-metadata, about this issue.

musicEnfanthen:
Issues and discussions related to this: #579 and #546

doerners:
Another older issue related to this: #478

pe-ro:
I think it's important to recognize the difference between bibliographic **citation** and bibliographic **description**.  The `<bibl>` and `<biblStruct>` elements address bibliographic citation; that is, a bibliographic reference.  In-depth description of the thing referred to is not generally accommodated in a reference, e.g., references in a paper don't contain the dimensions of the referenced item.  By allowing `<bibl>` and `<biblStruct>`, `<source>` is permitting citation. On the other hand, the FRBR elements, `<work>`, `<expression>`, `<manifestation>`, and `<item>`, allow for description.

That being said, the issue suggests a need for description where currently only citation is allowed.  Allowing additional descriptive elements within `<bibl>` and `<biblStruct>` would permit `<source>` to keep its current model, but would create additional problems elsewhere, since it would make it possible to put bibliographic description virtually everywhere.

I think the better path is to insert a variant of the content model of `<manifestation>` into the current model of `<source>`:
```
<content>
  <rng:zeroOrMore>
    <rng:ref name="model.headLike"/>
  </rng:zeroOrMore>
  <rng:zeroOrMore>
    <rng:choice>
      <rng:ref name="locus"/>
      <rng:ref name="locusGrp"/>
    </rng:choice>
  </rng:zeroOrMore>
  <rng:choice>
    <rng:zeroOrMore>
      <rng:choice>
        <rng:ref name="model.biblLike"/>
      </rng:choice>
    </rng:zeroOrMore>
    <rng:group>
      <rng:optional>
        <rng:ref name="model.identifierLike"/>
      </rng:optional>
      <rng:ref name="titleStmt"/>
      <rng:optional>
        <rng:ref name="model.workIdent"/>
      </rng:optional>
      <rng:ref name="macro.bibldescPart"/>
      <rng:optional>
        <rng:ref name="creation"/>
      </rng:optional>
      <rng:optional>
        <rng:ref name="history"/>
      </rng:optional>
      <rng:optional>
        <rng:ref name="langUsage"/>
      </rng:optional>
      <rng:optional>
        <rng:ref name="contents"/>
      </rng:optional>
      <rng:optional>
        <rng:ref name="notesStmt"/>
      </rng:optional>
      <rng:optional>
        <rng:ref name="classification"/>
      </rng:optional>
      <rng:zeroOrMore>
        <rng:ref name="relatedItem"/>
      </rng:zeroOrMore>
      <rng:zeroOrMore>
        <rng:ref name="extMeta"/>
      </rng:zeroOrMore>
    </rng:group>
  </rng:choice>
</content>
```
Then the encoder has a choice of using a bibliographic citation (i.e., `<bibl>` or `<biblStruct>`), or a more extensive bibliographic description.

bwbohl:
@ahankinson & @lpugin is could you provide some samples of the

>  incipits in MEI, with some minimal RISM data in the MEI header

for clarification. Thanks

bwbohl:
just a link or a plaintext with the metadata that should go into the MEI would be sufficient ;-)

bwbohl:
ping

pe-ro:
I think the issue this PR attempts to address is resolved by adding `<biblFull>` to `<source>`, which I believe is under discussion in the metadata IG.  I would like to table this PR in favor of #1366 and #1367.

riedde:
> I think the issue this PR attempts to address is resolved by adding `<biblFull>` to `<source>`, which I believe is under discussion in the metadata IG. I would like to table this PR in favor of #1366 and #1367.

Yes, I also think this is resolved by adding `<biblFull>` to `<source>`, which is already done-

PR #1130, opened 2023-05-23
spread out constraints to affected elements

### LINKED ISSUE #1129: not all Schematron processed in Guidelines generation
On https://music-encoding.org/guidelines/dev/elements/typeNote.html, there is no Schematron displayed, even though there should be some, according to the sources of `<typeNote>` (see below). When processing locally, the XSLT is also dropping that – I'll investigate.

https://github.com/music-encoding/music-encoding/blob/7386533bf85c70c13d4f399018d34949d291eb62/source/modules/MEI.msDesc.xml#L859-L880.

kepper:
fixes #1129

This is not the most elegant solution, but it works. Basically, the Schematron rule added on watermark doesn't do anything when msDesc is turned off, but also doesn't do any harm either. When msDesc is turned on, it works as expected.

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -2915,6 +2915,15 @@
     <content>
       <rng:ref name="macro.struc-unstrucContent"/>
     </content>
+    <constraintSpec ident="check_watermark_inline" scheme="isoschematron">
+      <constraint>
+        <sch:rule
+          context="mei:watermark">
+          <sch:assert test="ancestor::mei:physDesc">The watermark element may only appear as a
+            descendant of the physDesc element.</sch:assert>
+        </sch:rule>
+      </constraint>
+    </constraintSpec>
     <remarks xml:lang="en">
       <p>The <att>facs</att> attribute may be used to record the location of the watermark in a
         facsimile image.</p>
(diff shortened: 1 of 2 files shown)


ahankinson (approved):

source/modules/MEI.header.xml
```diff
@@ -2915,6 +2915,15 @@
     <content>
       <rng:ref name="macro.struc-unstrucContent"/>
     </content>
+    <constraintSpec ident="check_watermark_inline" scheme="isoschematron">
+      <constraint>
+        <sch:rule
+          context="mei:watermark">
+          <sch:assert test="ancestor::mei:physDesc">The watermark element may only appear as a
```
musicEnfanthen, line 2922: Use `<sch:name/>` instead of element name in all occasions?

pe-ro: @musicEnfanthen , good idea, but would require a lot of human review. Better covered under a different issue.

musicEnfanthen: OK, after discussion at Dev Workshop, we decided not to go with `<sch:name/>` for now.

musicEnfanthen (approved):

PR #1134, opened 2023-05-23
fix(build): move schemata into correct folder and add canonicalized.xml

### LINKED ISSUE #1133: build: schemata are deployed into wrong (nested) folder
The deploy action deploys the schemata, built with the new build workflow, into a wrong folder. It creates a `schemata` folder inside `dev` instead of pushing the files directly to the `dev` folder.

### LINKED ISSUE #1138: Add canonicalized source to build assets
The canonicalized source of MEI should be added to the release assets as it forms the ideal starting point for creating customizations. Compared to the source in the repo – which is split into multiple files for housekeeping purposes – the canonicalized source resolves all the xincludes and thus is a real **O**ne-**D**ocument-**D**oes-it-All.

musicEnfanthen:
This PR fixes the step of the deploy action that copies the built schema into the expected folder in the schema repo. Previously, the schemata folder itself, created during the build, was copied to the dev folder. Now only the content of the schemata folder should be found in dev.

It additionally copies the missing mei-source_canonicalized.xml to the schema folder.

Fixes #1133 
Fixes #1138

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -59,7 +59,10 @@ jobs:
           path: ${{ env.SCHEMA_DIR }}
 
       - name: Copy built schema to SCHEMA_DIR
-        run: cp -r dist/schemata/ $SCHEMA_DIR/dev/
+        run: | 
+          cp -r dist/schemata/ $SCHEMA_DIR/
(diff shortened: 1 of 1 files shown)


kepper (approved):
looks good to me

bwbohl (approved):

PR #1142, opened 2023-05-24
Fixed: Add glyph.auth and glyph.uri to rend

ahankinson:
Addresses result of conversation in #704.

The att.extSym class was broken into two sub-classes so that we could only add rend to one of them, without also pulling in glyph.name and glyph.num. No other elements should be affected by this change.

### DIFF
--- a/source/modules/MEI.externalsymbols.xml
+++ b/source/modules/MEI.externalsymbols.xml
@@ -24,7 +24,7 @@
       </attDef>
       <attDef ident="glyph.uri" usage="opt">
         <desc xml:lang="en">The web-accessible location of the controlled vocabulary from which the value of
-          <att>glyph.name</att> or <att>glyph.num</att> is taken, or the textual content of the element</desc>
+          <att>glyph.name</att> or <att>glyph.num</att> is taken, or the textual content of the element.</desc>
         <datatype>
           <rng:ref name="data.URI"/>
         </datatype>
--- a/source/modules/MEI.externalsymbols.xml
+++ b/source/modules/MEI.externalsymbols.xml
@@ -7,13 +7,12 @@
   <moduleSpec ident="MEI.externalsymbols">
     <desc xml:lang="en">External symbols component declarations.</desc>
   </moduleSpec>
-  <classSpec ident="att.extSym" module="MEI.externalsymbols" type="atts">
-    <desc xml:lang="en">Attributes used to associate MEI features with corresponding glyphs in an
-      externally-defined standard such as SMuFL.</desc>
+  <classSpec ident="att.extSym.auth" module="MEI.externalsymbols" type="atts">
+    <desc xml:lang="en">Attributes that point to an external symbol authority.</desc>
     <attList>
       <attDef ident="glyph.auth" usage="opt">
         <desc xml:lang="en">A name or label associated with the controlled vocabulary from which the value of
-            <att>glyph.name</att> or <att>glyph.num</att> is taken.</desc>
+          <att>glyph.name</att> or <att>glyph.num</att> is taken, or the textual content of the element.</desc>
         <datatype>
           <rng:data type="NMTOKEN"/>
         </datatype>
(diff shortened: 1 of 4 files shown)


jcdevaney:
This looks like a good solution to #704

source/modules/MEI.shared.xml
```diff
@@ -7282,7 +7283,9 @@
     </attList>
     <remarks xml:lang="en">
       <p>When an entire element should be rendered in a special way, a style sheet function should
-        be used instead of the <gi scheme="MEI">rend</gi> element.</p>
+        be used instead of the <gi scheme="MEI">rend</gi> element. The <att>glyph.auth</att> and <att>glyph.uri</att>
```
musicEnfanthen, line 7286: Could it be helpful to add this last addition about the use of glyph.auth and glyph.uri to the textual guidelines as well (chapter [9.2.2](https://music-encoding.org/guidelines/dev/content/textencoding.html#sharedTextRendition)) together with an example of how to use it, e.g., with SMuFL?

bwbohl:
I like the proposal but wondered whether we should harmonise the attribute names with att.authorized (`@auth`, `@auth.uri`, `@codedval`)?
I think the attributes can be compared as given in this table:
| glyph |  auth |
|--------|--------|
| glyph.auth | auth |
| glyph.name | ? if allowd: label ? |
| glyph.num | codedval |
| glyph.uri | auth.uri | 

Harmonising these could mean any of the following:
* change `@glyph.name` to `@glyph.label` 
* change `@glyph.num` to `@glyph.codedval` as this would allow for non-numeric identifiers

Or even 
* introduce something like `@auth.name`?

More important maybe would be harmonising the attribute definition, e.g. in the case of the `.uri`attributes

| glyph.uri | auth.uri |
|--------|--------|
| The web-accessible location of the controlled vocabulary from which the value of glyph.name or glyph.num is taken. Value conforms to [data.URI](https://music-encoding.org/guidelines/dev/data-types/data.URI.html). | A web-accessible location of the controlled vocabulary or other authoritative source of identification or definition for this element or its content. This attribute may contain a complete URI or a partial URI which is completed by the value of the codedval attribute. Value conforms to [data.URI](https://music-encoding.org/guidelines/dev/data-types/data.URI.html). | 

The definition or documentation of glyph attributes should also indicate wheter I can take the glyph.num form a vocabulary and reference that with glyph.uri but pick a glyph.name deliberately or if that one also hast to come from the _same_ authority.

ahankinson:
@bwbohl The URI / name is specifically about the glyphs in the tag. `@auth.uri` is more about external authorities for other things (people, etc.) So we should stick with the `@glyph` attributes for now.

If you're suggesting that we also change the names of the `@glyph` attributes, I think this should be handled in another issue.

bwbohl:
I think we can go on with the PR. To not forget about the useful Guidelines comments I also put it in the MEI Guidelines project.

bwbohl (dismissed):

jcdevaney (dismissed):
Looks good, I approve

ahankinson:
Wait!

ahankinson:
I’ve got a paragraph for the guidelines almost ready to go.

source/modules/MEI.externalsymbols.xml
```diff
@@ -23,6 +22,18 @@
           </valItem>
         </valList>
       </attDef>
+      <attDef ident="glyph.uri" usage="opt">
+        <desc xml:lang="en">The web-accessible location of the controlled vocabulary from which the value of
+          <att>glyph.name</att> or <att>glyph.num</att> is taken, or the textual content of the element</desc>
```
musicEnfanthen:
```suggestion
          <att>glyph.name</att> or <att>glyph.num</att> is taken, or the textual content of the element.</desc>
```

bwbohl (approved):

musicEnfanthen (approved):

PR #1148, opened 2023-05-25
Cleanup Verovio examples

### LINKED ISSUE #1146: Warnings from Verovio and Prince on live rendered example file
The file `source/examples/verovio/tempo-01.mei` ([line 46](https://github.com/music-encoding/music-encoding/blob/develop/source/examples/verovio/tempo-01.mei#L46)) triggers a warning when rendered with Verovio to be included in the guidelines during the build process:

`[Warning] Using rend@fontname with 'VerovioText' is deprecated. Use 'rend@fontfam="smufl"' instead`

Would simple replacement be sufficient, or would we need to add the font-family "smufl" to the [source/assets/css/print/fonts.css](https://github.com/music-encoding/music-encoding/blob/develop/source/assets/css/print/fonts.css) as well, @kepper?

The same file and line are triggering another warning in Prince, since the used unicode for the half note in the tempo indication is not correctly resolved. Is there a better way to do it with SMuFL, @lpugin ?

lpugin:
Does the following:
* Update the Schema to 5.0.0-dev
* Make formatting consistent
* Add `@n` to all `layer` elements
* Change `rend@fonfname="VerovioText"` to `rend@fontfam="smufl"` (Fixes #1146)

PS `rend@fontfam` will have to be changed to `rend@glyph.auth` once #1142 is implemented in Verovio

### DIFF
--- a/source/examples/verovio/accid-03.mei
+++ b/source/examples/verovio/accid-03.mei
@@ -1,5 +1,8 @@
-<?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="http://music-encoding.org/schema/4.0.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="4.0.0">
-   <meiHead meiversion="4.0.0">
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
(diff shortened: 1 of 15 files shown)


musicEnfanthen (approved):
Before merging this one, we should update the docker image with PR https://github.com/music-encoding/docker-mei/pull/11 to have the latest Verovio version in the build pipeline. 

Tested locally with the updated local docker image, works perfectly.

musicEnfanthen:
@bwbohl Can you please double check:` ant generate-images` runs fine for me with this PR. However, when I run `ant build-guidelines-html` with this PR, java output throws:

`ERROR: Unable to parse the following XML snippet, which is apparently not well-formed:` (And then listing all the mei example files with html-escaped angled brackets (`&lt;` instead of `<`)

It later generates and copies the images as usual, so seems to be only an internal warning? But adds a lot of lines to the already very verbose output.

musicEnfanthen:
@kepper Do we need to add smufl as font-face in [source/assets/css/print/fonts.css](https://github.com/music-encoding/music-encoding/blob/develop/source/assets/css/print/fonts.css) ?

ahankinson (approved):

PR #1150, opened 2023-05-25
Add "fences" as a notehead modifier

ahankinson:
This PR changes the notehead modifier list to add the value of `fences` (and removes `dblwhole`). This is as a result of the discussions in #867 and discussions at the MEI Developers Meeting 25.05.2023. 

Refs #865.

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -3014,7 +3014,7 @@
         <valItem ident="circle">
           <desc xml:lang="en">Enclosing circle.</desc>
         </valItem>
-        <valItem ident="dblwhole">
+        <valItem ident="fences">
           <desc xml:lang="en">Enclosing "fences".</desc>
         </valItem>
       </valList>


pe-ro (approved):
Looks good to me.

lpugin (approved):

rettinghaus:
As said before, this is a note head by its own. You cannot add randomly "fences" to all note heads.

-1 from side.

ahankinson:
There are other notehead shapes that have fences on them. 

https://w3c.github.io/smufl/latest/tables/noteheads.html

"Triangle notehead up double whole"
"X notehead double whole"

So using the `fences` notehead modifier we can have:

```xml
<note dur="breve" head.shape="x" head.modifier="fences" />
```

To represent this head shape. But we cannot do this if `head.shape` is just `dblwhole`.

rettinghaus:
You don't need to be able, because these shape note heads only have a singular glyph for breve. So 

`<note dur="breve" head.shape="x" />`

should be sufficient. I want semantic differentiation between "old style" and "new style" breves.

lpugin:
In that case may I ask why doing:
```xml
<note dur="breve" glyph.auth="smufl" glyph.name="noteheadDoubleWhole"/>
```
and
```xml
<note dur="breve" glyph.auth="smufl" glyph.name="noteheadDoubleWholeSquare"/>
```
is not enough for making this distinction? I am not sure there is much "semantics" difference between the two - a breve is a breve. It looks like it is essentially a visual, or engraving style, difference, isn't it?

bwbohl (approved):

ahankinson:
Discussion 25.05.2023: Agree that @rettinghaus' comments are valid, and that not all noteheads need to be enclosed in fences. The suggestion of @lpugin is a very useful direction, but we observed that a `<note>` is composed of several SMuFL glyphs, so having one `glyph.name` value isn't sufficient for capturing all possible glyphs that could compose a single note.

What was proposed was that we might explore changing `@head.shape` to take SMuFL codepoint names instead of MEI-specific values. This would seem to address @rettinghaus' desire to specify a specific notehead shape ("old style" vs. "new style"). This change is out of scope for MEI 5, but a new issue will be opened for MEI 6. 

We also discussed a possible set of values for `@glyph.auth`. `smufl` is the obvious one, but we observed that also `unicode` could be another valid authority, since Unicode does contain note heads and other valid glyphs that could be used as a glyph codepoint. We also recognized that it wouldn't be possible to mix glyph authorities for different noteheads and stems since only one `glyph.auth` is available.

The limited nature of this specific PR, changing `dblwhole` to `fences` to align the values of the attribute with its description is taken as a simple bugfix, and while not sufficient for addressing the whole problem, does remove the confusion of what `@head.mod="dblwhole"` is supposed to encode. 

So the result of the discussion is:
 - approve this PR as a limited bugfix, but recognize that it doesn't actually address the full problem
 - open a new issue to explore changing `head.shape` to accept SMuFL in order to specify the visual shape of the notehead from an external authority.

Sound OK?

pe-ro:
Seems fine to me.

Issue #767, opened 2021-02-04
@tstamp2 is missing in metaMark


krHERO:
while `@tstamp` is allowed, `@tstamp2` is missing in `metaMark` .
btw: since `@startid` and `@endid` are allowed, `@tstamp2` should also be added.
https://music-encoding.org/guidelines/v4/elements/metamark.html

rettinghaus:
In latest develop `@tstamp2` is available (though `@tstamp2.ges` and `@tstamp2.real` are not). 
What needs to be done here?

kepper:
has been solved in the past. If someone needs `@tstamp2.ges` or `@tstamp2.real` in the future, this can be opened again.

PR #1151, opened 2023-05-25
allow tstamp2.ges and tstamp2.real on metaMark

kepper:
In #797 and #767 we were too fast in removing `@tstamp2.ges` and `@tstamp2.real` from `<metaMark>`, which was found when evaluating the diff of v4 and v5. Adding them back in, as agreed on the Charlottesville dev meeting.


musicEnfanthen (approved):

rettinghaus (approved):

pe-ro (approved):
Looks good to me.

PR #1155, opened 2023-05-26
Fixed: instrDef was misspelled

### LINKED ISSUE #1112: Adding `instrDef` to MEI basic
As it stand, `instrDef` is not available in MEI basic since nothing from the MIDI module has been included.

I can see basic use cases where having `instrDef` would be useful, typically for providing more information about to which instrument a `staffDef` or a `staffGrp` relates and to encode some MIDI channel information.

My questions are:
* What speaks against having `instrDef` in MEI basic?
* If not, is there anything else from the MIDI module that should be included too?

ahankinson:
The instrDef element was requested as part of MEI Basic, but it was misspelled so it was not included.

Fixes #1112


rettinghaus (approved):

musicEnfanthen (approved):
Good catch

lpugin (approved):

PR #1157, opened 2023-05-26
schema: resolve att def duplications

kepper:
There are more than 100 attributes sharing their names with other attDefs. Most of them are perfectly sensible, because they are used in different contexts, and have quite different models – they just share their names, which are perfectly appropriate in their context. Sometimes, however, `<attDef>`s have just been copied over to not interfere with the class system. These cases are maintenance issues, which this PR tries to resolve. None of these changes should affect MEI instances.


rettinghaus:
This is super, thanks. Please give me the time to have a proper look at this.

musicEnfanthen (dismissed):
Thanks @kepper for this consolidation work.

kepper:
Just for reference: I had another look at mensural stems, and I believe it's best to leave the situation as it is for now. We can revise as soon as we have the discussion around #1159 (which I'll push to someday…)

source/modules/MEI.gestural.xml
```diff
@@ -607,6 +569,30 @@
...
+          <rng:ref name="data.OCTAVE"/>
+        </datatype>
+      </attDef>
+      <attDef ident="pname.ges" usage="opt">
+        <desc xml:lang="en">Contains a performed pitch name that differs from the written value.</desc>
+        <datatype>
+          <rng:ref name="data.PITCHNAME.GES"/>
+        </datatype>
+      </attDef>
+      <attDef ident="pnum" usage="opt">
```
rettinghaus, line 611: Shouldn't this be in the MIDI module? Something like `att.pitch.midi`?

source/modules/MEI.cmn.xml
```diff
@@ -1140,6 +1116,22 @@
       </attDef>
     </attList>
   </classSpec>
+  <classSpec ident="att.tremForm" module="MEI.cmn" type="atts">
```
rettinghaus, line 1130: Why not `att.tremolo.log`? `att.tremForm` looks weird somehow.

source/modules/MEI.shared.xml
```diff
@@ -1317,6 +1317,17 @@
       <memberOf key="att.pitched"/>
     </classes>
   </classSpec>
+  <classSpec ident="att.keyMode" module="MEI.shared" type="atts">
```
rettinghaus, line 1349: Thinking about the proposed changes for `keySig`/`key` it might be better to name it `att.key.anl`.

rettinghaus (changes requested):
Looks great, thanks for the work! Some ideas for improvements.

rettinghaus:
Also I'm a bit hesitant to extract `att.coordinated.ul`. Why is `att.coordinated` inappropriate there?

source/modules/MEI.gestural.xml
ahankinson: It's not just a MIDI number; could also be base-40 as per the description.

source/modules/MEI.cmn.xml
ahankinson: While I agree in principle with @rettinghaus, we actually don't have any `att.tremolo.*` classes, and the only other one is `att.tremMeasured`. So I would say this change is at least consistent with what we currently have for tremolo, even if it isn't completely consistent with the rest of the spec.

source/modules/MEI.shared.xml
ahankinson: We currently don't have any other `att.key.*` attributes, but the other `att.key*` classes we have (`att.keyAccid`, `att.keySig`, `att.keySigDefault`) all have the domain appended. So perhaps `att.keyMode.anl`?

source/modules/MEI.neumes.xml
```diff
@@ -185,6 +177,42 @@
       <memberOf key="att.sylText"/>
     </classes>
   </classSpec>
+  <classSpec ident="att.neumetype" module="MEI.neumes" type="atts">
```
musicEnfanthen: Following other nomenclature this should probably be `neumeType` ?

source/modules/MEI.analytical.xml
```diff
@@ -353,6 +346,7 @@
       <memberOf key="att.harmonicFunction"/>
       <memberOf key="att.intervalMelodic"/>
       <memberOf key="att.melodicFunction"/>
+      <memberOf key="att.neumetype"/>
```
musicEnfanthen: See comment below for `neumeType`

musicEnfanthen:
@kepper Would you have time to address conflicts and comments?

source/modules/MEI.gestural.xml
kepper: I agree with @ahankinson, this seems to be broader than MIDI…

source/modules/MEI.cmn.xml
kepper: To be honest, I hadn't invested a whole lot of time into considering the best name for this. With a closer look now, I come to the conclusion that there is actually way more to do, and thus to discuss, than just this name. The (proposed) `att.tremForm` is decribed as _Attributes describing the form of a tremolo_. Then we have an attribute name _form_, which is described as _describes the style of the tremolo_. Putting aside the slight inconsistencies, that seems to indicate a _visual_ statement. The purpose, however, is a _logical_ one. So, the situation seems to require much more consideration than just the name of the att class, and I don't think we should do that _now_ – we clearly have many other situations like this, which we would want to fix as well, as soon as we run into them.

ahankinson: Visual and logical can be the same… I don’t think we need to complicate this. Can we proceed with this change and open a new issue for the additional post-5 changes?

source/modules/MEI.shared.xml
rettinghaus: Please compare #1210

source/modules/MEI.gestural.xml
bwbohl: Also agree

source/modules/MEI.cmn.xml
bwbohl: I think this is ok for now as it has a local consistency, similar to what was acceptable for us in #1173 and resulting in #1194 for MEI6

source/modules/MEI.shared.xml
bwbohl: To be resolved in context of #1194 in MEI 6

source/modules/MEI.neumes.xml
bwbohl:
```suggestion
  <classSpec ident="att.neumeType" module="MEI.neumes" type="atts">
```

source/modules/MEI.analytical.xml
bwbohl:
```suggestion
      <memberOf key="att.neumeType"/>
```

bwbohl: @kepper

source/modules/MEI.neumes.xml
bwbohl: @kepper

bwbohl:
ping @kepper

kepper:
> Also I'm a bit hesitant to extract `att.coordinated.ul`. Why is `att.coordinated` inappropriate there?

`<graphic>` has it's own `@ulx` and `@uly` definitions, which seems inappropriate. As I didn't want to change the footprint of the the schema, using `att.coordinated` on `<graphic>` would be incorrect, so separating out the upper left coordinates in a class of their own seems to be the better way. Of course it is also possible to discuss whether `<graphic>` should use other attributes, but that seems like a bigger change than what we have so far. Also, it seems like combining the new `att.coordinated.ul` with `@width` and `@height` might allow other approaches to describe rectangles in the future (not saying this is good or bad – just saying it becomes possible…).

bwbohl:
@kepper could you please resolve conflicts and address the comments above?

source/modules/MEI.neumes.xml
bwbohl: yes probably it would be best to leave this as is and fix in MEI 6

source/modules/MEI.analytical.xml
```diff
@@ -362,6 +356,9 @@
   </classSpec>
   <classSpec ident="att.neume.anl" module="MEI.analytical" type="atts">
     <desc xml:lang="en">Analytical domain attributes.</desc>
+    <classes>
+      <memberOf key="att.neumetype"/>
```
bwbohl:
```suggestion
      <memberOf key="att.neumeType"/>
```

source/modules/MEI.neumes.xml
bwbohl:
```suggestion
  <classSpec ident="att.neumeType" module="MEI.neumes" type="atts">
```

ahankinson (dismissed):

bwbohl (approved):

ahankinson (approved):

PR #1161, opened 2023-05-26
Fixed: Use value of "schematron" for constraintSpec

### LINKED ISSUE #1158: Change "isoschematron" to "schematron"
Currently the value of "isoschematron" gives us validation errors in our source since it has been changed in the TEI stylesheets. So we should update our source files to change all values of "isoschematron" to "schematron".

ahankinson:
The TEI has changed the value of "schematron" in its RNG, but the MEI RNG allows both "isoschematron" and "schematron".

If we're thinking about moving towards pure ODD this change will mean one less thing that breaks when we update.

Also fixed an issue where the customizations were using the TEI RNG, but the sources were pointing to the (older) MEI RNG. These were normalized to the MEI RNG so that all source and customizations use the same ODD schema.

Fixes #1158

### DIFF
--- a/customizations/mei-CMN.xml
+++ b/customizations/mei-CMN.xml
@@ -16,8 +16,8 @@
   
   CONTACT: info@music-encoding.org
 -->
-<?xml-model href="http://www.tei-c.org/release/xml/tei/custom/schema/relaxng/tei_odds.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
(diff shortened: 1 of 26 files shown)


musicEnfanthen (approved):
Looks good to me, thanks

ahankinson:
FYI: I've also built it with the ant build script and it seems to work OK

kepper (approved):
Looks very good to me, many thanks @ahankinson 👍

PR #1163, opened 2023-05-26
Fixed: Add copyright notice and update dates

### LINKED ISSUE #1144: Update copyright / license dates
Most of the copyright and license dates currently say 2017 / 2018, etc. Before the 5.0.0 release we should update these dates.

ahankinson:
This PR fixes the dates for the copyright notices and also adds the notice to the MEI source files.

It also fixes some formatting issues with the guidelines documents and adds an XML declaration (as noticed by @rettinghaus).

Fixes #1144

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-NOTICE: Copyright (c) 2017 by the Music Encoding Initiative (MEI) Board (formerly known as "MEI Council").
+NOTICE: Copyright (c) 2017-2023 by the Music Encoding Initiative (MEI) Board (formerly known as "MEI Council").
 
 Licensed under the Educational Community License, Version 2.0 (the "License"); you may
 not use this file except in compliance with the License. You may obtain a copy of the License
--- a/build.xml
+++ b/build.xml
@@ -1,4 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!-- 
+  NOTICE: Copyright (c) 2017-2023 by the Music Encoding Initiative (MEI) Board (formerly known as "MEI Council").
+
+  Licensed under the Educational Community License, Version 2.0 (the "License"); you may
+  not use this file except in compliance with the License. You may obtain a copy of the License
+  at https://opensource.org/licenses/ECL-2.0.
+  
+  Unless required by applicable law or agreed to in writing, software distributed under the
+  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS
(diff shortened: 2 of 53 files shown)


musicEnfanthen:
@ahankinson May I ask you to resolve the merge conflict?

LICENSE
```diff
@@ -1,4 +1,4 @@
-NOTICE: Copyright (c) 2017 by the Music Encoding Initiative (MEI) Board (formerly known as "MEI Council").
+NOTICE: Copyright (c) 2017-2023 by the Music Encoding Initiative (MEI) Board (formerly known as "MEI Council").
```
rettinghaus, line 1: Please use typographically better en dash instead of a hyphen.

build.xml
```diff
@@ -1,4 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!-- 
+  NOTICE: Copyright (c) 2017-2023 by the Music Encoding Initiative (MEI) Board (formerly known as "MEI Council").
```
rettinghaus, line 3: Is this `build.xml` this old? I think it showed up at a later point.

rettinghaus (changes requested):
Thanks for the work. Some minor requests from my side. Also thanks for updating and securing the links, which would have justified a separate PR.

build.xml
musicEnfanthen: You are right, @rettinghaus . It was created in late 2020, cf. https://github.com/music-encoding/music-encoding/pull/740/commits/b7a3e0ae1ee7ca86a4c93e3f8362775fa01fc178

ahankinson: This is a blanket copyright statement for all the files in the project; many of the guidelines files didn't exist in 2017 either. This statement is the same one as the one in `LICENSE`.

LICENSE
ahankinson: I would like to keep it as plain ASCII text.

ahankinson:
I've resolved the merge conflict (I think)

musicEnfanthen (approved):

LICENSE
bwbohl: plain ascii is good for the general license file

build.xml
bwbohl: Having the same statement in all files is fine. Maintaining the exact date of creation in every file could turnout a maintenance nightmare.

ahankinson: I'm going to suggest we don't have file-specific license statements, for the reasons @bwbohl mentions. So I will resolve this.

bwbohl (approved):

PR #1170, opened 2023-05-27
Fixed: Key signature changes

### LINKED ISSUE #525: Harmonize key.sig et al. attributes on scoreDef, staffDef, key and keySig

ahankinson:
This PR contains changes to the key signature attributes on `<keySig>` and `<staffDef> / <scoreDef>`. It clarifies how key signature changes should be drawn, and makes the naming of the attributes more consistent.

Fixes #525


ahankinson:
This is marked draft until I can write a paragraph or two in the guidelines about key signature changes.

source/modules/MEI.visual.xml
```diff
@@ -824,10 +824,10 @@
     <desc xml:lang="en">Used by staffDef and scoreDef to provide default values for attributes in the visual
       domain related to key signatures.</desc>
     <attList>
-      <attDef ident="keysig.showchange" usage="opt">
+      <attDef ident="keysig.cancelaccid" usage="opt">
```
rettinghaus, line 827: The description should be updated, as it is no longer only if, but rather where.

source/modules/MEI.xml
```diff
@@ -510,6 +510,25 @@
...
+        <valItem ident="none">
+          <desc xml:lang="en">Do not show cancellation accidentals.</desc>
+        </valItem>
+        <valItem ident="before">
+          <desc xml:lang="en">Show cancellation accidentals before the new key accidentals.</desc>
+        </valItem>
+        <valItem ident="after">
+          <desc xml:lang="en">Show cancellation accidentals after the new key accidentals ("Old style" or "French")</desc>
+        </valItem>
+        <valItem ident="before-bar">
```
rettinghaus, line 526: I don't think this value makes much sense here, as `keySig` is a layer element and does not have do appear next to a measure.

Random example from Ravel's _Le tombeau de Couperin_

Le tombeau de Couperin (excerpt)

If you need to split it up, you should encode the cancellation at the end of one measure and the new one at the beginning of the next. (Maybe connect them with a linking attribute like `@precedes` and `@follows`). 

Or what to do if you have a `barLine` element next to it, but in the following order: 

```xml
<note />
<keySig keysig.cancelaccid="before-bar" keysig="3f" />
<barLine />
<note />
```

source/modules/MEI.xml
```diff
@@ -510,6 +510,25 @@
       </valList>
     </content>
   </macroSpec>
+  <macroSpec ident="data.CANCELACCID" module="MEI" type="dt">
+    <desc xml:lang="en">Indicates the location of the tonic in the circle of fifths.</desc>
+    <content>
+      <valList type="closed">
+        <valItem ident="none">
+          <desc xml:lang="en">Do not show cancellation accidentals.</desc>
```
rettinghaus, line 518: Maybe better to take the encoder's perspective than the renderer's: "Without cancellation accidentals."? Possibly with a "Written" or "Printed" in front of that.

rettinghaus (changes requested):
Thanks, mostly looking fine.

source/modules/MEI.visual.xml
ahankinson: Fixed! Thanks.

source/modules/MEI.xml
```diff
@@ -510,6 +510,25 @@
       </valList>
     </content>
   </macroSpec>
+  <macroSpec ident="data.CANCELACCID" module="MEI" type="dt">
+    <desc xml:lang="en">Indicates the location of the tonic in the circle of fifths.</desc>
```
ahankinson:
```suggestion
    <desc xml:lang="en">Indicates where cancellation accidentals are shown in a key signature.</desc>
```

source/modules/MEI.xml
ahankinson: I think "Do not show..." is clearer, as it aligns with the other definitions that all say "Show ...".

source/modules/MEI.xml
ahankinson: (Sorry, I had written this response before I resolved it, but it didn't save for some reason.)

> Random example from Ravel's Le tombeau de Couperin

You can still do this:

`<keySig keysig.cancelaccid="after" ... />`

We make no guarantees in MEI that all values are valid in all circumstances.

source/modules/MEI.xml
rettinghaus: I meant that for all descriptions.

source/modules/MEI.xml
```diff
@@ -510,6 +510,25 @@
...
+          <desc xml:lang="en">Do not show cancellation accidentals.</desc>
+        </valItem>
+        <valItem ident="before">
+          <desc xml:lang="en">Show cancellation accidentals before the new key accidentals.</desc>
+        </valItem>
+        <valItem ident="after">
+          <desc xml:lang="en">Show cancellation accidentals after the new key accidentals ("Old style" or "French")</desc>
+        </valItem>
+        <valItem ident="before-bar">
+          <desc xml:lang="en">Show cancellation accidentals before the barline.</desc>
```
lpugin:
```suggestion
          <desc xml:lang="en">Show cancellation accidentals before the barline (also known as "Russian").</desc>
```

source/docs/04-cmn.xml
```diff
@@ -231,8 +231,8 @@
                   </p>
                </div>
                <div xml:id="cmnReDef" type="div3">
-                  <head>Redefinition of Score Parameters</head>
-                  <p>Sometimes it is necessary to provide the parameters of a score or a staff with new values. For example. a score may change keys, gain or lose staves, use different layout settings at any point, etc. Likewise, a staff may change its clef, gain or lose layers, or become invisible, and so on. To accommodate these circumstances, in CMN <gi scheme="MEI">staffDef</gi> is allowed to occur in the following locations:</p>
+                  <head>Re-definition of Score Parameters</head>
+                  <p>Sometimes it is necessary to re-define the parameters of a score or a staff. For example. a score may change keys, the number of staves, or use different layout settings. Likewise, a staff may change its clef, change the number of layers, or become invisible. To accommodate these changes, <gi scheme="MEI">staffDef</gi> is allowed to occur in the following locations:</p>
```
rettinghaus:
```suggestion
                  <p>Sometimes it is necessary to re-define the parameters of a score or a staff. For example, a score may change keys, the number of staves, or use different layout settings. Likewise, a staff may change its clef, change the number of layers, or become invisible. To accommodate these changes, <gi scheme="MEI">staffDef</gi> is allowed to occur in the following locations:</p>
```

source/modules/MEI.xml
rettinghaus: At least there should be a scheamtron warning when using this value while `keySig` is a `descendant` of measure.

rettinghaus (changes requested):

source/docs/04-cmn.xml
```diff
@@ -244,7 +244,14 @@
                      <item>within sections and endings; that is, inside <gi scheme="MEI">section</gi> and <gi scheme="MEI">ending</gi> elements; and</item>
                      <item>between sections and endings; that is, directly within a <gi scheme="MEI">score</gi> or <gi scheme="MEI">part</gi>.</item>
                   </list>
-                  <p>The possibility also exists to include <gi scheme="MEI">scoreDef</gi> and <gi scheme="MEI">staffDef</gi> in staves and layers when the mei-all schema is in use; however, this practice is not recommended for the CMN repertoire.</p>
+                  <p>It is also possible to include <gi scheme="MEI">scoreDef</gi> and <gi scheme="MEI">staffDef</gi> in staves and layers when the MEI All schema is in use; however, this practice is not recommended for the CMN repertoire.</p>
+                  <p>The following example shows how to change the key and meter signatures within a score. The <att>keysig.cancelaccid</att> attribute may be used to control the position of the cancellation accidentals, while the <att>keysig.visible</att> can be used to hide the key signature entirely.</p>
```
```suggestion
                  <p>The following example shows how to change the key and meter signatures within a score. The <att>keysig.cancelaccid</att> attribute may be used to control the position of the cancellation accidentals of the key signature change, while the <att>keysig.visible</att> can be used to hide the key signature entirely.</p>  
```

musicEnfanthen: Suggested the clarification, since the last reference was to "key and meter signatures".

lpugin (approved):

bwbohl (approved):

PR #1182, opened 2023-05-30
Fixed: MEI Basic Schema not validating

ahankinson:
The MEI.critapp module was not included, so the app.rdg.log class deletion was giving an error because it could not be found.

Additionally, added Laurent to the contributors and an entry in the revision log to reflect the changes from the MEI Developers Conference.


ahankinson:
Superseded by #1183

PR #1184, opened 2023-05-30
Initial script for generating images using a python script instead of…

lpugin:
This PR adds a Python script offering an alternative to the [node script ](https://github.com/music-encoding/docker-mei/blob/main/index.js) for generating images with Verovio outside Docker.

The Python scripts dependencies are standard ones plus:
```python
import verovio
```
Both can be installed with `pip`

The script can be run through a new `ant` a new target `generate-images-py` (guidelines should have been built beforehand):
```bash
ant generate-images-py
```

### DIFF
--- a/BUILD_COMMANDLINE.md
+++ b/BUILD_COMMANDLINE.md
@@ -8,7 +8,7 @@
    |----|----------------|
    |Java| 17 (at least Java 8 at runtime to support Apache Ant™)|
    |Apache Ant|1.10.13|
-   |Verovio Toolkit|3.9|
+   |Verovio Toolkit|3.15|
    |Prince XML|15.1|
    |Saxon HE*| 11.4 |
    |Xerces*|Synchrosoft patched version 25.1.0.1|
@@ -54,6 +54,14 @@
 
      We recommend using the latest stable release of Apache Ant™. If your system has an older version of Apache Ant™ installed you might still give it a try though. If the prompt returns an empty string, please refer to the [Apache Ant™ Installation Instructions](https://ant.apache.org/manual/install.html).
 
+  * Is Verovio installed for generating example images locally?
+
+    To build the images with Verovio, you need Python3 to be installed with the `verovio` and `lxml` modules. These can be installed with
+
+    ```shell
+    pip install verovio lxml
+    ```
+
 2. Initialize the build process
 
    * Switch to your clone’s directory:
@@ -77,6 +85,11 @@
      ```
 
      The results of this build can be found in the web folder (`music-encoding/dist/guidelines/dev/web`). The guidelines are stored in the `index.html` file.
+
+     To generate the example images with Verovio, you need to run:
+     ```shell
+     ant generate-images-py
+     ```
 
    * Build a specific customization's RNG schema:
 
--- a/BUILD_COMMANDLINE.md
+++ b/BUILD_COMMANDLINE.md
@@ -56,7 +56,7 @@
 
   * Is Verovio installed for generating example images locally?
 
-    To build the images with Verovio, you need Python3 to be installed with the `verovio` and `lxml` modules. These can be installed with
+    To build the images with Verovio, you need Python3 to be installed with the `verovio` module. These can be installed with
 
     ```shell
     pip install verovio lxml
--- a/BUILD_COMMANDLINE.md
+++ b/BUILD_COMMANDLINE.md
@@ -56,12 +56,22 @@
 
   * Is Verovio installed for generating example images locally?
 
-    To build the images with Verovio, you need Python3 to be installed with the `verovio` module. These can be installed with
+    Optional: If you wish, you can use a Python virtual environment to manage your dependencies. Before installing
+    Verovio, create and activate a virtual environment. 
+    
+    ```shell
+    python3 -m venv ./.venv
+    source ./.venv/bin/activate
+    ```
+    
+    This will install your Python libraries in the local `.venv` directory. Once your virtual environment is active you can continue to installing Verovio.
+
+    To build the images with Verovio, you need Python3 to be installed with the `verovio` module. This can be installed with:
 
     ```shell
-    pip install verovio lxml
+    pip install verovio
     ```
-
+    
 2. Initialize the build process
 
    * Switch to your clone’s directory:
@@ -90,6 +100,12 @@
      ```shell
      ant generate-images-py
      ```
+
+     **Note:** If you have installed your dependencies in a virtual environment, be sure to activate it prior to calling the Ant task. Activate it using:
+     
+     ```shell
+     source ./.venv/bin/activate
+     ```     
 
    * Build a specific customization's RNG schema:
 
--- a/build.xml
+++ b/build.xml
@@ -269,6 +269,13 @@
             <arg value="index.js"/>
         </exec>
     </target>
+    
+    <target name="generate-images-py">
+        <exec executable="python3" logerror="true" dir="${basedir}/utils/examples" failifexecutionfails="true" failonerror="true">
+            <arg value="generate-examples.py"/>
+            <arg value="--clean"/>
+        </exec>
+    </target>
 
     <target name="copy-generated-images" description="copy generated images to dist after they're created by running Docker image">
         <copy todir="${dir.dist.guidelines}/assets/images/GeneratedImages">
--- a/build.xml
+++ b/build.xml
@@ -276,6 +276,7 @@
         <exec executable="python3" logerror="true" dir="${basedir}/utils/examples" failifexecutionfails="true" failonerror="true">
             <arg value="generate-examples.py"/>
             <arg value="--clean"/>
+            <arg value="--verbose"/>
         </exec>
     </target>
 
--- a/utils/examples/generate-examples.py
+++ b/utils/examples/generate-examples.py
@@ -64,7 +64,7 @@ if __name__ == "__main__":
 
     for file in os.listdir(IMAGES_PATH):
         # Skip everything that is not an .mei or .xml file
-        if not file.endswith(".mei") and not file.endswith(".xml"): 
+        if not file.endswith(".mei"): 
             continue
 
         mei_file = os.path.join(IMAGES_PATH, file)
--- a/utils/examples/generate-examples.py
+++ b/utils/examples/generate-examples.py
@@ -103,7 +103,7 @@ if __name__ == "__main__":
         if meta:
             # Overwrite any pre-defined options with the options from the MEI file.
             log.info("Found some locally-defined meta options: %s", meta)
-            metaOptions = meta
+            meta_options = meta
             options.update(metaOptions)
 
         log.debug("Running Verovio with the following options: %s", pprint.pformat(options))
--- a/utils/examples/generate-examples.py
+++ b/utils/examples/generate-examples.py
@@ -85,7 +85,7 @@ if __name__ == "__main__":
 
         log.debug("Running Verovio with the following options: %s", pprint.pformat(options))
         tk.setOptions(options)
-        if not(tk.loadData(mei_example)):
+        if not tk.loadData(mei_example):
             log.error("Failed to load %s", mei_file)
             success = False
             continue
--- a/utils/examples/generate-examples.py
+++ b/utils/examples/generate-examples.py
@@ -71,7 +71,8 @@ if __name__ == "__main__":
 
     for file in os.listdir(IMAGES_PATH):
         # Skip everything that is not an .mei or .xml file
-        if not(file.endswith(".mei")) and not(file.endswith(".xml")): continue
+        if not(file.endswith(".mei")) and not(file.endswith(".xml")): 
+            continue
         
         options: Dict = defaultOptions.copy()
 
--- a/utils/examples/generate-examples.py
+++ b/utils/examples/generate-examples.py
@@ -62,7 +62,7 @@ if __name__ == "__main__":
     log.info("Using Verovio %s", tk.getVersion())
 
     # keep all the options to be able to reset them for each example
-    defaultOptions: Dict = tk.getDefaultOptions()
+    default_options: dict = tk.getDefaultOptions()
     # Overwrite the default options with our locally-defined options
     defaultOptions.update(VRV_OPTIONS)
 
--- a/utils/examples/generate-examples.py
+++ b/utils/examples/generate-examples.py
@@ -12,7 +12,7 @@ IMAGES_PATH = "../../build/assets/images/GeneratedImages"
 
 MEI_NS: Dict = {'mei': 'http://www.music-encoding.org/ns/mei'}
 
-VRV_OPTIONS: Dict = {
+VRV_OPTIONS: dict = {
    'scale': 10,
    'pageWidth': 1500,
    'adjustPageHeight': True,
--- a/utils/examples/generate-examples.py
+++ b/utils/examples/generate-examples.py
@@ -1,0 +1,119 @@
+import argparse
+import os
+import sys
+import pprint
+from typing import Dict
+import logging
+
+import verovio
+from lxml import etree
+
+IMAGES_PATH = "../../build/assets/images/GeneratedImages"
+
+MEI_NS: Dict = {'mei': 'http://www.music-encoding.org/ns/mei'}
+
+VRV_OPTIONS: Dict = {
+   'scale': 10,
+   'pageWidth': 1500,
+   'adjustPageHeight': True,
+   'adjustPageWidth': True,
+   'footer': 'none',
+   'header':'none',
+   'openControlEvents': True,
+   'outputFormatRaw': True,
+   'removeIds': True,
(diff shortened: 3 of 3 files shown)


utils/examples/generate-examples.py
```diff
@@ -0,0 +1,119 @@
...
+    defaultOptions: Dict = tk.getDefaultOptions()
+    # Overwrite the default options with our locally-defined options
+    defaultOptions.update(VRV_OPTIONS)
+
+    # Will remove extraneous whitespace
+    et_parser = etree.XMLParser(remove_blank_text=True)
+
+    for file in os.listdir(IMAGES_PATH):
+        # Skip everything that is not an .mei or .xml file
+        if not(file.endswith(".mei")) and not(file.endswith(".xml")): continue
```
ahankinson:
```suggestion
        if not(file.endswith(".mei")) and not(file.endswith(".xml")): 
            continue
```

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,119 @@
...
+        # parse the MEI file to XML
+        log.debug("Parsing downloaded text to XML")
+        tree = None
+        meta = None
+        try:
+            tree = etree.fromstring(bytes(mei_example.encode("utf-8")), parser=et_parser)
+            # try to get the extMeta tag and load the options if existing
+            meta = tree.findtext(".//mei:meiHead/mei:extMeta", namespaces=MEI_NS)
+        except:
+            log.info("Could not parse file")
```
ahankinson:
```suggestion
            log.info("Could not extract extMeta from %s", mei_file)
```

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,119 @@
...
+
+        # parse the MEI file to XML
+        log.debug("Parsing downloaded text to XML")
+        tree = None
+        meta = None
+        try:
+            tree = etree.fromstring(bytes(mei_example.encode("utf-8")), parser=et_parser)
+            # try to get the extMeta tag and load the options if existing
+            meta = tree.findtext(".//mei:meiHead/mei:extMeta", namespaces=MEI_NS)
+        except:
```
ahankinson: Can you make this exception less broad? What exceptions do you expect will be raised here?

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,119 @@
...
+            options.update(metaOptions)
+
+        log.debug("Running Verovio with the following options: %s", pprint.pformat(options))
+        tk.setOptions(options)
+        tk.loadData(mei_example)
+
+        svg: str = tk.renderToSVG(1)
+
+        with open(svg_file, 'w') as f:
+            f.write(svg)
```
ahankinson: Do you need to check that `svg` is not an empty string before writing it?

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,120 @@
...
+    log = logging.getLogger(__name__)
+    log.info("Running at logging level %s", level_msg)
+
+    verovio.enableLog(verovio.LOG_DEBUG)
+    tk = verovio.toolkit()
+    # version of the toolkit
+    log.info("Using Verovio %s", tk.getVersion())
+
+    # keep all the options to be able to reset them for each example
+    defaultOptions: Dict = tk.getDefaultOptions()
```
ahankinson:
```suggestion
    default_options: dict = tk.getDefaultOptions()
```

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,120 @@
...
+    # Overwrite the default options with our locally-defined options
+    defaultOptions.update(VRV_OPTIONS)
```
ahankinson:
```suggestion
    default_options.update(VRV_OPTIONS)
```

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,120 @@
...
+
+    # Will remove extraneous whitespace
+    et_parser = etree.XMLParser(remove_blank_text=True)
+
+    for file in os.listdir(IMAGES_PATH):
+        # Skip everything that is not an .mei or .xml file
+        if not(file.endswith(".mei")) and not(file.endswith(".xml")): 
+            continue
+        
+        options: Dict = defaultOptions.copy()
```
ahankinson:
```suggestion
        options: dict = default_options.copy()
```

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,120 @@
...
+            # try to get the extMeta tag and load the options if existing
+            meta = tree.findtext(".//mei:meiHead/mei:extMeta", namespaces=MEI_NS)
+        except:
+            log.info("Could not parse file")
+
+        # This is currently not used be in-place of having rendering options in <extMeta>
+        if meta:
+            # Overwrite any pre-defined options with the options from the MEI file.
+            log.info("Found some locally-defined meta options: %s", meta)
+            metaOptions = meta
```
ahankinson:
```suggestion
            meta_options = meta
```

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,120 @@
...
+            options.update(metaOptions)
```
ahankinson:
```suggestion
            options.update(meta_options)
```

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,120 @@
+import argparse
+import os
+import sys
+import pprint
+from typing import Dict
```
ahankinson:
```suggestion
```

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,120 @@
...
+import logging
+
+import verovio
+from lxml import etree
+
+IMAGES_PATH = "../../build/assets/images/GeneratedImages"
+
+MEI_NS: Dict = {'mei': 'http://www.music-encoding.org/ns/mei'}
+
+VRV_OPTIONS: Dict = {
```
ahankinson:
```suggestion
VRV_OPTIONS: dict = {
```

ahankinson (changes requested):

BUILD_COMMANDLINE.md
```diff
@@ -54,6 +54,14 @@
 
      We recommend using the latest stable release of Apache Ant™. If your system has an older version of Apache Ant™ installed you might still give it a try though. If the prompt returns an empty string, please refer to the [Apache Ant™ Installation Instructions](https://ant.apache.org/manual/install.html).
 
+  * Is Verovio installed for generating example images locally?
+
+    To build the images with Verovio, you need Python3 to be installed with the `verovio` and `lxml` modules. These can be installed with
+
+    ```shell
+    pip install verovio lxml
```
ahankinson: Note that this means that Verovio and LXML need to be installed globally, and not inside a virtual environment. This may cause problems if you have conflicting global setups for Python.

lpugin: Do you want to add this as a note here?

ahankinson: Done.

ahankinson:
@lpugin looks like there's a merge conflict now. Could you please resolve this?

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,107 @@
...
+        if not(tk.loadData(mei_example)):
+            log.error("Failed to load %s", mei_file)
+            success = False
+            continue
+
+        svg: str = tk.renderToSVG(1)
+
+        if len(svg):
+            with open(svg_file, 'w') as f:
+                f.write(svg)
```
musicEnfanthen, line 91:
```suggestion
                f.write(svg)
                log.info("Successfully rendered %s", svg_file)
```

build.xml
```diff
@@ -271,6 +271,13 @@
             <arg value="index.js"/>
         </exec>
     </target>
+    
+    <target name="generate-images-py">
+        <exec executable="python3" logerror="true" dir="${basedir}/utils/examples" failifexecutionfails="true" failonerror="true">
+            <arg value="generate-examples.py"/>
+            <arg value="--clean"/>
```
musicEnfanthen, line 278:
```suggestion
            <arg value="--clean"/>
            <arg value="--verbose"/>
```

musicEnfanthen: Suggestion to run on verbose mode by default (= log level INFO) to see Verovio version and rendered file names in output

utils/examples/generate-examples.py
musicEnfanthen: Add a info message which file was successfully rendered

musicEnfanthen (commented):
Thanks @lpugin. Tested locally and works great. Some minor suggestions below.

build.xml
```diff
@@ -271,6 +271,14 @@
             <arg value="index.js"/>
         </exec>
     </target>
+    
+    <target name="generate-images-py">
+        <exec executable="python3" logerror="true" dir="${basedir}/utils/examples" failifexecutionfails="true" failonerror="true">
```
musicEnfanthen, line 276: Would it be possible to use just `python` instead of `python3`? (So I mean the alias, the version should of course be python 3)

ahankinson: In some cases `python` is still aliased to Python 2. `python3` is generally favoured by most package managers as the preferred name for Python 3 installations. (That said, if you have a virtual environment you will have access to both the `python` and `python3` executables, since `python` is an alias to `python3`.)

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,108 @@
...
+    log.info("Using Verovio %s", tk.getVersion())
+
+    # keep all the options to be able to reset them for each example
+    default_options: dict = tk.getDefaultOptions()
+    # Overwrite the default options with our locally-defined options
+    default_options.update(VRV_OPTIONS)
+
+    for file in os.listdir(IMAGES_PATH):
+        # Skip everything that is not an .mei or .xml file
+        if not(file.endswith(".mei")) and not(file.endswith(".xml")):
```
ahankinson:
```suggestion
        if not file.endswith(".mei") and not file.endswith(".xml"): 
```

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,108 @@
...
+    default_options: dict = tk.getDefaultOptions()
+    # Overwrite the default options with our locally-defined options
+    default_options.update(VRV_OPTIONS)
+
+    for file in os.listdir(IMAGES_PATH):
+        # Skip everything that is not an .mei or .xml file
+        if not(file.endswith(".mei")) and not(file.endswith(".xml")): 
+            continue
+        
+        options: dict = default_options.copy()
```
ahankinson: Not sure why you need to make a copy of the options dict on every pass, or reset the options on the toolkit. It should be enough to set `tk.setOptions` once.

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,108 @@
...
+            continue
+
+        # Download the MEI file from the given url
+        mei_example: str = ""
+        with open(mei_file) as f:
+            mei_example = f.read()
+
+        log.debug("Running Verovio with the following options: %s", pprint.pformat(options))
+        tk.setOptions(options)
+        if not(tk.loadData(mei_example)):
```
ahankinson:
```suggestion
        if not tk.loadData(mei_example):
```

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,108 @@
...
+            log.error("Failed to load %s", mei_file)
+            success = False
+            continue
+
+        svg: str = tk.renderToSVG(1)
+
+        if len(svg):
```
ahankinson:
```suggestion
        if svg:
```

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,108 @@
...
+            with open(svg_file, 'w') as f:
+                f.write(svg)
+                log.info("Successfully rendered %s", svg_file)
+        else:
+            log.error("Failed to render %s", mei_file)
+            success = False
+
+        log.debug("Finished processing %s", mei_file)
+    
+    if not(success):
```
ahankinson:
```suggestion
    if not success:
```

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,108 @@
...
+    verovio.enableLog(verovio.LOG_DEBUG)
+    tk = verovio.toolkit()
+    # version of the toolkit
+    log.info("Using Verovio %s", tk.getVersion())
+
+    # keep all the options to be able to reset them for each example
+    default_options: dict = tk.getDefaultOptions()
+    # Overwrite the default options with our locally-defined options
+    default_options.update(VRV_OPTIONS)
+
```
ahankinson, line 64: I would define `success: bool = True` here... (see below)

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,108 @@
...
+            log.error("Failed to load %s", mei_file)
+            success = False
+            continue
+
+        svg: str = tk.renderToSVG(1)
+
+        if len(svg):
+            with open(svg_file, 'w') as f:
+                f.write(svg)
+                log.info("Successfully rendered %s", svg_file)
```
ahankinson, line 92: .... and here do `success &= True` ...

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,108 @@
...
+        else:
+            log.error("Failed to render %s", mei_file)
+            success = False
```
ahankinson, line 95:
```suggestion
            success &= False
```

The reason is that you're constantly resetting the `success` variable to the outcome of the last run, rather than tracking whether all renders were executed successfully. `True & True & False & True = False`, but `True & True & True & True = True`

ahankinson (changes requested):

utils/examples/generate-examples.py
lpugin: This remains from when we had (potentially) specific options read from the example. It can now be moved up and set only once.

utils/examples/generate-examples.py
lpugin: No. `success` is changed (i.e., set to `False`) only when a file fails, and it is never reset otherwise. This means that at the end, it indicates if at least one file fails.

utils/examples/generate-examples.py
lpugin: See below too

utils/examples/generate-examples.py
lpugin: See below

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,102 @@
+import argparse
+import os
```
rettinghaus, line 2: Could we drop os and instead use 
```suggestion
from pathlib import Path
```

Makes everything easier and more readable.
We don't use `with open` and don't have to go over all the files if we use `glob`.

utils/examples/generate-examples.py
```diff
@@ -0,0 +1,102 @@
...
+    tk = verovio.toolkit()
+    # version of the toolkit
+    log.info("Using Verovio %s", tk.getVersion())
+
+    log.debug("Running Verovio with the following options: %s", pprint.pformat(VRV_OPTIONS))
+    tk.setOptions(VRV_OPTIONS)
+
+    for file in os.listdir(IMAGES_PATH):
+        # Skip everything that is not an .mei or .xml file
+        if not file.endswith(".mei") and not file.endswith(".xml"):
```
rettinghaus: I would prefer to clean up the files and having a unified extension.

lpugin: Sounds good. They are all `.mei`, I think.

musicEnfanthen: Yes, they are all `.mei`. Checking for only that extension should be fine.

build.xml
musicEnfanthen: I see your point, thanks.

utils/examples/generate-examples.py
bwbohl: @ahankinson @lpugin @musicEnfanthen please resolve this suggestion

bwbohl:
@musicEnfanthen is this also tested in the docker-mei container?

musicEnfanthen:
@bwbohl I was able to execute the target from within docker-mei, yes.

utils/examples/generate-examples.py
musicEnfanthen: I am inclined to leave this suggestion for later improvement. The script runs now and does what is expected of it. We will definitely be able to improve the script further later.

musicEnfanthen:
@ahankinson Are your requested changes addressed?

musicEnfanthen (approved):

utils/examples/generate-examples.py
bwbohl: referenced this in a new issue #1213 and added it to Milestone 5.1

ahankinson (approved):

bwbohl:
Many thanks for this I think it's much better to have a script for generating the images inside this repo and should even replace the other ant target and node.js method.

bwbohl (approved):

Issue #987, opened 2022-08-16
Visual appearance of meterSigGrp


rettinghaus:
Currently `meterSigGrp` requires a `@func` attribute, but this is not sufficient to specify the visual appearance of the group, as there are different forms for "interchanging" and "alternating" meter signatures.

Sometimes they are printed just next to each other (Tchaikovsky):

![without](https://upload.wikimedia.org/wikipedia/commons/8/87/Tchaikovsky_String_Quartet_in_F_excerpt.png)

Also there is the bracketed form (Liszt): 

![brackets](https://user-images.githubusercontent.com/7693447/184843884-2c157ac7-bc07-4ac1-8267-73cbf04cc811.png)

(This could be encoded by using `@enclose` on the second `meterSig`, but is a characteristic of the `meterSigGrp`.)

And often you'll find an equal sign (Debussy):

![equal](https://user-images.githubusercontent.com/7693447/184843947-ee5f43fc-64d4-40b9-a98c-78fbef3f9f47.png)

Probably there are more forms out there (like the following with a dash/minus):

![dash](https://steinberg.help/dorico/v2/en/_shared_picts/picts/dorico/notation_reference/time_signatures_types_interchangeable.png)

(Dorico Pro let's you choose the separator style (see [Types of time signatures](https://steinberg.help/dorico/v2/en/dorico/topics/notation_reference/notation_reference_time_signatures_types_c.html)).)

Maybe a `@form` attribute could be added to `meterSigGrp` to specify it's appearance? 

Relates to #970 
Relates to #322

rettinghaus:
Another thing: is there any particular reason why `meterSigGrp` cannot be made invisible as whole? 
Each part `meterSig` can be rendered `invis`, but there is no way to say this for the whole group, so there is a formal ambiguity about each separator glyph. 
Probably both, a `@form="invis"` and a `@visible="false"` would be needed to have a complete formal description of visibility.

PR #1190, opened 2023-06-01
schema: enhance meterSigGrp

### LINKED ISSUE #970: Nested meterSigGrps
Currently it is not possible to nest `meterSigGrp` within `meterSigGrp`. 

Take this example from the _Fitzwilliam Virginal Book_ edition from 1899:

image

One way to encode this (visually) would be

```xml
<meterSigGrp>
   <meterSig sym="cut" />
   <meterSigGrp enclose="brack">
      <meterSig sym="cut" />
      <meterSig count="3" form="num" />
   </meterSigGrp>
</meterSigGrp>
```

Making it part of `att.enclosingChars` would also allow us to place brackets around a complete `meterSigGrp`. 

Probably then the func attribute shouldn’t be necessary (or should allow a value like “unknown”). 

Although it works a bit different than `staffGrp` it seems to make sense to allow nesting of `meterSigGrp`s accordingly.

### LINKED ISSUE #1430: Issue with listing 128 at the Guidelines MEI 5.0
Whole encoding question can be found here  https://github.com/music-encoding/music-encoding/blob/develop/source/examples/verovio/04-score-redefinition.mei (chapter 4.2.4: https://music-encoding.org/guidelines/dev/content/cmn.html#cmnReDef), at listing 128. 

Firstly, the numbering scheme of the measures seems disorganized to me (the first measure is labeled as "1," the second as "4," then after the three rest measures, the next one is labeled as "5"...). Secondly, regarding the accidentals in the first chord, why is C flat? And finally, why is it possible to hold a whole chord in the third measure within a "cut" measure bar? These are examples of cases that happened more than once. I hope you can help me.

Here you have the extract of the code I am refering to, y marked with **these** symbols the examples I talk about above.

<music>
    <body>
      <mdiv>
        <score>
          <?edit-start?>
          <scoreDef keysig="4f" meter.sym="common">
            <staffGrp>
              <staffDef n="1" lines="5" clef.shape="G" clef.line="2" />
            </staffGrp>
          </scoreDef>
          <section>
            <measure right="dbl" n="1">
              <staff n="1">
                <layer n="1">
                  **<chord dur="1">**
                    <note oct="4" pname="a" accid.ges="f" />
                    <note oct="5" pname="c" **accid.ges="f"** />
                    <note oct="5" pname="e" accid.ges="f" />
                  </chord>
                </layer>
              </staff>
            </measure>
            <scoreDef keysig="0" keysig.cancelaccid="none" />
            <measure right="dbl" **n="4"**>
              <staff n="1">
                <layer n="1">
                  <chord dur="1">
                    <note oct="4" pname="a" />
                    <note oct="5" pname="c" />
                    <note oct="5" pname="e" />
                  </chord>
                </layer>
              </staff>
            </measure>
            <scoreDef keysig="2s" keysig.cancelaccid="before" meter.sym=**"cut"** />
            <measure n="2">
              <staff n="1">
                <layer n="1">
                  **<chord dur="1">**
                    <note oct="4" pname="b" />
                    <note oct="5" pname="d" />
                    <note oct="5" pname="f" accid.ges="s" />
                  </chord>
                </layer>
              </staff>
            </measure>
            <measure right="dbl" n="3">
              <staff n="1">
                <layer n="1">
                  <multiRest num="3" />
                </layer>
              </staff>
            </measure>
            <scoreDef keysig.visible="false" keysig="5f" meter.count="4" meter.unit="4" />
            <measure right="dbl" n="5">
              <staff n="1">
                <layer n="1">
                  <chord dur="1">
                    <note oct="4" pname="g" />
                    <note oct="4" pname="b" accid.ges="f" />
                    <note oct="5" pname="d" />
                  </chord>
                </layer>
              </staff>
            </measure>
            <scoreDef keysig="2s" keysig.cancelaccid="before-bar" />
            <measure right="end" n="2">
              <staff n="1">
                <layer n="1">
                  <chord dur="1">
                    <note oct="4" pname="b" />
                    <note oct="5" pname="d" />
                    <note oct="5" pname="f" accid.ges="s" />
                  </chord>
                </layer>
              </staff>
            </measure>
          </section>
          <?edit-end?>
        </score>
      </mdiv>
    </body>
  </music>
</mei>

rettinghaus:
This PR allows nesting of `meterSigGrp` elements, and makes `meterSigGrp` a part of `att.enclosingChars` and `att.visibility`. 

It also removes the schematron assertion rule and makes the rule for needing at least two children part of the schema itself.

A small example for `meterSigGrp` has been added to the documentation.

Closes #970

See https://github.com/music-encoding/music-encoding/issues/987#issuecomment-1221933906

Includes fix to resolve #1430

### DIFF
--- a/source/docs/04-cmn.xml
+++ b/source/docs/04-cmn.xml
@@ -238,7 +238,7 @@
                      <p>
                         <figure>
                            <head>Alternating meters in Tchaikovsky's String Quartet in F major</head>
-                           <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="verovio code" xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/verovio/Tchaikovsky_scherzo.mei" parse="text"/></egXML>
+                           <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="verovio code" xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/verovio/tchaikovsky_scherzo.mei" parse="text"/></egXML>
                         </figure>
                      </p>
                </div>
--- a/source/docs/04-cmn.xml
+++ b/source/docs/04-cmn.xml
@@ -234,6 +234,13 @@
                         <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve" valid="true"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/cmn/cmn-layerMeter.txt" parse="text"/></egXML>
                      </figure>
                   </p>
+                     <p>When multiple time signatures appear next to each other the <gi scheme="MEI">meterSigGrp</gi> element has to be used.</p>
+                     <p>
+                        <figure>
+                           <head>Alternating meters</head>
+                           <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="verovio code" xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/verovio/tchaikovsky_scherzo.mei" parse="text"/></egXML>
+                        </figure>
+                     </p>
                </div>
                <div xml:id="cmnReDef" type="div3">
                   <head>Re-definition of Score Parameters</head>
--- a/source/docs/04-cmn.xml
+++ b/source/docs/04-cmn.xml
@@ -923,7 +923,7 @@
                      <p>For the repetition of a single note or chord, MEI offers the <gi scheme="MEI">bTrem</gi> (bowed tremolo) element, which is a member of the <ident type="class">model.eventLike.cmn</ident> class, meaning it is encoded following the normal course of musical events within a <gi scheme="MEI">layer</gi>. It holds exactly one <gi scheme="MEI">note</gi> or <gi scheme="MEI">chord</gi> element that is to be repeated.</p>
                      <p>
                         <figure>
-                           <head>Bowed tremolandi in Schubert’s last string quartet</head>
+                           <head>Bowed tremolandi in Schubert’s last String Quartet</head>
                            <graphic url="../images/ExampleImages/SchubertD887.png"/>
                         </figure>
                         <figure>
@@ -989,7 +989,7 @@
                      <p>In MEI, fermatas may be encoded using an attribute on elements like <gi scheme="MEI">note</gi>, <gi scheme="MEI">chord</gi> or <gi scheme="MEI">rest</gi>. This attribute allows placement of a fermata above or below the element to which it’s attached.</p>
                      <p>
                         <figure>
-                           <head>Fermatas in Mozart’s String quartet K. 428 indicating general pauses</head>
+                           <head>Fermatas in Mozart’s String Quartet K. 428 indicating general pauses</head>
                            <graphic url="../images/ExampleImages/MozartK428.4.png"/>
                         </figure>
                      </p>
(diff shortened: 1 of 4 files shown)


pe-ro:
The current model doesn't require 2 children.  It requires 2 **meterSig** children.  The model you're proposing requires at least 2 children, any number of which may be `<meterSig>` or `<meterSigGrp>`.  This would validate the following --
```
<meterSigGrp>
  <meterSig/>
  <meterSigGrp>
    <meterSig/>
    <meterSig/>
  </meterSigGrp>
</meterSigGrp>
```
which is fine.  It's just not the model that was there before.

I believe your model also results in an ambiguous content model, which RelaxNG tolerates, but XMLSchema does not.  If translation to XSD is ever envisioned, further investigation is required.

pe-ro (dismissed):
I just checked -- your model is NOT ambiguous. So, I don't find a reason not to implement it.

lpugin:
Could you give the complete encoding you have in mind for the example in https://github.com/music-encoding/music-encoding/issues/970 since what you gave is essentially visual ?

rettinghaus:
@pe-ro looking at similar groups ([clefGrp](https://music-encoding.org/guidelines/dev/elements/clefGrp.html), [instrGrp](https://music-encoding.org/guidelines/dev/elements/instrGrp.html), [staffGrp](https://music-encoding.org/guidelines/dev/elements/staffGrp.html)) we need at least one child. I think this should be harmonized.

bwbohl:
@lpugin & @pe-ro can we move forward with this for MEI 5.0 or are there things that should be fixed concerning the similar groups? Is there a need to add anything to the guidelines as well?

pe-ro:
@bwbohl , I think we can move ahead and return to the harmonization and Guidelines questions later.

bwbohl:
We pushed this to Milestone 5.x yesterday. This allows for adjusting the guidelines accordingly.

rettinghaus:
@bwbohl What needs to be adjusted in the Guidelines? `meterSigGrp` is only mentioned as possible child of `staffDef`, and that didn't change …

rettinghaus:
pinging @bwbohl

bwbohl:
Maybe we could include an example in the "special cases section", e.g., one similar to the one you provided in #970 or even a less special example just to give a hint one can do such things in MEI?

rettinghaus:
@bwbohl should be ready now

musicEnfanthen:
@rettinghaus Can you please outsource the fix for #1430 in commit aa6a4430e6f096817548dc6f709c89db5a1dbe14 into a separate PR?

bwbohl (dismissed):

source/docs/04-cmn.xml
```diff
@@ -234,6 +234,13 @@
                         <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve" valid="true"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/cmn/cmn-layerMeter.txt" parse="text"/></egXML>
                      </figure>
                   </p>
+                     <p>When multiple time signatures appear next to each other the <gi scheme="MEI">meterSigGrp</gi> element has to be used.</p>
+                     <p>
+                        <figure>
+                           <head>Alternating meters in Tchaikovsky's String Quartet in F major</head>
+                           <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="verovio code" xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/verovio/Tchaikovsky_scherzo.mei" parse="text"/></egXML>
```
```suggestion
                           <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="verovio code" xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/verovio/tchaikovsky_scherzo.mei" parse="text"/></egXML>
```

bwbohl (approved):
thx for the fix @musicEnfanthen

musicEnfanthen (approved):

PR #1196, opened 2023-06-02
Improve init target

bwbohl:
This PR improves the init target of the build.xml file as follows:

Nice for local development with xerces and saxon on java.classpath
* adds a classpath search to check if saxon is available
* adds a classpath search to check if xerces is available
* only call the targets to download saxon or xerces if not available --> reduces verbosity of terminal output
* adds human readable feedback of the availability checks
* removes the creation of lib folder as xerces or saxon download create it if needed
* Makes description more precise

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -134,17 +134,15 @@
         <antcall target="clean"/>
     </target>
     
+    <target name="init" description="Initialize: checks if saxon and xerces are available, if not downloads them." unless="${docker}">
         <available property="xerces-available" file="${dir.lib.xerces}/${xerces.jar.file}"/>
         <available property="xerces-available" classname="org.apache.xerces.jaxp.SAXParserFactoryImpl"/>
+        <echo message="xerces available: ${xerces-available}"/>
         <available property="saxon-available" file="${dir.lib.saxon}/${saxon.jar.file}"/>
-        <mkdir dir="${dir.lib}"/>
-        <antcall target="saxon-prepare"/>
         <available property="saxon-available" classname="net.sf.saxon.Transform"/>
         <echo message="saxon available: ${saxon-available}"/>
-        <available file="${dir.lib}" property="lib-available"/>
-        <echo message="lib directory available: ${lib-available}"/>
-        <mkdir dir="${dir.lib}" unless:true="${lib-available}"/>
         <antcall target="saxon-prepare" unless:true="${saxon-available}"/>
+        <antcall target="xerces-download" unless:true="${xerces-available}"/>
         <echo>initialized</echo>
     </target>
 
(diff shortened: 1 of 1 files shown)


build.xml
```diff
@@ -134,12 +134,15 @@
         <antcall target="clean"/>
     </target>
     
-    <target name="init" description="initializes the build environment, e.g., downloads saxon" unless="${docker}">
+    <target name="init" description="Initialize: checks if saxon and xerces are available, if not downloads them." unless="${docker}">
```
rettinghaus:
```suggestion
    <target name="init" description="Initialize: checks if saxon and xerces are available, and if not, downloads them." unless="${docker}">
```

rettinghaus (dismissed):

build.xml
```diff
@@ -134,12 +134,15 @@
...
         <available property="xerces-available" file="${dir.lib.xerces}/${xerces.jar.file}"/>
+        <available property="xerces-available" classname="org.apache.xerces.jaxp.SAXParserFactoryImpl"/>
```
musicEnfanthen, line 139: Is it correct that this line addresses the same property name as the line before?

Can you briefly indicate why this additional check is needed (same for line 142 with Saxon)?

build.xml
```diff
@@ -134,12 +134,15 @@
...
+        <echo message="xerces available: ${xerces-available}"/>
         <available property="saxon-available" file="${dir.lib.saxon}/${saxon.jar.file}"/>
-        <mkdir dir="${dir.lib}"/>
```
musicEnfanthen, line 140: Is it for sure that the lib folder is always created by the sub targets even when it does not exist?

build.xml
bwbohl: Currently, the mei.classpath is composed of the jars available in the lib folder and the system java.classpath. The double check checks for the lib folder first – whenever a property in Apache Ant is first initialized with a value, it is immutable afterward – so the second check for the classpath availability checks if the java.classpath already contains the desired classes; if so, it is unnecessary to download the jars.

build.xml
bwbohl: tested it locally, although only on mac

musicEnfanthen: Double-checked on win, seems to work. Thanks

bwbohl: Thx, did you check native windows or gitshell?

build.xml
musicEnfanthen: I see, thanks. Just found this: "If the resource is present, the property value is set to "true" by default; otherwise, the property is not set." (https://ant.apache.org/manual/Tasks/available.html) So if the first check does not find a jar file, there is no property set at all, and it can safely be reused for the second check.

musicEnfanthen (approved):

build.xml
musicEnfanthen: both ;) (cmd and git bash for win)

Issue #648, opened 2020-02-25
beam slope


rettinghaus:
The `@slope` attribute on `<beam>` currently allows decimal values, although the schema doesn't say anything about the meaning of this. Do the allowed values make sense at all? I guess 
`data.DEGREES` would be better here.

craigsapp:
Yes, fractional values would be required, since the endpoints of a slur are typically locked to specific positions in relation to the staff lines (they are usually tangent to the staff line, or centered on a staff line). In order to do this, the angle needs to be a continuous floating-point number.

The `@slope` attribute is not very useful and should be calculated as a byproduct of the stem lengths of the notes attached at the endpoints.  Here is a demo:

<img width="1011" alt="Screen Shot 2020-02-25 at 08 01 40" src="https://user-images.githubusercontent.com/3487289/75265542-e9aef880-57a5-11ea-8fcc-ca5e982b36b8.png">

Notice that the stem lengths of the two notes remain fixed at one octave in length, but the slope of the beam between them changes based on the distance between the notes.  The only particularly interesting value is 0.0 to indicate a horizontal beam, where the distance between the notes is no longer fixed.   There will be cases where the slope is too large because the notes are too close together, but to reduce the slope, the stem lengths should be adjusted: either one stem should be lengthened, or another should be shortened.  In other words, if you provide the slope, then also one stem length would also need to be provided in order to calculate the stem length of the notes at the other endpoint.

Another thought: The endpoints of the beams are (usually) quantized to specific vertical positions.  If you use a beam slope to set these positions, the endpoints will be approximately at the quantized positions, but not exactly.   In other words, there will be rounding error in converting the slope to vertical positions and any change in the horizontal layout would also mess up alignment with these quantized vertical positions.

So the only useful application of @slope would be when one of the endpoints of the beam does not attach to a note.


<img width="683" alt="Screen Shot 2020-02-25 at 08 30 04" src="https://user-images.githubusercontent.com/3487289/75268430-575e2300-57ac-11ea-8c74-e9fe4f1cc3f1.png">

rettinghaus:
If we could refine `@slope` and extract it into a specific attribute class (e.g. `att.sloped`) we could use this in multiple places like hairpins (see #733) and slurs. 

What about something like 

> Defines the slope of the object by specifying the relative vertical distance of its start and end points in virtual units. 

This would allow to encode unusual steep beams or give better impression of the slur design. 
[...]

PR #1197, opened 2023-06-02
schema: add @bracket.slope in att.tuplet.vis

### LINKED ISSUE #1195: tuplet@slope
It would be useful to add `@slope` to `<tuplet>` and `<tupletSpan>` to allow for explicit horizontal tuplet brackets.  This would be similar to the existing `beam@slope` and `beamSpan@slope`.

See https://github.com/rism-digital/verovio/pull/3442 for discussion.

pe-ro:
To avoid duplicate attribute definitions, att.tuplet.vis is referenced by att.tupletSpan.vis.

Fixes #1195


source/modules/MEI.visual.xml
```diff
@@ -1933,6 +1933,13 @@
           <rng:ref name="data.STAFFREL.basic"/>
         </datatype>
       </attDef>
+      <attDef ident="bracket.slope" usage="opt">
+         <desc xml:lang="en">Records the slope of the bracket independently of the slope of the beam (if any)
+           encompassing the events of the tuplet.</desc>
```
craigsapp: It might be good to explicitly state the units (degrees presumably).

craigsapp (dismissed):

source/modules/MEI.visual.xml
rettinghaus: refers to #648

rettinghaus (dismissed):

source/modules/MEI.visual.xml
pe-ro: Degrees sounds good. Other thoughts, @lpugin, @rettinghaus, @kepper, others?

craigsapp: The problem I have in #648, is that a non-zero degree value is not overly useful since it would only apply to one specific layout (that was specifically for beams, but also applies to tuplet brackets and brackets in general).   

A better unit would be the vertical offset of the endpoints, which remains constant over all layout spacings (more or less, sine beams may be flattened when the beam angle is too high).

In other words, perhaps something better than degrees would be in diatonic steps (MEI units).   0 means horizontal, 2 = move up by a staff line from the starting point to the ending point.

pe-ro: Is this description better --

Records the slope of a bracket in degrees of deviation from horizontal. Positive values indicate upward slope, negative values measure slope downward, and a zero value indicates the bracket is horizontal.  The value here is independent of the slope of the beam (if any) encompassing the events of the tuplet.

rettinghaus: Degrees is problematic, because the angle depends on the spacing of the music (as seen in #648). Also you probably would want to limit it then (-90 to 90)? Slope itself is ∆y/∆x, and thus also dependent on spacing. We could define it in MEI as the difference of the distances between the endpoints relative to the staff lines. "3" would mean the end of the bracket is 3 vus higher than its start (at a horizontal staff), "-2.5" it would go down two and a half units.

pe-ro: Ok,  how about modeling `@bracket.slope` on `@vo`:

```
<attDef ident="slope" usage="opt">
  <desc xml:lang="en">Records the slope of a bracket in terms of staff 
interline distance; that is, in units of 1/2 the distance between adjacent staff 
lines. Positive values indicate upward slope, negative values measure slope 
downward, and a zero value indicates the bracket is horizontal. The value here 
is independent of the slope of the beam (if any) encompassing the events of 
the tuplet.</desc>
  <datatype>
    <rng:ref name="data.MEASUREMENTSIGNED"/>
  </datatype>
</attDef>
```
To keep the description short, the text starting with "Positive values indicate ..." can go in the `<remarks>` section.

ahankinson: Would `beam@slope` also change accordingly? Would be annoying to have two attributes with the same name and different types of values.

ahankinson:
Given that this is new functionality I would suggest holding off merging this until after 5.0.

source/modules/MEI.visual.xml
craigsapp: > Ok, how about modeling @bracket.slope on `@vo`

That sounds good since a degree value will not be useful for dynamic layout control.

One thought is that `beam@slope` and similar could be changed to `@vo2` in a similar manner to `@tstamp` and `@tstamp2`:  `@vo2` would be the vertical offset of the endpoint compared to the starting point's vertical placement.   This would be clearer than using the name "slope" since that mathematically implies units of degrees.

Slur currently has `@startvo` and `@vo`, so perhaps change these to `@vo` and `@vo2`.  This would match `@tstamp` attribute naming system as well as to `@x`/`@x2` and `@y`/`@y2`.

rettinghaus: @craigsapp please stop mixing things up again. `vo` is a vertical offset, accompanied by `@startvo` and `@endvo`. 
@pe-ro yes, please change beam@slope.

rettinghaus:
@ahankinson actually #648 is open for over three years. It would be nice to close this now with 5.0.

ahankinson:
@rettinghaus an open issue doesn’t cause problems, but a hasty solution committed in a rush might, particularly if the conversation involves new ways of measurement and isn’t unified across the spec.

rettinghaus:
@ahankinson generally I agree with you. But the older issue pointed to the problem, which is, that the values in `@slope` didn't have a particular _meaning_, making it basically unusable. Changing it to `data.MEASUREMENTSIGNED` would only fill it with meaning without making existing encodings invalid. So this seems to me not to be a "hasty solution committed in a rush". 
But as you pointed out (and I emphasized) `@bracket.slope` and `beam@slope` need to use the same `data type`.

pe-ro:
I assigned this to the 5.x milestone, but I'd be happy to move it forward to 5.0.

pe-ro:
PR updated.

craigsapp:
> I assigned this to the 5.x milestone, but I'd be happy to move it forward to 5.0.

I don't see a need to rush.  It only became interesting due to refinements in verovio layout (and such refinements can go on forever :-)

source/modules/MEI.cmn.xml
```diff
@@ -291,9 +291,12 @@
         </datatype>
       </attDef>
       <attDef ident="slope" usage="opt">
-        <desc xml:lang="en">Records the slope of the beam.</desc>
+        <desc xml:lang="en">Records the slope of a beam; e.g., in terms of staff interline distance; 
+          that is, in units of 1/2 the distance between adjacent staff lines. Positive values indicate 
+          upward slope, negative values measure slope downward, and a zero value indicates the beam
```
ahankinson:
```suggestion
          upward slope, negative values indicate slope downward, and a zero value indicates the beam 
```

source/modules/MEI.visual.xml
```diff
@@ -238,9 +238,12 @@
         </valList>
       </attDef>
       <attDef ident="beam.slope" usage="opt">
-        <desc xml:lang="en">Captures beam slope.</desc>
+        <desc xml:lang="en">Records the slope of a beam; e.g., in terms of staff interline distance; 
+          that is, in units of 1/2 the distance between adjacent staff lines. Positive values indicate 
+          upward slope, negative values measure slope downward, and a zero value indicates the beam
```
ahankinson:
```suggestion
          upward slope, negative values indicate slope downward, and a zero value indicates the beam 
```

source/modules/MEI.visual.xml
```diff
@@ -1933,6 +1936,15 @@
           <rng:ref name="data.STAFFREL.basic"/>
         </datatype>
       </attDef>
+      <attDef ident="bracket.slope" usage="opt">
+         <desc xml:lang="en">Records the slope of a bracket; e.g., in terms of staff interline distance; 
+          that is, in units of 1/2 the distance between adjacent staff lines. Positive values indicate 
+          upward slope, negative values measure slope downward, and a zero value indicates the bracket
```
ahankinson:
```suggestion
          upward slope, negative values indicate slope downward, and a zero value indicates the bracket 
```

source/modules/MEI.cmn.xml
```diff
@@ -291,9 +291,12 @@
         </datatype>
       </attDef>
       <attDef ident="slope" usage="opt">
-        <desc xml:lang="en">Records the slope of the beam.</desc>
+        <desc xml:lang="en">Records the slope of a beam; e.g., in terms of staff interline distance;
```
ahankinson: We should refer to these as virtual units (`vu`) instead of re-explaining what a VU is in MEI.

ahankinson (changes requested):

pe-ro:
I think I incorporated @ahankinson's changes, but along the way I managed to create an unnecessary branch I don't know how to remove.

source/modules/MEI.cmn.xml
```diff
@@ -291,9 +291,11 @@
         </datatype>
       </attDef>
       <attDef ident="slope" usage="opt">
-        <desc xml:lang="en">Records the slope of the beam.</desc>
+        <desc xml:lang="en">Records the slope of a beam; e.g., in MEI virtual units. Positive values indicate
```
ahankinson, line 306: Is it really "e.g., in MEI virtual units" or just "Records the slope of a beam in MEI virtual units." ? If the latter, this should be changed in the other descriptions as well.

rettinghaus: `data.MEASUREMENTSIGNED` allows all sorts of units.

ahankinson: Hm. Right. OK

ahankinson (approved):
The `<abbr>` elves might make a visit, but it looks good to me!

bwbohl:
Generally I do agree but would love to see some examples and documentation added to the guidelines ;-)

musicEnfanthen:
Still needs work, moving to 6.

bwbohl:
I think adding some minimal mention of bracket.slope to the guidelines would do the job of a minimum documentation

bwbohl:
@pe-ro, am I assuming correctly that in the context of this pull request, the following branch can be removed: https://github.com/music-encoding/music-encoding/tree/pe-ro-patch-1?

pe-ro:
@bwbohl , you're correct

rettinghaus (approved):

PR #1200, opened 2023-06-05
build.xml: introduce property `verbose`

bwbohl:
Before this commit the generation of schemata and compiled ODDs gave very verbose feedback. This resulted in a very long and hardly graspable output on the commandline. Especially in the GitHub Action context this made the build process hard to debug and required a lot of scrolling. This commit introduces the following changes:
* The property `verbose` will analyze whether Ant was called with the `-verbose` or `-v` option and store this as boolean; defaults to false.
* The `verbose` property will be propagated to the  `tei:anttasks` to make the verbosity of the schema and compiled-odd generation controllable.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -72,6 +72,9 @@
     <property name="xerces.jar.file" value="oxygen-patched-xerces-${xerces.version}.jar"/>
     <property name="oxygen.basic.utilites.jar.file" value="oxygen-basic-utilities-${xerces.version}.jar"/>
     <path id="mei.classpath">
+    <condition property="verbose" value="true" else="false">
+        <contains string="${sun.java.command}" substring="-v" />
+    </condition>
         <fileset dir="lib" erroronmissingdir="false">
             <include name="**/*.jar" />
         </fileset>
@@ -180,7 +183,7 @@
             <property name="inputFile" value="${customization.path}"/>
             <property name="outputFile" value="${dir.dist.schemata}/${odd.basename}_compiled.odd"/>
             <property name="defaultSource" value="${canonicalized.source.path.cross_platform}"/>
-            <property name="verbose">true</property>
+            <property name="verbose" value="${verbose}"/>
             <property name="summaryDoc" value="false"/>
             <property name="suppressTEIexamples" value="true"/>
         </ant>
@@ -195,7 +198,7 @@
         <ant dir="submodules/Stylesheets/html5" antfile="build-to.xml" target="odd" inheritall="true">
             <property name="inputFile" value="${dir.dist.schemata}/${odd.basename}_compiled.odd"/>
             <property name="outputFile" value="${dir.dist.schemata}/${odd.basename}_compiled.html"/>
-            <property name="verbose">true</property>
+            <property name="verbose" value="${verbose}"/>
             <property name="processODD">true</property>
         </ant>
     </target>
@@ -209,7 +212,7 @@
             <property name="inputFile" value="${customization.path}"/>
             <property name="outputFile" value="${dir.dist.schemata}/${odd.basename}.rng"/>
             <property name="defaultSource" value="${canonicalized.source.path.cross_platform}"/>
-            <property name="verbose">true</property>
+            <property name="verbose" value="${verbose}"/>
         </ant>
     </target>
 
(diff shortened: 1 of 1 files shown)


build.xml
```diff
@@ -72,6 +72,9 @@
     <property name="xerces.jar.file" value="oxygen-patched-xerces-${xerces.version}.jar"/>
     <property name="oxygen.basic.utilites.jar.file" value="oxygen-basic-utilities-${xerces.version}.jar"/>
     <path id="mei.classpath">
+    <condition property="verbose" value="true" else="false">
+        <contains string="${sun.java.command}" substring="-v" />
+    </condition>
```
musicEnfanthen, line 76: Placement inside the classpath element correct? If yes, indentation correct?

musicEnfanthen:
Good thing, thanks.

Can we use this for the `loadresource` stuff too which adds quite some extra output? There is a `quiet` option for `loadresource` that could consume the value of `verbose` property?

rettinghaus (dismissed):

build.xml
bwbohl: uups something apparently went wrong when moving this to a new branch, thx for spotting!

rettinghaus (dismissed):
Thank you!

musicEnfanthen:
We could also use a looger to write all the output to a file instead of the console? Cf. https://stackoverflow.com/a/20643091

bwbohl:
> We could also use a looger to write all the output to a file instead of the console? Cf. https://stackoverflow.com/a/20643091

Yes, but probably overkill for now. Maybe open an issue for a solid logging design.

bwbohl:
> Good thing, thanks.
> 
> Can we use this for the `loadresource` stuff too which adds quite some extra output? There is a `quiet` option for `loadresource` that could consume the value of `verbose` property?

The `-quiet` option can also be submitted on the ant call but that would override `-verbose`. So if you call it with `quite` `level="verbose"`will not be printed only warnings and errors. The logging level can be set on all echo task and some other.

musicEnfanthen (dismissed):

musicEnfanthen:
OK, should be fine for the moment. We can fine-tune later on.

bwbohl:
> Good thing, thanks.
> 
> Can we use this for the `loadresource` stuff too which adds quite some extra output? There is a `quiet` option for `loadresource` that could consume the value of `verbose` property?

Just read the documentation of the `loadresource/@quiet` will add it to this PR.

build.xml
```diff
@@ -29,23 +29,23 @@
     <property name="github.ref" value="">
         <!-- string    external    The branch or tag ref that triggered the workflow run. For branches this in the format refs/heads/<branch_name>, and for tags it is refs/tags/<tag_name> -->
     </property>
-    <loadresource property="github.ref-type">
+    <loadresource property="github.ref-type" quiet="true">
```
musicEnfanthen, line 32: When running ant with `verbose` option, the quiet mode is overriden. So I think it is fine to hardcode "`true`" here, right? (Otherwise, we could have used the value of `$verbose` here too.)

build.xml
```diff
@@ -71,6 +71,9 @@
     <property name="xerces.version" value="25.1.0.1"/>
     <property name="xerces.jar.file" value="oxygen-patched-xerces-${xerces.version}.jar"/>
     <property name="oxygen.basic.utilites.jar.file" value="oxygen-basic-utilities-${xerces.version}.jar"/>
+    <condition property="verbose" value="true" else="false">
+        <contains string="${sun.java.command}" substring="-v" />
```
musicEnfanthen, line 75: This should check for substring `-verbose` as well, since you can run ant with both flags in verbose mode.

musicEnfanthen: Or will the check find that already? (`-v` as substring of `-verbose`)

quite (commented):
quite a typo in commit message ;)

build.xml
bwbohl: Yes, the check for `-v` will also evaluate true for `-verbose`.

build.xml
bwbohl: Yes, `-v` and `-verbose` respectively seem to override `@quiet`.

bwbohl:
> quite a typo in commit message ;)

sorry for the hassle

rettinghaus (approved):

musicEnfanthen (approved):

musicEnfanthen:
@ahankinson May I ask you to have a quick look at this PR, too?

kepper (approved):
thanks

musicEnfanthen:
Thank you!

PR #1203, opened 2023-06-05
build: improve gh param handling

### LINKED ISSUE #960: build: redesign guidelines build to match target folder structure
When building the guidelines locally the resulting folder structure is not identical to the structure in the guidelines repo. Consequently local testing does not work 100%. We should inspect this and maybe move the copy actions executed by the GitHub Action script to be run in the ant build, instead.

bwbohl:
might fix #960


musicEnfanthen:
@bwbohl Can you revisit if this is still needed?

utils/guidelines_xslt/odd2html/config.xsl
```diff
@@ -136,8 +136,8 @@
             <xd:p>The dist folder, in which the final results of the XSLT are stored.</xd:p>
         </xd:desc>
     </xd:doc>
-    <xsl:variable name="dist.folder" select="$cleaned.basedir || 'dist/guidelines/web/'" as="xs:string"/>
-
+    <xsl:variable name="dist.folder" select="$cleaned.basedir || 'dist/guidelines/' || $revision || '/web/'" as="xs:string"/>
```
bwbohl, line 139: this is the interesting part, as it would allow to build for different revisions. But maybe it would be better to build into `'dist/' || $revision`?

rettinghaus:
@bwbohl how to proceed with this?

PR #1207, opened 2023-06-12
Adjust MEI- Basic customization to use att.duration.log for rest

lpugin:
This PR adjusts `att.rest.log` in MEI-Basic to use `att.duration.log` instead of `att.restduration.log`. 

The reason for this change is that `att.restduration.log` is meaningless in MEI-Basic since it exists essentially to allow for additional rest durations in mensural notation which are not available in MEI-Basic.

The change in the customization yields no modification to the resulting MEI-Basic schema and is only intended to simplify handling of rests with code generated directly from the MEI-Basic ODD.

### DIFF
--- a/customizations/mei-basic.xml
+++ b/customizations/mei-basic.xml
@@ -53,7 +53,7 @@
             <change n="2" when="2019-10-26" who="#JK">
                 <desc>Revisions following the discussions on GitHub, and at the Nashville Hackathon.</desc>
             </change>
-            <change n="3" when="2023-05-24" who="#LP">
+            <change n="3" when="2023-06-12" who="#LP">
                 <desc>Revisions at the MEI Developers Conference, Charlottesville.</desc>
             </change>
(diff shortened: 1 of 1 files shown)


customizations/mei-basic.xml
```diff
@@ -53,7 +53,7 @@
             <change n="2" when="2019-10-26" who="#JK">
                 <desc>Revisions following the discussions on GitHub, and at the Nashville Hackathon.</desc>
             </change>
-            <change n="3" when="2023-05-24" who="#LP">
+            <change n="3" when="2023-06-12" who="#LP">
```
rettinghaus: No need to update this here. If you want the changes to have a common entry, it should be:
```suggestion
            <change n="3" from="2023-05-24" to="2023-06-12" who="#LP">
```

rettinghaus (approved):
Thanks!

bwbohl (approved):

PR #1208, opened 2023-06-13
Add `ending` to MEI-Basic

lpugin:
The PR adds `<ending>` element to MEI-Basic. It is currently missing, probably because it has been overlooked at some point.

I did not think this change needed preliminary discussions since it is a quite common feature in CWMN. Please say so if you think it does need to be discussed.


kepper (approved):
sure! good catch 👍

rettinghaus (approved):

bwbohl (approved):

Issue #1194, opened 2023-06-02
Harmonize naming scheme for data types and attribute classes


bwbohl:
As surfaced in #1173 the naming scheme of data types could be harmonized.

>he general pattern for datatypes is data.UPPERCASE.lowercase. The 3rd part of the name functions as a qualifier, sometimes by repertoire, sometimes signifying membership in a higher order class, sometimes for distinction between two similarly named datatypes that couldn't be differentiated by name alone.

>The exceptions to this that I could find with a quick visual scan are data.OCTAVE.DIS, which I would rename to data.OCTAVEDIS, and data.PERCENT.LIMITED and data.PERCENT.LIMITED.SIGNED, which can be renamed to data.PERCENT.limited and data.PERCENT.limited.signed (or data.PERCENT.limitedsigned), respectively. It would be good to search for other exceptions with a regex.

>In the case of note head modifier datatypes, data.NOTEHEADMODIFIER.list can be renamed as data.NOTEHEADMODIFIERLIST, and data.NOTHEADMODIFIER.pat can be renamed to data.NOTHEADMODIFIERPATTERN with no ill effects.

Also from #1157:

>@rettinghaus: Thinking about the proposed changes for keySig/key it might be better to name it att.key.anl.
>
> @ahankinson: We currently don't have any other att.key.* attributes, but the other att.key* classes we have (att.keyAccid, att.keySig, att.keySigDefault) all have the domain appended. So perhaps att.keyMode.anl?

bwbohl:
- [ ] `att.neumetype` should probably be `att.neumeType` as surfaced in #1157

PR #1210, opened 2023-06-14
schema: fix mixing visual and gestural values in keySig

### LINKED ISSUE #590: key signatures: written vs. gestural

rettinghaus:
This PR aligns `att.keySig.anl` with `att.keySigDefault.anl` (as it is supposed to be). 

Now `key@accid`, `keySig@accid`, `scoreDef@key.accid`, and `staffDef@key.accid` all require  a value from `data.ACCIDENTAL.GESTURAL.basic`.

closes #590

refers to #1157

### DIFF
--- a/source/modules/MEI.analytical.xml
+++ b/source/modules/MEI.analytical.xml
@@ -245,10 +245,24 @@
   <classSpec ident="att.keySig.anl" module="MEI.analytical" type="atts">
     <desc xml:lang="en">Analytical domain attributes.</desc>
     <classes>
-      <memberOf key="att.accidental"/>
-      <memberOf key="att.keyMode"/>
       <memberOf key="att.pitch"/>
     </classes>
+    <attList>
+      <attDef ident="accid" usage="opt">
+        <gloss versionDate="2023-06-14" xml:lang="en">accidental</gloss>
+        <desc xml:lang="en">Contains an accidental for the tonic key, if one is required, <abbr>e.g.</abbr>, if <att>pname</att>
+          equals <val>c</val> and <att>accid</att> equals <val>s</val>, then a tonic of C# is indicated.</desc>
+        <datatype>
+          <rng:ref name="data.ACCIDENTAL.GESTURAL.basic"/>
+        </datatype>
+      </attDef>
+      <attDef ident="mode" usage="opt">
+        <desc xml:lang="en">Indicates major, minor, or other tonality.</desc>
+        <datatype>
+          <rng:ref name="data.MODE"/>
+        </datatype>
+      </attDef>
+    </attList>
   </classSpec>
   <classSpec ident="att.keySigDefault.anl" module="MEI.analytical" type="atts">
     <desc xml:lang="en">Used by staffDef and scoreDef to provide default values for attributes in the analytical
@@ -258,7 +272,7 @@
         <desc xml:lang="en">Contains an accidental for the tonic key, if one is required, <abbr>e.g.</abbr>, if <att>key.pname</att>
           equals <val>c</val> and <att>key.accid</att> equals <val>s</val>, then a tonic of C# is indicated.</desc>
         <datatype>
-          <rng:ref name="data.ACCIDENTAL.GESTURAL"/>
+          <rng:ref name="data.ACCIDENTAL.GESTURAL.basic"/>
         </datatype>
       </attDef>
       <attDef ident="key.mode" usage="opt">
--- a/source/modules/MEI.analytical.xml
+++ b/source/modules/MEI.analytical.xml
@@ -249,7 +249,7 @@
     </classes>
     <attList>
       <attDef ident="accid" usage="opt">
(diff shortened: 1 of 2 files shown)


source/modules/MEI.analytical.xml
```diff
@@ -242,17 +242,28 @@
   </classSpec>
   <classSpec ident="att.keySig.anl" module="MEI.analytical" type="atts">
     <desc xml:lang="en">Analytical domain attributes.</desc>
-    <classes>
-      <memberOf key="att.accidental"/>
-      <memberOf key="att.pitch"/>
-    </classes>
```
ahankinson, line 251: I don’t think we should duplicate the definition of `pname` locally here. There doesn’t seem to be any difference between the attribute class that is being removed and the attribute definition that is being added.

source/modules/MEI.analytical.xml
```diff
@@ -242,17 +242,28 @@
...
     <attList>
+      <attDef ident="accid" usage="opt">
+        <gloss versionDate="2023-06-14" xml:lang="en">accidental</gloss>
+        <desc xml:lang="en">Contains an accidental for the tonic key, if one is required, <abbr>e.g.</abbr>, if <att>pname</att>
+          equals <val>c</val> and <att>accid</att> equals <val>s</val>, then a tonic of C# is indicated.</desc>
+        <datatype>
+          <rng:ref name="data.ACCIDENTAL.GESTURAL"/>
```
ahankinson: Perhaps it would make more sense to only allow `data.ACCIDENTAL.GESTURAL.basic` here, since allowing the full datatype also allows extended, aeu, and persian which I don't think is necessary?

source/modules/MEI.header.xml
```diff
@@ -1635,22 +1635,13 @@
...
       <memberOf key="att.bibl"/>
-      <memberOf key="att.pitch"/>
+      <memberOf key="att.keySig.anl"/>
       <memberOf key="model.workIdent"/>
     </classes>
     <content>
       <rng:text/>
     </content>
-    <attList>
-      <attDef ident="mode" usage="opt">
```
ahankinson: Maybe I'm missing something, but without `@mode` how do you say something is in "F# minor"?

rettinghaus:
Aligned also key element. Still, maybe better to rename the class into `att.key`?

ahankinson (commented):
Added a question about `@mode`

source/modules/MEI.header.xml
rettinghaus: It's part of `att.keySig.anl`.

ahankinson: So it's a change from `@mode` to `@key.mode`? In general I think this is a good thing, but I didn't see it mentioned anywhere in an issue or description, so wanted to check.

pe-ro (approved):

pe-ro (dismissed):

bwbohl:
Edit: [The renaming] should be tackled in the context of #1194, moving to MEI 6

rettinghaus:
After more than ten months without any further comments, I believe this is ready to be merged?

bwbohl:
I’m entirely with you @rettinghaus, that this should get merged. Nevertheless, I pushed it to MEI 6 milestone because that’s when we planned to rename classes consistently. Please excuse this strategic delay ;-)

bwbohl (dismissed):

source/modules/MEI.analytical.xml
```diff
@@ -245,10 +245,24 @@
...
+        <datatype>
+          <rng:ref name="data.ACCIDENTAL.GESTURAL.basic"/>
+        </datatype>
+      </attDef>
+      <attDef ident="mode" usage="opt">
+        <desc xml:lang="en">Indicates major, minor, or other tonality.</desc>
+        <datatype>
+          <rng:ref name="data.MODE"/>
+        </datatype>
+      </attDef>
```
musicEnfanthen, line 264: Isn't this identical to `att.keyMode`?

rettinghaus: I think the idea was to remove `att.keyMode` in favor of the inline definition in `att.keySig.anl`.

musicEnfanthen: I guess than `att.keyMode` could be deleted since it is not used anywhere else?

On the other hand, `att.pitch` and `att.accidental` hold only one attribute, too.

bwbohl: Let’s have a proper review of unused classes and other definitions before we finalise the release version

bwbohl: cf. #1726

bwbohl (approved):

riedde (approved):
Looks good! Time to merge it.

PR #1212, opened 2023-06-14
schema: fix visibility of mRest and multiRest

rettinghaus:
`rest` is not a member of `att.visibility`, because we have [`space`](https://music-encoding.org/guidelines/dev/content/shared.html#sharedNoteSpacing). So that `mRest` actually is a member of `att.visibility` seems wrong, because we have `mSpace`.

`multiRest` on the other hand was indeed missing the `visible` attribute.

### DIFF
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -1253,6 +1253,7 @@
       <memberOf key="att.staffLoc.pitched"/>
       <memberOf key="att.typography"/>
       <memberOf key="att.width"/>
+      <memberOf key="att.visibility"/>
     </classes>
     <attList>
       <attDef ident="block" usage="opt">
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -1252,8 +1252,8 @@
       <memberOf key="att.staffLoc"/>
       <memberOf key="att.staffLoc.pitched"/>
       <memberOf key="att.typography"/>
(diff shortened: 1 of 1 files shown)


source/modules/MEI.visual.xml
```diff
@@ -1254,6 +1253,7 @@
       <memberOf key="att.staffLoc.pitched"/>
       <memberOf key="att.typography"/>
       <memberOf key="att.width"/>
+      <memberOf key="att.visibility"/>
```
```suggestion
      <memberOf key="att.visibility"/>
      <memberOf key="att.width"/>
```

musicEnfanthen: Alphabetical order

musicEnfanthen: Out of curiosity, what's the point of an invisible multiRest? It would not cover any more measure space than an mSpace, right?

musicEnfanthen:
Out of curiosity, what's the point of an invisible multiRest? It would not cover any more measure space than an mSpace, right?

rettinghaus:
Not really. Some publishers don't bother to show the exact length of a `mutliRest` and print e.g. "TACET" instead of the rest.  But still you may want / need to encode how many measures are hidden behind the text.

bwbohl (approved):

musicEnfanthen (approved):
Thanks @rettinghaus. We should make sure that this change is reflected in the XSLT

PR #1188, opened 2023-05-30
Remove "italic" and "oblique" values from data.TEXTRENDITIONLIST

pe-ro:
The values "italic" and "oblique" duplicate values available in data.FONTSTYLE.

Fixes #1102

PR #1214, opened 2023-06-14
Revert "Remove "italic" and "oblique" values from data.TEXTRENDITIONLIST"

bwbohl:
Reverts music-encoding/music-encoding#1188


musicEnfanthen (approved):

musicEnfanthen:
Explanation for reversion given in https://github.com/music-encoding/music-encoding/pull/1188#issuecomment-1592045421. We should address this topic more broadly with the next release after v5.

PR #1218, opened 2023-06-15
guidelines: fix and unify examples

rettinghaus:
This PR fixes all encoding errors in the new examples in the Verovio folder and add some more, which have been used for Chapter 4 in the Guidelines. 

All files have now the same `pubStmt` with proper license and a unified `seriesStmt`.

### DIFF
--- a/source/examples/verovio/alteration.mei
+++ b/source/examples/verovio/alteration.mei
@@ -1,65 +1,77 @@
 <?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
 <?xml-model href="https://music-encoding.org/schema/dev/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
 <mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0.0-dev">
-   <meiHead>
-      <fileDesc>
-         <titleStmt>
-            <title>Example of 'alteration'</title>
-         </titleStmt>
-         <pubStmt>
-            <date>2019-11-15</date>
-         </pubStmt>
-         <notesStmt>
-            <annot>The bottom staff, together with the dotted barlines, is used here to help visualizing the durational values of the notes in the upper staff.</annot>
-         </notesStmt>
-      </fileDesc>
-   </meiHead>
-   <music>
-      <body>
-         <mdiv>
-            <score>
-               <scoreDef>
-                  <staffGrp>
-                     <staffDef label="voice" n="1" notationtype="mensural.white" lines="5" clef.shape="G" clef.line="2" modusminor="3" />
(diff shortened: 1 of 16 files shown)


bwbohl (commented):
Many thanks for your work, one question though, why are there new examples that are not referenced from the guidelines?

musicEnfanthen:
@rettinghaus Can you remove the newly added exmaples from this PR, please? They are not referenced from anywhere and would just be built with every commit to no avail. But please don't throw them away, will be great to see those included after release of v5 .

bwbohl (dismissed):

source/examples/verovio/alteration.mei
```diff
@@ -1,65 +1,77 @@
...
+  <meiHead>
+    <fileDesc>
+      <titleStmt>
+        <title>Example of 'alteration'</title>
+      </titleStmt>
+      <pubStmt>
+        <date isodate="2023">2023</date>
+        <respStmt>
+          <corpName>Music Encoding Initiative (MEI) Board</corpName>
+        </respStmt>
```
musicEnfanthen, line 13: There seems to be some inconsistency here: Sometimes `respStmt` is added, in other cases it is not present.

rettinghaus: Should be better now.

source/examples/verovio/implicit-mensuration.mei
```diff
@@ -1,82 +1,98 @@
...
+                clef.line="4" mensur.sign="C" mensur.slash="1" prolatio="2" tempus="2" />
+              <staffDef n="3" label="Altus" lines="5" notationtype="mensural.white" clef.shape="C"
+                clef.line="3" prolatio="2" tempus="2" />
+              <staffDef n="4" label="Bassus" lines="5" notationtype="mensural.white" clef.shape="F"
+                clef.line="4" prolatio="2" tempus="2" />
+            </staffGrp>
+          </scoreDef>
+          <section>
+            <staff n="1">
+              <layer>
```
musicEnfanthen, line 53: `n` attribute missing on all layers in this file

musicEnfanthen: So it still gives a  `[Warning] Missing @n on <layer>, filled by order `

rettinghaus: That's not on me, that is as it was before. It's optional in MEI, but needed for Verovio. If it bothers you, please ask the creator of the file (or the software).

musicEnfanthen: I see. 

@martha-thomae @music-encoding/ig-mensural Would you be fine with adding `n` attributes to the layer elements here?

rettinghaus: @musicEnfanthen Please open a separate issue for that, as it is not connected to this PR.

musicEnfanthen: See #1255

musicEnfanthen (approved):

bwbohl (approved):

PR #1219, opened 2023-06-15
utils: fix guidelines html

### LINKED ISSUE #1199: Undefined variable on elements pages
Using Chrome on MacOS, I get a javascript error when opening an element page:

https://music-encoding.org/guidelines/dev/elements/slur.html

Screenshot 2023-06-03 at 8 44 50 AM

This prevents changing the display style of the attributes list.

Here is the file line causing the problem:

Screenshot 2023-06-03 at 8 45 49 AM

```javascript
        const newTab = document.getElementById(style + '_tab');
        newTab.classList.add('active');
```

where `newTab` is undefined.   `style` is an input parameter to the function (style is either null or there is no tab ID with the name `style + '_tab'`.

A similar problem (likely the same one) has occurred previously, related to the `localStorage` variables.  Here is what I see for the contents of `localStorage`:

Screenshot 2023-06-03 at 8 50 18 AM

When I delete these variables and reload the page, the problem is fixed.  It seems that `localStorage.meiSpecs_attributes_display` is somehow causing the problem (and/or its value of `full`).

`localStorage` after deleting as well as reloading the page:

Screenshot 2023-06-03 at 9 09 41 AM

The value of `meiSpecs_attributes_display` is now changed from `full` to `attributes_full`.  Perhaps this is an interaction between older and newer versions of the guidelines (currently 3, 4, dev).   This seems the most likely problem, since viewing the version 4 page:

https://music-encoding.org/guidelines/v4/elements/slur.html#attributes

causes me a similar problem (but now in reverse) since the MEI4 version uses `full` for tab display and MEIdev uses `attributes_full`.

The solution to avoid this conflict  can be implemented in several ways.  And easy way would be to do something like `meiSpecs_mayContain_display_dev` where `_dev` is the development version of the guidelines.   Then v4 (and v3) can use the variable `meiSpecs_mayContain_display` or they could be updated to use `meiSpecs_mayContain_display_v4`, where there is a version tag in the variable name (derived by the URL piece that comes between `guidelines` and `elements`.

Another solution would be to either keep using `full` instead of `attributes_full`.  The `attributes_` tag is not necessary in the data string since the variable already contains it: `meiSpecs_attributes_display`, so the "attributes" scope is already understood.

Part of the solution should be to check if `newTab` is defined before using it, and also the solution would be mostly to keep track of the value of `style` when it is generated to see if it remains `null` after generating it.   This could also be related to the final solution where values of `full` and `attributes_full` could be automatically fixed based on the version (although above solutions are better).

Such as on line 1624 of the same page (for the current compiling of the website):

```javascript
const defaultValue = facetUl.children[0].children[0].getAttribute('data-display');
```

`defaultValue` should be checked to see if it is defined after this line, and deal with an undefined value (does not seem to be a problem in this case). `localStorage` is checked near this and if access is falsy the above code is run, but the final attributes style string is not checked for validity if it is not null (which is resulting in the cause of this problem).

musicEnfanthen:
This PR tries to tackle issue #1199 . It introduces some null checks in the creation steps of the facet tabs of the html guidelines and adjusts the values used for the localStorage to be compatible again to those in v4.

It also adjust the anchor links of the tabs to not jump down to the tab content (and thus moving the tab headers out of the screen).

Fixes #1199

### DIFF
--- a/utils/guidelines_xslt/odd2html/generateWebsite.xsl
+++ b/utils/guidelines_xslt/odd2html/generateWebsite.xsl
@@ -677,7 +677,7 @@
                             <li class="tab-item">
                                 <a data-display="{@id}" 
                                     id="{@id}_tab"
-                                    href="#{@id}"
+                                    href="#{@id}_tab"
                                     class="displayTab{if(position() = 1) then(' active') else()}"><xsl:value-of select="@data-label"/></a>
                             </li>
                         </xsl:for-each>
--- a/utils/guidelines_xslt/odd2html/htmlFile.xsl
+++ b/utils/guidelines_xslt/odd2html/htmlFile.xsl
@@ -253,10 +253,10 @@
                             const tabbedFacets = document.querySelectorAll('.facet ul.tab');
                             
                             const tabClick = function(e) {
-                                const style = e.target.getAttribute('data-display');
+                                const targetDataDisplay = e.target.getAttribute('data-display');
                                 const facetId = e.target.parentNode.parentNode.parentNode.parentNode.id;
-                                //console.log('clicked at ' + facetId + ' with style ' + style)
-                                setTabs(facetId,style)
+                                //console.log('clicked at ' + facetId + ' with data display ' + targetDataDisplay);
+                                setTabs(facetId,targetDataDisplay);
                             }
                             
                             console.log('[INFO] Javascript initialized')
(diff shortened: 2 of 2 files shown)


kepper (approved):
as far as I can see, all this is good. At least it should be a(nother) step in the right direction ;-) Many thanks, @musicEnfanthen 👍

bwbohl (approved):

PR #1226, opened 2023-06-16
guidelines: fix caption for curve shape

ahankinson:
I noticed that this caption was missing the numeric values. I found the original in the MEI 4 documentation. Not sure if this is an anomaly or a systematic problem.


rettinghaus (approved):

musicEnfanthen (approved):

musicEnfanthen:
Thanks for the catch

bwbohl (approved):

PR #1230, opened 2023-06-16
guidelines: remove reference to ineume and uneume in chapter 6

vigliensoni:
This pull request removes the references to interrupted and uninterrupted neumes (`ineume` and `uneume`) from the guidelines.


source/docs/06-neumes.xml
```diff
@@ -244,7 +244,7 @@
...
-               <p>Broadly speaking, Western early notations belong to two categories. On one side we have notations where two or more notes were represented by a single pen-stroke, while on the other side there are notations where the notes are graphically separated by means of discrete dots or short pen-strokes. These distinctions have been described even within single notational styles as interrupted neumes `&lt;ineume&gt;` or uninterrupted `&lt;uneume&gt;` (Morent &amp; Schräder 2008; Morent 2011; MEI Guidelines v3, ch. 6: Neume Notation), and now as gapped or not gapped (Behrendt, Bain, &amp; Helsen 2017; MEI Guidelines v4, ch. 6: Neume Notation). To date, the MEI Neumes Module has been tested mainly on stroke notations (St. Gall, Old Hispanic, etc.), but also on Aquitanian point-notation.</p>
```
musicEnfanthen, line 247: Since the reference to the literature is now removed, do you want to keep the 3 mentioned titles for uneumes and ineumes in the bibliography 6.5?

musicEnfanthen:
Thank you for this PR and fix, @vigliensoni. Just a minor question above.

source/docs/06-neumes.xml
fujinaga: Yes, please remove the 3 titles in 6.5. Thank you.

fujinaga (approved):
Looks good to me.

source/docs/06-neumes.xml
```diff
@@ -244,7 +244,7 @@
...
+               <p>Broadly speaking, Western early notations belong to two main categories. On one side we have notations where two or more notes were represented by a single pen-stroke, while on the other side there are notations where the notes are graphically separated by means of discrete dots or short pen-strokes. These distinctions have been described even within single notational styles as gapped or not gapped (Behrendt, Bain, &amp; Helsen 2017). To date, the MEI Neumes Module has been tested mainly on square notations and stroke notations (St. Gall, Old Hispanic, etc.), but also on Aquitanian point-notation.</p>
```
ahankinson, line 247:
```suggestion
               <p>Broadly speaking, Western neume notations belong to two main categories. On one side we have notations where two or more notes were represented by a single pen-stroke, while on the other side there are notations where the notes are graphically separated by means of discrete dots or short pen-strokes. These distinctions have been described even within single notational styles as gapped or not gapped (Behrendt, Bain, &amp; Helsen 2017). To date, the MEI Neumes Module has been tested mainly on square notations and stroke notations (St. Gall, Old Hispanic, etc.), but also on Aquitanian point-notation.</p>
```

ahankinson: It's OK, but I don't see the point in the paragraph now. Since we've removed the distinction between "connected" and "unconnected" neumes, then this seems to be a bit of a dead-end paragraph, describing something that no longer matters for MEI.

Would it be problematic to just remove the paragraph entirely?

fujinaga: But what about the concept of gapped or not gapped?

ahankinson: Perhaps a sentence pointing out how to record the difference in MEI? Is there a `@gap` attribute? Which of the notations mentioned at the end are "gapped"? Which are "not gapped"?

I think this paragraph is lacking the "so what" part... It's OK that some notations are defined as gapped or ungapped, but what's the specific impact on MEI so that this is worth mentioning to the readers?

fujinaga: There is a `@con` attribute... Let me ask Elsa, who originally wrote the paragraph, to see if we can remove it.

musicEnfanthen: I would not remove it. 

There is some valuable information about the two different main categories of early notations, and that the current encoding approach (as described in the given literature incl. the guidelines) is mostly tested with stroke and square notations, right?

Examples for encodings of different types are given in the following section 6.4.

But I agree with @ahankinson that maybe one or two examples at the end of this paragraph (or references to existing examples) would help to illustrate the concept of gapped/not gapped for the uninformed reader.

ahankinson: > ...that the current encoding approach (as described in the given literature incl. the guidelines) is mostly tested with stroke and square notations, right?

In the previous MEI encoding method, there was a strong separation between "gapped" and "ungapped" as indicated by the different elements, `<ineume>` and `<uneume>`. You couldn't make an MEI file without choosing one or the other.

Now the difference is much more slight -- there is the optional `@con` attribute where you *may* indicate gapped (the default being "ungapped"), but you don't have to.

> To date, the MEI Neumes Module has been tested mainly on square notations and stroke notations (St. Gall, Old Hispanic, etc.), but also on Aquitanian point-notation.

This sentence is a non-sequitur, for a couple reasons. 

  1. The reader has no connection between the previous sentence about "gapped" vs. "ungapped" and this one so they're left uncertain about why it's being mentions. Are all of the notations mentioned gapped, or ungapped? Or does the list represent a set of neume notations that represent both gapped and ungapped styles?
  2. I did a `cmd-f` search in the neumes chapter, and save for one caption this is the only mention of 'gapped' or 'ungapped' in the whole chapter; the one exception being an example of Old Hispanic notation, which has quite specialized connections anyway ("looped", "extended".). I would expect somewhere to mention the availability of `con="g"`, and that *no* `@con` attribute present means that they are connected (at least, that's what I understand from not having any explicit "ungapped" value on `@con`). So if this was fundamental to understanding Western neume notation and the design of the module, I would expect it to be followed-up later on with some further explanation of its impact on the specific encoding.

musicEnfanthen: Yes, I totally see your point that after the recent changes the content of the paragraph appears quite "gapped". 

> Are all of the notations mentioned gapped, or ungapped? Or does the list represent a set of neume notations that represent both gapped and ungapped styles?

As far as I understand the concept, there is not a single neume notational style that is one OR the other, but they all include gapped or ungapped neumes to different extents (e.g., in St. Gall notation Clivis, Torculus and Porrectus would be examples of ungapped neumes, climacus and scandicus for gapped ones).

> save for one caption this is the only mention of 'gapped' or 'ungapped' in the whole chapter

> [musicEnfanthen:] maybe one or two examples at the end of this paragraph (or references to existing examples) would help to illustrate the concept of gapped/not gapped for the uninformed reader.

I previously missed that there is indeed a prominent example right at the beginning of the chapter in section [6.1.1 Basic four elements](https://music-encoding.org/guidelines/dev/content/neumes.html#basicfourelements) that illustrates the concept, but here called connected vs. non-connected (Fig. 58 & 59).

So while I agree that there are some inconsistencies, I wonder if fixing that wouldn't require a broader review that goes beyond the scope of this PR (and then possibly beyond version 5)?

bwbohl: I propose to move this discussion to a new issue for MEI 5.x and get this PR done

bwbohl: please see #1251

ahankinson (approved):
I'll approve to move it forward, but this section needs work.

musicEnfanthen:
Agreed. Thank you @ahankinson for approval and @vigliensoni and @fujinaga for the PR.

musicEnfanthen (approved):

PR #1235, opened 2023-06-20
customizations: add `<symbol>` to MEI Basic

lpugin:
This PR adjusts the MEI-Basic customization by adding `<symbol>`. The element is missing to encode text featuring SMuFL glyphs, such as tempo indications or similar text with music symbols.

The PR also removes erroneous references to `att.altSym`, which is not available in the customization.


customizations/mei-basic.xml
```diff
@@ -759,6 +761,20 @@
...
+                
+                <classSpec ident="att.symbol.vis" module="MEI.visual" type="atts" mode="change">
+                    <desc xml:lang="en">Visual domain attributes.</desc>
+                    <classes>
+                        <!--<memberOf key="att.altSym"/>-->
+                        <memberOf key="att.color"/>
+                        <memberOf key="att.extSym"/>
+                        <memberOf key="att.scalable"/>
+                        <!--<memberOf key="att.typography"/>-->
+                        <!--<memberOf key="att.visualOffset"/>-->
```
ahankinson: It seems strange that we would remove the offset attributes here, when they're available on other elements and attribute classes. 

Given that one of the purposes of MEI Basic is to promote notation interchange between software, and one of the things that users find most frustrating when they interchange things is poor preservation of symbol layout, then I think I would be in favour of re-adding the `@ho`, `@vo`, (and maybe? `@to`?) back to the symbols.

lpugin: Yes, this makes sense. I will add them back.

ahankinson (approved):

musicEnfanthen (approved):

PR #1239, opened 2023-06-21
build: only run prince if command found

### LINKED ISSUE #1201: build: only run prince if command found
When running the `dist` target locally without having Prince XML installed, the build-guidelines-pdf subtask fails. Maybe this should fail more gently, e.g., by adding console feedback or adding a text file describing the reason. Or maybe the dist, or init targets should precheck for Prince XML, too?

bwbohl:
This should automatically decide whether to generate the pdf or not, depending on the availability of `prince` on `$PATH`.

Please test both in docker-mei and on a local system. Report back on the local system OS, whether Saxon, xerces or prince were available on your `$PATH`.


rettinghaus (approved):

musicEnfanthen:
@bwbohl This PR is missing the full section of the old PR where the `prince-available` command was introduced, so at the moment it is never initiated at all. Also, what is the env variable used for? Is it never called?

https://github.com/music-encoding/music-encoding/pull/1221/files#diff-766797f233c18114f9499750cf1ffbf3829aeea50283850619c01bd173132021R137-R148

build.xml
```diff
@@ -139,14 +139,19 @@
         <delete dir="lib"/>
         <antcall target="clean"/>
     </target>
-    
-    <target name="init" description="Initialize: checks if saxon and xerces are available, and if not, downloads them." unless="${docker}">
+
+    <target name="init" description="Initialize: checks if saxon and xerces are available, and if not, downloads them Checks if prince is available." unless="${docker}">
```
rettinghaus: Please double check this description.

bwbohl: thx

bwbohl:
```suggestion
    <target name="init" description="Initialize: checks if saxon and xerces are available, and if not, downloads them. Checks if prince is available." unless="${docker}">
```

rettinghaus (approved):

bwbohl:
@musicEnfanthen reintroduced missing section, please have a second look ;-)

musicEnfanthen (approved):
Tested locally and with docker, thanks.

PR #1246, opened 2023-06-30
schema: create a `att.runningtext` attribute class

lpugin:
This PR adds a `att.pg.log` class for the `@func` attribute that is currently duplicated as direct attribute in `pgHead` and `pgFoot`.

No change in the resulting Schema.

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -2501,17 +2501,6 @@
       <memberOf key="att.alignment"/>
     </classes>
   </classSpec>
-  <classSpec ident="att.pg.log" module="MEI.shared" type="atts">
-    <desc xml:lang="en">Attributes that record the function (i.e., placement) of page text elements.</desc>
-    <attList>
-      <attDef ident="func" usage="opt">
-        <desc xml:lang="en">Records the function (i.e., placement) of a page header or footer.</desc>
-        <datatype>
-          <rng:ref name="data.PGFUNC"/>
-        </datatype>
-      </attDef>
-    </attList>
-  </classSpec>
   <classSpec ident="att.phrase.log" module="MEI.shared" type="atts">
     <desc xml:lang="en">Logical domain attributes.</desc>
     <classes>
@@ -2815,6 +2804,17 @@
           data.DURATIONRESTS datatype.</desc>
         <datatype>
           <rng:ref name="data.DURATIONRESTS"/>
+        </datatype>
+      </attDef>
+    </attList>
+  </classSpec>
+  <classSpec ident="att.runningtext" module="MEI.shared" type="atts">
+    <desc xml:lang="en">Attributes that record the function (i.e., placement) of page running text elements.</desc>
+    <attList>
+      <attDef ident="func" usage="opt">
+        <desc xml:lang="en">Records the function (i.e., placement) of a page header or footer.</desc>
+        <datatype>
+          <rng:ref name="data.PGFUNC"/>
         </datatype>
       </attDef>
     </attList>
@@ -6951,9 +6951,9 @@
     <desc xml:lang="en">A running footer.</desc>
     <classes>
       <memberOf key="att.common"/>
-      <memberOf key="att.pg.log"/>
       <memberOf key="att.facsimile"/>
       <memberOf key="att.lang"/>
+      <memberOf key="att.runningtext"/>
     </classes>
     <content>
       <rng:zeroOrMore>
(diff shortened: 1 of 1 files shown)


ahankinson:
Could you please provide a bit of a justification for why this change is necessary? Thanks!

lpugin:
Sure. It avoids the duplication of the `@func` attribute. Now nearly identical in two places:

https://github.com/music-encoding/music-encoding/blob/048832a1126389a031081a40672aeae4cdd2bff1/source/modules/MEI.shared.xml#L6966-L6971

https://github.com/music-encoding/music-encoding/blob/048832a1126389a031081a40672aeae4cdd2bff1/source/modules/MEI.shared.xml#L7011-L7016

Furthermore, having a attribute class is useful for Verovio since as it is, only the data type is generated but no C++ attribute class.

So the change is not strictly necessary, but since it actually does not change anything to MEI it is not really a change ;-)

ahankinson:
Great, thanks! Looks like an OK change to me.

ahankinson (dismissed):

rettinghaus:
I'm not happy about the name `att.pg.log` as it may lead to confusion ("function of the page"?). But AFAIK there is no generic term for header _and_ footer. So that's probably the best we can come up with.

lpugin:
A term found in typography context about these are "running" text elements. We could rename it to `att.runningtext.log` if you prefer

rettinghaus:
Yes, I like that much better.

lpugin:
@musicEnfanthen ? @bwbohl shall I rename it?

musicEnfanthen:
~Renaming sounds good to me.~ EDIT: looking at it again are you sure that "running text" is used to address header and footer? I read from its definition that is rather addressing the text body itself? https://www.merriam-webster.com/dictionary/running%20text

Also, do we need the .log suffix or would it be sufficient without it (cf. att.pitch)?

ahankinson:
I’d be cool with dropping “log”. It’s a cross domain attribute.

source/modules/MEI.shared.xml
```diff
@@ -2501,6 +2501,17 @@
       <memberOf key="att.alignment"/>
     </classes>
   </classSpec>
+  <classSpec ident="att.pg.log" module="MEI.shared" type="atts">
```
rettinghaus:
```suggestion
  <classSpec ident="att.runningtext" module="MEI.shared" type="atts">
```

source/modules/MEI.shared.xml
```diff
@@ -2501,6 +2501,17 @@
...
+    <desc xml:lang="en">Attributes that record the function (i.e., placement) of page text elements.</desc>
```
rettinghaus:
```suggestion
    <desc xml:lang="en">Attributes that record the function (i.e., placement) of running text elements.</desc>
```

source/modules/MEI.shared.xml
```diff
@@ -6940,6 +6951,7 @@
     <desc xml:lang="en">A running footer.</desc>
     <classes>
       <memberOf key="att.common"/>
+      <memberOf key="att.pg.log"/>
```
rettinghaus:
```suggestion
      <memberOf key="att.runningtext"/>
```

source/modules/MEI.shared.xml
```diff
@@ -6985,6 +6991,7 @@
     <desc xml:lang="en">A running header.</desc>
     <classes>
       <memberOf key="att.common"/>
+      <memberOf key="att.pg.log"/>
```
rettinghaus:
```suggestion
      <memberOf key="att.runningtext"/>
```

lpugin:
@musicEnfanthen "running text" in this context is a shortcut for "running header text" and "running footer text". I am OK with it, but if you find it too confusing, we can look for something else.

musicEnfanthen:
No, I am fine with it. Just wanted to doublecheck that meaning fits for human readers 😉

musicEnfanthen (approved):

ahankinson (approved):

ahankinson:
@rettinghaus are you ok with it now? If so I'll merge.

rettinghaus:
@ahankinson please go ahead.

pe-ro:
Sorry for joining the party late.  The generic term for running headers, footers, catchwords, etc. on a printed page is "forme work".  We got rid of the `<fw>` element quite some time ago, but I think this would be a good opportunity to re-introduce the concept by naming this new class `att.formework`.

rettinghaus:
For reference: https://tei-c.org/release/doc/tei-p5-doc/de/html/ref-fw.html

PR #1248, opened 2023-06-30
schema: allow repeated pgHead and pgFoot in scoreDef

lpugin:
Fix `scoreDef` following refactoring of page headers and footers. The problem is that it does not allow `pgHead` and `pgFoot` to be repeated. This is a problem since we have removed `pgHead2` and `pgFoot2` and replaced them with `@func` on `pgHead` and `pgFoot`. So instead of having before
```xml
<scoreDef>
   <pgHead/>
   <pgHead2/>
</scoreDef>
```
We now want to be able to do
```xml
<scoreDef>
   <pgHead func="first"/>
   <pgHead func="all"/>
</scoreDef>
```

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -7613,12 +7613,12 @@
       <rng:optional>
         <rng:ref name="model.meterSigLike"/>
       </rng:optional>
-      <rng:optional>
+      <rng:zeroOrMore>
         <rng:ref name="pgHead"/>
-      </rng:optional>
-      <rng:optional>
+      </rng:zeroOrMore>
+      <rng:zeroOrMore>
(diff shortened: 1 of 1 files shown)


rettinghaus (approved):

musicEnfanthen (commented):
Makes sense. @lpugin Can you please check in the documentation if there are any references to the old elements/behaviour?

bwbohl:
@lpugin please report back, also if no documentation has to be updated.

lpugin:
There is a PR being discussed for the documentation https://github.com/music-encoding/music-encoding/pull/1249

musicEnfanthen (approved):

bwbohl (approved):

PR #1249, opened 2023-07-02
guidelines: add some documentation about att.formework

lpugin:
Minimal documentation for att.formework with a reference to the attribute class.

### DIFF
--- a/source/docs/02-shared.xml
+++ b/source/docs/02-shared.xml
@@ -322,8 +322,10 @@
                      <specList>
                         <specDesc key="pgHead"/>
                         <specDesc key="pgFoot"/>
+                        <specDesc key="att.formework" atts="func"/>
                      </specList>
                   </p>
+                  <p>Both <gi scheme="MEI">pgHead</gi> and <gi scheme="MEI">pgFoot</gi> have a <att>func</att> that allows to specify to which page(s) the forme work element applies.</p>
                   <p>Columned layout can be captured with the following elements:</p>
                   <p>
                      <specList>
--- a/source/docs/02-shared.xml
+++ b/source/docs/02-shared.xml
@@ -325,7 +325,7 @@
                         <specDesc key="att.formework" atts="func"/>
                      </specList>
                   </p>
-                  <p>Both <gi scheme="MEI">pgHead</gi> and <gi scheme="MEI">pgFoot</gi> have a <att>func</att> that allows to specify to which page(s) the forme work element applies.</p>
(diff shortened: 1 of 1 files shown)


source/docs/02-shared.xml
```diff
@@ -322,8 +322,10 @@
                      <specList>
                         <specDesc key="pgHead"/>
                         <specDesc key="pgFoot"/>
+                        <specDesc key="att.formework" atts="func"/>
                      </specList>
                   </p>
+                  <p>Both <gi scheme="MEI">pgHead</gi> and <gi scheme="MEI">pgFoot</gi> have a <att>func</att> that allows to specify to which page(s) the forme work element applies.</p>
```
ahankinson:
```suggestion
                  <p>"Forme work" is the name for running elements (page headers and footers). Both <gi scheme="MEI">pgHead</gi> and <gi scheme="MEI">pgFoot</gi> have a <att>func</att> attribute that allows encoders to specify the placement and repeating nature of a given header or footer.</p>
```

lpugin: Looks good, even though I find "placement" confusing. It is the placement of the forme within the document, but it could be understood as the placement of the fome content on the page.

ahankinson: Yeah, I wasn't super happy with that either, but the previous wording made it seem like it specified the pages themselves, rather than the repeating nature the header / footer (even / odd, etc.).

ahankinson: So I guess "placement" in this sense is "placed on the even pages". We could drop it and just make it "...specify the repeating nature of..." but that seems to not be quite descriptive enough somehow.

lpugin: I think saying "to which page(s) the forme work element applies" is good enough. Then encoders can look at the definition in `att.formework` to understand how `@func` works.

source/docs/02-shared.xml
```diff
@@ -322,8 +322,10 @@
                      <specList>
                         <specDesc key="pgHead"/>
                         <specDesc key="pgFoot"/>
+                        <specDesc key="att.formework" atts="func"/>
                      </specList>
                   </p>
+                  <p>Forme work" is the name for running elements (page headers and footers). Both <gi scheme="MEI">pgHead</gi> and <gi scheme="MEI">pgFoot</gi> have a <att>func</att> attribute that allows encoders to specify to which page(s) the forme work element applies. This includes alternating patterns.</p>
```
musicEnfanthen:
```suggestion
                  <p>"Forme work" is the name for running elements (page headers and footers). Both <gi scheme="MEI">pgHead</gi> and <gi scheme="MEI">pgFoot</gi> have a <att>func</att> attribute that allows encoders to specify to which page(s) the forme work element applies. This includes alternating patterns.</p>
```

musicEnfanthen (approved):

bwbohl (approved):

PR #1253, opened 2023-07-07
build: make init-mei-classpath depend on init target

bwbohl:
Before this commit, the properties set in the init target were only available in this target. Consequently, they would not be available in the dist target, so the build-guidelines-pdf target was never called.


musicEnfanthen (approved):

musicEnfanthen:
Yes, works again :)

PR #1259, opened 2023-07-12
build: add comparison of MEI versions

### LINKED ISSUE #1256: create and test version comparison

musicEnfanthen:
This PR adds an Ant target to run the MEI version comparison from build.xml. This prepares to include the comparison in the CI build process and gives the ability to include the comparison output file with releases and commits (for the dev branch).
By default, the target compares the current version (dev) to the previous version (MEI 4.0.1), but any other version can be compared via customized inputs.

New ant target tested locally and with docker.

Closes #1256

### DIFF
--- a/BUILD_COMMANDLINE.md
+++ b/BUILD_COMMANDLINE.md
@@ -152,6 +152,6 @@ The following targets can be called using `ant <target>`:
 | `build-guidelines-pdf` | Builds the PDF version of the MEI guidelines. (Calls `build-guidelines-html` before execution.) |
 | `init` | Initializes the build environment, i.e., checks if saxon and xerces are available, and if not, downloads jar files for Saxon, Xerces and adds them to the `lib` folder. Checks if prince is available. |
 | `init-mei-classpath` | Initializes the mei.classpath, which is essential for the schema generation, by prepending the jar files contained in the lib directory to the java classpath. Will be called automatically if needed. |
-| `compare-versions` | Compares the canonicalized sources of the MEI dev version with the previous stable version and creates an HTML output; custom versions and output folder can be set via `-Dsource`, `-Dold` and `-Doutput` input params. <br> NB: path of old file is relative to `comparison.xsl`, while others are relative to `build.xml`. |
+| `compare-versions` | Compares the canonicalized sources of the MEI dev version with the previous stable version and creates an HTML output; custom versions and output folder can be set via `-Dsource`, `-Dold` and `-Doutput` input params. |
 | `clean` | Deletes the following directories: `build`, `dist` and `temp`. |
 | `reset` | Resets the build environment. Same as `clean`, but additionaly deletes the `lib` directory with the Saxon and Xerces jar files. |
--- a/BUILD_COMMANDLINE.md
+++ b/BUILD_COMMANDLINE.md
@@ -152,5 +152,6 @@ The following targets can be called using `ant <target>`:
 | `build-guidelines-pdf` | Builds the PDF version of the MEI guidelines. (Calls `build-guidelines-html` before execution.) |
 | `init` | Initializes the build environment, e.g., downloads jar files for Saxon, Xerces and adds them to the `lib` folder. |
 | `init-mei-classpath` | Initializes the mei.classpath which is essential for the schema generation. Will be called automatically if needed. |
+| `compare-versions` | Compares the canonicalized sources of the MEI dev version with the previous stable version and creates an HTML output; custom versions and output folder can be set via `-Dsource`, `-Dold` and `-Doutput` input params. <br> NB: path of old file is relative to `comparison.xml`, while others are relative to `build.xml`. |
 | `clean` | Deletes the following directories: `build`, `dist` and `temp`. |
 | `reset` | Resets the build environment. Same as `clean`, but additionaly deletes the `lib` directory with the Saxon and Xerces jar files. |
--- a/BUILD_COMMANDLINE.md
+++ b/BUILD_COMMANDLINE.md
@@ -152,6 +152,6 @@ The following targets can be called using `ant <target>`:
 | `build-guidelines-pdf` | Builds the PDF version of the MEI guidelines. (Calls `build-guidelines-html` before execution.) |
 | `init` | Initializes the build environment, i.e., checks if saxon and xerces are available, and if not, downloads jar files for Saxon, Xerces and adds them to the `lib` folder. Checks if prince is available. |
 | `init-mei-classpath` | Initializes the mei.classpath, which is essential for the schema generation, by prepending the jar files contained in the lib directory to the java classpath. Will be called automatically if needed. |
-| `compare-versions` | Compares the canonicalized sources of the MEI dev version with the previous stable version and creates an HTML output; custom versions and output folder can be set via `-Dsource`, `-Dold` and `-Doutput` input params. <br> NB: path of old file is relative to `comparison.xml`, while others are relative to `build.xml`. |
+| `compare-versions` | Compares the canonicalized sources of the MEI dev version with the previous stable version and creates an HTML output; custom versions and output folder can be set via `-Dsource`, `-Dold` and `-Doutput` input params. <br> NB: path of old file is relative to `comparison.xsl`, while others are relative to `build.xml`. |
 | `clean` | Deletes the following directories: `build`, `dist` and `temp`. |
 | `reset` | Resets the build environment. Same as `clean`, but additionaly deletes the `lib` directory with the Saxon and Xerces jar files. |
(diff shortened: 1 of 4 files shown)


bwbohl:
many thanks @musicEnfanthen, this looks all good but maybe we should accompany this with adding the compiled source to  https://github.com/music-encoding/schema/tree/main/4.0.1?

musicEnfanthen:
Yes, agree, at least the canonicalized source for 4.0.1 that is needed for the comparison. I also created "semi-manually" a canonicalized source for v3 (at least it gives the same results as the comparison that is online for v3 vs. v4)

build.xml
```diff
@@ -390,7 +390,50 @@
...
+            <arg value="-xi:on"/>
+            <arg value="output.folder=${output.folder}"/>
+            <arg value="old.version.filename=${old.file}"/>
+        </java>
+        <echo>Comparison done. See ${output.folder}/comparison.html for results.</echo>
+    </target>
+
     <target name="dist" depends="init-mei-classpath, canonicalize-source">
+
         <classfileset refid="mei.classpath"/>
```
```suggestion
    <target name="dist" depends="init-mei-classpath, canonicalize-source">
        <classfileset refid="mei.classpath"/>
```

musicEnfanthen:
Accompanying PR on schema repo is available: https://github.com/music-encoding/schema/pull/7

build.xml
```diff
@@ -390,6 +390,48 @@
...
+        <echo level="verbose">source.default: ${source.default}</echo>
+        <echo level="verbose">source user input: ${source}</echo>
+
+        <condition property="source.file" value="${source}" else="${source.default}">
+            <isset property="source"/>
+        </condition>
+        <echo>Using source.file: ${source.file}</echo>
+
+        <!-- set old file to compare with (relative to comparison.xml) -->
+        <property name="old.default" value="MEI_v4.0.1_canonicalized.xml"/>
```
```suggestion
        <property name="old.default" value="mei-source_canonicalized_v4.0.1.xml"/>
```

bwbohl:
@musicEnfanthen could you resolve the tiny conflict, please?

musicEnfanthen:
@bwbohl Done, thanks for the pointer.

rettinghaus (approved):

BUILD_COMMANDLINE.md
```diff
@@ -152,5 +152,6 @@ The following targets can be called using `ant <target>`:
 | `build-guidelines-pdf` | Builds the PDF version of the MEI guidelines. (Calls `build-guidelines-html` before execution.) |
 | `init` | Initializes the build environment, i.e., checks if saxon and xerces are available, and if not, downloads jar files for Saxon, Xerces and adds them to the `lib` folder. Checks if prince is available. |
 | `init-mei-classpath` | Initializes the mei.classpath, which is essential for the schema generation, by prepending the jar files contained in the lib directory to the java classpath. Will be called automatically if needed. |
+| `compare-versions` | Compares the canonicalized sources of the MEI dev version with the previous stable version and creates an HTML output; custom versions and output folder can be set via `-Dsource`, `-Dold` and `-Doutput` input params. <br> NB: path of old file is relative to `comparison.xml`, while others are relative to `build.xml`. |
```
bwbohl:
```suggestion
| `compare-versions` | Compares the canonicalized sources of the MEI dev version with the previous stable version and creates an HTML output; custom versions and output folder can be set via `-Dsource`, `-Dold` and `-Doutput` input params. <br> NB: path of old file is relative to `comparison.xsl`, while others are relative to `build.xml`. |
```

bwbohl: Moreover, the paths could probably also be absolute.

build.xml
```diff
@@ -390,6 +390,48 @@
...
+        <property name="source.default" value="${dir.build}/mei-source_canonicalized.xml"/>
+        <echo level="verbose">source.default: ${source.default}</echo>
+        <echo level="verbose">source user input: ${source}</echo>
+
+        <condition property="source.file" value="${source}" else="${source.default}">
+            <isset property="source"/>
+        </condition>
+        <echo>Using source.file: ${source.file}</echo>
+
+        <!-- set old file to compare with (relative to comparison.xml) -->
```
bwbohl:
```suggestion
        <!-- set old file to compare with (relative to comparison.xsl) -->
```

build.xml
```diff
@@ -390,6 +390,48 @@
...
+        <property name="old.default" value="mei-source_canonicalized_v4.0.1.xml"/>
```
bwbohl: Why do you want this relative to `comparison.xsl`? The ant file will treat such paths relative to itself. So in terms of, e.g. checking the availability of $old file, it would make more sense to make this relative to build.xml or silently add the folder to a check (cf. https://github.com/musicEnfanthen/music-encoding/pull/1)

bwbohl:
Generally, it works great if the source-files are in place. But I wonder whether the $old.default should probably be fetched to temp or build instead of next to compariosn.xsl?
Moreover, we could think about moving the output to the dist.folder if one runs dist. Which makes me wonder why not by default put the build result of the taret in the build folder?

build.xml
musicEnfanthen: It was relative to comparison.xsl because the XSLT did not find it otherwise. But I think it should work with absolute path.

musicEnfanthen:
> Moreover, we could think about moving the output to the dist.folder if one runs dist. Which makes me wonder why not by default put the build result of the taret in the build folder?

The comparions.html needs to be in the same folder as the resources folder to be displayed correctly (with css and js). That's the only reason why I moved the output file next to the XSLT.

build.xml
```diff
@@ -390,6 +390,62 @@
...
+
+        <echo level="verbose">old.default: ${old.default}</echo>
+        <echo level="verbose">old user input: ${old}</echo>
+
+        <condition property="old.file" value="${old}" else="${old.default}">
+            <isset property="old"/>
+        </condition>
+        <echo>Using old.file: ${old.file}</echo>
+
+        <!-- check if old.file is available -->
```
```suggestion
        <!-- check if old.file is available, otherwise fail early -->
```

musicEnfanthen:
Thanks for the additional improvements @bwbohl . Looks great.

Reworked your suggestions for the ant target. As sugested, it now defines the location of the default old file in the build folder (since we run canonicalize-source beforehand the build folder will be created in any case and can be used). If it does not find a file there, it will download it from the schema repo. It will fail early now when the old source file cannot be found (which is probably only possible if there is a custom user input since downloaded default file should be available). So the checks if the file is available can be skipped with later steps. I left the output folder in the utils/compare_versions directory since the html file will need the resources to be displayed correctly. We can improve and move that in the future, but would keep it for now.

In the XSLT, I refactored the version variables and guideline links so that it works now with dev and major versions.

build.xml
```diff
@@ -390,6 +390,62 @@
...
+        </condition>
+        <echo>Using output folder: ${output.folder}</echo>
+
+        <!-- compare versions -->
+        <java classname="net.sf.saxon.Transform" classpathref="mei.classpath" failonerror="true" fork="true">
+            <arg value="-s:${source.file}"/>
+            <arg value="-xsl:${basedir}/utils/compare_versions/comparison.xsl"/>
+            <arg value="-xi:on"/>
+            <arg value="output.folder=${output.folder}"/>
+            <arg value="old.version.filename=/${old.file}"/>
```
bwbohl: Of course you're right that an absolute path will make the transformation work, no matter, where `${old.source}` is located. Running `ant compare-versions` natively from the music-encoding folder fails though. Moreover, setting it absolute here will prevent submitting an arbitrary file path. 

```suggestion
            <arg value="old.version.filename=${old.file}"/>
```

The absolute path for `${old.default}` could be achieved by prepending `${basedir}` in its definition (see above).

build.xml
```diff
@@ -390,6 +390,62 @@
...
+        <echo level="verbose">source.default: ${source.default}</echo>
+        <echo level="verbose">source user input: ${source}</echo>
+
+        <condition property="source.file" value="${source}" else="${source.default}">
+            <isset property="source"/>
+        </condition>
+        <echo>Using source.file: ${source.file}</echo>
+
+        <!-- set old file to compare with -->
+        <property name="old.default" value="${dir.build}/mei-source_canonicalized_v4.0.1.xml"/>
```
bwbohl, line 405:
```suggestion
        <property name="old.default" value="${basedir}/${dir.build}/mei-source_canonicalized_v4.0.1.xml"/>
```

see comment below

bwbohl (changes requested):

bwbohl:
with the two mentioned changes it works fine locally.

musicEnfanthen:
Unfortunately, your suggestions will not work on Win. The XSLT does not find the old.file then.

bwbohl:
In the current state, it does not work on macOS, though…

build.xml
bwbohl:
```suggestion
            <arg value="old.version.filename=file:${old.file}"/>
```

bwbohl (approved):

PR #1264, opened 2023-07-12
build: avoid multiple runs of initial targets and improve logging

### LINKED ISSUE #1262: build: target canonicalize source runs multiple times
The canonicalize-source targets gets triggered every time a sub target is started due to the dependency chains. It should be run only once since the canonicalized file is created then and ready to be used.

### LINKED ISSUE #1263: build: saxon or xerces downloaded with every build step when not on path
When saxon or xerces are not on path locally, build.xml will download it with every build step. This was tried to address in #1196 and #1209. But due to the immutability of ant properties, the recently introduced checks for `xerces-available` or `saxon-available` will be set to false (and stay false) when Saxon or Xerces are not on path. As a consequence, the properties will never be set to true and so the download targets will be triggered from any sub target.

musicEnfanthen:
This PR improves performance and logging of the build.xml by running `canonicalize-source` only once and not with every sub build target again. It also does improve the handling of the availability checks in `init` which are immutable ant properties and did not update on subsequent runs. This means, the condition itself cannot be set to true OR false, but will be set to true if a condition is met, otherwise stays unset. With `if:set` resp. `unless:set`, the output of the echo message and the content of the property are then updated correctly.

Fixes #1262
Fixes #1263

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -138,31 +138,36 @@
     
     <target name="init" description="Initialize: checks if saxon and xerces are available, and if not, downloads them. Checks if prince is available.">
         <!-- check if xerces is available -->
-        <condition  property="xerces-available" value="true" else="false">
+        <condition property="xerces-available">
             <or>
                 <available file="${dir.lib.xerces}/${xerces.jar.file}"/>
                 <available classname="org.apache.xerces.jaxp.SAXParserFactoryImpl"/>
             </or>
         </condition>
-        <echo message="xerces available: ${xerces-available}"/>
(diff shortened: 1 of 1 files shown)


musicEnfanthen:
To be more precise: The `init` and `canonicalize-source` targets still get triggered, but they are not running their transformations or subtargets any more as soon as the properties in question are available. They just give a friendly logging message that everything is set.

bwbohl (approved):

rettinghaus (approved):
Looks nice!

PR #1267, opened 2023-07-13
schema: fix typo in MEI.msDesc.xml

bwbohl:


rettinghaus (approved):
Nice catch!

ahankinson (approved):
No need to keep this open, so I'll merge.

PR #1268, opened 2023-07-13
schema: add missing values for double accidentals

rettinghaus:
From the Gould arrow quartertone accidentals (with arrows) the double sharps and double flats were missing. This PR adds these values to `data.ACCIDENTAL.WRITTEN.extended` and introduces corresponding values for five-quartertones to the gestural range.

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -107,6 +107,18 @@
         <valItem ident="nd">
           <desc xml:lang="en">Natural note lowered by quarter tone (natural modified by arrow).</desc>
         </valItem>
+        <valItem ident="xu">
+          <desc xml:lang="en">Double sharp note raised by quarter tone (double sharp modified by arrow).</desc>
+        </valItem>
+        <valItem ident="xd">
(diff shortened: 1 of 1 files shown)


source/modules/MEI.xml
```diff
@@ -227,6 +239,12 @@
         <valItem ident="fd">
           <desc xml:lang="en">Three quarter-tones flat.</desc>
         </valItem>
+        <valItem ident="xu">
+          <desc xml:lang="en">Five quarter-tones sharp.</desc>
+        </valItem>
+        <valItem ident="ffd">
+          <desc xml:lang="en">Five quarter-tones flat.</desc>
+        </valItem>
```
musicEnfanthen, line 247: Would we need `xd` and `ffu` as alternatives for three quarter-tone sharp/flat as well?

rettinghaus: We already have three-quarter values in the gestural range.

musicEnfanthen (approved):

bwbohl (approved):
thx @rettinghaus

Issue #1273, opened 2023-07-19
`@lform` and `lwidth` removed from `ending` in MEI Basic


lpugin:
It seems that [this](https://github.com/music-encoding/music-encoding/pull/1157) PR removed the attributes in MEI-Basic.

I tried to understand what happened but narrowing down to which PR caused the change is all I could do.

Any thoughts?

musicEnfanthen:
Thanks @lpugin for catching this. 

It seems that the commit dd345d25b8c5814486ae34165d476ccf143860b4 in PR #1157 replaced the direct `lform` and `lwidth` attributes in `att.lineRend.base` with membership in `att.curveRend` (which contains those two attributes). In consequence, elements like `ending` could access those attributes only via `att.curveRend` but that class was deleted in MEI basic.

PR #1279, opened 2023-07-22
schema: remove curveRend in favor of lineRend.base

rettinghaus:
This is another follow up for #1157 (and #1273).

`att.curveRend` and `att.lineRend.base` is used now in places where the former was in use. 

`att.curveRend` was using `@lform` and `@lwidth` (instead of `@cform` and `@cwidth`), yet in the attribute description the term "curve" was used. Furthermore it is unclear why you could say a curve is dashed, but were not allowed to encode how many dashes there are actually (@lsegs). 

This intends to improve consistency in the schema a bit. Everything should work as before with the addition of `@lsegs` to `bend`, `curve`, `lv`, `phrase`, `slur`, and `tie`.


lpugin (approved):

bwbohl (approved):

musicEnfanthen:
Thanks @rettinghaus.

musicEnfanthen:
@rettinghaus @lpugin @bwbohl Could you have a look at (kind of) related issue #1282 as well?

PR #1281, opened 2023-07-23
customizations: adjust att.harm.vis in MEI Basic

lpugin:
Current, `att.harm.vis` is deleted in MEI Basic. However, most of its attributes should be there, e.g., `@place`.

This PR preserve `att.harm.vis` in MEI Basic and adjust it to be inline with other control events.


rettinghaus (approved):
Makes sense, thank you.

musicEnfanthen (approved):

musicEnfanthen:
Thank you, @lpugin

bwbohl (approved):

PR #1286, opened 2023-07-27
schema: remove att.phrase.vis.cmn

rettinghaus:
This is (hopefully) the last PR connected to the clean-up around `lineRend` attributes. It removes `att.phrase.vis.cmn` and aligns `att.phrase.vis` with `att.slur.vis` and `att.tie.vis`.

As a follow up it should be discussed if the `elementSpec` for `phrase` should be moved from the `shared` to the `cmn` module like `slur` and `tie`.

### DIFF
--- a/source/modules/MEI.cmn.xml
+++ b/source/modules/MEI.cmn.xml
@@ -955,13 +955,6 @@
       </attDef>
     </attList>
   </classSpec>
-  <classSpec ident="att.phrase.vis.cmn" module="MEI.cmn" type="atts">
-    <desc xml:lang="en">Visual domain attributes.</desc>
-    <classes>
-      <memberOf key="att.curvature"/>
-      <memberOf key="att.lineRend.base"/>
-    </classes>
-  </classSpec>
   <classSpec ident="att.pianoPedals" module="MEI.cmn" type="atts">
     <desc xml:lang="en">Used by scoreDef and staffDef to provide default description of piano pedal
       rendition.</desc>
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -1446,11 +1446,12 @@
     <desc xml:lang="en">Visual domain attributes.</desc>
     <classes>
       <memberOf key="att.color"/>
+      <memberOf key="att.curvature"/>
+      <memberOf key="att.lineRend.base"/>
       <memberOf key="att.visualOffset"/>
       <memberOf key="att.visualOffset2"/>
       <memberOf key="att.xy"/>
       <memberOf key="att.xy2"/>
-      <memberOf key="att.phrase.vis.cmn"/>
     </classes>
   </classSpec>
   <classSpec ident="att.plica.vis" module="MEI.visual" type="atts">


musicEnfanthen (approved):
Seems plausible to me. `att.phrase.vis.cmn` is at the moment only available at `phrase` and a redundant extra layer providing the both classes `att.curvature` and `att.lineRend.base`

pe-ro:
Because in CMN phrase marks and slurs share the same visual mark, some people tend to confuse them and use their names interchangeably.  However, they are not the same thing. A curved line over a string passage is not a slur, it’s a phrase mark – take these notes on one bow.

The slur and tie elements belong in the CMN module.  The phrase element, on the other hand, is in the shared module because phrases can be marked in repertoires other than CMN.  When phrase marks occur in CMN, then they get visual attributes from the CMN repertoire; that is, att.phrase.vis.cmn.  Used some other repertoire, they might get attributes from att.phrase.vis.somethingElse.

Simplification is fine, but over simplification is not.


From: Klaus Rettinghaus ***@***.***>
Sent: Thursday, July 27, 2023 11:26 AM
To: music-encoding/music-encoding ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [music-encoding/music-encoding] Remove att.phrase.vis.cmn (PR #1286)


This is (hopefully) the last PR connected to the clean-up around lineRend attributes. It removes att.phrase.vis.cmn and aligns att.phrase.vis with att.slur.vis and att.tie.vis.

As a follow up it should be discussed if the elementSpec for phrase should be moved from the shared to the cmn module like slur and tie.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/music-encoding/music-encoding/pull/1286

Commit Summary

  *   0586852<https://github.com/music-encoding/music-encoding/pull/1286/commits/0586852ead4dbba63d34a7f50e1e55c18f57f765> remove att.phrase.vis.cmn

File Changes

(2 files<https://github.com/music-encoding/music-encoding/pull/1286/files>)

  *   M source/modules/MEI.cmn.xml<https://github.com/music-encoding/music-encoding/pull/1286/files#diff-93b6e3a58ee4217c9aa9036cc204887c84f6fd41243e4d9d541469e4239028e6> (7)
  *   M source/modules/MEI.visual.xml<https://github.com/music-encoding/music-encoding/pull/1286/files#diff-63466462567b6bd91b66683e13c7d6c235593f5b56825404def7d54fa34723c8> (3)

Patch Links:

  *   https://github.com/music-encoding/music-encoding/pull/1286.patch
  *   https://github.com/music-encoding/music-encoding/pull/1286.diff

rettinghaus:
Thank you @pe-ro. Could you please elaborate a bit more on what "other repertoires" you had in mind and share an example? 

My impression is that all repertoires that can be encoded with MEI already have specialized elements for something "phrase-like", e.g., `<ligature>` element in mensural. Also it is still not possible to use control elements in un-measured encodings, correct?

ahankinson:
A `<ligature>` isn't a phrase mark; it's a paleographic convention where multiple independent symbols are connected with a single "pen" stroke. In some repertoires (e.g., neumes) this *may also* indicate a way of phrasing a syllable, but it also may not, and multiple ligated symbols could combine under a single phrase. 

That said, other repertoires like tablature can have phrase markings. They may also be an analytical observation, where multiple annotators may differ on how to phrase passages in repertoires that may or may not be applied to repertoires that are not CMN. (e.g., applying phrasing annotations to chant to indicate possible ways of performing a passage). 

If anything, it could be moved to the `analytical` module; but I'm inclined to leave it alone in `shared`.

rettinghaus:
Thanks. I didn't say a ligature is a phrase mark. Still it's due to my lack of knowledge and imagination how a phrase mark would look in a non-CMN repertoire, that's why I asked for an example. So it would then be definitely not a line-like symbol, because that's strictly CMN? And how should it then be encoded?

pe-ro:
@ahankinson has it right.  Tablature is a good example of another repertoire where phrase indications may be necessary and where the symbols employed may or may not be line-like.  But I had in mind notational repertoires not yet covered by MEI, such as Byzantine chant and Asian notations (portions of which are tablature-like).  The graphical rendition of these phrase marks is unknown, but that's exactly the reason for the under-specification of `<phrase>`.  The current design allows `<phrase>` to take on whatever rendition the repertoire requires, including no rendition at all.

PR #1287, opened 2023-07-27
schema: fix double entries in att.neumeTypes

rettinghaus:
This PR removes doubled entries from `att.neumeType`.


musicEnfanthen (approved):
Good catch, thanks.

bwbohl (approved):

PR #1289, opened 2023-07-28
schema: remove doubled namespace declaration

rettinghaus:
This PR removes the doubled namespace declaration for TEI. We do not use `tei:` namespace prefix anywhere, so this is obsolete.

### DIFF
--- a/source/mei-source.xml
+++ b/source/mei-source.xml
@@ -127,8 +127,7 @@
       </body>
       <back>
          <div>
-            <schemaSpec xmlns:tei="http://www.tei-c.org/ns/1.0"
-                        ident="mei"
+            <schemaSpec ident="mei"
                         ns="http://www.music-encoding.org/ns/mei"
                         start="mei">
                <xi:include href="modules/MEI.xml"/>
--- a/source/modules/MEI.analytical.xml
+++ b/source/modules/MEI.analytical.xml
@@ -19,8 +19,7 @@
 -->
 <?xml-model href="../validation/mei_odds.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
 <?xml-model href="../validation/mei_odds.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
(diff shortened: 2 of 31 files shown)


ahankinson:
If we remove it does it mean that our schema validation will fail on other people embedding TEI in MEI?

rettinghaus:
@ahankinson how could it? This is our schema. Please explain.

kepper:
I might be wrong (as always), but technically, `xmlns="http://www.tei-c.org/ns/1.0"` and `xmlns:tei="http://www.tei-c.org/ns/1.0"` are two different statements: One is making a statement about the default namespace from the current node onwards, the other one is binding a prefix to a namespace, which may then be used elsewhere. Maybe we're not using that prefix by ourselves (haven't checked), but it could be used by others' customizations. It is _probably_ safe to remove those declarations, but as it's technically a different statement than the one we keep, I lean to keep it.

ahankinson:
@rettinghaus it was a question. Given that there is no statement on the impact or motivation of this PR other than "cleaning" I thought it was worthwhile to ask a clarification question. I seem to remember we've had problems in the past when mucking around with namespace declarations in the schema, and it broke things. 

So I thought it was worthwhile asking.

rettinghaus:
Ok, I've did some extensive testing. The only thing that changes is, that the 3938 extra copies of the namespace in the compiled `mei-all.odd` are removed (similar in all other compiled ODDs). The RNGs aren't changed a bit. 

@kepper If other customizations used this, it isn't a problem (on our side). The stylesheets handle everything fine (just tested with:
[mei-basic.xml.zip](https://github.com/music-encoding/music-encoding/files/12229467/mei-basic.xml.zip)).

riedde:
Approved at the MEI Developer Meeting 2024 (by @pe-ro) ;-).

pe-ro (approved):
pe-ro approves!

riedde (approved):
Discussed at the MEI DEv Workshop.

musicEnfanthen:
Looks good to me, will let @bwbohl have a final look ;)

bwbohl (approved):

PR #1285, opened 2023-07-27
schema: align att.arpeg.vis with other line-like elements

rettinghaus:
This PR makes `att.arpeg.vis` member of `att.lineRend.base` to replace the solitary `@line.form` and `@line.width` with the common `@lform` and `@lwidth` attributes.

PR #1291, opened 2023-07-28
customizations: clean up of line.form and line.width

musicEnfanthen:
As a follow-up on #1285, this PR fixes the last remnants of `line.form` and `line.width` in the specs. So it syncs the model of `att.arpeg.vis` in MEI Basic with that in MEI all. (It is out commented in Basic, but should be in sync nonetheless.) Additionally, it fixes the long-deprecated line attributes for `gliss` described in the CMN guidelines.


rettinghaus (approved):
Thanks for fixing this!

bwbohl (approved):

PR #1292, opened 2023-07-29
guidelines: update GitHub user name

rettinghaus:
This is a small update to the GitHub user name returned by the API.

### DIFF
--- a/source/contributors/curated.contributors.json
+++ b/source/contributors/curated.contributors.json
@@ -18,6 +18,15 @@
         "suppress":true
     },
     {
+        "login": "dinamix1",
+        "name": "dinamix",
+        "github": "https://github.com/dinamix1",
+        "viaf": "",
+        "orcid": "",
+        "avatar": "https://avatars.githubusercontent.com/u/9153739?v=4",
+        "suppress":true
+    },
+    {
         "login": "doerners",
         "name": "Sophia Dörner",
         "github": "https://github.com/doerners",
--- a/source/docs/01-introduction.xml
+++ b/source/docs/01-introduction.xml
@@ -37,7 +37,6 @@
                <!--TODO: To be filled automatically via GitHub API-->
                <name ref="https://api.github.com/users/bwbohl">Benjamin W. Bohl</name>
                <name ref="https://api.github.com/users/BruxDDay">BruxDDay</name>
-               <name ref="https://api.github.com/users/dinamix">Pablo Alvarez</name>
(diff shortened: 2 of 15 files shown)


musicEnfanthen:
Thanks @rettinghaus but it seems that the whole reference has been corrupted somehow in the past. There is no contribution by a user handle `@dinamix` (cf. https://github.com/music-encoding/music-encoding/commits?author=dinamix). The correct handle of the user would be `@dinamix1` instead (see https://github.com/music-encoding/music-encoding/commits?author=dinamix1). And the GH user handle `@dinamix1` uses `dinamix` as his GH name.

However, the contributions of `@dinamix1` refer exclusively to the example encodings (which were still combined with the main repo in 2015). I am not sure if it is correct then to list the user as contributor in all chapters of the guidelines?

@kepper How is your guideline here?

kepper:
It's hard to draw lines here that will need to apply to everything else as well. My first idea would be to have `@dinamix1` listed among the contributors list at the end of the (PDF) Guidelines, but not necessarily in each file. Both places aren't reflected to the web version anyway – which seems another, and perhaps even bigger issue…

musicEnfanthen:
Thanks @kepper for clarification. Yes, we will definitely have to enhance and improve the (semi)automatic collation of contributors in the future.

For now, @rettinghaus could you take care of removing the references to `@dinamix` from all the guideline chapters, and include it in https://github.com/music-encoding/music-encoding/blob/develop/source/contributors/curated.contributors.json instead with the correct handle `@dinamix1`? (There is a sorting issue with the entries at the beginning of that file, so cleaning it up a bit would also be great.)

rettinghaus:
done.

rettinghaus:
Perhaps it would make sense to have an overview over all commits, because I just noticed that @apacha is not mentioned (bd646f2).

musicEnfanthen:
@rettinghaus Yes, you are right. @apacha is missing (https://github.com/music-encoding/music-encoding/commits?author=apacha). In the "insights" tab on GH I can also see @pfefferniels (https://github.com/music-encoding/music-encoding/commits?author=pfefferniels) and @alastair (https://github.com/music-encoding/music-encoding/commits?author=alastair) who seem to not be mentioned in the chapter contributions. Would you like to add these in the chapters preambles?

musicEnfanthen (approved):

bwbohl (approved):

musicEnfanthen:
Never mind @rettinghaus , will prepare a follow-up.

PR #1294, opened 2023-07-31
guidelines: add missing contributors

musicEnfanthen:
As a follow-up of #1292, this PR adds the missing contributors to the guidelines chapters and the curated json list which is now sorted alphabetically.

In addition, Dependabot is removed from the curated list.


rettinghaus (approved):
Thank you!

bwbohl (approved):

PR #1295, opened 2023-07-31
schema: fix schematron phys desc ancestors

### LINKED ISSUE #1276: anyStart throws schematron warnings for required ancestor, e.g. `heraldry`, `watermark`, `locus`
The `heraldry`, `watermark`, `locus` have Schematron rules testing for an ancestor `physDesc`. While this is true in a complete `meiHead`, it does not make sense in the anyStart schema if a files starts below the `physDesc` (e.g., in the sample-encodings collection).
There might be other elements with parent or ancestor checks in Schematron rules.
There are alternative solutions:
* remove those Schematron checks in the anyStart customisation
* rework them in the customisation to not throw an error but a warning or info
* rework them generally to warning or info level.

Is there no other way for those three elements mentioned to ensure they are only used in `physDesc`?
What would be an easy solution for MEI 5?
Opinions? Especially from @music-encoding/ig-metadata?

bwbohl:
closes #1276

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -2929,15 +2929,6 @@
     <content>
       <rng:ref name="macro.struc-unstrucContent"/>
     </content>
-    <constraintSpec ident="check_watermark_inline" scheme="schematron">
-      <constraint>
-        <sch:rule
-          context="mei:watermark">
-          <sch:assert test="ancestor::mei:physDesc">The watermark element may only appear as a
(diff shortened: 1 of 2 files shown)


musicEnfanthen (approved):

pe-ro (approved):

riedde (approved):

musicEnfanthen:
Thanks everyone!

PR #1300, opened 2023-08-01
schema: reduce version number

### LINKED ISSUE #1174: build: reduce version number to 2 levels
As discussed in #https://github.com/music-encoding/music-encoding/issues/1125#issuecomment-1564512842, the version number should be reduced to only 2 levels. Semantic versioning with three levels does not really fit for schema development.

I'd suggest to do the transformation during the release of v5, and not before, since reducing `5.0.0-dev` to `5.0-dev` could potentially break already existing data etc.

So during the release, we would merge `5.0.0-dev` into `5.0`, and then make `6.0-dev` the new development version.

musicEnfanthen:
This PR is a first step towards v5.0 as it reduces the current version number to two levels (`5.0-dev`).

Fixes #1174

### DIFF
--- a/customizations/mei-all_anyStart.xml
+++ b/customizations/mei-all_anyStart.xml
@@ -161,8 +161,8 @@
             <attDef ident="meiversion" usage="rec" mode="change">
               <desc>Specifies a generic MEI version label.</desc>
               <valList type="closed">
-                <valItem ident="5.0.0-dev">
-                  <desc>Development version of MEI 5.0.0</desc>
+                <valItem ident="5.0-dev">
+                  <desc>Development version of MEI 5.0</desc>
                 </valItem>
               </valList>
             </attDef>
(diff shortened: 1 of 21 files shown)


musicEnfanthen:
Do we need to change the version numbers for the files in [tests](https://github.com/music-encoding/music-encoding/tree/release/v5.0/tests) as well?

rettinghaus:
I think the tests need to be updated accordingly. (There is still a version 3.0.0, so I wonder if we even do testing …)

bwbohl:
testing was stopped quite a while ago but it still sits there to keep nagging that we should do test ;-)

bwbohl:
updating them of, course would be good

bwbohl:
Why not leave away the `-dev` part everywhere on this branch?

musicEnfanthen:
@bwbohl I think it is better to go in two steps. First consolidate all the version numbers to 5.0-dev (which allows testing properly since it points to dev version), and then change to 5.0 for the release.

bwbohl (approved):

bwbohl (approved):

PR #1301, opened 2023-08-01
guidelines: fix syntax errors

### LINKED ISSUE #1299: guidelines: encoding errors
* `01-introduction.xml`
The last two paragraphs in `aboutVersion` are not allowed at this point.
* `05-mensural.xml`
In `mensuralSigns` attribute values are encoded within `att`, which is not allowed. Also the id (`mensuralSigns`) differ from the name (`Mensuration`)
* `08-lyricsperfdir.xml` and `09-textencoding.xml`
`respStmt` is missing an editor name. At least an empty `name` element should be added.

musicEnfanthen:
This PR fixes some syntax errors in the guidelines. There was also a closing bracket missing in the curated contributors list.

Fixes #1299


musicEnfanthen:
The PR does not fix the two paragraphs in chapter 1 mentioned in #1299 since I do not get why `p` should not be allowed there.

source/docs/05-mensural.xml
```diff
@@ -269,7 +269,7 @@
...
-                  <p><hi rend="bold">Sesquialtera</hi> is frequently used to change the mensuration. The effect of the sesquialtera on the mensuration can be encoded by using the <att>tempus</att> and <att>prolatio</att> attributes of <gi scheme="MEI">mensur</gi> (for example, when changing the tempus to perfect, the effect can be encoded in <att>tempus="3"</att>). The actual sesquialtera can be encoded using <att>num="3"</att>, <att>numbase="2"</att>, and <att>level</att> to define the note level the sesquialtera is applied to (<abbr>e.g.</abbr>, <att>level="semibrevis"</att>).</p>
+                  <p><hi rend="bold">Sesquialtera</hi> is frequently used to change the mensuration. The effect of the sesquialtera on the mensuration can be encoded by using the <att>tempus</att> and <att>prolatio</att> attributes of <gi scheme="MEI">mensur</gi> (for example, when changing the tempus to perfect, the effect can be encoded in <att>tempus</att>="3"). The actual sesquialtera can be encoded using <att>num</att>="3", <att>numbase</att>="2", and <att>level</att> to define the note level the sesquialtera is applied to (<abbr>e.g.</abbr>, <att>level</att>="semibrevis").</p>
```
rettinghaus, line 272: Parallel cases use `<hi rend="italic">` for the complete attribute+value. Although I don't think that this is the best solution, we should keep consistency here.

rettinghaus (changes requested):
Thanks! One remark.

source/docs/05-mensural.xml
musicEnfanthen: I see. Can you point me to some instances of this usage? I found the approach used here applied in the CMN chapter and double checked with the rendering in the PDF.

`<att>tempus</att>="3"` renders as `@tempus="3"`
while `<hi rend="italic">tempus="3"` renders as _tempus="3"_

rettinghaus: [Here](https://github.com/music-encoding/music-encoding/blob/3efe83de4dba12c96caec61b23c973f361b396ec/source/docs/05-mensural.xml#L192C1345-L192C1345) and [here](https://github.com/music-encoding/music-encoding/blob/3efe83de4dba12c96caec61b23c973f361b396ec/source/docs/05-mensural.xml#L202C1-L202C1). But, as you pointed out, in chapters 4 and 9 it's your proposed way, so maybe the others should be changed …

rettinghaus: I forgot line 272: https://github.com/music-encoding/music-encoding/blob/3efe83de4dba12c96caec61b23c973f361b396ec/source/docs/05-mensural.xml#L272

musicEnfanthen: Thanks for the pointers. Adjusted it in chapter 5.

rettinghaus (approved):
Great, thank you!

rettinghaus (approved):

bwbohl (approved):

bwbohl:
> The PR does not fix the two paragraphs in chapter 1 mentioned in #1299 since I do not get why `p` should not be allowed there.

see comments in #1299

musicEnfanthen:
Fixed as described in #1299

bwbohl (approved):

PR #1305, opened 2023-08-01
guidelines: correction of correction

rettinghaus:
Something happened during some conversion. This `div` was completely messed up. I've added missing `att` elements and used correct `ptr`. One sentence was removed, because the target in there completely changed.


source/docs/03-metadata.xml
```diff
@@ -585,15 +585,15 @@
...
+                        <item>Indicates the extent of normalization or regularization of the original source carried out in converting it to electronic form. <att>method</att> indicates the method employed to mark corrections and normalizations. Was the text normalized, for example by regularizing any non-standard enharmonic spellings, etc.? If so, were normalizations performed silently or are they marked using the tags described in chapter <ptr target="#editTrans"/>? What authority was used for the regularization?</item>
```
musicEnfanthen: I am not in favour of removing the sentence. The section numbering refers to MEI v3 (https://music-encoding.org/guidelines/v3/content/shared.html#sharedNamesNumbersDates), and the content was moved to chapter 9 in 9.2.8 and following sections (cf. https://music-encoding.org/guidelines/dev/content/textencoding.html#sharedNames). Maybe adjust like that:

```suggestion
                        <item>Indicates the extent of normalization or regularization of the original source carried out in converting it to electronic form. <att>method</att> indicates the method employed to mark corrections and normalizations. Was the text normalized, for example by regularizing any non-standard enharmonic spellings, etc.? If so, were normalizations performed silently or are they marked using the tags described in chapter <ptr target="#editTrans"/>? What authority was used for the regularization? Also, what principles were used when normalizing numbers to provide the standard values for the value attribute described in sections <ptr target="#sharedNames"/>–<ptr target="sharedAddresses"/> and what format is used for them?</item>
```

musicEnfanthen:
Thanks for catching this! One remark above

rettinghaus:
There are none.

musicEnfanthen:
Sorry, forgot to submit

musicEnfanthen (approved):

musicEnfanthen:
Thanks

PR #1307, opened 2023-08-01
guidelines: add attribute changes to "about version" section

musicEnfanthen:
This PR includes some more statements on major attribute changes for MEI 5 in the "about version" section of chapter 1 of the guidelines.

### DIFF
--- a/source/docs/01-introduction.xml
+++ b/source/docs/01-introduction.xml
@@ -236,15 +236,15 @@
                      <gi scheme="MEI">pgHead</gi> and <gi scheme="MEI">pgFoot</gi> respectively. 
                      Most other changes affect more specific aspects in the model of MEI, usually 
                      expressed in attributes. These include the refinement of the encoding of key 
-                     signatures, with @key.sig moved to <att>keysig</att>, @keysig.show moved to 
-                     <att>keysig.visible</att>, and @keysig.showchange and @sig.showchange moved to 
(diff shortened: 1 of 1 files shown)


rettinghaus (approved):

source/docs/01-introduction.xml
```diff
@@ -233,17 +233,38 @@
...
-                     reference classes not available anymore. If you have advanced local customizations
-                     based on MEI v4 or older releases, please check that the rules provided still work as
-                     expected under v5. A very helpful addition for this task may be the validation for MEI
-                     customizations, which is now available and used for all customizations officially
-                     provided by MEI.</p>
+                     <gi scheme="MEI">pgHead</gi> and <gi scheme="MEI">pgFoot</gi> respectively. 
+                     Most other changes affect more specific aspects in the model of MEI, usually 
+                     expressed in attributes. These include the refinement of the encoding of key 
+                     signatures, with @key.sig moved to <att>keysig</att>, @keysig.show moved to 
+                     <att>keysig.visible</att>, and @keysig.showchange and @sig.showchange moved to
```
bwbohl: There is probably an XSLT-reason why the deprecated attributes are not wrapped in `<att>…</att>`?

musicEnfanthen: Yes, these are the removed attributes. I was wrong about these being linked to the current guidelines (which wouldn't work), but this only happens for the elements. You are right, I will add these att tags.

bwbohl (approved):

PR #1309, opened 2023-08-13
customizations: fix missing `@n` in verse in MEI Basic

lpugin:
One or more `verse` is allowed, but no `@n`. It is semantically problematic and since this is a critical bug in MEI Basic I am making a PR to the release candidate.


rettinghaus (approved):
I'm not totally sure we need label in Basic, but apart from that it looks good. Thank you.

bwbohl (approved):

PR #1310, opened 2023-08-15
schema: add stem.split to att.stems

### LINKED ISSUE #721: split stem notation
How to encode split stem notation? 

split stem

For "easy" cases like altered unions in chords you can avoid split stem notation. [Dorico](https://steinberg.help/dorico/v2/en/dorico/topics/notation_reference/notation_reference_stems_altered_unisons_split_stems_c.html) lets you choose the display of these cases. 

But there are more complicated examples like in Bartók's Violin Sonata (with a triplet unison)
split-stem

Could the proposed `@stem.sameas` in #586 be used for this somehow or do we need even more control like relative positions of the noteheads?

rettinghaus:
It seems we reached consensus on possible encodings for split stem notation. 
This PR offers an implementation with `@stem.split` as part of `att.stems`. 
Alternatively it could be a direct child of `note` to simplify the schematron rules.

closes #721

- [x] modify schema
- [ ] add example to the guidelines

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -3258,9 +3258,6 @@
           <valItem ident="right">
             <desc xml:lang="en">The stem branches off to the right.</desc>
           </valItem>
-          <valItem ident="straight">
-            <desc xml:lang="en">The stem is a straight line.</desc>
-          </valItem>
         </valList>
       </attDef>
       <attDef ident="stem.visible" usage="opt">
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -3237,6 +3237,32 @@
           </constraint>
         </constraintSpec>
       </attDef>
+      <attDef ident="stem.split" usage="opt">
+        <desc xml:lang="en">Determines if and how a stem should be split within a chord.</desc>
+        <constraintSpec ident="check_stem.split" scheme="schematron">
+          <constraint>
+            <sch:rule context="@stem.split">
+              <sch:assert role="error" test="not(parent::node()/local-name() = 'chord')">
+                @stem.split must not be used on chords.
+              </sch:assert>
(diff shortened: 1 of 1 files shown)


riedde (dismissed):
Reviewed and tested at the MEI Developer Workshop 2024

pe-ro (dismissed):
Approved

source/modules/MEI.shared.xml
```diff
@@ -3237,6 +3237,29 @@
           </constraint>
         </constraintSpec>
       </attDef>
+      <attDef ident="stem.split" usage="opt">
+        <desc xml:lang="en">Determines if and how a stem should be split within a chord.</desc>
+        <constraintSpec ident="check_stem.split" scheme="schematron">
+          <constraint>
+            <sch:rule context="@stem.split">
+              <sch:assert role="error" test="not(parent::node()/local-name() = 'chord')">
+                @stem.split must not be used on chords.
```
```suggestion
                @stem.split must not be used directly on chord elements.
```

musicEnfanthen (changes requested):
Can you please add a short example and explanation in the guidelines for this use case.

bwbohl:
as mentioned above please add a corresponding example to the guidelines, thx.

bwbohl:
- [x] modify schema
- [ ] add example to the guidelines

musicEnfanthen:
Still needs work, moving to 6

martha-thomae:
Hi @rettinghaus. I think this only needs guidelines examples, and it will be done. Do you think you can include these examples by the next ODD Meeting (March 27th)? That would be great as we can merge it right away (either during or before the meeting). Thank you!

martha-thomae:
@rettinghaus, could you add the guidelines' examples for this? We have a milestone now open for this kind of fixes (guidelines), so if you can add these examples in, we can merge this. The next milestone will be to move to PureODD, so if these changes can be done soon, that would be awesome! Thank you!

Issue #595, opened 2019-05-30
MEI Basic: allow @key.mode on <scoreDef>


kepper:
`<scoreDef>` and `<staffDef>` should allow @key.mode

musicEnfanthen:
Pinging @lpugin

rettinghaus:
There is a contradiction in the spec. We explicitly include `att.keySigDefault.anl` (including `@key.mode`) and then we delete it again.

https://github.com/music-encoding/music-encoding/blob/dffa5e038aa1f8a752ce789ec0b9c2496a7a9c73/customizations/mei-basic.xml#L50

https://github.com/music-encoding/music-encoding/blob/dffa5e038aa1f8a752ce789ec0b9c2496a7a9c73/customizations/mei-basic.xml#L169

https://github.com/music-encoding/music-encoding/blob/dffa5e038aa1f8a752ce789ec0b9c2496a7a9c73/customizations/mei-basic.xml#L177

ahankinson:
May I question why we need `@key.mode` in basic? It seems to me that marking something "major" or "minor" has very little to do with the rendering or playback of the notation, and is primarily an analytical observation.

I'm not necessarily opposed to it, but I don't really see the justification for it to be included in the original issue.

rettinghaus:
I agree with @ahankinson here. Just because it's in MusicXML, it doesn't have to be in MEI basic. 

Still, the customization should be cleaned up.

bwbohl:
also agree with @ahankinson. @rettinghaus could you please check the current status of the customization considering your mentioned cleanup?

rettinghaus:
@bwbohl Seems to be fine now. I guess this can be closed then, @kepper ?

kepper:
Seems we prefer to move on without `@key.mode` ;-)

PR #1313, opened 2023-08-23
Fix MEI Basic

rettinghaus:
This PR brings some fixes to the Basic customization:

* fix availability of `@color`
* fix availability of `@enclose`
* fix availability of `@loc`, `@oloc`, and `@ploc` for `<rest>`
* comment out previously deleted attribute classes (clean-up including leftovers from #595)

### DIFF
--- a/customizations/mei-basic.xml
+++ b/customizations/mei-basic.xml
@@ -637,7 +637,7 @@
                         <memberOf key="att.cue"/>
                         <memberOf key="att.duration.log"/><!-- changed from att.restduration.log -->
                         <memberOf key="att.event"/>
-                        <memberOf key="att.rest.log.cmn"/>
+                        <!--<memberOf key="att.rest.log.cmn"/>-->
                     </classes>
                 </classSpec>
 
@@ -696,7 +696,7 @@
                         <memberOf key="att.extSym"/>
                         <memberOf key="att.stems"/>
                         <!--<memberOf key="att.typography"/>-->
-                        <memberOf key="att.visibility"/>
+                        <!--<memberOf key="att.visibility"/>-->
                         <memberOf key="att.visualOffset.ho"/>
                         <memberOf key="att.visualOffset.to"/>
                         <!--<memberOf key="att.xy"/>-->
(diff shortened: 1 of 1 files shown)


ahankinson:
Can you please describe your motivation for adding these? I think the `*loc` attributes were deliberately left out, so some indication of why you're adding them back in. Is this to fix something that's broken? Or is it to add something new to MEI Basic?

rettinghaus:
@ahankinson The `*loc` attributes are explicitly commented out on all elements in the customization, except for `rest` (see `att.rest.vis`), which makes sense, because otherwise it would not be possible to encode the vertical placement of rests.

ahankinson:
OK, so it's adding something new to MEI Basic. Do we need all of them, or is just `@loc` enough? Remember the whole point of MEI Basic is to reduce the number of options available. 

I guess that's the same with the others? I know `@enclose` is the subject of some debate, since it could be seen as "Editorial" -- I think @lpugin has some ideas about this.

rettinghaus:
It's not adding something new, it just fixes contradictions that have been there from the first draft on, similar to #1244. 

Without `@enclose` it is not possible to encode cautionary accidentals with parentheses, which MEI Basic should absolutely allow. 
In many places `att.color` and `att.enclosingChars` were added by @lpugin in #1242.

lpugin:
In my opinion these are not necessarily critical issues, even though I agree that making sure the use of `att.color` is uniform makes sense.

Rest position: there is already `rest@vo`, so there is already a way to encode the vertical position of a rest.

Cautionary accidentals: there is `accid@func`, so there too the need for `@enclose` is debatable.

bwbohl:
As I see potential for discussion: can we push this to `5.x`?

rettinghaus:
@bwbohl I've removed the changes on rest positioning and `@enclose` for a different PR. So this one contains only fixes all agree on.

customizations/mei-basic.xml
```diff
@@ -59,6 +63,9 @@
             <change n="3" from="2023-05-24" to="2023-06-12" who="#LP">
                 <desc>Revisions at the MEI Developers Conference, Charlottesville.</desc>
             </change>
+            <change n="4" when="2023-08-23" who="#KR">
+                <desc>Minor fixes.</desc>
```
lpugin: This is not a very useful description in my opinion.
```suggestion
                <desc>Fixing color attribute and cleanup attribute class references.</desc>
```

lpugin (approved):

ahankinson (approved):

rettinghaus:
Thanks.

bwbohl:
THX all!

PR #1318, opened 2023-08-28
schema: align dedicatee element with composer and similar

rettinghaus:
In reaction to https://github.com/music-encoding/music-encoding/issues/555#issuecomment-1694513365 this remodels `dedicatee` to be in line with `composer`, `funder` and other similar elements, mainly to make it available as child of `work`.

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -5201,9 +5201,10 @@
     <classes>
       <memberOf key="att.common"/>
       <memberOf key="att.bibl"/>
+      <memberOf key="att.evidence"/>
       <memberOf key="att.facsimile"/>
       <memberOf key="att.lang"/>
-      <memberOf key="model.textPhraseLike.limited"/>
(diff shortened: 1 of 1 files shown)


lpugin:
The issue has a `MEI 6` milestone, so I think the PR should be made to develop?

ahankinson:
It looks ok to me, but yes: It should go to MEI 6.

rettinghaus:
Changing into draft until MEI 5 is out.

pe-ro (approved):
Thanks for the PR, @rettinghaus.

pe-ro:
The removal of `<dedicatee>` from [model.textPhraseLike.limited](https://music-encoding.org/guidelines/dev/model-classes/model.textPhraseLike.limited.html) will also need to be addressed in the ~~MEI 4 -> MEI 5~~ MEI 5 -> MEI 6 [EDITED by @bwbohl ] XSLT.

rettinghaus:
Basically addressed in #1596.

PR #1319, opened 2023-08-28
customizations: implement customization specific version numbers with suffix

### LINKED ISSUE #1174: build: reduce version number to 2 levels

### LINKED ISSUE #1290: implement customization-specific version numbers with suffix
e.g., mei-basic in the dev version implements `5.0.0-dev+basic` at the moment. This is a good idea, because it definitely tells which customization was used when creating a file, and there might be cases when a customization allows things that would not be valid against mei-all.

bwbohl:
closes #1174

### DIFF
--- a/customizations/mei-CMN.xml
+++ b/customizations/mei-CMN.xml
@@ -109,6 +109,9 @@
           <attList>
             <attDef ident="meiversion" usage="rec" mode="change">
               <valList type="closed">
+                <valItem ident="5.0-dev">
+                  <desc xml:lang="en">Development version of MEI 5.0.</desc>
+                </valItem>
                 <valItem ident="5.0-dev+CMN">
                   <desc>Development version of MEI 5.0+CMN.</desc>
                 </valItem>
--- a/customizations/mei-CMN.xml
+++ b/customizations/mei-CMN.xml
@@ -108,6 +108,7 @@
         <classSpec ident="att.meiVersion" module="MEI.shared" type="atts" mode="change">
           <attList>
             <attDef ident="meiversion" usage="rec" mode="change">
+              <defaultVal>5.0-dev+CMN</defaultVal>
               <valList type="closed" mode="change">
                 <valItem ident="5.0-dev+anyStart" mode="delete"/>
                 <valItem ident="5.0-dev+basic" mode="delete"/>
--- a/customizations/mei-CMN.xml
+++ b/customizations/mei-CMN.xml
@@ -108,12 +108,12 @@
         <classSpec ident="att.meiVersion" module="MEI.shared" type="atts" mode="change">
           <attList>
             <attDef ident="meiversion" usage="rec" mode="change">
-              <defaultVal>5.0-dev+CMN</defaultVal>
+              <defaultVal>5.0+CMN</defaultVal>
               <valList type="closed" mode="change">
-                <valItem ident="5.0-dev+anyStart" mode="delete"/>
-                <valItem ident="5.0-dev+basic" mode="delete"/>
-                <valItem ident="5.0-dev+Mensural" mode="delete"/>
-                <valItem ident="5.0-dev+Neumes" mode="delete"/>
+                <valItem ident="5.0+anyStart" mode="delete"/>
+                <valItem ident="5.0+basic" mode="delete"/>
+                <valItem ident="5.0+Mensural" mode="delete"/>
+                <valItem ident="5.0+Neumes" mode="delete"/>
             </valList>
             </attDef>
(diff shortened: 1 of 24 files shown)


source/modules/MEI.shared.xml
```diff
@@ -1894,7 +1894,22 @@
         <defaultVal>5.0-dev</defaultVal>
         <valList type="closed">
           <valItem ident="5.0-dev">
-            <desc xml:lang="en">Development version of MEI 5.0</desc>
+            <desc xml:lang="en">Development version of MEI 5.0.</desc>
+          </valItem>
+          <valItem ident="5.0-dev+anyStart">
```
rettinghaus: This doesn't work, because `anyStart` isn't compatible with any other schema. See following example.

```xml
<editorialDecl xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0-dev+anyStart">
  <p>A test to make a point.</p>
</editorialDecl>
```

This validates with `anyStart` schema, but not with `all`.

rettinghaus (changes requested):

customizations/mei-all_anyStart.xml
```diff
@@ -159,10 +159,9 @@
         <classSpec ident="att.meiVersion" module="MEI.shared" type="atts" mode="change">
           <attList>
             <attDef ident="meiversion" usage="rec" mode="change">
-              <desc>Specifies a generic MEI version label.</desc>
               <valList type="closed">
-                <valItem ident="5.0-dev">
-                  <desc>Development version of MEI 5.0</desc>
```
lpugin, line 165: Didn't we agree that "5.0" should also be accepted as a version value for any customizations that only reduce the elements available?

source/modules/MEI.shared.xml
bwbohl: good point

bwbohl: Nevertheless, it would work for all files that have `meiversion="5.0+anyStart"` and start with one of the allowed start elements of mei-all…

customizations/mei-all_anyStart.xml
bwbohl: Ah yes, missed that (cf. https://github.com/music-encoding/music-encoding/issues/1290#issuecomment-1655795460)

source/modules/MEI.shared.xml
bwbohl: What do you think, should anyStart then also disallow any other value for `@meiversion`?

customizations/mei-CMN.xml
```diff
@@ -105,6 +105,21 @@
           </content>
         </moduleRef>
 
+        <classSpec ident="att.meiVersion" module="MEI.shared" type="atts" mode="change">
+          <attList>
+            <attDef ident="meiversion" usage="rec" mode="change">
+              <valList type="closed">
+                <valItem ident="5.0-dev">
```
ahankinson: As this is on the `release/5.0` branch, it's probably safe to just change this to `5.0` directly, and remove the `-dev`.

Would save an additional PR, and would give us a bit more time to find any weird issues that might arise when changing this.

source/modules/MEI.shared.xml
bwbohl: I think 5.0+anyStart can safely be available in mei-all, because  `@meiversion` is only allowed on the start-element in mei-all and thus would be valid against both schemata.

bwbohl: > ```xml
><editorialDecl xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0-dev+anyStart">
>  <p>A test to make a point.</p>
></editorialDecl>```
>This validates with anyStart schema, but not with all.

This is absolutely okay because the error reported is not solely due to `@meiversion` but because `editorialDecl` is not a valid start element in mei-all. Disallowing `meiversion="5.0"` in anyStart would make more sense, or adding a schematron rule that reports if the root element of an anyStart-document is not one that is also allowed in `5.0` and `meiversion` is not anyStart.

rettinghaus (approved):

musicEnfanthen (approved):

PR #1320, opened 2023-08-28
Fix: #1314 mei basic unclear status of attribute classes

### LINKED ISSUE #1314: MEI Basic: unclear status of attribute classes
The following attribute classes are explicitly added to some elements in the MEI Basic customization, although they are deleted from the schema:

- att.bibl
- att.metadataPointing
- att.measure.log 
- att.measure.vis
- att.responsibility

It needs to be decided wether these should be actually available or not. 
This is a follow up to #1313.

bwbohl:
* comment out references to deleted attribute classes as mentioned in #1314

### DIFF
--- a/customizations/mei-basic.xml
+++ b/customizations/mei-basic.xml
@@ -1004,7 +1004,7 @@
                         MEI-conformant text.</desc>
                     <classes mode="replace">
                         <memberOf key="att.basic"/>
-                        <memberOf key="att.bibl"/>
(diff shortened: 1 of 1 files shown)


rettinghaus:
@bwbohl I think it points to the wrong target branch?

bwbohl:
> @bwbohl I think it points to the wrong target branch?

fixed

rettinghaus (approved):

PR #1332, opened 2023-09-01
gh-actions: fix release workflow

musicEnfanthen:
This PR tries to fix the failing release workflow. It imrpoves and simplifies the exit and if conditions, and adds some validation steps that will make it easier to debug failing workflows in the future.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -72,12 +72,12 @@ jobs:
 
       ### SET TARGET_DIR environment variable for dev version
       - name: Set TARGET_DIR for dev
-        if: ${{ (github.repository_owner == env.REPO_OWNER) && (github.ref_type == 'branch') && (github.event.base_ref == 'develop') }}
+        if: github.repository_owner == env.REPO_OWNER && github.ref == 'refs/heads/develop'
         run: echo "TARGET_DIR=dev" >> $GITHUB_ENV
 
       ### SET TARGET_DIR environment variable for tagged release
       - name: Set TARGET_DIR for tagged release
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -76,17 +81,26 @@ jobs:
...
       - name: Set TARGET_DIR for tagged release
-        if: ${{ (github.repository_owner == env.REPO_OWNER) && (github.ref_type == 'tag')  && (github.event.base_ref == 'stable') }}
+        if: github.repository_owner == env.REPO_OWNER  && startsWith(github.ref, 'refs/tags/')
         run: echo "TARGET_DIR=${{ github.ref_name }}" >> $GITHUB_ENV
 
+      ### VALIDATE TARGET NAME
+      - name: Validate TARGET_DIR
+        run: echo "TARGET_DIR=$TARGET_DIR"
+
+      ### Exit if ref is not develop branch or release tag
```
ahankinson: This could probably be moved up the chain, since it doesn't require anything else to run (I don't think?). There's no point getting this far if we're just going to exit...

ahankinson:
You're a star, @musicEnfanthen ! GH Actions are not fun to write.

.github/workflows/deploy.yml
musicEnfanthen: We wanted to run the build and artifact upload before exiting (on each branch that triggers the workflow). But you are right, we could probably exit before setting the target_dir variable.

musicEnfanthen:
@ahankinson If you are OK with the latest changes, you can merge this PR. It goes to the release branch.

ahankinson:
Oh I’m fine with them; I just thought you or Benni should have the honors. Consider this message to be permission to merge so that you guys get your names in the commit history.

PR #1342, opened 2023-09-05
gh-actions: Remove zip step

bwbohl:
in order to fix the workflow from failing due to a permission error of `run: zip …`


musicEnfanthen (approved):
Thanks

PR #1352, opened 2023-09-25
schema: revise description of `<catRel>`

pe-ro:


source/modules/MEI.header.xml
```diff
@@ -681,7 +681,8 @@
   </elementSpec>
   <elementSpec ident="catRel" module="MEI.header">
     <gloss versionDate="2022-05-18" xml:lang="en">category relationship</gloss>
-    <desc xml:lang="en">Contains the name of a related category.</desc>
+    <desc xml:lang="en">(category relationship) Contains the name; i.e., label, of a related
```
ahankinson: Doesn't this repeat what is in the `gloss` element above?

pe-ro: You're right, it does.  Another cut-and-paste error.  The parenthetical part of `<desc>` should be removed.

ahankinson:
```suggestion
    <desc xml:lang="en">Contains the name; i.e., label, of a related
```

bwbohl:
```suggestion
    <desc xml:lang="en">Contains the name, i.e., label, of a related
```

musicEnfanthen (approved):
Thank you.

bwbohl (approved):

PR #1353, opened 2023-09-25
schema: allow `<p>` in `<sourceDesc>`

pe-ro:
Allow `<p>` in `<sourceDesc>` to accommodate textual descriptions, rather than bibliographic references or descriptions.


bwbohl (dismissed):

bwbohl:
This is duly needed for making `sourceDesc` usable more broadly by, e.g., critical editions with specific guidelines for writing source descriptions, both in born-digital and retro-digitized contexts.

bwbohl:
This should also be reflected in the guidelines and sample-encoding, though. Thus converting to draft.

pe-ro:
Not very good at GitHub, so can someone else take this up please?


I've attached 

- a revision of 03-metadata.xml, which updates the metadata chapter of the guidelines re: sourceDesc/p (div beginning at line 861; and
- an example file that will need to be referenced in the revision above.

The revision also moves the "headerAssociatingMetadataAndData" div to "headerusecases".

[03-metadata.txt](https://github.com/music-encoding/music-encoding/files/13980498/03-metadata.txt)
[sourceDesc-paragraph_example.txt](https://github.com/music-encoding/music-encoding/files/13980504/sourceDesc-paragraph_example.txt)

pe-ro:
Perhaps this should be part of #1413 ?

source/modules/MEI.header.xml
```diff
@@ -2550,9 +2550,12 @@
...
       </rng:zeroOrMore>
-      <rng:oneOrMore>
-        <rng:ref name="source"/>
-      </rng:oneOrMore>
+      <choice>
+        <ref name="p"/>
+        <oneOrMore>
+          <ref name="source"/>
+        </oneOrMore>
+      </choice>
```
musicEnfanthen:
```suggestion
      <rng:choice>
        <rng:ref name="p"/>
        <rng:oneOrMore>
          <rng:ref name="source"/>
        </rng:oneOrMore>
      </rng:choice>
```

musicEnfanthen:
OK, there is now additional documentation for the guidelines and an example. Seems to be ready for review?

(Above I added suggestion since rng namespace seems to be missing?)

source/modules/MEI.header.xml
riedde: @musicEnfanthen I can't accept this suggestion, because I do not have writing access here. But you are right. `<choice>` is taken from the rng-namespace, so this has to be added.

bwbohl (approved):

riedde (dismissed):
Thank you @pe-ro for this improvement. Just a small question that shouldn't be critical: Was it intended not to use the term  "bibliographic description" (as the element definition of source does)?

source/docs/03-metadata.xml
```diff
@@ -525,12 +543,7 @@
                            <specDesc key="biblStruct"/>
                         </specList>
                      </p>
-                     <div xml:id="headerAssociatingMetadataAndData" type="div5">
-                        <head>Associating Metadata and Data</head>
-                        <p>In the MEI header, the <att>data</att> attribute may be used to associate metadata with related notational elements.</p>
-                        <p>Similarly, in the body of the MEI document, the <att>decls</att> attribute may be used to associate parts of the encoded text with related metadata.</p>
-                        <p>The most useful associations of this type are between the bibliographic description of a source and the material taken from it.</p>
-                     </div>
+                     <p/>
```
musicEnfanthen: @pe-ro Was the deletion of this div about associating metadata and data intended?

musicEnfanthen: Sorry, just seeing that the change was commited by @riedde . @riedde , did you delete this section by intention?

bwbohl: Aside the deletion here, I think it would be better suited as info at a higher level because it’s not about `sourceDesc` exclusively but true for the whole header.

bwbohl (changes requested):
Lost section "Associating Metadata and Data"

source/docs/03-metadata.xml
musicEnfanthen: @riedde Could you have a look at this please?

riedde: I would suggest to keep the div but to delete the last paragraph. Because I'm not sure if this is still the most useful association.

musicEnfanthen: ODD meeting 11/2024: Proposal to move the section to 3.3 (Metadata concepts).

musicEnfanthen:
ODD meeting 11/2024: Moving to v6 (and dev workshop 2025) since there seems to be need for some more discussion.

Proposal to move section "Associating Metadata and Data" to the end of section 3.3 (still for v5.1).

source/docs/03-metadata.xml
riedde: Done in #1567

riedde:
Closes #1496

riedde:
> Lost section "Associating Metadata and Data"

@bwbohl should be resolved in #1567 but the PR was merged into develop and not into here.

source/docs/03-metadata.xml
```suggestion
                     <div xml:id="headerAssociatingMetadataAndData" type="div5">
                     <head>Associating Metadata and Data</head>
                        <p>In the MEI header, the <att>data</att> attribute may be used to associate metadata with related notational elements.</p>
                        <p>Similarly, in the body of the MEI document, the <att>decls</att> attribute may be used to associate parts of the encoded text with related metadata.</p>
                        <p>The most useful associations of this type are between the bibliographic description of a source and the material taken from it.</p>
                     </div>
```

riedde: @bwbohl this will fix it, doesn't it?

musicEnfanthen: @riedde No, we moved the section "Associating Metadata ...." to 3.3. Probably the deletion of the empty paragraph should be fine here.

riedde: Okay, then the comment by @bwbohl is resolved and my suggestion can be discarded.

bwbohl (changes requested):
@riedde can you take care of implementing the changes to the guidelines that @pe-ro appended

riedde (approved):
Even If I did some small changes (only updating branches and fixing code), I reviewed and built it. It works fine.

bwbohl (approved):

PR #1354, opened 2023-09-25
schema: add "parallel" to title types

pe-ro:


source/modules/MEI.shared.xml
```diff
@@ -8374,6 +8374,9 @@
           <valItem ident="alternative">
             <desc xml:lang="en">Alternate title by which the item is also known.</desc>
           </valItem>
+          <valItem ident="parallel">
+            <desc xml:lang="en">title in another language or script.</desc>
```
ahankinson:
```suggestion
            <desc xml:lang="en">Title in another language or script.</desc>
```

source/modules/MEI.shared.xml
```diff
@@ -8387,20 +8390,22 @@
...
+    for convenience in analysing titles and processing them according to their type; where
+    such specialized processing is not necessary, there is no need for such analysis, and
+    the entire title, including subtitles and any parallel titles, may be enclosed within
+    a single <gi scheme="MEI">title</gi> element. Title parts may be encoded in <gi
+      scheme="MEI">titlePart</gi> sub-elements. The name of the list from which a
+    controlled value is taken may be recorded using the <att>auth</att> attribute. The
+    number of initial characters (such as those constituting an article or preposition)
+    that should not be used for sorting a title or name may be indicated in the
+      <att>nonfiling</att> attribute.</p>
+</remarks>
```
ahankinson: I think the formatting and indentation is off here.

pe-ro: It looks ok to me, but I can't figure out how to see the change in the context of the entire file.  I also don't know how to revise the PR.

ahankinson: This should work to see it? https://github.com/music-encoding/music-encoding/blob/949464601e80914856d3467a6132e061b7e29c51/source/modules/MEI.shared.xml#L8392-L8408

ahankinson: To change it you should have a local branch, `develop-title-type`, on your computer. If so, you can change to that branch, fix it, then commit and push. The updates should appear here automatically.

pe-ro: I think I fixed the indentation.  Have a look-see now?

ahankinson: Indentation looks good to me. (Not sure if others here will want the `</p>` on a new line or not, but I'm not bothered.)

There's also an outstanding suggestion for using a capital "T" to start -- you can just click "commit suggestion" above if you agree with it.

bwbohl:
I fixed the indentation issue

bwbohl (approved):

bwbohl:
@pe-ro, can you maybe give an example for a parallel title, either here or ideally in the guidelines?

pe-ro:
I would like to withdraw this PR.  Adding "parallel" to `title/@type` when we already allow the value "translated" is overkill.  Sorry for the noise/thrashing.

PR #1357, opened 2023-09-26
Relax content order of `mei:work`

pe-ro:
[@bwbohl EDIT] This PR removes the predefined order of the contents in `mei:work`.


bwbohl:
As this is part of a bigger change I changed the PR base to the [develop-relax-FRBR](https://github.com/music-encoding/music-encoding/tree/develop-relax-FRBR) branch

PR #1358, opened 2023-09-26
schema: relax model of work, expression, manifestation, and item

pe-ro:
[@bwbohl EDIT] This PR relaxes the content model of the mentioned FRBR elements by removing the predefined order of their child elements. Doing so it accommodates broader use in different editorial contexts / following different editorial guidelines.

Issue #1350, opened 2023-09-25
Relax current models of work, expression, manifestation, and item


pe-ro:
The current models of `<work>`, `<expression>`, `<manifestation>`, and `<item>` require their content to be ordered.  This is unduly restrictive, especially when transcoding metadata from other formats, but was considered necessary to control the number of occurrences of some child elements.

However, the RNG `<interleave>` element can be used to control for number of some child elements *without* having to also specify a particular order of the children.

For example, the new content model for `<work>` can be --
```xml
<elementSpec ident="work" module="MEI.header">
  <desc xml:lang="en">Provides a detailed description of a work — a distinct intellectual
    or artistic creation — specifically its history, language use, and high-level musical
    attributes ( <abbr>e.g.</abbr> , key, tempo, meter, medium of performance, and
    intended duration).</desc>
  <classes>
    <memberOf key="att.common"/>
    <memberOf key="att.authorized"/>
    <memberOf key="att.bibl"/>
    <memberOf key="att.dataPointing"/>
    <memberOf key="att.pointing"/>
    <memberOf key="att.targetEval"/>
    <memberOf key="model.workLike"/>
  </classes>
  <content>
    <rng:zeroOrMore>
      <rng:ref name="model.headLike"/>
    </rng:zeroOrMore>
    <interleave>
      <rng:zeroOrMore>
        <rng:ref name="model.identifierLike"/>
      </rng:zeroOrMore>
      <rng:oneOrMore>
        <rng:ref name="model.titleLike"/>
      </rng:oneOrMore>
      <rng:zeroOrMore>
        <rng:ref name="model.respLikePart"/>
      </rng:zeroOrMore>
      <rng:zeroOrMore>
        <rng:ref name="model.workIdent"/>
      </rng:zeroOrMore>
      <rng:zeroOrMore>
        <rng:ref name="otherChar"/>
      </rng:zeroOrMore>
      <rng:optional>
        <rng:ref name="creation"/>
      </rng:optional>
      <rng:optional>
        <rng:ref name="history"/>
      </rng:optional>
      <rng:optional>
        <rng:ref name="langUsage"/>
      </rng:optional>
      <rng:optional>
        <rng:ref name="perfMedium"/>
      </rng:optional>
      <rng:optional>
        <rng:ref name="perfDuration"/>
      </rng:optional>
      <rng:optional>
        <rng:ref name="audience"/>
      </rng:optional>
      <rng:optional>
        <rng:ref name="contents"/>
[...]

PR #1359, opened 2023-09-26
schema: relax model of work

pe-ro:


pe-ro:
Addresses #1350

bwbohl:
already addressed in #1358

Issue #1348, opened 2023-09-25
Add `<geogLoc>` element for geographic coordinates


pe-ro:
Currently MEI has no element for marking geographic coordinates. These could be useful for creating visualizations representing places of composition, birth, death, etc.  So, I propose adding a `<geog>` element --

```xml
<elementSpec ident="geog" module="MEI.namesdates">
  <gloss versionDate="2022-05-18" xml:lang="en">geographical coordinates</gloss>
  <desc xml:lang="en">Contains numeric cartographic data that represents a point, line, or
    area on the surface of the earth.</desc>
  <classes>
    <memberOf key="att.common"/>
    <memberOf key="att.authorized"/>
    <memberOf key="att.bibl"/>
    <memberOf key="att.edit"/>
    <memberOf key="att.facsimile"/>
    <memberOf key="model.textPhraseLike.limited"/>
  </classes>
  <content>
    <rng:text/>
  </content>
  <attList>
    <attDef ident="datum" usage="opt">
      <desc xml:lang="en">indicates the datum to which the data conforms.</desc>
      <valList type="semi">
        <valItem ident="WGS84">
          <desc xml:lang="en">(World Geodetic System) a pair of numbers to be interpreted
            as latitude followed by longitude according to the World Geodetic
            System.</desc>
        </valItem>
        <valItem ident="MGRS">
          <desc xml:lang="en">(Military Grid Reference System) the values supplied are
            geospatial entity object codes.</desc>
        </valItem>
        <valItem ident="OSGB36">
          <desc xml:lang="en">(ordnance survey great britain) the value supplied is to be
            interpreted as a British National Grid Reference.</desc>
        </valItem>
        <valItem ident="ED50">
          <desc xml:lang="en">(European Datum coordinate system) the value supplied is to
            be interpreted as latitude followed by longitude according to the European
            Datum coordinate system.</desc>
        </valItem>
      </valList>
    </attDef>
  </attList>
  <remarks xml:lang="en">
    <p>When the <att>datum</att> attribute is not provided, the content of this element is
      assumed to be compliant with the World Geodetic System (WGS84).</p>
  </remarks>
  <remarks xml:lang="en">
    <p>The model of this element is based on the Text Encoding Initiative (TEI) <ref
        target="https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-geoDecl.html"
        >geo</ref> element.</p>
  </remarks>
</elementSpec>
```
[...]

PR #1360, opened 2023-09-26
Add <geog> element

pe-ro:
Address issue #1348


pe-ro:
Only half-baked, closing.

pe-ro:
See discussion of #1348 for a better proposal

Issue #1297, opened 2023-08-01
Remodel inline elements with required ancestor, e.g. heraldry, watermark, locus


musicEnfanthen:
Follow-up of #1276 for MEI 6:

As @bwbohl pointed out in https://github.com/music-encoding/music-encoding/issues/1276#issue-1815405658:

> The `heraldry`, `watermark`, `locus` have Schematron rules testing for an ancestor `physDesc`. While this is true in a complete `meiHead`, it does not make sense in the anyStart schema if a files starts below the `physDesc` (e.g., in the sample-encodings collection). There might be other elements with parent or ancestor checks in Schematron rules. There are alternative solutions:
> 
>     * remove those Schematron checks in the anyStart customisation
> 
>     * rework them in the customisation to not throw an error but a warning or info
> 
>     * rework them generally to warning or info level.
> 
> 
> Is there no other way for those three elements mentioned to ensure they are only used in `physDesc`?

For MEI 5, we removed these Schematron rules in all customizations. The elements should be remodeld to be constrained to their required ancestors by the model, not just by Schematron rules.

PR #1361, opened 2023-09-26
schema: remove watermark's membership in msInline

pe-ro:
Partially addresses 
issue #1276 
issue #1297

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -2886,7 +2886,6 @@
       <memberOf key="att.bibl"/>
       <memberOf key="att.facsimile"/>
       <memberOf key="att.lang"/>
-      <memberOf key="model.msInline"/>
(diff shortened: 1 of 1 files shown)


riedde (approved):

musicEnfanthen (approved):

bwbohl:
I can agree that it’s not necessary to allow watermark "inline" within any text. Nevertheless I’d consider `<supportDesc>` or `<support>` as valid containers.

pe-ro:
It would be appropriate to add `<watermark>` to `<support>`, but not `<supportDesc>`.

riedde:
I'm open to this idea. There is also a proposal for developing watermark (#1632), just to avoid a duplicate development.

musicEnfanthen (approved):

bwbohl (approved):

PR #1364, opened 2023-09-26
schema: modify content model of monogr

pe-ro:
This PR brings the content model of `<monogr>` closer to that of its TEI counterpart.  It's also part of a larger effort to clarify the use of `<bibl>` and `<biblStruct>`, and 
accommodate non-FRBR description using a proposed `<biblFull>` element.


source/modules/MEI.shared.xml
```diff
@@ -6548,66 +6548,58 @@
         <rng:optional>
           <rng:choice>
             <rng:group>
-              <rng:choice>
-                <rng:ref name="model.respLikePart"/>
-                <rng:ref name="respStmt"/>
-              </rng:choice>
+              <rng:ref name="mei_model.respLikePart"/>
```
```suggestion
              <rng:ref name="model.respLikePart"/>
```

bwbohl: I think the prefix`mei_` should be removed here and also in the below references

pe-ro: Yes, they should be removed.

rettinghaus (dismissed):

source/modules/MEI.shared.xml
```diff
@@ -6635,11 +6628,10 @@
           <rng:choice>
             <rng:ref name="imprint"/>
             <rng:ref name="extent"/>
-            <rng:ref name="biblScope"/>
           </rng:choice>
         </rng:zeroOrMore>
-      </rng:group>
-    </content>
+      </rng:group>      
+      </content>
```
```suggestion
      </rng:group>
    </content>
```

source/modules/MEI.shared.xml
```diff
@@ -6635,7 +6628,6 @@
           <rng:choice>
             <rng:ref name="imprint"/>
             <rng:ref name="extent"/>
-            <rng:ref name="biblScope"/>
```
musicEnfanthen, line 6708: @pe-ro Is the removal of biblScope here correct? Is it not needed in monogr anymore?

pe-ro: @musicEnfanthen, good eye!  However, I think the removal of `<biblScope>` is appropriate.

musicEnfanthen (approved):

rettinghaus (approved):

musicEnfanthen:
Thank you!

PR #1365, opened 2023-09-26
schema: revise model of `<respStmt>`

pe-ro:
This PR changes the content model of respStmt to 
- make it more TEI-like ( resp+, nameLike+ or nameLike+, resp+);
- clarify use of respStmt

This is a backward-compatibility breaking change.  However, if the existing model of respStmt has been used to contain only names, each with a role attribute, then that usage can easily be converted via XSLT to the new model; that is, the content of the role attribute can be placed in a new resp element.

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -7513,7 +7513,7 @@
       <rng:zeroOrMore>
         <rng:ref name="annot"/>
       </rng:zeroOrMore>
-    </rng:content>
+    </content>
     <remarks xml:lang="en">
       <p>The model of this element is based on the <ref target="https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-respStmt.html">respStmt</ref> element of the Text Encoding Initiative (TEI).</p>
     </remarks>
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -7482,26 +7482,38 @@
       <rng:zeroOrMore>
         <rng:ref name="model.headLike"/>
       </rng:zeroOrMore>
+      <rng:choice>
+        <rng:group>
+          <rng:oneOrMore>
+            <rng:oneOrMore>
+              <rng:ref name="resp"/>
+            </rng:oneOrMore>
+            <rng:oneOrMore>
+              <rng:choice>
+                <rng:ref name="model.nameLike.agent"/>
+                <rng:ref name="name"/>
+              </rng:choice>
+            </rng:oneOrMore>
+          </rng:oneOrMore>
+        </rng:group>
+        <rng:group>
+          <rng:oneOrMore>
+            <rng:oneOrMore>
+              <rng:choice>
+                <rng:ref name="model.nameLike.agent"/>
+                <rng:ref name="name"/>
+              </rng:choice>
+            </rng:oneOrMore>
+            <rng:oneOrMore>
+              <rng:ref name="resp"/>
+            </rng:oneOrMore>
+          </rng:oneOrMore>
+        </rng:group>
+      </rng:choice>
       <rng:zeroOrMore>
-        <rng:choice>
-          <rng:ref name="name"/>
-          <rng:ref name="resp"/>
-          <rng:ref name="model.nameLike.agent"/>
-        </rng:choice>
+        <rng:ref name="annot"/>
       </rng:zeroOrMore>
-    </content>
-    <constraintSpec ident="check_respStmt" scheme="schematron">
-      <constraint>
-        <sch:rule context="mei:respStmt[not(ancestor::mei:change)]">
-          <sch:assert
-            test="(mei:resp and (mei:name or mei:corpName or mei:persName)) or
-            count(mei:*[@role]) = count(mei:*) and count(mei:*) &gt; 0"
-            role="warning">At least one element pair (a resp element and a name-like element) is
-            recommended. Alternatively, each name-like element may have a @role
-            attribute.</sch:assert>
-        </sch:rule>
-      </constraint>
-    </constraintSpec>
+    </rng:content>
     <remarks xml:lang="en">
       <p>The model of this element is based on the <ref target="https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-respStmt.html">respStmt</ref> element of the Text Encoding Initiative (TEI).</p>
     </remarks>
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -7509,6 +7509,12 @@
             </rng:oneOrMore>
           </rng:oneOrMore>
         </rng:group>
+        <rng:oneOrMore>
+          <rng:choice>
+            <rng:ref name="model.nameLike.agent"/>
+            <rng:ref name="name"/>
+          </rng:choice>
+        </rng:oneOrMore>
       </rng:choice>
       <rng:zeroOrMore>
         <rng:ref name="annot"/>


rettinghaus (dismissed):
Looks good to me, thanks!

riedde (approved):
Thanks you for this improvement. @pe-ro quick question: Have you checked if this has side-effects to the guidelines? Is there documentation that has to be changed?

pe-ro:
@riedde, I have not checked for any effects in the Guidelines.  I'll do so in the next couple of days.

pe-ro:
I just realized that the backward compatibility problem with my last proposal can be eliminated by a model for `<respStmt>` consisting of --
```xml
<content>
  <zeroOrMore>
    <ref name="model.headLike"/>
  </zeroOrMore>
  <choice>
    <group>
      <oneOrMore>
        <oneOrMore>
          <ref name="resp"/>
        </oneOrMore>
        <oneOrMore>
          <choice>
            <ref name="model.nameLike.agent"/>
            <ref name="name"/>
          </choice>
        </oneOrMore>
      </oneOrMore>
    </group>
    <group>
      <oneOrMore>
        <oneOrMore>
          <choice>
            <ref name="model.nameLike.agent"/>
            <ref name="name"/>
          </choice>
        </oneOrMore>
        <oneOrMore>
          <ref name="resp"/>
        </oneOrMore>
      </oneOrMore>
    </group>
    <oneOrMore>
      <choice>
        <ref name="model.nameLike.agent"/>
        <ref name="name"/>
      </choice>
    </oneOrMore>
  </choice>
  <zeroOrMore>
    <ref name="annot"/>
  </zeroOrMore>
</content>
```
This will continue to allow `<respStmt>` to contain name-like elements only.  It will likely also result in an ambiguous content model in an XSD version of the schema.  But, since that's not been a concern for a long time now, it seems inconsequential.

source/modules/MEI.shared.xml
```diff
@@ -7454,26 +7454,38 @@
...
+                <rng:ref name="model.nameLike.agent"/>
+                <rng:ref name="name"/>
+              </rng:choice>
+            </rng:oneOrMore>
+            <rng:oneOrMore>
+              <rng:ref name="resp"/>
+            </rng:oneOrMore>
+          </rng:oneOrMore>
+        </rng:group>
+      </rng:choice>
```
bwbohl: If I get it right the intention is not to predetermine the order of `resp`, `name`etc. Just out of curiosity:
Shouldnt' this do the same job?

```xml
<interleave>
    <oneOrMore>
       <ref name="resp"/>
    </oneOrMore>
    <oneOrMore>
       <choice>
          <ref name="name"/>
          <ref name="model.nameLike.agent"/>
       </choice>
    </oneOrMore>
</interleave>
```

bwbohl: or in pure-ODD:
```xml
<sequence preserveOrder="false">
    <elementRef key="resp" minOccurs="1" maxOccurs="unbounded" />
    <alternate minOccurs="1" maxOccurs="unbounded">
        <elementRef key="name" />
        <classRef key="model.nameLike.agent" />
    </alternate>
</sequence>
```

pe-ro:
@bwbohl, the objective is to ensure that `<respStmt>` allows only the following possibilities each time it is used:
- one or more `<resp>` elements followed by one or more "name-like" elements;
- one or more "name-like" elements followed by one or more `<resp>` elements;
- one or more "name-like" elements.

I don't think a model that includes `<interleave>` will do the job.  `<interleave>` would make all 3 available simultaneously.

pe-ro:
@riedde, following up on your question 

> Have you checked if this has side-effects to the guidelines? Is there documentation that has to be changed?

The short answer is, yes, changes will need to be made in the Guidelines everywhere `<respStmt>` is discussed or an example is provided.  That is, 

[3.3.1 Title Statement](https://music-encoding.org/guidelines/v5/content/metadata.html#headerTitleStatement), [3.3.2 Responsibility Attribution](https://music-encoding.org/guidelines/v5/content/metadata.html#headerrespstatement), [3.4.1.1 Edition Statement](https://music-encoding.org/guidelines/v5/content/metadata.html#headerEditionstmt), [3.4.1.3 Publication, Distribution, etc.](https://music-encoding.org/guidelines/v5/content/metadata.html#headerPublicationDistribution), [3.4.1.4 Series Statement](https://music-encoding.org/guidelines/v5/content/metadata.html#headerSeriesStatement), [3.4.3 Revision Description](https://music-encoding.org/guidelines/v5/content/metadata.html#headerRevisionDescription), [3.8.3 Minimal and Recommended Header Information](https://music-encoding.org/guidelines/v5/content/metadata.html#headerMinimalRecommendedHeader), [9.2.12 Bibliographic Citations and References](https://music-encoding.org/guidelines/v5/content/textencoding.html#sharedBibliographicCitations), [11.3.1 Encoding Genetic States](https://music-encoding.org/guidelines/v5/content/scholarlyediting.html#geneticStates)

I haven't checked to see what the extent of the changes might be.  I suspect that only light modification will be necessary.

bwbohl:
> @bwbohl, the objective is to ensure that `<respStmt.` allows only the following possibilities each time it is used:
> 
> * one or more `<resp>` elements followed by one or more "name-like" elements;
> * one or more "name-like" elements followed by one or more `<resp>` elements;
> * one or more "name-like" elements.
> 
> I don't think a model that includes `<interleave>` will do the job. `<interleave>` would make all 3 available simultaneously.

@pe-ro yes, absolutely right, interleave will allow them in any order!
Thanks for clarifying the intention.

My interleave example is almost MEI v4 and v5, except that it requires at least one child in `respStmt`. So yes, your proposal is definitely an improvement, as it requires at least a "name-like" child; what I like most is that it requires sticking to one pattern of resp-nameLike or nameLike-resp. Many thanks!

bwbohl:
Still have to check the guidelines, though…

pe-ro:
The idea of a responsibility statement comes from the notion of transcribing that part of a bibliographic item's title that says who's responsible for its content.  The most common form looks like --

```The post-Beethoven symphonists. Symphony writers since Beethoven. By Felix Weingartner. From the German by Arthur Bles. With twelve portraits.```

which one should encode in TEI, and with this change in MEI, as
```
<titleStmt xmlns="http://www.music-encoding.org/ns/mei" 
   meiversion="5.0+anyStart">
  <title>
     <titlePart type="main">The post-Beethoven symphonists.</titlePart>
    <titlePart type="subordinate">Symphony writers since 
       Beethoven.</titlePart>
  </title>
  <respStmt>
    <resp>By</resp>
    <persName>Felix Weingartner.</persName>
    <resp>From the German by</resp>
    <persName>Arthur Bles.</persName>
    <annot>With twelve portraits.</annot>
  </respStmt>
</titleStmt>
```
Though I can't put my hand on one at the moment, there are other responsibility statements that reverse the order of `<resp>` and the name(s) -- something like --
```
<titleStmt xmlns="http://www.music-encoding.org/ns/mei" 
   meiversion="5.0+anyStart">
  <title>Completely made-up title.</title>
  <respStmt>
    <persName>Willy Makeit</persName>
    <resp>editor</resp>
    <persName>Betty Dont</persName>
     <resp>illustrator</resp>
  </respStmt>
</titleStmt>
```
Finally, there are those statements that omit the `<resp>` altogether --
```
<titleStmt xmlns="http://www.music-encoding.org/ns/mei" 
    meiversion="5.0+anyStart">
  <title>
    <titlePart type="main">Beethoven :</titlePart>
    <titlePart type="subordinate">studies in the creative process /</titlePart> 
 </title>
  <respStmt>
    <persName role="author">Lewis Lockwood.</persName>
  </respStmt>
</titleStmt>
```
Since there's no indication of what Lewis Lockwood's role is on the item itself, MEI requires, in a departure from TEI, that it be supplied in the `@role` attribute on `<persName>`.

The model of `<respStmt>` used up until now allowed the free mixture of `<resp>` and name-like elements to accommodate the various forms that a responsibility statement can take.  However, it makes more sense to adopt a model closer to that of TEI and traditional library practice.

musicEnfanthen (approved):

bwbohl (approved):

PR #1366, opened 2023-09-26
Add biblFull element

pe-ro:


pe-ro:
This PR addresses in part issue #1078

PR #1367, opened 2023-09-26
schema: add `biblFull` element and `profileDescPart` model class

pe-ro:
Addresses in part issue #1078


bwbohl:
merged #1368 as these should go together
> This PR adds a profileDescPart class to which elements within the proposed `<biblFull>` element subscribe.

bwbohl:
merged #1366 as creating biblFull was the motivation for creating model.profileDescPart

source/modules/MEI.shared.xml
```diff
@@ -4598,7 +4601,7 @@
         the use of <gi scheme="MEI">barLine</gi> is not necessary within <gi scheme="MEI">measure</gi> elements in
         CMN.</p>
     </remarks>
-  </elementSpec>
+  </elementSpec>
```
```suggestion
  </elementSpec>
```

source/modules/MEI.shared.xml
```diff
@@ -4145,6 +4145,9 @@
       <memberOf key="model.textComponentLike"/>
     </classes>
   </classSpec>
+  <classSpec ident="model.profileDescPart" type="model">
+    <desc xml:lang="en">Groups elements that capture a portion of a bibliographic
+      description.</desc>
```
```suggestion
      description.</desc>
  </classSpec>
```

thx for spotting @musicEnfanthen

source/modules/MEI.shared.xml
```diff
@@ -4685,6 +4688,65 @@
...
+      <memberOf key="att.common"/>
+      <memberOf key="att.bibl"/>
+      <memberOf key="att.facsimile"/>
+      <memberOf key="att.lang"/>
+      <memberOf key="att.pointing"/>
+      <memberOf key="model.biblLike"/>
+    </classes>
+    <content>
+      <rng:zeroOrMore>
+        <rng:ref name="mei_model.identifierLike"/>
```
```suggestion
        <rng:ref name="model.identifierLike"/>
```

`mei_` should be deleted here and everywhere else

bwbohl:
converted to draft as:

1. some work still has to be done, and
2. most important: this should be seen as a proposal to resolve the greater discussion around responsibilities / agents /  FRBRized MEI vs. flat source description

pe-ro:
I'm returning to this PR after quite some time, and I'm a little confused.  I'm pretty sure that the PR included a new `<biblFull>` element, but I don't see it now.  There was some confusion around this PR and a some shuffling of things.  I'm wondering if `<biblFull>` got lost along the way and, if so, if there is a way to recover it.  I can't find my original work on this, and I'd prefer not to have to re-create it.

pe-ro:
Never mind.  I see that `<biblFull>` is actually addressed in #1366.  Sorry for the churn.

riedde:
I think we should thake and modify the guidelines proposal made in #1122 to have a description for a non-FRBR way of encoding the physical discription.

riedde:
This PR closes #1122

pe-ro:
Not sure how to proceed here.  I think this PR is ready for review, but not sure what to do about updating the branch.  @riedde , @bwbohl , @musicEnfanthen ?

musicEnfanthen:
Supersedes #1122

source/modules/MEI.shared.xml
musicEnfanthen: @riedde Can you add this suggested change please?

source/docs/03-metadata.xml
```diff
@@ -124,6 +125,18 @@
...
+               <div xml:id="nonfrbrsourcedescription" type="div3">
+                  <head>Non-FRBR Source Description</head>
+                  <p>MEI allows two methodologies to be used when doing metadata capture and source description. The <abbr>FRBR</abbr> methodology (<ptr target="#FRBR"/>) gives encoders a mechanism for building detailed descriptions of bibliographic entities. Most bibliographic-based projects should favor the FRBR method.</p>
+                  <p>However, there are cases where detailed bibliographic analysis at the level required by the FRBR module is not possible or desirable. These may include situations such as retro-conversion of non-FRBR catalogues, or capturing the output of notation encoding software where only minimal source description capabilities are available. In these cases, encoders may capture metadata in the MEI header without using FRBR encoding entities and instead choose a simplified methodology for source description.</p>
+                  <p>The <gi scheme="MEI">sourceDesc</gi> element as a child of the <gi scheme="MEI">fileDesc</gi> element may be used to capture simple bibliographic descriptions. An example is given below.</p>
```
musicEnfanthen, line 127: @pe-ro @riedde Is the idea to always use the new `biblFull` element for this purpose. Maybe `biblFull` should be mentioned here, so this would fix #1619 as well?

pe-ro: `<source>` may use `<bibl>`, `<biblStruct>`, or `<biblFull>`.  The first two permit bibliographic *citation* for the souce, while `<biblFull>` allows bibliographic *description*.  In other words, `<bibl>` and `<biblStruct>` are not as fully-featured as `<biblFull>`.

I think you're right that it would be advantageous to modify the text here so that it also closes #1619.  But I think revising 3.4.1.6 will also need to be done.

source/docs/03-metadata.xml
```diff
@@ -105,6 +105,7 @@
...
                <p>This chapter thus addresses the description of an encoded item so that the musical text, as well as its sources, encoding, and revisions are all thoroughly documented. Such documentation is necessary for scholars using the texts, for software processing them, and for catalogers in libraries and archives. Together these descriptions and declarations provide an electronic analog to the title page attached to a printed work. They also constitute an equivalent for the content of the code books or introductory manuals customarily accompanying electronic data sets.</p>
+               <p>For large encoding projects, the Functional Requirements for Bibliographic Records (<abbr>FRBR</abbr>) capabilities of MEI provides most flexibility and descriptive capabilities. However, MEI also provides simpler, non-FRBR methods for source description for those cases where the FRBR capbilities are not needed. These are further discussed in <ptr target="#nonfrbrsourcedescription" />.</p>
```
pe-ro, line 103: @musicEnfanthen, @riedde, How about consolidating this paragraph with the "Non-FRBR Source Description" text below and putting the whole thing at the end of section 3.4.1.6 Source Description?  For instance,

```
<p>MEI allows two procedures for source description -- a FRBR-based approach and a more traditional, 
non-FRBR method.  The <abbr>FRBR</abbr> mechanism (discussed at<ptr target="#FRBR"/>) gives 
encoders a way to create detailed hierarchical descriptions of bibliographic entities. Most 
bibliographically-based projects should favor the FRBR method.</p>

<p>However, when detailed bibliographic analysis at the level required by the FRBR module is not 
possible or desirable -- for example, for retrospective conversion of non-FRBR catalogs or for 
capturing the output of software where only minimal source description capabilities are 
available -- the <gi scheme="MEI">source</gi> element may contain <gi scheme="MEI">
bibl</gi> or <gi scheme="MEI">biblStruct</gi> elements, which permit the encoding of a
 bibliographic <emph> citation</emph> for source material.  On the other hand, the 
<gi scheme="MEI">biblFull</gi> element facilitates the structured <emph>description</emph> 
of a resource, including its physical properties, physical location, the identification of the work 
it embodies, and additional information not available in a structured way in the 
<gi scheme="MEI">bibl</gi> or <gi scheme="MEI">biblStruct</gi> elements.</p>
```
With this in place, I believe the "Non-FRBR Source Description" section can be eliminated. Some "jiggering" may be necessary to work in the examples, though.  I need to see something in place before I can go further.

source/modules/MEI.shared.xml
riedde: @bwbohl I would do this, but I do not have writing permission to this branch.

ahankinson:
Is it too late to suggest an alternative to `<biblFull>` or even question the necessity of it? 

It's clear in TEI that `<biblFull>`, like `<bibl>`, are for bibliographic _citation_. But here `<biblFull>` is fundamentally different, as the documentation states, but the similarities in the name make for some confusion in how they could be used. So possibly give it a name reflecting this utility, something along the lines of `<structDesc>`?

However, I also see an opportunity to simplify the source description method further. 

```xml
<sourceDesc>
         <source>
                 <titleStmt>
                     <title>La clé du caveau</title>
                 </titleStmt>
                 <physDesc>
                     <dimensions>
                         <width unit="cm">9.5</width>
                         <height unit="cm">14</height>
                     </dimensions>
                     <extent unit="page">376p</extent>
                 </physDesc>
                 <physLoc>
                     <identifier>XA Li 1811/1</identifier>
                     <repository>
                         <corpName>Universität Bern, Institut für Musikwissenschaft</corpName>
                         <geogName>Bern</geogName>
                         <country>Switzerland</country>
                     </repository>
                 </physLoc>
         </source>
     </sourceDesc>
```

That is, make all the children of the proposed `<biblFull>` element a direct child of `<source>`. Since `<source>` is not re-used in other contexts other than `<sourceDesc>`, there is no risk of child elements being inappropriately used in other contexts.

pe-ro:
@ahankinson, it's never too late. Your suggestion was considered, but not adopted because it doesn't play well with the current model of `<source>`.  (I also considered giving `<source>` the same content model as `<manifestation>`.) 

Using your suggestion, the new model of `<source>` would either have to disallow the current children (`<bibl>` and `<biblStruct>`) or continue to allow them but in alternation with your proposal.  Something like 
```
source (bibl? | biblStruct? | (titleStmt,  ...)?)
```
which has a couple of side-effects: there's now an unbalanced tree and the content of `<source>` would likely have to be limited to a single description.  (Currently, `<source>` permits multiple descriptions of the same source. Granted, not a common need, but changing this would cause a compatibility problem.)

For these reasons, it's better to wrap the elements you propose with another. I'm not wedded to the name "biblFull".  If you can think of a better name, I'd be happy to change it.

pe-ro:
Would changing the name of `<biblFull>` to `<biblDesc>` be beneficial?

lpugin:
I like it better. It is also more in-line with other elements in the header. But I don't have the full picture of what is being discussed here.

ahankinson:
> ... it doesn't play well with the current model of `<source>`.

That's sort of the problem, though, isn't it?  `<source>` was gutted with the FRBR work, and ideally it would  be restored to give an alternative means of source description. This is the core of the issue in #1078 .

In MEI 2012, `<source>` could contain the following child elements: 

> classification contents editionStmt history key langUsage mensuration meter notesStmt perfMedium physDesc pubStmt relatedItem seriesStmt titleStmt identifier incip tempo

> ... your new model of `<source>` would either have to disallow the current children (`<bibl>` and `<biblStruct>`) or continue to allow them but in alternation with your proposal. 

I've been trying to figure out how `<bibl>` and `<biblStruct>` fit in to source description. These are bibligraphic citation elements, not bibligraphic description elements. But would you need a separate `<source>` element just to encode a single `<bibl>` entry? Why not move them up a level into `<sourceDesc>`? 

This is based on the assumption that a `biblList`  would anyway be a better container for bibliographic citations to multiple sources, whereas `<source>` would be an actual analytical catalogue record for each source.

Maybe something like: 

```xml
<sourceDesc>
    <biblList>
        <bibl ... >
        <bibl ... >
        ... etc. 
    </biblList>
</sourceDesc>
```

So the content model would simply change to:

`sourceDesc: (source? | biblList?)`

Then `<source>` can be reserved for bibliographic description, but shorter forms of citation can also be used.

(Not to mention that `<bibl>` allows much more structured content than `<biblStruct>` as child elements... But that's a can of worms for another day.)

> Would changing the name of `<biblFull>` to `<biblDesc>` be beneficial?

I think my objection is the `bibl` bit, since it is getting conflated between citation and description. `biblList / bibl / biblStruct` all fit together, but `biblFull` (or `biblDesc`) would be different. That's why I proposed `<structDesc>` (structured description) since it makes a clear separation from the existing `bibl...` family of elements.

pe-ro:
Sorry, I missed your suggestion of `<structDesc>`.  I see your point, but `<structDesc>` is too generic -- it doesn't communicate what it's a structured description of.

> (Not to mention that `<bibl>` allows much more structured content than `<biblStruct>` as child elements... But that's a can of worms for another day.)

Not opening another can of worms, but I think you're misusing the term "structured".  The content model of `<bibl>` contains a lot of elements, but that's because its purpose is to capture unpredictable data.  One can freely mix text and sub-elements in any order.  This mix of text and elements is typically referred to as "unstructured content".  On the other hand, the content models of `<biblStruct>` and `<biblFull>` are structured; that is, these elements exclude a lot of the stuff allowed by `<bibl>` (especially the text between sub-elements) and their content is (mostly) ordered.

pe-ro:
OK, renaming `<sourceDesc>` to `<sourceList>` and `<biblFull>` to `<sourceDesc>` results in
```
sourceList (head*, source+)
source (head*, (locus|locusGrp)*, (model.biblLike|sourceDesc)*)
sourceDesc (identifier*, titleStmt, macro.biblDescPart, notesStmt?, 
    macro.profileDescPart?, relatedItem*)
```
Renaming `<sourceDesc>` to `<sourceList>` breaks with TEI practice somewhat, but I don't think that's a huge issue at this point because MEI's use of `<sourceDesc>` was already a departure from TEI practice.

Since `<sourceDesc>` is now descriptive and not citational in nature, its content model can be relaxed in the same way proposed for the FRBR elements.  We may also want to begin to think about harmonizing `<sourceDesc>` and `<manifestation>`.

@riedde, thoughts?

ahankinson:
> ... renaming `<sourceDesc>` to `<sourceList>` and `<biblFull>` to `<sourceDesc>` ... Since `<sourceDesc>` is now descriptive and not citational in nature... 

Hmmm... I didn't suggest that? I suggested `<sourceDesc>` takes `<biblList>` or `<source>` to handle both citation and descriptive methods. Then `<source>` effectively gets reverted to its pre-FRBR module state.

Somehow that seems less drastic than renaming `<sourceDesc>` etc.?

pe-ro:
@ahankinson, I was trying to deal with the problem of accurately naming the function of the `<source>` element in your proposal.  As I see it, `sourceDesc/biblList/bibl` and `sourceDesc/source` are both describing a source but with different names and structures.  Also, what's the purpose of `<biblList>` here?  Why not just `sourceDesc/bibl` or `sourceDesc/biblStruct`?

Unlike your proposal, `sourceList/source/bibl`,  `sourceList/source/biblStruct`, and `sourceList/source/sourceDesc` all make it clear that one is describing a source.

Reverting `<source>` to an earlier state seems to me to be adopting a "convenience" that causes more problems.  We've made a lot of progress in the header in general and simply plopping a chunk of the "old" schema into the next version seems regressive to me.  Moving forward with a new structure won't be any more work than determining whether the "old" structure still does what we want..

Renaming elements isn't drastic at all.  It's easily accomplished in both the schema and in existing instances.  Changing structures is more difficult -- that's why my suggestion creates only minimal structural changes (compared with v. 5.1) at the cost of changing a couple of names.

ahankinson:
> As I see it, `sourceDesc/biblList/bibl` and `sourceDesc/source` are both describing a source but with different names and structures.

Exactly! I'm trying to square the difference of where, currently, in `<source>` you would use `<bibl>` and `<biblStruct>` and where you might use `<biblFull>`. You will see that earlier I posited that despite the similarities in name, they have fundamentally different functions. 

So I was trying to allow for the "citation" method; that is, here are a list of references we used to make up the musical source in this file; vs. catalogue description;  that is, here is a full catalogue record of the source. This seems to be a current distinction being made in `<source>` so I was simply trying to preserve that.

This is why I suggested the `<bibl*>` elements get moved up one level to `<sourceDesc>` since they effectively contain a citation list. This then preserves the `<source>` element ONLY for doing actual source description.

I also proposed `<biblList>` instead of making `<bibl*>` direct children of `<sourceDesc>`. This is because you would probably have one or more citations, and `<biblList>` seems like it's the ideal container for several `<bibl>` and / or `<biblStruct>` elements.

> Reverting `<source>` to an earlier state seems to me to be adopting a "convenience" that causes more problems...

Let's not lose track that my suggestion above actually isn't to bring back the old schema, but simply to remove the extra hierarchy that you're currently proposing -- that is, to forego `<biblFull>` and make the child elements that you are suggesting for `<biblFull>` direct children of `<source>`. It just so happens that this is, more-or-less, what `<source>` used to be. 

As one of the original authors of #1078 I can say with a fair amount of certainty that this is what we had in mind when writing that

pe-ro:
> Exactly! I'm trying to square the difference of where, currently, in <source> you would use `<bibl>` and `<biblStruct>` and where you might use `<biblFull>`. You will see that earlier I posited that despite the similarities in name, they have fundamentally different functions.

But moving `<bibl>` and `<biblStruct>` up a level leads to problems. The following markup
```
<sourceDesc>
  <biblList>
    <bibl/>
  </biblList>
  <source/>
</sourceDesc>
```
makes a distinction between sources cited and sources described that isn't necessary or desirable.  The `<bibl>` element is "describing" a source and the `<source>` element is "describing" a source, but they're doing so at different places in the hierarchy. I think this will be very confusing to the casual user. Also, if I want to have a citation to a source and description of the same source, I can't link the citation and the description together without some heavy-weight pointing mechanism.  

In my opinion, it's better to use
```
<sourceList>
  <source>
    <bibl/> | <biblStruct/> | <sourceDesc/>
  </source>
</sourceList>
```
where `<source>` may contain any mix of `<bibl>`, `<biblStruct>`, or `<sourceDesc>` elements related to the source item.

By changing the currently-named `<sourceDesc>` to `<sourceList>`  the current structure remains the same, and renaming `<biblFull>` to `<sourceDesc>` accommodates the difference between bibliographic citation and bibliographic description.

ahankinson:
Whoa. I think you just completely inverted the hierarchy (`sourceDesc / source` to `source / sourceDesc`), and added a new element `sourceList`. So much for not changing the model too much! :-D

> ... makes a distinction between sources cited and sources described that isn't necessary or desirable. 

but this distinction already exists in this current proposal, and I was trying to preserve that while also reserving `source` for "original cataloguing". (that is, the use of `<title>` in citation, but `<titleStmt>` in original cataloguing.)

It exists because `<bibl>` and `<biblStruct>` are children of `<source>` which, as I keep pointing out, have fundamentally different content models from the proposed `<biblFull>`, yet they will co-exist so I thought it **was** necessary and desirable. It's just that it's moved up a level to make the distinction between the citation list of sources used in creating the file, and the original cataloguing description of the source.

I'm not sure I see the utility in changing `<sourceDesc>` to `<sourceList>`. That seems like a distinction without a difference. Maybe it's something we can come back to?

> Also, if I want to have a citation to a source and description of the same source, I can't link the citation and the description together without some heavy-weight pointing mechanism.

Why would you want to do that? You can usually derive the citation from the description, but you can't do it the other way around. But if you did want to do that, isn't it just a `<bibl sameas="#desc"> ... `? Or possibly `<bibl target="#desc">...` with `<source xml:id="desc"> ...`? That doesn't seem too heavy?

riedde:
This issue seems to be stucked. After I read the whole story I think there are a few options: Expanding the content model of `<source>` by:
- creating an element for bibliographic description
- restructuring the content model
- adding more (preexisting) elements

But as @ahankinson mentioned, I remember that this issue derives from #1078 and should be a step of resolving the problem that the FRBR-Module is not independend and that some parameters, needed for a bibliographic description, are not possible at the moment without using FRBR.

One idea from the comments above that I like to push is the idea of giving `<source>` the same content as `<manifestation>`. This could be a compromise...
However, I do not propose to have two elements for the same thing, but I'm open to think about ways of joining these two elements. E.g., renaming `<manifestationList>` to `<sourceList>` (might be more coherent than `sourceDesc`) and renaming `<manifestation>` to `<source>`. I do not want to undermine the FRBR-Module, but if the element that represents a manifestation is called `manifestation` or `source` does not break the model. It is just a question of defining the element and explaning the implementation. (Thus, the otherway round would also be possible `source` to `manifestation`.)

I'm not sure if I got every detail in this discussion, but perhaps this approach can make the introduction of (multiple) new elements and intensive restructuring superfluous.

pe-ro:
@riedde, while they can be "harmonized" (brought closer together), the `<source>` and `<manifestation>` elements can't use exactly the same content model because they serve different purposes.  A "source" in the non-FRBR context may describe any of the FRBR entities "work", "expression", "manifestation", and/or "item".  In other words, "source" is a broader concept than "manifestation". A "manifestation" is a particular kind of "source"; i.e., a production run made up of duplicate items. A "singleton" is a manifestation with a production run of one, a single item usually produced as an autograph.  A manifestation only makes sense in the FRBR context. Giving it the name "source" would be confusing.  Likewise, renaming `<source>` to `<manifestation>` in the sourceDesc context would be just as confusing, if not more to those who want to avoid FRBR.  Finally, using a single element for both `<source>` and `<manifestion>` would make it more difficult to provide FRBR as a stand-alone module.

The attached files illustrate how I think `<source>` and `<manifestation>` ought to work. The sourceDesc_testing.mei file shows all the possible children of `source/biblDesc`, `source/bibl`, and `source/biblStruct`.  While one probably wouldn't use more than one of `<biblDesc>`, `<bibl>`, or `<biblStruct>` within `<source>`, allowing all three provides the greatest flexibility.  As described above, these 3 child elements each contain a different kind of "description". `<bibl>` is the least, and `<biblStruct>` is the most,  restrictive of the three.  These elements are designed for capturing traditional bibliographic references.  `<biblDesc>` on the other hand is similar to TEI's `<biblFull>` because it allows for more of a "description" than a "reference".  Its content model is also closest to that of the `<manifestation>` element.  Principally, the models differ in content (`<manifestation>` allowing more and different  kinds of relationships to be encoded).

The file manifestation_testing.mei is provided for comparision with sourceDesc_testing.mei. It demonstrates the possible children of `<manifestation>`.

The file mei-all_anyStart_testing.rng is an edited version of mei-all_anyStart.rng used to validate the other 2 files. It's probably not in sync with the latest development version of the schema.

[meiSourceManifestationTesting.zip](https://github.com/user-attachments/files/20195411/meiSourceManifestationTesting.zip)

riedde:
> (I also considered giving <source> the same content model as <manifestation>.)

I reacted to this comment (in https://github.com/music-encoding/music-encoding/pull/1367#issuecomment-2665850544)

Thank you for the clarification and the files. I'm getting an idea of the dimension of this topic, slowly.

At todays meeting of @music-encoding/ig-metadata we had a look into this issue and we like to ask for some use-cases and examples for deeper insights. It is hard to understand and to reflect the need of all elements you propose as content for `biblDesc`, e.g., `perfMedium` or `context`, `audience` (which have a predetermined place in the mei hierachy) without having examples, especially for me coming from a different research context.

pe-ro:
@riedde, thanks for helping to get this issue "unstuck". :-)

As was noted earlier in the discussion, `<source>` was reduced in power and flexibility as more attention was given to the FRBR module.  The main thrust of the current issue is to restore `<source>` by giving it the same descriptive power as the FRBR tree, i.e., work, expression, manifestation, and item, just in a flatter (but not completely flat) structure.  Mostly, this means allowing multiple children of `<source>` -- `<bibl>`, `<biblStruct>`, and `<biblDesc>` -- and giving `<biblDesc>` most (but not all) the same elements as the FRBR entities.

I don't understand your comment about allowing `<perfMedium>`, `<context>`, and `<audience>` in `<biblDesc>`, mostly because of your parenthetical comment -- "(which have a predetermined place in the mei hierachy)".  I can only say that because these elements are allowed in `<manifestation>`, etc., they're also allowed by `<biblDesc>`.  Please let me know if this answer isn't specific enough.

I found some problems with the RNG and sourceDesc example file I sent yesterday, so I'm including updated versions here.  The zip file also contains new examples for FRBR entities other than manifestation.

[meiSourceManifestationTesting.zip](https://github.com/user-attachments/files/20230273/meiSourceManifestationTesting.zip)

pe-ro:
Now with new and improved examples!

[meiSourceTesting.zip](https://github.com/user-attachments/files/20234614/meiSourceTesting.zip)

The bibliographic_examples.mei file demonstrates how bibliographic entities can be described using `<bibl>`, `<biblStruct>`, or `<biblDesc>`.  For bibliographic **references**, the differences are mostly stylistic.  `<biblDesc>`, however, allows much more in-depth **description** than the other two.  

There's fairly wide leeway in the art of bibliographic citation and description, so the examples are meant to show possibilities, not how these things must be done.

riedde:
@pe-ro 
> I don't understand your comment about allowing `<perfMedium>`, `<context>`, and `<audience>` in `<biblDesc>`, mostly because of your parenthetical comment -- "(which have a predetermined place in the mei hierachy)". I can only say that because these elements are allowed in `<manifestation>`, etc., they're also allowed by `<biblDesc>`. Please let me know if this answer isn't specific enough.

Yes, giving `<source>` the same content as `<manifestation>` answers my question. I see your argument, but my point was that e.g., `<perfMedium>`, `<context>` and `<audience>`, are allowed in `<work>` (not `<manifestation>`) and `<work>` is NOT part of the FRBR-Module. It's definded in `MEI.header` and always available as part of a core module. Maybe I just didn't get the wideness of bibliographic description, but what is the benefit of giving `<source>` also the content of `<work>` when `<work>` is present? (I know that there were some overlappings in MEI2012)

pe-ro:
This looks good so far.  There are additional PRs that pertain to this one, but in order not to confuse things more, I think this PR should be merged first.

source/docs/03-metadata.xml
riedde: inserted in 88464ed76408f14e011258c5b4b14d17901b57c1

bwbohl:
> This looks good so far. There are additional PRs that pertain to this one, but in order not to confuse things more, I think this PR should be merged first.

Which PRs should be part of this one?

riedde:
After reviewing this PR I can't figure out which are the PRs mentions  by @pe-ro and @bwbohl

This could be a candidate for MEI 6.

bwbohl (dismissed):

bwbohl:
> This looks good so far. There are additional PRs that pertain to this one, but in order not to confuse things more, I think this PR should be merged first.

@pe-ro: Can you point us to those?

pe-ro:
> This looks good so far. There are additional PRs that pertain to this one, but in order not to confuse things more, I think this PR should be merged first.
>
> @pe-ro: Can you point us to those?

No, I can't because I've lost that train of thought.  In any case, this PR should be merged soon.

bwbohl:
@pe-ro I guess it doesn’t really matter to have other PR’s included here because the changes to me seem sound and self-contained. @music-encoding/maintainers please review and if you see it like me, approve.

bwbohl (approved):

riedde (approved):
According to https://github.com/music-encoding/music-encoding/pull/1367#issuecomment-4018860181 depending PRs may come up afterwards.  For the moment this looks solid and stable.

pe-ro:
I believe the merging of this PR has caused a problem; that is, the commits I made just before the merge introduced an error that causes the schema not to be valid.  I'm not sure if the problem is in the PR itself or in the teiodds.xsl stylesheet --

/home/runner/work/music-encoding/music-encoding/lib/stylesheets/rng/build-to.xml:30: The following error occurred while executing this line:
/home/runner/work/music-encoding/music-encoding/lib/stylesheets/common/teianttasks.xml:355: Fatal error during transformation using /home/runner/work/music-encoding/music-encoding/lib/stylesheets/profiles/default/rng/to.xsl: An empty sequence is not allowed as the value of variable $to_be_sequenced; SystemID: file:/home/runner/work/music-encoding/music-encoding/lib/stylesheets/odds/teiodds.xsl; Line#: 2102; Column#: 1

riedde:
Thanks you @pe-ro for this hint. @bwbohl and me fixed that. the problem was a missing `rng:` namespace and a prefix before element names `mei_` which have been removed.

PR #1370, opened 2023-09-26
schema: remodel role-specific elements in work and expression

pe-ro:
This PR is one of a number of replacements for PR #799

### DIFF
--- a/source/modules/MEI.frbr.xml
+++ b/source/modules/MEI.frbr.xml
@@ -185,7 +185,15 @@
         <rng:ref name="model.titleLike"/>
       </rng:oneOrMore>
       <rng:zeroOrMore>
-        <rng:ref name="model.respLikePart"/>
+        <rng:choice>
+          <rng:ref name="model.respLikePart"/>
+          <ref name="arranger"/>
+          <ref name="contributor"/>
+          <ref name="funder"/>
+          <ref name="librettist"/>
+          <ref name="lyricist"/>
+          <ref name="sponsor"/>
+        </rng:choice>
       </rng:zeroOrMore>
       <rng:zeroOrMore>
         <rng:ref name="model.workIdent"/>
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -2962,6 +2962,11 @@
       </rng:oneOrMore>
       <rng:zeroOrMore>
         <rng:ref name="model.respLikePart"/>
+        <rng:ref name="contributor"/>
+        <rng:ref name="funder"/>
+        <rng:ref name="librettist"/>
+        <rng:ref name="lyricist"/>
+        <rng:ref name="sponsor"/>
       </rng:zeroOrMore>
       <rng:zeroOrMore>
         <rng:ref name="model.workIdent"/>
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -2961,12 +2961,14 @@
         <rng:ref name="model.titleLike"/>
       </rng:oneOrMore>
       <rng:zeroOrMore>
-        <rng:ref name="model.respLikePart"/>
-        <rng:ref name="contributor"/>
-        <rng:ref name="funder"/>
-        <rng:ref name="librettist"/>
-        <rng:ref name="lyricist"/>
-        <rng:ref name="sponsor"/>
+        <rng:choice>
+          <rng:ref name="model.respLikePart"/>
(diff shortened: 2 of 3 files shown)


rettinghaus:
Why can't we continue with #799? 
It was noted before, that ["Reviewers review proposals. They do not suggest their own because they don't like the original."](https://github.com/music-encoding/music-encoding/pull/625#issuecomment-540963413)

pe-ro:
@rettinghaus, by all means continue with PR #799. I hope you'll note my comments there -- comments which led to creation of a this PR which has the same motivation as  #799 but accomplishes the goal with different means.

Of course reviewers review, but a PR should not be granted an automatic approval either.

ahankinson:
Yeah, I think this duplicates the work done in #799 ... This PR removes these elements without also replacing them with something, meaning that merging this will leave MEI in a state where we remove the ability to describe a composer or an author, without also having a replacement method. I know this is only in development, but it would be good to have removals and replacement solutions together, IMO.

pe-ro:
Sorry for the confusion.  The separation of this PR from #1371 is due to 2 things:
- an attempt (perhaps misguided) to create requests that are as atomic as possible; and
- my lack of knowledge of how to create a single PR that includes changes to multiple files.

bwbohl:
> Sorry for the confusion. The separation of this PR from #1371 is due to 2 things:
> 
> * an attempt (perhaps misguided) to create requests that are as atomic as possible; and
> * my lack of knowledge of how to create a single PR that includes changes to multiple files.

Ok, I understand. So, the duplication is due to workflow issues. If I analyze the situation correctly, you, @pe-ro, are using the GitHub web interface for making your modifications to the files in the `develop` branch. This results in a new branch and pull request for every file you change. This is a bit cumbersome to maintain for reviewers, as they have to keep track of all the branches and pull requests that should go together.

The best option in case of an existing PR that wants to achieve a similar thing would be to make use of the comment and, most importantly, code-line comments (which, btw. also allow direct alternative code suggestions that could be merged into the PR after the discussion is resolved).

In general, if you want to modify multiple files, the most common way would be to clone the repository to your machine, create a new branch there, and then make granular commits to that branch for every change you make, e.g., commit 1 : remove author and composer; commit 2: relax model of work, and so forth.

If you want or have to use GitHub’s web interface, you could also create a branch first, then switch to that branch (instead of develop) make your changes to a single file and commit it directly to that newly created branch (each with sound granularity and change description). After you have completed all your changes for the goal you want to achieve, you can open a new pull request from your branch against develop. Consequently, there would be just one pull request, with all the changes needed to make it a sound and round code-suggestions^.

Given the situation we have right now I would propose the following workflow.

1. We mark this PR a draft.
2. We merge all your PRs that should go together into one single branch that is not develop; If I get it right this would mean merging #1371, #1372, and #1373 into the branch of #1370 (being this PR; code of which is in the branch `develop-agents`). After that, the two concurring PRs can be compared more easily.
2. We transfer the changes of this (your) PR to comments and code-suggestions in #799; this way all the proposed modifications and the discussion are in the same place and easier to review.

This sound good?

pe-ro:
@bwbohl, sorry for not responding to this sooner.  Thanks for helping me out.  I think it will be easier for everyone, including me, to remove/mark-as-a-draft/whatever the pull requests I submitted earlier and for me to follow your instructions on using the GitHub interface to submit a single, comprehensive pull request.  There will be a large number of revisions, but the logic will be easier to follow overall.

bwbohl:
Hey @pe-ro, thanks for coming back to this. Never mind any delays, we all are doing this with the time we can spare ;-)
I changed the base of the PRs mentioned above to the `develop-agents` branch so we can merge them.

pe-ro:
Sorry, now I'm more confused.  Should I create a new pull request from a new branch or leave things as they are?

bwbohl:
We can just merge these PRs to other branches than develop, namely the `develop-agents` branch. This way, everything comes together in one, and we can pimp up the explanation here. How does that sound?

pe-ro:
If that's easier for you, it's fine by me.  Is anything else required of me?

bwbohl:
I could just merge them

pe-ro:
OK, but there will be conflicts between my suggestions and those of #799.

bwbohl:
Yeah, that's why I merged your other develo-agentX branches here. Now we have your proposal in one place and can proceed e.g. by comparing this branch to the one of #799

pe-ro:
Got it. Thanks again.

PR #1371, opened 2023-09-26
Modify role specific elements

pe-ro:
Add new creator element

Remove membership in respLikePart for
- arranger;
- funder;
- librettist;
- lyricist; and
- sponsor

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -4515,7 +4515,7 @@
       <memberOf key="att.evidence"/>
       <memberOf key="att.facsimile"/>
       <memberOf key="att.lang"/>
-      <memberOf key="model.respLikePart"/>
+      <memberOf key="att.metadataPointing"/>
     </classes>
     <content>
       <rng:zeroOrMore>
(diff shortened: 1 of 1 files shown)


ahankinson:
Is this an alternate solution to the creation of `<agent>`?

pe-ro:
Yes, this is an alternative to the creation of `<agent>`.  In my opinion, the proposal to create  `<agent>` as a replacement for the various role-specific elements 
- places too much of a burden on the encoder to specify the nature of the agent's relationship to the bibliographic item; and
- doesn't help clarify that some agents are privileged in some styles of bibliographic description.  Allowing some roles to be "singled out" makes it easier to transcode from existing markup like MARC and MODS.

`<creator>` and `<editor>` can mark those privileged roles and other roles can go into `<respStmt>`.  As I've said elsewhere, if treating some roles as special offends the encoder's sensibilities, then all roles can go into `<respStmt>` since `<respStmt>` retains its former children, i.e., `<resp>` and name-lIke elements.

`<creator>`, `<editor>` and `<resp>` are all places where authority attributes (`@auth` and `@auth.uri`) can be used to provide pointers to standard values for the agent-to-item relationship.  While the name-like elements (`<name>`, `<persName>`, and `<corpName>`) can use these same attributes for pointers to standardized / variant names for these entities.

PR #1375, opened 2023-09-28
Way to Pure ODD

rettinghaus:
This is a big step towards purification of our ODD schema. 
Every step taken here doesn't change the produced RNG at all.
Some things can be simplified even more, but my goal was to not change the generated RNG.

### DIFF
--- a/source/modules/MEI.analytical.xml
+++ b/source/modules/MEI.analytical.xml
@@ -192,7 +192,7 @@
           alteration, <val>1</val>, <val>v7</val>, <val>^1</val>, or <val>v5+</val>, for example. 
           The amount of chromatic alternation is not indicated.</desc>
         <datatype>
-          <rng:ref name="data.SCALEDEGREE"/>
+          <dataRef key="data.SCALEDEGREE"/>
         </datatype>
       </attDef>
     </attList>
@@ -213,7 +213,7 @@
         <gloss xml:lang="en">interval harmonic</gloss>
         <desc xml:lang="en">Encodes the harmonic interval between pitches occurring at the same time.</desc>
         <datatype maxOccurs="unbounded">
-          <rng:ref name="data.INTERVAL.HARMONIC"/>
+          <dataRef key="data.INTERVAL.HARMONIC"/>
         </datatype>
       </attDef>
     </attList>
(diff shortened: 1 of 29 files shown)


rettinghaus:
Things still missing:

* adapting xslt for generating the Guidelines

pe-ro:
@rettinghaus, @raffazizzi, how does Pure ODD handle RNG's `<interleave>` concept?  As I've said elsewhere, if it doesn't handle it at all, then we'd be missing a significant advantage of RNG.  If Pure ODD can't handle `<interleave>`, then can we go with a "mostly Pure ODD" where those models that don't use `<interleave>` can be "pure", while those that do can be expressed in RNG itself?

bwbohl:
@pe-ro the interleave shoud be representable by something similar to:

```xml
<content>
 <alternate minOccurs="1"
  maxOccurs="unbounded">
  <elementRef key="ptr"/>
  <elementRef key="ref"/>
 </alternate>
</content>
```
(cf. https://www.tei-c.org/release/doc/tei-p5-doc/en/html/TD.html#index-egXML-d54e151588)

peterstadler:
Regarding `rng:interleave`: the intuitive ODD way to do this is via `<tei:sequence preserveOrder="false">`. See https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-sequence.html 

That said, there is some discomfort with it, see e.g. this [thread on TEI-L](https://listserv.brown.edu/cgi-bin/wa?A2=TEI-L;d4f23489.2303)

peterstadler:
@rettinghaus how did process the files and is your script (don't expect you to do this by hand) available? Did you have a look at the [TEI scripts](https://github.com/TEIC/TEI/tree/dev/P5/Scripts)? 

In general, I think we should rather review the script to pureOddify the MEI sources rather than the outcome.

pe-ro:
@peterstadler, Thanks for the reminder/update.  I'm not concerned about the PureODD syntax, only that the concept of controlling how many elements may occur without controlling their order is accommodated somehow.

Frankly, I don't see the point of inventing new syntax (PureODD) for something easily represented in RNG when one is allowed to incorporate RNG into ODD.  But as a great philosopher once said, "Who am I to blow against the wind?"  :-)

rettinghaus:
@peterstadler No, this was thus far manual work.

rettinghaus:
In order to clear things up I've rebased this branch and it MUST not be merged before #1501.

PR #1378, opened 2023-10-04
guidelines: fix documentation for optimize

### LINKED ISSUE #1377: Add "not" to the definition of `@optimize="true"`
Am I wrong, or is the definition of `@optimize="true"` in the MEI guidelines wrong? According to the MEI guidelines (versions 3, 4 and 5.0 checked): "when the value is 'true', empty staves **are displayed**." But it should be the other way round: "when the value is 'true', empty staves are **not** displayed **[i.e. hidden, 'optimized']**." 

In Verovio it is rendered correctly the other way round (example from W. A. Mozart, String Quartet K. 168/4, empty staves in the cello at the beginning, MEI-code downloadable in: https://dme.mozarteum.at/movi/navigator/168/004/01/nma):

- empty staves (cello from m. 6) are **not** displayed, when `@optimize="true"`:

KV_168_04_01_optimize
//
KV_168_04_06_optimize_true

- empty staves (cello from m. 6) are displayed, when `@optimize="false"`:
KV_168_04_01_optimize
//
KV_168_04_06_optimize_false

I think the word "not" was just left out in the definition...

ahankinson:
Fixes #1377

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -2334,7 +2334,7 @@
     <attList>
       <attDef ident="optimize" usage="opt">
         <desc xml:lang="en">Indicates whether staves without notes, rests, etc. should be displayed. When the
-          value is 'true', empty staves are displayed.</desc>
+          value is 'true', empty staves are not displayed.</desc>
         <datatype>
           <rng:ref name="data.BOOLEAN"/>
         </datatype>


rettinghaus (approved):
Thank you. 
This should be backported, any idea how?

bwbohl:
@rettinghaus do you mean, fixing this in the previous versions of MEI?
I think that is dangerous because the "simple fix" just turns around the definition of the attribute 180°. Changing that would render all MEI files not referencing our future schema unpredictable in their meaning.

rettinghaus:
@bwbohl I thought of the guidelines, but this has never been mentioned there, only in the definition.
But this is something for 5.0.1. 😉

pe-ro (approved):
Approved

bwbohl:
We probably put this in the top of the release notes!

Issue #1347, opened 2023-09-25
Use of a:defaultValue causes problems with schematron for <tempo>


pe-ro:
The use of a:defaultValue in the definition of the meiversion attribute causes the schematron associated with `<tempo>` to issue an error for all uses of `<tempo>` in any metadata context; in `<work>` and `<expression>`, for example.

It also violates the principal that default attribute values are to be avoided because they insert values into the document model that aren't explicitly in a document's markup.  MEI adhered to this principal until v. 5.

So, instead of modifying the constraints applied to `<tempo>` in the bibliographic context, I propose to --
- remove a:defaultValue in the definition of `@meiversion`; and 
- replace the schematron dealing the document element and @meiversion with the following constraints:
```xml
<sch:rule xmlns="http://www.tei-c.org/ns/1.0"
  xmlns:rng="http://relaxng.org/ns/structure/1.0"
  xmlns:sch="http://purl.oclc.org/dsdl/schematron" context="/mei:*">
  <sch:assert test="/mei:*[@meiversion]">The document element must have the
@meiversion attribute in order to indicate the MEI version in use.</sch:assert>
</sch:rule>
```
```xml
<sch:rule xmlns="http://www.tei-c.org/ns/1.0"
  xmlns:rng="http://relaxng.org/ns/structure/1.0"
  xmlns:sch="http://purl.oclc.org/dsdl/schematron" context="/mei:*//*[@meiversion]">
  <sch:assert test="not(@meiversion)">Only the document element can have an
@meiversion attribute.</sch:assert>
</sch:rule>
```
and
```xml
<sch:rule xmlns="http://www.tei-c.org/ns/1.0"
  xmlns:rng="http://relaxng.org/ns/structure/1.0"
  xmlns:sch="http://purl.oclc.org/dsdl/schematron" context="/mei:*">
  <sch:report role="error"
    test="not(local-name(.) = ('mei')) and not(@meiversion = '5.0+anyStart')">If the
    document element is any other than mei, the value of @meiversion must be
    '5.0+anyStart'.</sch:report>
</sch:rule>
```

bwbohl:
@pe-ro, thanks for reporting, will have a look into this.

pe-ro:
While we're on this topic, the current Schematron for `<tempo>` --
```xml
<pattern xmlns="http://purl.oclc.org/dsdl/schematron"
  id="mei-tempo-tempo_start-type_attributes_required-constraint-rule-184">
[...]

PR #1386, opened 2023-10-26
schema: replace version 5.0 with dev

musicEnfanthen:
Following the 5.0 release, the content of the develop branch was identical to that of the stable branch. This pull request restores the dev version for the develop branch.

The proposed solution involves using `dev+...` as the version name/number, eliminating the need to definitely determine the next version number.

However, it is still up for discussion whether to use `6.0-dev+...`, `5.1-dev+...`, or another naming convention.

### DIFF
--- a/customizations/mei-all.xml
+++ b/customizations/mei-all.xml
@@ -108,7 +108,7 @@
         <classSpec ident="att.meiVersion" module="MEI.shared" type="atts" mode="change">
           <attList>
             <attDef ident="meiversion" usage="rec" mode="change">
-              <defaultVal>dev</defaultVal>
+              <defaultVal>5.1-dev</defaultVal>
               <valList type="closed" mode="change">
                 <valItem ident="5.1-dev+anyStart" mode="delete"/>
                 <valItem ident="5.1-dev+basic" mode="delete"/>
--- a/customizations/mei-all.xml
+++ b/customizations/mei-all.xml
@@ -110,11 +110,11 @@
             <attDef ident="meiversion" usage="rec" mode="change">
               <defaultVal>dev</defaultVal>
               <valList type="closed" mode="change">
-                <valItem ident="dev+anyStart" mode="delete"/>
-                <valItem ident="dev+basic" mode="delete"/>
-                <valItem ident="dev+CMN" mode="delete"/>
-                <valItem ident="dev+Mensural" mode="delete"/>
-                <valItem ident="dev+Neumes" mode="delete"/>
+                <valItem ident="5.1-dev+anyStart" mode="delete"/>
+                <valItem ident="5.1-dev+basic" mode="delete"/>
+                <valItem ident="5.1-dev+CMN" mode="delete"/>
+                <valItem ident="5.1-dev+Mensural" mode="delete"/>
+                <valItem ident="5.1-dev+Neumes" mode="delete"/>
             </valList>
             </attDef>
           </attList>
--- a/customizations/mei-all.xml
+++ b/customizations/mei-all.xml
@@ -108,13 +108,13 @@
         <classSpec ident="att.meiVersion" module="MEI.shared" type="atts" mode="change">
           <attList>
             <attDef ident="meiversion" usage="rec" mode="change">
-              <defaultVal>5.0</defaultVal>
+              <defaultVal>dev</defaultVal>
               <valList type="closed" mode="change">
-                <valItem ident="5.0+anyStart" mode="delete"/>
-                <valItem ident="5.0+basic" mode="delete"/>
-                <valItem ident="5.0+CMN" mode="delete"/>
-                <valItem ident="5.0+Mensural" mode="delete"/>
-                <valItem ident="5.0+Neumes" mode="delete"/>
+                <valItem ident="dev+anyStart" mode="delete"/>
+                <valItem ident="dev+basic" mode="delete"/>
+                <valItem ident="dev+CMN" mode="delete"/>
+                <valItem ident="dev+Mensural" mode="delete"/>
+                <valItem ident="dev+Neumes" mode="delete"/>
             </valList>
             </attDef>
           </attList>
--- a/customizations/mei-all_anyStart.xml
+++ b/customizations/mei-all_anyStart.xml
@@ -171,10 +171,10 @@
             <attDef ident="meiversion" usage="rec" mode="change">
               <defaultVal>dev+anyStart</defaultVal>
               <valList type="closed" mode="change">
-                <valItem ident="dev+basic" mode="delete"/>
-                <valItem ident="dev+CMN" mode="delete"/>
-                <valItem ident="dev+Mensural" mode="delete"/>
-                <valItem ident="dev+Neumes" mode="delete"/>
+                <valItem ident="5.1-dev+basic" mode="delete"/>
+                <valItem ident="5.1-dev+CMN" mode="delete"/>
+                <valItem ident="5.1-dev+Mensural" mode="delete"/>
+                <valItem ident="5.1-dev+Neumes" mode="delete"/>
             </valList>
             </attDef>
           </attList>
--- a/customizations/mei-all_anyStart.xml
+++ b/customizations/mei-all_anyStart.xml
@@ -163,13 +163,13 @@
           <constraintSpec ident="meiVersion.warning.anyStart" scheme="schematron" mode="add">
             <constraint>
               <sch:rule context="/mei:*">
-                <sch:report role="error" test="not(local-name(.) = ('mei', 'meiHead', 'meiCorpus', 'music')) and not(@meiversion = 'dev+anyStart')">If the root element is any other than mei, meiHead, meiCorpus, or music the value of @meiversion must be 'dev+anyStart'.</sch:report>
+                <sch:report role="error" test="not(local-name(.) = ('mei', 'meiHead', 'meiCorpus', 'music')) and not(@meiversion = '5.1-dev+anyStart')">If the root element is any other than mei, meiHead, meiCorpus, or music the value of @meiversion must be '5.1-dev+anyStart'.</sch:report>
               </sch:rule>
             </constraint>
           </constraintSpec>
           <attList>
             <attDef ident="meiversion" usage="rec" mode="change">
-              <defaultVal>dev+anyStart</defaultVal>
+              <defaultVal>5.1-dev+anyStart</defaultVal>
               <valList type="closed" mode="change">
                 <valItem ident="5.1-dev+basic" mode="delete"/>
                 <valItem ident="5.1-dev+CMN" mode="delete"/>
--- a/customizations/mei-all_anyStart.xml
+++ b/customizations/mei-all_anyStart.xml
@@ -169,7 +169,6 @@
           </constraintSpec>
           <attList>
             <attDef ident="meiversion" usage="rec" mode="change">
-              <defaultVal>5.1-dev+anyStart</defaultVal>
               <valList type="closed" mode="change">
                 <valItem ident="5.1-dev+basic" mode="delete"/>
                 <valItem ident="5.1-dev+CMN" mode="delete"/>
--- a/customizations/mei-all_anyStart.xml
+++ b/customizations/mei-all_anyStart.xml
@@ -163,18 +163,18 @@
           <constraintSpec ident="meiVersion.warning.anyStart" scheme="schematron" mode="add">
             <constraint>
               <sch:rule context="/mei:*">
-                <sch:report role="error" test="not(local-name(.) = ('mei', 'meiHead', 'meiCorpus', 'music')) and not(@meiversion = '5.0+anyStart')">If the root element is any other than mei, meiHead, meiCorpus, or music the value of @meiversion must be '5.0+anyStart'.</sch:report>
+                <sch:report role="error" test="not(local-name(.) = ('mei', 'meiHead', 'meiCorpus', 'music')) and not(@meiversion = 'dev+anyStart')">If the root element is any other than mei, meiHead, meiCorpus, or music the value of @meiversion must be 'dev+anyStart'.</sch:report>
               </sch:rule>
             </constraint>
           </constraintSpec>
           <attList>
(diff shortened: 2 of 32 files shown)


musicEnfanthen:
Reviewers: Whatever version name we'll choose, please double check if I missed some instance of the version in the repository.

lpugin:
Is the plan to have things like `@meiversion="dev+all"`? I am not fully convinced that this is a good idea. In an ideal world, they should be no MEI files using a development schema outside development sandboxes. However, we know this is not the case, and with this change we will be having MEI files with no indication which development phase they relate or related to.

source/modules/MEI.shared.xml
```diff
@@ -1891,25 +1891,25 @@
     <attList>
       <attDef ident="meiversion" usage="opt">
         <desc xml:lang="en">Specifies a generic MEI version label.</desc>
-        <defaultVal>5.0</defaultVal>
+        <defaultVal>dev</defaultVal>
         <valList type="closed">
-          <valItem ident="5.0">
-            <desc xml:lang="en">MEI 5.0</desc>
+          <valItem ident="dev">
```
```suggestion
          <valItem ident="5.1-dev">
```

source/modules/MEI.shared.xml
```diff
@@ -1891,25 +1891,25 @@
...
+            <desc xml:lang="en">Development version of MEI 6.0</desc>
           </valItem>
-          <valItem ident="5.0+anyStart">
-            <desc>MEI 5.0+anyStart</desc>
+          <valItem ident="dev+anyStart">
```
```suggestion
          <valItem ident="5.1-dev+anyStart">
```

ahankinson:
Not to be "that guy" but I thought we had agreed that the next release is 6?

musicEnfanthen:
Valid point as well. The reasoning we discussed today during the ODD meeting is the following:

The idea behind 5.1-dev would be to keep a door open for a possible interim release, but if that is not needed, we could go from 5.1-dev directly to 6.0. But with 6.0-dev it would be harder to go for a possible 5.1 release (whether it is needed or not).

What do you think?

ahankinson:
I think the likelihood that we’ll do a 5.1 release is pretty low. Even if we want to keep with SemVer (I’m not convinced it’s a good fit for what we’re doing), @pe-ro has already filed a number of PRs, many of which would necessitate a major version release when they’re merged. I don’t think it’s useful to artificially delay their acceptance after review simply because we need to make our version numbers make sense. As I’ve said before, tracking compatibility changes is really tricky, and the solution is just to agree that we don’t actually need to do it. 

To make it really simple, the next merge of develop to main can be 6. Then the next merge is 7. And the next is 8. That’s it. Do we really need anything more complicated than that? Or, put another way, has anyone really done anything with the previous point releases that they wouldn’t have done with a major release? Have point releases added any value at all? 

4.0.1 fixed a bug that was found in the schema that prevented the schema from actually working, but we didn’t actually gain anything with needing to update all the documents `meiversion` string to 4.0.1. It was all still MEI4.

And MEI 4.1 became MEI 5 once we realized that it was really hard to separate out the “breaking” changes retroactively.

rettinghaus:
We had an extensive discussion and a plethora of arguments on this topic. We can continue this indefinitely or just move forward. I'm fine with the decision made during the ODD meeting.

rettinghaus (dismissed):

ahankinson:
Am I misremembering a different conclusion at the previous meeting?

ahankinson:
> The idea behind 5.1-dev would be to keep a door open for a possible interim release, but if that is not needed, we could go from 5.1-dev directly to 6.0. But with 6.0-dev it would be harder to go for a possible 5.1 release (whether it is needed or not).

> What do you think?

OK, I've been thinking about this all weekend. @rettinghaus is right -- we do seem to keep circling the drain here. But at the same time, I honestly think that our adherence to SemVer is causing more problems than it solves because we don't really know what we're versioning.

At the same time, I recognize that some wish to have a system in place to help manage the source code. [1] 

So what about this as an alternate proposal? Rather than `{major}.{minor}` we understand the version numbers to be `{compatibility}.{updates}`. That is, the `compatibility` value gives a digit for indicating document compatibility, while we use the second digit as a source control digit to indicate updates to the source code that do not break compatibility. 

**The key here is that `@meiversion` is changed to only take the `{compatibility}` value.** Any other place (e.g., GH tags, URLs?) we can use the full version. Given that the compatibility shouldn't change from a 5.0 to a 5.1 release, this should be no problem. (And if we **do** find the changes we have put in break compatibility, that would be a clear indicator that we need to bump up the compatibility version.)

The thinking behind this is that the `@meiversion` value is the *only* interface between a version number and a tagged release. Moving to a single digit here means that we're not "breaking" compatibility artificially by only changing this value with point updates. (As a reminder: Releasing "5.1" under the current scheme, regardless of whether it contains schema changes, would necessitate changing `@meiversion` values which, in itself, would break compatibility.)

Thus, if people are producing MEI 5 documents, and nothing actually changes in the validation, then their documents wouldn't actually have to change either. They can use a MEI 5.0 schema to validate an MEI 5.1 document, since the values in the schema wouldn't change either. 5.1 would be an administrative change only.

This also bypasses the whole conversation on what constitutes a "breaking" change. Any change that means that a "compatibility 5" schema is no longer able to validate a "compatibility 5" document is a "breaking" change.

So let's look at a few scenarios:

 - The ODD2ODD refactor. This shouldn't change any compatibility, so it would just be an `{update}` level change.
 - Updating documentation. Also shouldn't change any compatibility, so it would just be an `{update}` level change.
 - Adding a new value for an attribute. This *would* break compatibility, so would necessitate a `{compatibility}` level change. 
 - Adding / removing elements or attributes. Also are compatibility-breaking changes.
 - Adding / removing child or parent elements. This is also a compatibility-breaking change.
 - Adding / removing / changing a customization. I think we need to consider customizations part of the "core" since all schemas are expressed through customizations (IOW, we don't have the "MEI All" schema without the "MEI All" customization. So modifying any values in the customization that change elements / attributes and hierarchy would break this.
 - Adding schematron rules. This is a tricky one, but I think we should consider schematron part of the validation, and changes to the schematron rules that raise new **errors** are breaking changes. New **warnings** can be considered administrative changes. (This can also be managed with a guided upgrade, e.g., we can raise a warning and say that in the next version it will be an error.)

So, To avoid a breaking change, the current value of `@meiversion` can be expanded to include `5`, `5+anyStart`, `5+basic`, etc. Since any 5.0 schema would also be a valid 5.1 schema, I don't think we need to remove the existing `5.0...` values, but we simply never increment this number. This would then get normalized in 6.

---

[1] RE: Source code. To explain my thinking further, what might be happening is a divide between thinking of MEI as "source code" vs. thinking of it as "a set of rules."  For me, the XML is simply a way of expressing a set of rules, and those rules are what MEI is. If, tomorrow, we were to do away with XML and change MEI to a JSON format, the rules for encoding MEI would not actually change. So what I'm trying to communicate is that we can look at this as versioning the rules, not the actual expression of these rules in XML / ODD / RNG / Whatever.

It may be that others are thinking of the MEI Schema as a set of source code that needs to be managed as source code. If you think of the XML as being integral to MEI, and commits to the XML necessitate these sorts of changes, then this may explain why SemVer is an important system to follow. I personally don't think this way (anymore... I did once) but I recognize that some do.

So the thinking behind my proposal above is to try and reconcile the two: That is, we have a versioning for the "rules" of MEI, and then a second level of versioning for the "expression" of those rules in source code. The source code versioning can function in all the places the source code versioning is important (git, tags, etc.) while the compatibility versioning functions in the places where it is important (indicating whether a document is "valid" to a specific set of MEI rules.)

lpugin:
I am fine with the idea to have MEI version numbers being a single number. That is, `@meiversion="x"` and not `@meiversion="x.x"` as we have now or `@meiversion="x.x.x` before.

Since we are now on version `5.0`, it would not really work to add `5` as possible version number in the next update since then `@meiversion="5"` files would not be valid with the current (5.0) schema. If we go with @ahankinson's proposal, that would really work from version `6` onwards.

For the time being, I would suggest staying with `5.1-dev`. If we have changes merged into the schema that break compatibility in the way we understood it so far (all files valid with `5.0` should also be valid with `5.1`, but not necessary the opposite) then we would move to `6-dev` and drop the second number for the version at that point only. As @musicEnfanthen we do not necessary have to have a `5.1` and we can directly move to `6`. I would say that it will be to the technical co-chairs to decide if and when that should happen, namely for them to merge or not merge PRs depending on the consequences and the desired next release.

Does this sound reasonable?

ahankinson:
@lpugin yes, that sounds reasonable. 

On the understanding that moving to a single-digit `@meiversion` now would break compatibility, I'll approve the general direction of this change. I still have my reservations that 5.1 will ever see the light of day, but I guess that's a problem for future us?

source/modules/MEI.shared.xml
```diff
@@ -1891,25 +1891,25 @@
     <attList>
       <attDef ident="meiversion" usage="opt">
         <desc xml:lang="en">Specifies a generic MEI version label.</desc>
-        <defaultVal>5.0</defaultVal>
+        <defaultVal>dev</defaultVal>
```
ahankinson: `5.1-dev`?

customizations/mei-all.xml
```diff
@@ -108,13 +108,13 @@
         <classSpec ident="att.meiVersion" module="MEI.shared" type="atts" mode="change">
           <attList>
             <attDef ident="meiversion" usage="rec" mode="change">
-              <defaultVal>5.0</defaultVal>
+              <defaultVal>dev</defaultVal>
```
ahankinson: `5.1-dev`?

ahankinson (dismissed):
I think there were a few changes skipped? I marked them but didn't set them as a suggestion because I wasn't sure if it was intentional or not?

source/modules/MEI.shared.xml
bwbohl:
```suggestion
        <defaultVal>5.1-dev</defaultVal>
```

customizations/mei-all.xml
bwbohl:
```suggestion
              <defaultVal>5.1-dev</defaultVal>
```

customizations/mei-all_anyStart.xml
```diff
@@ -163,18 +163,18 @@
...
               <sch:rule context="/mei:*">
-                <sch:report role="error" test="not(local-name(.) = ('mei', 'meiHead', 'meiCorpus', 'music')) and not(@meiversion = '5.0+anyStart')">If the root element is any other than mei, meiHead, meiCorpus, or music the value of @meiversion must be '5.0+anyStart'.</sch:report>
+                <sch:report role="error" test="not(local-name(.) = ('mei', 'meiHead', 'meiCorpus', 'music')) and not(@meiversion = '5.1-dev+anyStart')">If the root element is any other than mei, meiHead, meiCorpus, or music the value of @meiversion must be '5.1-dev+anyStart'.</sch:report>
               </sch:rule>
             </constraint>
           </constraintSpec>
           <attList>
             <attDef ident="meiversion" usage="rec" mode="change">
-              <defaultVal>5.0+anyStart</defaultVal>
+              <defaultVal>5.1-dev+anyStart</defaultVal>
```
bwbohl:
```suggestion
```

bwbohl: We should remove the defaultVal for anyStart to accomodate / fix #1347

source/modules/MEI.shared.xml
bwbohl:
```suggestion
        <defaultVal>5.1-dev</defaultVal>
```

source/modules/MEI.shared.xml
```diff
@@ -1891,25 +1891,25 @@
...
         <valList type="closed">
-          <valItem ident="5.0">
-            <desc xml:lang="en">MEI 5.0</desc>
+          <valItem ident="5.1-dev">
+            <desc xml:lang="en">Development version of MEI 6.0</desc>
```
```suggestion
            <desc xml:lang="en">Development version</desc>
```

source/modules/MEI.shared.xml
```diff
@@ -1891,25 +1891,25 @@
...
           </valItem>
-          <valItem ident="5.0+anyStart">
-            <desc>MEI 5.0+anyStart</desc>
+          <valItem ident="5.1-dev+anyStart">
+            <desc>Development version of MEI 6.0+anyStart</desc>
```
```suggestion
            <desc>Development version</desc>
```

source/modules/MEI.shared.xml
```diff
@@ -1891,25 +1891,25 @@
...
           </valItem>
-          <valItem ident="5.0+basic">
-              <desc>MEI 5.0+basic</desc>
+          <valItem ident="5.1-dev+basic">
+              <desc>Development version of MEI 6.0+basic</desc>
```
```suggestion
              <desc>Development version</desc>
```

source/modules/MEI.shared.xml
```diff
@@ -1891,25 +1891,25 @@
...
           </valItem>
-          <valItem ident="5.0+CMN">
-            <desc>MEI 5.0+CMN</desc>
+          <valItem ident="5.1-dev+CMN">
+            <desc>Development version of MEI 6.0+CMN</desc>
```
```suggestion
            <desc>Development version</desc>
```

source/modules/MEI.shared.xml
```diff
@@ -1891,25 +1891,25 @@
...
           </valItem>
-          <valItem ident="5.0+Mensural">
-            <desc>MEI 5.0+Mensural</desc>
+          <valItem ident="5.1-dev+Mensural">
+            <desc>Development version of MEI 6.0+Mensural</desc>
```
```suggestion
            <desc>Development version</desc>
```

source/modules/MEI.shared.xml
```diff
@@ -1891,25 +1891,25 @@
...
           </valItem>
-          <valItem ident="5.0+Neumes">
-            <desc>MEI 5.0+Neumes</desc>
+          <valItem ident="5.1-dev+Neumes">
+            <desc>Development version of MEI 6.0+Neumes</desc>
```
```suggestion
            <desc>Development version</desc>
```

bwbohl (changes requested):

rettinghaus (approved):

bwbohl (approved):

PR #1393, opened 2023-11-30
schema: allow additive `@dur` on `<halfmRpt>`

th-we:
I think additive `@dur` makes sense for `<halfmRpt>`, because how would we otherwise encode situations like this:

grafik

This example can then be encoded like:
```xml
<measure>
  <staff>
    <layer>
      <beam>
        <note dur="8" pname="a" oct="4"/>
        <note dur="8" pname="b" oct="4"/>
        <note dur="8" pname="a" oct="4"/>
        <note dur="8" pname="g" oct="4"/>
        <note dur="8" pname="a" oct="4"/>
      </beam>
      <halfmRpt dur="2 8"/>
    </layer>
  </staff>
</measure>
<measure>
  <staff>
    <layer>
      <beam>
        <halfmRpt dur="2 8"/>
        <note dur="8" pname="a" oct="4"/>
        <note dur="8" pname="b" oct="4"/>
        <note dur="8" pname="a" oct="4"/>
        <note dur="8" pname="b" oct="4"/>
        <note dur="8" pname="a" oct="4"/>
      </beam>
    </layer>
  </staff>
</measure>
```

### DIFF
--- a/source/modules/MEI.cmn.xml
+++ b/source/modules/MEI.cmn.xml
@@ -578,7 +578,7 @@
     <desc xml:lang="en">Logical domain attributes.</desc>
     <classes>
       <memberOf key="att.event"/>
-      <memberOf key="att.duration.log"/>
+      <memberOf key="att.duration.additive"/>
     </classes>
   </classSpec>
   <classSpec ident="att.harpPedal.log" module="MEI.cmn" type="atts">


ahankinson:
I see your point, but as a counter-point, `<halfmRpt>` doesn't actually need any duration, because it is, by definition, always half a measure in any time signature.

th-we:
Yeah, maybe, but then `@dur` should be removed entirely.

ahankinson:
Looking at `<mRest>` I see that it uses the additive duration class, so I think this change makes sense.

ahankinson (approved):

rettinghaus (approved):

musicEnfanthen:
Merging since we have two aprovals.

Issue #1374, opened 2023-09-27
Add @select attribute in att.dataPointing class


pe-ro:
Currently to point from metadata elements to elements elsewhere in the file that they apply to, one must use the `@data` attribute.  I propose adding a `@select` attribute, also in the att.dataPointing class, which allows an XPath expression as its content.  This would make it possible to select the "pointed to" elements not just by their xml:id, but also by their position in the file hierarchy.  This would also make it possible to select a set of elements, even when none of the elements in the set have an id.  Such an attribute would reduce reliance on IDs and make MEI files smaller.

Also, the set of elements carrying the attributes from att.dataPointing can be expanded.  For instance, allowing `@select` on `<symbolDef>` would make it easy to associate an SVG shape with multiple elements, perhaps widely scattered throughout the encoding.  There may be other places where allowing att.dataPointing may be desirable.

I think `@select` will certainly be an integral part of any MEI style language that we might develop in the future.

ahankinson:
Questions:

 - Would `@select` be limited to XPath 1? XPath 2 is not supported in browsers, for example, so the content of `@select` would need to be targeted at the lowest common execution platform...
 - Do `@select` statements assume they start from the root? Or are they relative to the location of the element that contains the `@select` statement?
 - `@plist` already captures a list of IDs, and they don't have to be of a specific type. I guess the difference is that `@select` would get executed dynamically, and `@plist` would need to be captured at encoding time? And also `@plist` relies on IDs...
 - "Such an attribute would reduce reliance on IDs and make MEI files smaller." I didn't know this was a goal? IDs make a very useful reference system. I can see how they can make MEI files more aesthetically pleasing for humans, but between large HD storage, faster networks, and efficient compression, I'm not sure the size difference would make a dramatic difference in the larger picture. Also, we'll still need IDs for doing standoff markup, and they're a much more efficient pointing system than writing a whole XPath statement to target specific elements.
 - "...integral part of any MEI style language ..." What sort of MEI-style language are you picturing?

lpugin:
Maybe I can give some answers.

> - Would `@select` be limited to XPath 1? XPath 2 is not supported in browsers, for example, so the content of `@select` would need to be targeted at the lowest common execution platform...

This is a good point. We probably want to limit it to XPath 1.0 to start with. We can always extend to XPath 2.0, which would be not breaking backward compatibility. It would not if we change the other way around (allowing XPath 2.0 and then restricting to XPath 1.0)

[...]

PR #1398, opened 2023-12-15
schema: add att.dataSelecting and add it to symbolDef

lpugin:
This PR implements the result of the discussion in https://github.com/music-encoding/music-encoding/issues/1374

I am marking it a draft PR because I am not sure if the `datatype` for `@select` is right. I am actually pretty sure it is not. Any ideas what it should be?

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -867,7 +867,7 @@
       <attDef ident="select" usage="opt">
         <desc xml:lang="en">XPath used to select data to which an element or a property applies.</desc>
         <datatype maxOccurs="1">
-          <rng:ref name="data.URI"/>
+          <rng:data type="token"/>
         </datatype>
       </attDef>
     </attList>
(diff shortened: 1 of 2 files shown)


pe-ro:
I don't know of any regex or base data type for XPath expressions.  The XSLT 2.0 schema uses the following declaration:
```
<xs:simpleType name="expression">
  <xs:annotation>
    >xs:documentation>An XPath 2.0 expression</xs:documentation>
  </xsl:annotation>
  <xs:restriction base="xs:token">
    <xs:pattern value=".+"/>
  </xs:restriction>
```
which I think means "a sequence of characters".  For simplicity's sake, I think we can use:
```
<attDef ident="select" usage="opt">
  <desc xml:lang="en">XPath used to select data to which an 
    element or a property applies.</desc>
  <datatype>
    <rng:data type="token"/>
  </datatype>
</attDef>
```
The RNG token datatype is preferred over the string datatype because whitespace within the value of `@select` won't be significant.

bwbohl (approved):

musicEnfanthen (approved):

PR #1399, opened 2023-12-19
docs: update remarks for score element

pe-ro:
The current remarks for score are misleading. Removing the initial dependent clause will help fix the problem.


musicEnfanthen (approved):
Thanks!

rettinghaus (approved):

PR #1400, opened 2023-12-21
gh-actions: update labeler action

musicEnfanthen:
This PR updates the labeler github action that was failing recently. After update to labeler v5, there was quite some change in the action syntax that needed adjustment (cf. https://github.com/actions/labeler/tree/v5.0.0#breaking-changes-in-v5).

Some of the glob paths in the config file have been adjusted as well when they referenced non-existing files or folders.

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.repository_owner == 'music-encoding'
     steps:
-    - uses: actions/labeler@v5
+    - uses: actions/labeler@8558fd74291d67161a8a78ce36a881fa63b766a9 # ratchet:actions/labeler@v5.0.0
       with:
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
         sync-labels: true
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -16,6 +16,9 @@ on:
 
 jobs:
   label:
+    permissions:
+      contents: read
+      pull-requests: write
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-labeler.yml
```diff
@@ -16,10 +16,13 @@ on:
...
 jobs:
   label:
+    permissions:
+      contents: read
+      pull-requests: write
     runs-on: ubuntu-latest
     if: github.repository_owner == 'music-encoding'
     steps:
-    - uses: actions/labeler@v5
+    - uses: actions/labeler@8558fd74291d67161a8a78ce36a881fa63b766a9 # ratchet:actions/labeler@v5.0.0
```
rettinghaus, line 25: Why setting it to a specific commit?

rettinghaus (commented):
Looks great, thank you. One question.

.github/workflows/pr-labeler.yml
musicEnfanthen: It's just a matter of precaution and best practice recommendation. Cf. https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

> 
> Pin actions to a full length commit SHA
> 
> Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload. When selecting a SHA, you should verify it is from the action's repository and not a repository fork.

This precautionary step is recommended for third-party actions, so not actually applies to github's own actions, indeed. But on the other hand, there is no harm in following these recommendations for those in-house actions as well. Dependabot will recognize any necessary updates accordingly.

bwbohl (approved):

rettinghaus (approved):

PR #1403, opened 2024-01-18
docs: remove superfluous quotes on values

rettinghaus:
This PR removes quotes around `<val>` elements that lead to doubled ones in the guidelines.

spotted by @musicEnfanthen


musicEnfanthen (approved):
Thanks for figuring out and the fix.

bwbohl (approved):

PR #1405, opened 2024-01-18
docs: add remark about neume customization to neume element

### LINKED ISSUE #694: guidelines: documentation seems to be in conflict with valid Neume MEI file structure
Even though `<neume>` elements are not valid children of `<layer>`, the documentation [here](https://music-encoding.org/guidelines/v4/elements/neume.html) says that they are. 

This documentation is based on what it is written in the schema, so I don't understand from where it is getting this information that `<neume>` may be contained in `<layer>` because when I use the current schema for neumes, I get a validation error when using `<neume>` as a direct child of `<layer>` (see file attached).

[neume_example.mei.zip](https://github.com/music-encoding/music-encoding/files/4935142/neume_example.mei.zip)

riedde:
MEI developer workshop 2024

Fixes #694

### DIFF
--- a/source/modules/MEI.neumes.xml
+++ b/source/modules/MEI.neumes.xml
@@ -422,6 +422,9 @@
         </rng:choice>
       </rng:zeroOrMore>
     </content>
+    <remarks xml:lang="en">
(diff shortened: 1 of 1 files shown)


fujinaga (approved):
Thank you.

bwbohl (approved):

PR #1406, opened 2024-01-18
utils: improve logging output for build-guidelines-html step

musicEnfanthen:
This PR makes the logging output for the ant build-guidelines-html step a little bit more verbose. Previously, one could get the impression that the job got stuck.


bwbohl (approved):

rettinghaus (approved):
Looks great, thanks!

PR #1410, opened 2024-01-18
utils: fix element descriptions in guidelines ref

### LINKED ISSUE #1343: build: properly handle element descriptions
Currently we loose text of subelements in the [Guidelines](https://music-encoding.org/guidelines/v5/content/metadata.html#headerrespstatement), because we only handle the `text()` in `generatewebsite.xsl`.

image

riedde:
I tried `.//text()` at first to process all elements which occur, but it returned too much content (like attribute names). I guess that the script I edited is dealing with pre-processed data. After analyzing the MEI modules I figured out that adding `abbr` to the script should work (for now).
Thanks for feedback.

Fixes #1343

### DIFF
--- a/utils/guidelines_xslt/odd2html/generateWebsite.xsl
+++ b/utils/guidelines_xslt/odd2html/generateWebsite.xsl
@@ -1007,7 +1007,7 @@
                                 <xsl:apply-templates select=".//span[@class='specList-elementSpec']" mode="#current"/>
                             </div>
                             <div class="tile-subtitle text-gray">
-                                <xsl:apply-templates select="text()"/>
+                                <xsl:apply-templates select="text()|abbr/text()"/>
                             </div>
                         </div>
                     </div>
@@ -1048,7 +1048,7 @@
                                 <xsl:apply-templates select=".//a" mode="#current"/>
                             </div>
                             <div class="tile-subtitle text-gray">
-                                <xsl:apply-templates select="text()"/>
+                                <xsl:apply-templates select="text()|abbr/text()"/>
                             </div>
                         </div>
                     </div>
--- a/utils/guidelines_xslt/odd2html/generateWebsite.xsl
+++ b/utils/guidelines_xslt/odd2html/generateWebsite.xsl
@@ -1007,7 +1007,7 @@
                                 <xsl:apply-templates select=".//span[@class='specList-elementSpec']" mode="#current"/>
                             </div>
                             <div class="tile-subtitle text-gray">
-                                <xsl:apply-templates select="text()|abbr/text()"/>
+                                <xsl:apply-templates select=".//span[@class='specList-elementSpec-desc']"/>
                             </div>
                         </div>
                     </div>
--- a/utils/guidelines_xslt/odd2html/generateWebsite.xsl
+++ b/utils/guidelines_xslt/odd2html/generateWebsite.xsl
@@ -1048,7 +1048,7 @@
                                 <xsl:apply-templates select=".//a" mode="#current"/>
                             </div>
                             <div class="tile-subtitle text-gray">
-                                <xsl:apply-templates select="text()|abbr/text()"/>
+                                <xsl:apply-templates select="text()"/>
                             </div>
                         </div>
                     </div>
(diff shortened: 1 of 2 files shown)


utils/guidelines_xslt/odd2html/generateWebsite.xsl
```diff
@@ -1007,7 +1007,7 @@
                                 <xsl:apply-templates select=".//span[@class='specList-elementSpec']" mode="#current"/>
                             </div>
                             <div class="tile-subtitle text-gray">
-                                <xsl:apply-templates select="text()"/>
+                                <xsl:apply-templates select="text()|abbr/text()"/>
```
musicEnfanthen: Thank you. This will work, but might need further adjustments if there other elements not covered by this in the future.

Proposing a slightly cleaner solution by adding a class to the text content, so we can select it directly via xslt:

```suggestion
                                <xsl:apply-templates select=".//span[@class='specList-elementSpec-desc']"/>
```

Then [guidelines.xsl L200-L203](https://github.com/music-encoding/music-encoding/blob/3385e049e20b99c5d4fdfe28914d23b8d412c6e9/utils/guidelines_xslt/odd2html/guidelines.xsl#L200-L203) need to be adapted like this (add a span element with class around the `tei:desc/node()` :

```xml
                <xsl:when test="not($specDesc/@atts)">
                    <span class="specList-{local-name($spec)}"><a class="{tools:getLinkClasses($key)}" href="#{$key}"><xsl:value-of select="$key"/></a></span>
                    <span class="specList-{local-name($spec)}-desc"><xsl:apply-templates select="$spec/tei:desc/node()" mode="#current"/></span>
                </xsl:when>
```

This has the following effect: Before the change, the generated HTML looked like this:

```html
<li class="specDesc">
    <span class="specList-elementSpec">
        <a class="link_odd link_odd_elementSpec" href="#analytic">analytic</a>
    </span>Contains bibliographic elements describing an item (<abbr>e.g.</abbr>, an article or poem) published within a monograph or journal and not as an independent publication.
</li>
```

Now it comes like this:

```html
<li class="specDesc">
    <span class="specList-elementSpec">
        <a class="link_odd link_odd_elementSpec" href="#analytic">analytic</a>
    </span>
    <span class="specList-elementSpec-desc">Contains bibliographic elements describing an item (<abbr>e.g.</abbr>, an article or poem) published within a monograph or journal and not as an independent publication.</span>
</li>
```

riedde: Thank you @musicEnfanthen for the background! I'll work on that :)

bwbohl: alternatively we could introduce a default template for all elements returning their text content? Something in the flavour of:

```xml
<xsl:template match ="*" mode="guidelines">
  <xsl:value-of select="."/>
</xsl:template>
  ```
  
  This should be at the very top of the stylesheet in order to have lower priority than any other template, so it will be only applied if the other templates don’t match.

Comment (bwbohl) [reply utils/guidelines_xslt/odd2html/generateWebsite.xsl]: I have to add to that comment a bit…
As we're working in a `mode` the XSLT default templates don’t apply here so reworking those to `mode="#any"`or `mode="#default guidelines"` would be the way to go.

Comment (musicEnfanthen) [reply utils/guidelines_xslt/odd2html/generateWebsite.xsl]: The problem is/was that without a specific class in the (intermediary) HTML source the "text content" of the node() was more than just the content we needed:

```
<li class="specDesc">
    <span class="specList-elementSpec">
        <a class="link_odd link_odd_elementSpec" href="#analytic">analytic</a>
    </span>Contains bibliographic elements describing an item (<abbr>e.g.</abbr>, an article or poem) published within a monograph or journal and not as an independent publication.
</li>
```

So it always showed the "analytic" text content of the link as well. You would need something like: Give me the text content of li.specDesc, but ignore the text content coming from span.specList-elementSpec and its children. 

Introducing another global template could have side effects on other stuff we are not aware of?

musicEnfanthen (approved):
Thanks

bwbohl (approved):

rettinghaus:
Great! Thank you @riedde! 

image

PR #1417, opened 2024-01-19
guidelines: tag composer names in Guidelines

rettinghaus:
Small PR to tag mentioned composers with `name` and add a link to Wikipedia.

to be merged after #914


bwbohl (approved):

musicEnfanthen:
Thanks for the tagging. Could be helpful for some kind of register in the future.

At some point, we should discuss if we really want to link to Wikipedia or to any other authority identifier.

musicEnfanthen (approved):

PR #1164, opened 2023-05-26
pureODD: change rng:empty to tei:empty

rettinghaus:
In light of #1161 this now makes perhaps more sense. Supersedes #1050. 

Tiny change to move the `empty` element from rng into tei namespace.

PR #1421, opened 2024-01-20
schema: change rng:empty to tei:empty

rettinghaus:
Another try to make a small step towards pureODD. 

Doesn't affect anything, just a small namespace clean-up in our ODDs.
See #1050 and #1164

### DIFF
--- a/customizations/mei-Mensural.xml
+++ b/customizations/mei-Mensural.xml
@@ -131,7 +131,7 @@
             <memberOf key="model.eventLike.mensural"/>
           </classes>
           <content>
-            <rng:empty/>
+            <empty/>
           </content>
           <remarks>
             <p>This element provides an alternative to the <att>dots</att> attribute on <gi scheme="MEI"
--- a/customizations/mei-basic.xml
+++ b/customizations/mei-basic.xml
@@ -897,7 +897,7 @@
                         <memberOf key="att.placement" mode="add"/>
                     </classes>
                     <content autoPrefix="true">
-                        <rng:empty/>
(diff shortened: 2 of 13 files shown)


bwbohl (approved):

rettinghaus:
@musicEnfanthen?

musicEnfanthen:
Thank you @rettinghaus . In general, I am inclined to leave pureODD transformation for a time after a potential 5.1 release. 

Since your proposed changes are very focused in scope, they do not seem to do any harm, on the other hand. 

Just to double-check: Could this change lead to a conflicting situation for the next release where we end up with some parts of the schema not "cooperating" very well?

rettinghaus:
@musicEnfanthen No, there shouldn't be any problem at all. What were you thinking of specifically? Could you please elaborate a bit?

bwbohl:
> Thank you @rettinghaus . In general, I am inclined to leave pureODD transformation for a time after a potential 5.1 release.
> 
> Since your proposed changes are very focused in scope, they do not seem to do any harm, on the other hand.
> 
> Just to double-check: Could this change lead to a conflicting situation for the next release where we end up with some parts of the schema not "cooperating" very well?

@musicEnfanthen this change is at the very detail level of the ODD and since an empty element is just –empty– it does not make any difference whether emptiness is defined with `rng:empty` or `tei:empty`.

musicEnfanthen:
Perfect, thank you @rettinghaus and @bwbohl !

musicEnfanthen (approved):

Issue #1120, opened 2023-05-22
build: clean up "Source" folder


ahankinson:
The `source` folder currently contains a mixture of MEI source code and other assets for building the Guidelines (css, js, fonts, etc.) 

Ideally, the assets for the guidelines would be separated into a separate folder so that the MEI module source is available in its own location.

rettinghaus:
Why shouldn't these assets be seen as part of the _source_?

Proposal:

* rename `mei-source-xml` into `guidelines.xml`
* rename `docs` into `Guidelines`

like in the [TEI repo](https://github.com/TEIC/TEI/tree/dev/P5/Source). `mei-source` is very generic and misleading, as the customizations are located elsewhere.

musicEnfanthen:
Pinging @music-encoding/technical-team for thoughts.

bwbohl:
> The `source` folder currently contains a mixture of MEI source code and other assets for building the Guidelines (css, js, fonts, etc.)
> 
> Ideally, the assets for the guidelines would be separated into a separate folder so that the MEI module source is available in its own location.

@ahankinson, it’s been a while since you raised this issue; nevertheless, I’m currently working on refactoring build-related code. Code reorganisation, of course, would be an issue when doing so, consequently it should be addressed first. Can you explain a bit more about your original intention behind raising this issue?

PR #1423, opened 2024-01-20
repo: remove outdated lists

rettinghaus:
This PR removes the files with the lists of used and unused images, now outdated.

relates to #1120


pe-ro:
I assume that it's the lists that are outdated, not the images themselves.  I further assume that lists like this can be dynamically created.  If these assumptions are correct, then remove the lists.

musicEnfanthen:
Apologies for the dumb question: What were these lists used for?

rettinghaus:
@musicEnfanthen they were introduced with 901ffde0809ff79262f564484c0ac4542284ed7e on moving back to TEI for the Guidelines (see #739).

bwbohl (approved):

musicEnfanthen (approved):

PR #1425, opened 2024-01-21
validation: avoid shadowed schematron context in mei-source.sch

musicEnfanthen:
This Pull Request decomposes the bundled rules in mei-source.sch into distinct patterns since Schematron doesn't seem to like this kind of shadowing and suppresses all but the first rule sharing the same context within a single pattern:

```
WARNING: Rule for context "XY" shadowed by preceeding rule
```


rettinghaus (approved):
Looks good, thanks!

rettinghaus:
pinging @bwbohl

bwbohl (approved):

PR #1435, opened 2024-02-28
Develop agents fix Updates

riedde:
This is a PR for revising the code after a complex merge @bwbohl and I did.
closes #1434 

TODO:

- [x] update documentation (text) done at https://github.com/music-encoding/music-encoding/pull/1435/commits/4b38495463b67d40df9f6e36f7029dd6a7860bdc
- [x] update documentation (examples) done at https://github.com/music-encoding/music-encoding/pull/1435/commits/708cc66d5b558c15f53db100a953e16a93f35329


riedde:
For documentation purpose concerning https://github.com/music-encoding/music-encoding/pull/1435/commits/06ed7e324e0662ff0155108ad09178c8db3742fe

During the March-meeting of @music-encoding/ig-metadata it was noticed that deleting the `<editor>` element will have multiple and mostly bad side-effects. (starting by language caused issues like "editor" in german is not the same than "editor" in english, or is an editor a creator or contributor? and so on...)
The result of the discussion was that keeping the editor-element would be best.

riedde:
In todays meeting of @music-encoding/ig-metadata we figured out that the elements `<funder>` and `<sponsor>` should be kept also because they do not fit into the `<creator>/<contributor>` model at all. (Resored at https://github.com/music-encoding/music-encoding/pull/1435/commits/2e31a640801e70d1addc93a809bfc41c00a5c401)

PR #1440, opened 2024-03-20
schema: remove keySig and meterSig from scoreDef

### LINKED ISSUE #1439: keySig should not be a child of scoreDef
`keySig` should be disallowed as child on `scoreDef`, as explicitly encoded `keyAccid` would then not make sense within multi staff encodings with different clefs.

Also it is quite weird, that keySig only is allowed before `pgHead`, so if feels decoupled from all the `staffDef` elements.

rettinghaus:
closes #1439

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -7627,12 +7627,6 @@
       <rng:optional>
         <rng:ref name="model.symbolTableLike"/>
       </rng:optional>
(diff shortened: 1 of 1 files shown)


lpugin (commented):
This hasn't been really discussed. Since we have corresponding attributes in `scoreDef` it isnt' really clear why the elements should be removed

Issue #1331, opened 2023-09-01
Inconsistency of data.STAFFLOC usage


ahankinson:
The documentation on [`data.STAFFLOC`](https://music-encoding.org/guidelines/dev/data-types/data.STAFFLOC.html) says:

> Staff location. The value 0 indicates the bottom line of the current staff; positive values are used for positions above the bottom line and negative values for the positions below. For example, in treble clef, 1 = F4, 2 = G4, 3 = A4, etc. and -1 = D4, -2 = C4, and so on.

Implied here, of course, is that "0" = E4, or the bottom line of the staff.

However, looking at how it is described in [`att.barLine.vis/@place`](https://music-encoding.org/guidelines/dev/attribute-classes/att.barLine.vis.html):

> The location may include staff lines, the spaces between the lines, and the spaces directly above and below the staff. The value ranges between 0 (just below the staff) to 2 * number of staff lines (directly above the staff). For example, on a 5-line staff the lines would be numbered 1, 3, 5, 7, and 9 while the spaces would be numbered 0, 2, 4, 6, 8, and 10. So, a value of 9 puts the bar line through the top line of the staff.

This doesn't match. The lines, according to `data.STAFFLOC` should be 0, 2, 4, 6, 8, while the spaces would be numbered 1, 3, 5, 7, and 9.

However, in other places we definitely treat the bottom line as 1. (See [`data.CLEFLINE`](https://music-encoding.org/guidelines/dev/data-types/data.CLEFLINE.html)) 

So, I think we should make this consistent. My opinion is that we should be 1-based on our lines, so `data.STAFFLOC` should change to 1-based counting. (i.e., making 1 = E4 in treble clef, so 0? should be D4 (I guess? Or should we skip zero and go straight from +1 to -1?)

 - At the very least, the docs on `data.STAFFLOC` should align with the docs on the places where it is used.
 - If there are places where we are inconsistent on whether the bottom line is 0 or 1, we should decide on one or the other.
 - It should mention that this counts the MEI Virtual Units (1/2 lines).

pe-ro:
I don't know for sure, but I believe that once upon a time, `att.barLine.vis/@place` used a different datatype  (as did all the barline-related attributes); hence, the different textual explanation. Somewhere along the way, however, `att.barLine.vis/@place` acquired the data.STAFFLOC values.  I agree that harmonizing on a 1-based counting of the lines is the best path.

ahankinson:
[...]

PR #1445, opened 2024-04-02
customizations: bring back some visual attributes to staffDef in MEI Basic

### LINKED ISSUE #1444: MEI Basic: staves are not scalable
The `@scale` attribute on `staffDef` is not available in MEI Basic, which is an essential part of music engraving since the 19th century.

In piano scores of chamber music all parts except for the piano part are printed smaller: 

image

image

image

image

rettinghaus:
This PR brings back some attributes from `att.staffDef.vis`: 

* `@scale`
* ~~`@spacing`~~
* `@lines.color`
* `@lines.visible`

The first one and the last one I see as mandatory, the others are definitely nice to have. 

Closes #1444

### DIFF
--- a/customizations/mei-basic.xml
+++ b/customizations/mei-basic.xml
@@ -867,7 +867,7 @@
                 </classSpec>
                 
                 
-                <classSpec ident="att.staffDef.vis" module="MEI.visual" type="atts" mode="change">
+                <classSpec ident="att.staffDef.vis" module="MEI.visual" type="atts" mode="replace">
                     <desc xml:lang="en">Visual domain attributes for staffDef.</desc>
                     <classes>
                       <!--<memberOf key="att.barring"/>
--- a/customizations/mei-basic.xml
+++ b/customizations/mei-basic.xml
@@ -396,7 +396,7 @@
                 <classSpec type="atts" ident="att.rdg.vis" module="MEI.visual" mode="delete"/>
                 <classSpec type="atts" ident="att.score.vis" module="MEI.visual" mode="delete"/>
                 <classSpec type="atts" ident="att.scoreDef.vis" module="MEI.visual" mode="delete"/>
-                <classSpec type="atts" ident="att.staffDef.vis" module="MEI.visual" mode="delete"/>
+                <!--<classSpec type="atts" ident="att.staffDef.vis" module="MEI.visual" mode="delete"/>-->
                 <classSpec type="atts" ident="att.syllable.vis" module="MEI.visual" mode="delete"/>
 
                 <!--<classSpec type="atts" ident="att.arpeg.vis" module="MEI.visual" mode="change">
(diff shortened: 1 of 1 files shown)


rettinghaus:
pinging @lpugin

kepper:
I can see a need for `@scale` and `@spacing` in MEI Basic. `@lines.color` and `@lines.visible` seem pretty out of scope to me, though. However, I would like to see a strategic discussion on where to go with MEI Basic before creeping in more and more things. We all agree that full MEI has wonderful features, but they shouldn't be reflected into Basic to not kill the original idea behind it. Of course, just my random 2c, though ;-)

craigsapp:
I am wondering what the scope (use) of MEI Basic is (since I only use/need the full MEI), and whatever the scope is will determine what features should be included.

"MEI Basic" as an isolated name would imply simple encodings (which in turn would imply for educational use or training people to work in MEI), in which case `@lines.color` does not seem simple.

If MEI Basic is intended as a data interchange format (i.e., the MusicXML-like parts of MEI), there could be two different focuses: either as a union or intersection of various graphical music editors' capabilities.   MusicXML is a union of features (and transferring between various notation programs is not always a smooth process).

If MEI Basic is an intersection of music editors, then all target editors would be able to color staff lines in order to include `@lines.color`.

But if MEI Basic is a union of music editors (meaning some can color staff lines but others cannot), then `@lines.color` should be included.

And presumably the main list of graphical editors would be Musescore, Finale, Sibelius and Dorico.

> MEI Basic before creeping in more and more things. We all agree that full MEI has wonderful features, but they shouldn't be reflected into Basic to not kill the original idea behind it.

That seems to imply the intersection model rather than union model.   It would be useful to have some metric to decide what should be allowed or not (such as the features is available in a sample set of graphical music editors).

(just my 2 hemitetartemorion, of course ;-)

rettinghaus:
@kepper Without `@lines.visible` it is not possible to properly encode rhythmic examples without a staff, and that is imho a very *basic* feature. `@lines.color` is in there, because `@color` is basically available everywhere else.

kepper:
My understanding was always to use `@lines="0"` for this, but I just noticed that this only allows positive integers. I believe that this is the better fix – having to have a line which then needs to be made invisible again feels strange. Admittedly, we'd have to check for side effects with other attributes (thinking about `@loc`…), but that seems better than claiming the presence of an invisible thing. 
When it comes to `@color`/ `@lines.color`, I wonder if those need to be available in MEI Basic at all. Is this really actively used / necessary here?

ahankinson:
> but that seems better than claiming the presence of an invisible thing.

An argument could be made that the notes are on a line (they are aligned with each other), you just can't see it...

lpugin:
I think allowing `@lines="0"` would make sense, and I don't think this would have any side effects with `@loc`. We currently already have in `@loc` values that are outside the boundary of `@lines`, including negative ones.

I know that allowing `0` was also discussed for the German tablature module, but since the discussion went back and forth I am not sure if this is currently still on the wish list or not. Pinging @musicog and @DILewis

rettinghaus:
As discussed [elsewhere](https://github.com/music-encoding/music-encoding/issues/1331#issuecomment-1703807450) the staff lines are building a reference system for putting note heads et al. When they are completely missing, where would be the origin. So -1 for `@lines="0"` from my side.
Also note that `@loc` currently is not available in MEI Basic.

ahankinson:
Yeah, I'm with @rettinghaus -- I'll also note that `@lines` is in the `att.staffDef.log` attribute group, so it's a structural attribute, not a visual one. (`@lines.visible` is in `att.staffDef.vis`).

musicEnfanthen:
@ahankinson  @lpugin @kepper do we have consensus on the proposed changes? If so, please approve. ;)

ahankinson:
Has the customization been tested?

The linked issue that it closes is only a very specific request for `@scale`, and it lacks any indication of motive... "Why" is having `@scale` "essential" for notation engraving. "Explain it to me like I'm five" as @pe-ro would say... 

In general it's OK to propose to add things to MEI Basic, but I think the bar for acceptance should be slightly higher so that we don't just end up with all of MEI in Basic... This means not simply tacking on "nice-to-haves" in PRs...

lpugin:
I am actually fine to have `@lines.visible`.

I am not convinced by introducing `@spacing` without additional thoughts about the implications of introducing such layout indications in MEI Basic. We deliberately kept `scoreDef` and `staffDef` as light way as possible. If we start cherry picking attributes, we will quickly end up with something not very helpful any more.

rettinghaus:
@lpugin is right. `@spacing` should be brought back whenever we make a general decision on page layout option for MEI Basic.

musicEnfanthen:
@lpugin @ahankinson @kepper Any further objections concerning this PR?

bwbohl (approved):

lpugin:
Not from my side

rettinghaus:
@lpugin then please approve 😉

lpugin (approved):

PR #1455, opened 2024-04-24
schema: make repeatMark part of att.color

### LINKED ISSUE #1454: Not possible to color repeatMark
From the description of `repeatMark` it is explicitly allowed to use it without further content, but there is no possibility to color the mark without an additional `rend` child holding a `symbol`. 
The `@color` attribute should be available on `repeatMark`.

rettinghaus:
closes #1454


bwbohl:
+1, as the remarks in the spec, also indicate that it is allowed to provide an empty element with `@func`; in those cases, this is the only way to supply a colour.

bwbohl (approved):

rettinghaus:
Closing in favor of #1457.

PR #1457, opened 2024-04-25
schema: make all members of att.controlEvent also part of att.color

### LINKED ISSUE #1086: dynam not a member of att.color
I noticed that `<dynam>` has no `@color`. Any reason for not adding it?

### LINKED ISSUE #1454: Not possible to color repeatMark

rettinghaus:
As discussed in #1086, all control elements should have `@color` available.

This affects 

- `attacca`
- `dir`
- `dynam`
- `f`
- `fing`
- `fingGrp`
- `harm`
- `metaMark`
- `repeatMark`

Closes #1086.
Closes #1454.

### DIFF
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -137,6 +137,7 @@
   <classSpec ident="att.attacca.vis" module="MEI.visual" type="atts">
     <desc xml:lang="en">Visual domain attributes.</desc>
     <classes>
+      <memberOf key="att.color"/>
       <memberOf key="att.extender"/>
       <memberOf key="att.placementRelStaff"/>
       <memberOf key="att.verticalGroup"/>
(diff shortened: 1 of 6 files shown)


bwbohl (dismissed):

source/modules/MEI.visual.xml
```diff
@@ -480,6 +481,7 @@
     <desc xml:lang="en">Visual domain attributes.</desc>
     <classes>
       <memberOf key="att.extender"/>
+      <memberOf key="att.color"/>
```
```suggestion
      <memberOf key="att.color"/>
      <memberOf key="att.extender"/>
```

musicEnfanthen:
Thank you @rettinghaus. Just a little sorting propsal above. 

While you are at it, maybe you can have a look at the order of `att.trill.vis` memberships, too? It is not changed in this PR because it already had the `att.color` membership, but as it is also one of the members of `att.controlevent`, it may be reasonable to fix it along the way.

I also realized that `<attaca>`, `<beamSpan>`, and `<tupletSpan>` are the only controlevents without direct `att.color` class now. While the two span elements inherit it via their membership in `att.beam.vis` and `att.tuplet.vis`, it seems that `<attaca>` is simply missing it?

rettinghaus:
@musicEnfanthen Thanks for pointing out `attacca`. I made it also member of `att.color` and integrated now #1455, so everything is in one place.

Additionally I did some more spring cleaning with the sorting.

musicEnfanthen (approved):

lpugin (approved):

PR #1458, opened 2024-04-25
build: drop os and use pathlib in generate-examples.py

### LINKED ISSUE #1213: build: generate images – python  drop os and use pathlib
Could we drop os and instead use 
```suggestion
from pathlib import Path
```

Makes everything easier and more readable.
We don't use `with open` and don't have to go over all the files if we use `glob`.

_Originally posted by @rettinghaus in https://github.com/music-encoding/music-encoding/pull/1184#discussion_r1211217227_

rettinghaus:
This PR updates the Python script used for example generation by replacing the os module with the more modern path lib module.

closes #1213

### DIFF
--- a/utils/examples/generate-examples.py
+++ b/utils/examples/generate-examples.py
@@ -73,7 +73,7 @@ if __name__ == "__main__":
         mei_example: str = mei_file.read_text()
 
         if not tk.loadData(mei_example):
-            log.error("Failed to load %s", mei_file)
+            log.error("Failed to load %s", mei_file.name)
             success = False
             continue
 
--- a/utils/examples/generate-examples.py
+++ b/utils/examples/generate-examples.py
@@ -1,12 +1,12 @@
 import argparse
-import os
 import sys
 import pprint
 import logging
+from pathlib import Path
 
 import verovio
 
-IMAGES_PATH = "../../build/assets/images/GeneratedImages"
+IMAGES_PATH = Path("../../build/assets/images/GeneratedImages")
 
 VRV_OPTIONS: dict = {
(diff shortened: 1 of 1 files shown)


utils/examples/generate-examples.py
```diff
@@ -52,32 +52,25 @@
     logging.basicConfig(format="[%(asctime)s] [%(levelname)8s] %(message)s (%(filename)s:%(lineno)s)",
                         level=level)
     log = logging.getLogger(__name__)
-    log.info("Running at logging level %s", level_msg)
+    log.info(f"Running at logging level {level_msg}")
```
ahankinson: Using f-strings always computes the value, while printf-style defers computing it if the message wouldn't be printed. 

See: https://docs.python.org/3/howto/logging.html#optimization

utils/examples/generate-examples.py
```diff
@@ -62,22 +62,15 @@
...
             continue
 
         # Download the MEI file from the given url
-        mei_example: str = ""
-        with open(mei_file) as f:
-            mei_example = f.read()
+        mei_example: str = mei_file.read_text()
 
         if not tk.loadData(mei_example):
             log.error("Failed to load %s", mei_file)
```
musicEnfanthen:
```suggestion
            log.error("Failed to load %s", mei_file.name)
```

musicEnfanthen (commented):
Just a small addition, otherwise looks good to me

musicEnfanthen (approved):

bwbohl (approved):

musicEnfanthen:
Thank you!

PR #1456, opened 2024-04-25
build: update Saxon and Xerces to latest stable

rettinghaus:
This PR updates `build.xml` to the latest versions of all dependencies, fixes some spelling errors and updates a property name to play nicely with the TEI Stylesheets.

See https://github.com/Saxonica/Saxon-HE

Issue #1117, opened 2023-05-19
docs: rework workflow description build with oXygen to use MEI‘s TEI Stylesheets and build.xml


bwbohl:

o-sapov:
To make it more comfortable to use oXygen, it would be advisable to provide an oXygen project file (`.xpr`) containing the necessary scenarios and settings.
	
Alternatively, if the community would rather not propagate proprietary third-party software, it is possible to place oXygen-related items outside the repository, for example by packaging them as an add-on.

bwbohl:
I started to revive https://github.com/music-encoding/oXygen-MEI-addon may it would be a good idea to integrate this there

PR #1460, opened 2024-04-29
build: fetch stylesheets only for build

### LINKED ISSUE #1459: Checkout TEI Stylesheets during build
I was wondering what the benefits are for having the TEI Stylesheets as a submodule in the repo and if it would be possible to download them with the ANT task. Right now the tagged version of the stylesheets and those of all other needed components are handled differently. With moving the process for checking out the stylesheets to the ANT init would bring all dependencies into a single file. 
Another benefit would be to clean up the repo a bit further (#873).

rettinghaus:
This PR removes the Stylesheets submodule and integrates fetching an official release from GitHub into the ant init process. 
There are also some updates and corrections to the messages.

Closes #1459

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -74,8 +74,8 @@
     <property name="dir.lib.saxon" value="${dir.lib}/saxon"/>
     <property name="dir.lib.stylesheets" value="${dir.lib}/stylesheets"/>
     <property name="dir.lib.xerces" value="${dir.lib}/xerces"/>
-    <property name="saxon.version" value="SaxonHE12-4"/>
-    <property name="saxon.jar.file" value="saxon-he-12.4.jar"/>
+    <property name="saxon.version" value="SaxonHE12-5"/>
+    <property name="saxon.jar.file" value="saxon-he-12.5.jar"/>
     <property name="stylesheets.version" value="7.57.0"/>
     <property name="xerces.version" value="26.1.0.1"/>
     <property name="xerces.jar.file" value="oxygen-patched-xerces-${xerces.version}.jar"/>
--- a/build.xml
+++ b/build.xml
@@ -168,7 +168,7 @@
         <echo if:set="saxon-available" message="saxon available: ${saxon-available}"/>
         <echo unless:set="saxon-available" message="saxon available: false"/>
         <!-- check if Stylesheets are available -->
-        <fileset id="version.file" dir="${dir.lib.stylesheets}">
+        <fileset id="stylesheets.version.file" dir="${dir.lib.stylesheets}">
             <patternset>
                 <include name="VERSION"/>
             </patternset>
--- a/build.xml
+++ b/build.xml
@@ -168,8 +168,14 @@
         <echo if:set="saxon-available" message="saxon available: ${saxon-available}"/>
         <echo unless:set="saxon-available" message="saxon available: false"/>
         <!-- check if Stylesheets are available -->
+        <fileset id="version.file" dir="${dir.lib.stylesheets}">
+            <patternset>
+                <include name="VERSION"/>
+            </patternset>
+            <contains text="${stylesheets.version}" />
+        </fileset>
         <condition property="stylesheets-available">
-            <available file="${dir.lib.stylesheets}"/>
+            <resourcecount when="equal" count="1" refid="version.file"/>
         </condition>
         <echo if:set="stylesheets-available" message="stylesheets available: ${stylesheets-available}"/>
         <echo unless:set="stylesheets-available" message="stylesheets available: false"/>
--- a/build.xml
+++ b/build.xml
@@ -76,7 +76,7 @@
     <property name="dir.lib.xerces" value="${dir.lib}/xerces"/>
     <property name="saxon.version" value="SaxonHE12-4"/>
     <property name="saxon.jar.file" value="saxon-he-12.4.jar"/>
-    <property name="stylesheets.version" value="7.56.0"/>
+    <property name="stylesheets.version" value="7.57.0"/>
     <property name="xerces.version" value="26.1.0.1"/>
     <property name="xerces.jar.file" value="oxygen-patched-xerces-${xerces.version}.jar"/>
     <property name="oxygen.basic.utilites.jar.file" value="oxygen-basic-utilities-${xerces.version}.jar"/>
--- a/build.xml
+++ b/build.xml
@@ -72,9 +72,11 @@
     <property name="dir.dist.schemata" location="${dir.dist}/schemata/"/>
     <property name="dir.lib" value="lib"/>
     <property name="dir.lib.saxon" value="${dir.lib}/saxon"/>
+    <property name="dir.lib.stylesheets" value="${dir.lib}/stylesheets"/>
     <property name="dir.lib.xerces" value="${dir.lib}/xerces"/>
     <property name="saxon.download.version" value="SaxonHE12-4"/>
     <property name="saxon.jar.file" value="saxon-he-12.4.jar"/>
+    <property name="stylesheets.download.version" value="7.56.0"/>
     <property name="xerces.version" value="26.1.0.1"/>
     <property name="xerces.jar.file" value="oxygen-patched-xerces-${xerces.version}.jar"/>
     <property name="oxygen.basic.utilites.jar.file" value="oxygen-basic-utilities-${xerces.version}.jar"/>
@@ -114,7 +116,7 @@
     </target>
 
     <target name="echo-mei">
-      <loadfile srcfile="source/assets/images/meilogo_ascii_100.txt" property="mei"/>
+        <loadfile srcfile="source/assets/images/meilogo_ascii_100.txt" property="mei" />
         <echo message="${mei}"/>
     </target>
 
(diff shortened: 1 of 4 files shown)


rettinghaus:
@bwbohl @musicEnfanthen If we could agree on this, it would replace #1456.

bwbohl (dismissed):

build.xml
```diff
@@ -133,21 +135,21 @@
...
 
-    <target name="reset" description="Resets the build environment. Same as `clean`, but additionally deletes the `lib` directory with the Saxon and Xerces jar files.">
+    <target name="reset" description="Resets the build environment. Same as `clean`, but additionally deletes the `lib` directory with the TEI Stylesheets and the Saxon and Xerces jar files.">
         <delete dir="lib"/>
         <antcall target="clean"/>
     </target>
     
-    <target name="init" description="Initializes the build environment, i.e., checks if Saxon and Xerces are available, and if not, downloads jar files for Saxon, Xerces and adds them to the lib folder. Checks if Prince is available.">
-        <!-- check if Xerces is available -->
+    <target name="init" description="Initializes the build environment, i.e., checks if TEI Stylesheets, Saxon and Xerces are available, and if not, it downloads everything into the `lib` folder. Checks if prince is available.">
```
```suggestion
    <target name="init" description="Initializes the build environment, i.e., checks if TEI Stylesheets, Saxon and Xerces are available, and if not, it downloads everything into the `lib` folder. Checks if Prince is available.">
```

musicEnfanthen: Please change the description of the target in the table with the targets in BUILD_COMMANDLINE accordingly

build.xml
```diff
@@ -133,21 +135,21 @@
...
+    <target name="get-local-git-branch" description="Retrieves the current git branch via command  line.">
         <exec executable="git" outputproperty="github.branch.local">
             <arg value="branch"/>
             <arg value="--show-current"/>
         </exec>
         <echo>Local github.branch: ${github.branch.local}</echo>
     </target>
 
-    <target name="reset" description="Resets the build environment. Same as `clean`, but additionally deletes the `lib` directory with the Saxon and Xerces jar files.">
+    <target name="reset" description="Resets the build environment. Same as `clean`, but additionally deletes the `lib` directory with the TEI Stylesheets and the Saxon and Xerces jar files.">
```
musicEnfanthen, line 146: Please change the description of the target in the table with the targets in BUILD_COMMANDLINE accordingly

build.xml
```diff
@@ -165,13 +167,23 @@
         </condition>
         <echo if:set="saxon-available" message="saxon available: ${saxon-available}"/>
         <echo unless:set="saxon-available" message="saxon available: false"/>
+        <!-- check if stylesheets are available -->
```
```suggestion
        <!-- check if Stylesheets are available -->
```

build.xml
```diff
@@ -190,7 +202,9 @@
 
     <target name="saxon-download" unless="${saxon-available}">
         <mkdir dir="temp"/>
-        <get src="https://github.com/Saxonica/Saxon-HE/releases/download/${saxon.download.version}/${saxon.download.version}J.zip" dest="temp/download"/>
+        <get
+            src="https://github.com/Saxonica/Saxon-HE/releases/download/${saxon.download.version}/${saxon.download.version}J.zip"
```
musicEnfanthen: Remove line breaks here

build.xml
```diff
@@ -165,13 +167,23 @@
         </condition>
         <echo if:set="saxon-available" message="saxon available: ${saxon-available}"/>
         <echo unless:set="saxon-available" message="saxon available: false"/>
+        <!-- check if stylesheets are available -->
+        <condition property="stylesheets-available">
+            <or>
```
musicEnfanthen: I guess, since we have only one conditional expression the "or" is not needed here

build.xml
```diff
@@ -165,13 +167,23 @@
...
+                <available file="${dir.lib.stylesheets}"/>
+            </or>
+        </condition>
+        <echo if:set="stylesheets-available"
+            message="stylesheets available: ${stylesheets-available}"/>
```
musicEnfanthen: Remove line break here

build.xml
```diff
@@ -202,6 +216,30 @@
         <unzip src="temp/download" overwrite="true" dest="${dir.lib.saxon}"/>
     </target>
 
+    <target name="stylesheets-download" unless="${stylesheets-available}">
+        <mkdir dir="temp"/>
+        <get
+            src="https://github.com/TEIC/Stylesheets/releases/download/v${stylesheets.download.version}/tei-xsl-${stylesheets.download.version}.zip"
```
musicEnfanthen: remove line breaks here

build.xml
```diff
@@ -202,6 +216,30 @@
...
+            dest="temp/download"/>
+    </target>
+
+    <target name="stylesheets-prepare" depends="stylesheets-unzip" unless="${stylesheets-available}">
+        <delete dir="temp/"/>
+    </target>
+
+    <target name="stylesheets-unzip" depends="stylesheets-download"
+        unless="${stylesheets-available}">
```
musicEnfanthen: remove line break here

build.xml
```diff
@@ -72,9 +72,11 @@
     <property name="dir.dist.schemata" location="${dir.dist}/schemata/"/>
     <property name="dir.lib" value="lib"/>
     <property name="dir.lib.saxon" value="${dir.lib}/saxon"/>
+    <property name="dir.lib.stylesheets" value="${dir.lib}/stylesheets"/>
     <property name="dir.lib.xerces" value="${dir.lib}/xerces"/>
     <property name="saxon.download.version" value="SaxonHE12-4"/>
     <property name="saxon.jar.file" value="saxon-he-12.4.jar"/>
+    <property name="stylesheets.download.version" value="7.56.0"/>
```
musicEnfanthen: Please shorten this variable name to `stylesheets.version` like with Xerces. That's totally sufficient. We will need to shorten it for saxon too. (If you like you can include it here)

musicEnfanthen (dismissed):

build.xml
```diff
@@ -165,13 +167,20 @@
         </condition>
         <echo if:set="saxon-available" message="saxon available: ${saxon-available}"/>
         <echo unless:set="saxon-available" message="saxon available: false"/>
+        <!-- check if Stylesheets are available -->
+        <condition property="stylesheets-available">
+            <available file="${dir.lib.stylesheets}"/>
```
musicEnfanthen: Tested this PR locally and with Docker, everything runs fine. Great, many thanks @rettinghaus !

However, I am wondering if this check is sufficient here? At the moment, it only checks if there is a folder named stylesheets. It does not care about the content. So you could have either an empty stylesheets folder or one with an older or newer version of the stylesheets or even one with a totally different content, and the check would return true. 

So what could be the best way to check for the correct content?

PS: I see that there is a VERSION file, and we definitely need the `/odd/build-to.xml`, `/html5/build-to.xml`, `/rng/build-to.xml` files. These call the `/common/teianttasks.xml` which references various other directories and files.

musicEnfanthen: This also applies when it comes to version updates. When you have 7.56 in your lib folder, but the build.xml would have been updated to 7.57, the build script would not download the newer version.

rettinghaus: I could argue that this only was a _proof of concept_. 😉 But you're absolutely right. Thanks for pointing this out!

rettinghaus: I changed the check to check the `VERSION` file for the expected version. From my point of view should that be sufficient: if anything doesn't work you can simply run `ant reset`.

build.xml
```diff
@@ -165,13 +167,26 @@
         </condition>
         <echo if:set="saxon-available" message="saxon available: ${saxon-available}"/>
         <echo unless:set="saxon-available" message="saxon available: false"/>
+        <!-- check if Stylesheets are available -->
+        <fileset id="version.file" dir="${dir.lib.stylesheets}">
```
```suggestion
        <fileset id="stylesheets.version.file" dir="${dir.lib.stylesheets}">
```

build.xml
```diff
@@ -165,13 +167,26 @@
...
+            <patternset>
+                <include name="VERSION"/>
+            </patternset>
+            <contains text="${stylesheets.version}" />
+        </fileset>
+        <condition property="stylesheets-available">
+            <resourcecount when="equal" count="1" refid="version.file"/>
+        </condition>
```
```suggestion
        <condition property="stylesheets-available">
            <and>
                <available file="${dir.lib.stylesheets}"/>
                <resourcecount when="equal" count="1" refid="stylesheets.version.file"/>
            </and>
        </condition>
```

build.xml
musicEnfanthen: just to be concise

build.xml
musicEnfanthen: Thanks for the suggested check for the VERSION file. It failed when running `ant reset` beforehand, since it does not have a stylesheet directory then. So finally we can bring back the conditional check 🎉 

I also tested with a different VERSION number in the file. So when you have 7.55 specified in your VERSION file, it now downloads the version requested by the ant script. 👍

build.xml
musicEnfanthen: See comments below

musicEnfanthen:
A final question/thought: What happens if you build the guidelines/specs/etc. with Oxygen with the stylesheets submodule removed?

rettinghaus:
@musicEnfanthen don't have Oxygen currently. @bwbohl could you check please?

bwbohl:
Tested with different oXygen strategies. Will work but we should document how to setup a transformation scenario with the music-encoding build.xml
cf. #1117 and #1465

bwbohl (approved):

musicEnfanthen (approved):

PR #1471, opened 2024-05-23
utils: make retrieval of contributors optional

### LINKED ISSUE #1186: build: exclude updates of known contributors
Updating the contributors on build seems to take some time and, most of all, is a source for build failure if the allowed GitHub API request quota is reached. I recommend leaving this out of the regular build and instead making it available as an additional test via GitHub actions.

kepper:
passing in any value other than 'false' will retrieve the list from GitHub. This fixes #1186.


utils/guidelines_xslt/odd2html.xsl
```diff
@@ -130,6 +130,13 @@
     </xd:doc>
     <xsl:param name="basedir" select="''" as="xs:string"/>
     
+    <xd:doc>
+        <xd:desc>
+            <xd:p>Whether Contributors should be retrieved live from GitHub. Please consider that this may result in excessive API calls.</xd:p>
+        </xd:desc>
+    </xd:doc>
+    <xsl:param name="retrieve-contributors" select="'false'" as="xs:string"/>
```
rettinghaus: Why not 
```suggestion
    <xsl:param name="retrieve-contributors" select="false()" as="xs:boolean"/>
```

utils/guidelines_xslt/odd2html/functions.xsl
```diff
@@ -469,8 +469,10 @@
         
         <xsl:variable name="contributors">
             <xsl:variable name="raw.contributors" as="node()*">
-                <xsl:sequence select="tools:retrieveData($docs.repo.contributors)/child::json:array/json:map"/>
-                <xsl:sequence select="tools:retrieveData($spec.repo.contributors)/child::json:array/json:map"/>
+                <xsl:if test="$retrieve-contributors ne 'false'">
```
rettinghaus: Then this could be simplified into
```suggestion
                <xsl:if test="$retrieve-contributors">
```

rettinghaus (changes requested):
Thanks! One idea from my side.

utils/guidelines_xslt/odd2html/functions.xsl
kepper: Problem is that in my experience, it doesn't work reliably to pass in values other than strings. Particularly boolean values seem to have problems. I know it's clumsy to do it like this, but it was intentional to avoid nasty problems which are really hard to debug.

kepper: obviously, the same applies to the other suggestion above.

rettinghaus: The funny thing is, that almost every value passed evaluates to `true()`, e.g., non-empty strings, integers other than 0. If you insist in using string comparison, it should not mimic a boolean. Then better to use "no" or something similar.

kepper: I don't insist on anything, I just don't want to write broekn code that's hard to debug. I don't mind to call it "no", if that is more clear. Normally, I try to avoid boolean-like parameters completely instead, but this name was proposed already, so I did followed my normal procedure – which is to explicitly use "false", because I then have clearly expected values and legible code, which just behaves a bit oddly and looks a bit uninformed, where in fact it's just the opposite.

rettinghaus: From my perspective this _is_ broken code, but who am I to judge.

rettinghaus (dismissed):

bwbohl:
I know about the frustrations with trying to externally pass booleans into XSLTs. Nevertheless, defining the param with type `xs:boolean` should atleast fail the stylesheet when anything else than the strings `true`, `false`, `1`, or `0` are being passed. That is beacuse other strings cannot be cast to xs:boolean properly.

bwbohl:
When defined as xs:string, though, the behaviour described by @rettinghaus will apply.
I know we haven’t been strict about this in the past (and do not have to be it when drafting improvements), but we should definitely harden our XSLTs in the future.

utils/guidelines_xslt/odd2html.xsl
bwbohl: as commented in https://github.com/music-encoding/music-encoding/pull/1471#issuecomment-2128880033 this solution is preferable.

utils/guidelines_xslt/odd2html/functions.xsl
bwbohl: as commented in https://github.com/music-encoding/music-encoding/pull/1471#issuecomment-2128880033 this solution is preferable because it does not resolve some arbitrary string to true.

bwbohl:
@musicEnfanthen would you be so kind as to test it with @rettinghaus's   suggestions, thanks!

musicEnfanthen:
@bwbohl @kepper I tested locally with boolean values and the string version, and, as far as I can say, both work as expected. (It is a bit hard to test since one seems to run quite easily into an API limit rate while testing.)

Also, passing a parameter with true or false from the ant script to the XSLT works fine with both approaches. There are also some other occasions where boolean values are used in the odd2html scripts. 

So in the end, it seems, that it is mere a question of "code esthetics", but shouldn't end as a blocker for this feature.

bwbohl:
@kepper please double check this is what you have in https://github.com/music-encoding/music-encoding/compare/develop...customizationDocs

utils/guidelines_xslt/odd2html/functions.xsl
```diff
@@ -426,6 +420,16 @@
                 </div>
             </xsl:for-each>
         </section>
+    </xsl:function>
+
+    <xd:doc>
+        <xd:desc>
+            <xd:p>Generates a list of contributors for the back of the Guidelines PDF</xd:p>
+        </xd:desc>
+        <xd:return></xd:return>
```
bwbohl:
```suggestion
        <xd:return>an html:section containing the contributors</xd:return>
```

bwbohl (dismissed):

rettinghaus (approved):

bwbohl (approved):

Issue #1379, opened 2023-10-06
Allow `@vgrp` on `<harm>` and other `<dir>` derivative elements


pe-ro:
The `@vgrp` attribute allows one to group individual directive-like elements on their vertical offset without having to specify the exact amount of offset.  For example, when a set of `<dir>` elements is grouped ; that is, they all have the same value for `@vgrp`, then a renderer should place them all at the same height above/below the staff they pertain to.

The current members of att.verticalGroup are: attacca, dir, dynam, hairpin, pedal, repeatMark, and tempo.  Additional members to be added under this proposal are: harm, breath, fermata, fing, and fingGrp.

bwbohl:
I think, it's a good idea to allow `@vgrp` on those elements

lpugin:
Verovio currently aligns `harm` based on their `@n`. It would not be a problem also to support `@vgrp`.

pe-ro:
Addressed in #1472

musicEnfanthen:
Fixed in #1472. Closing.

PR #1472, opened 2024-05-23
schema: make directive-like elements member of att.verticalGrp

pe-ro:
Make bend, caesura, cpMark, metaMark, ornam, harm, fing, fingGrp, fermata, harpPedal, octave, mordent, trill, and turn members of att.vertical.Grp.

### DIFF
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -304,6 +304,7 @@
       <memberOf key="att.staffLoc"/>
       <memberOf key="att.staffLoc.pitched"/>
       <memberOf key="att.typography"/>
+      <memberOf key="att.verticalGrp"/>
       <memberOf key="att.visualOffset"/>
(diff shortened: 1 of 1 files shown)


pe-ro:
Closes #1379

musicEnfanthen:
@pe-ro You also mentioned `<breath>` in #1379, but it is not updated here? By accident?

pe-ro:
@musicEnfanthen, yep, `<breath>` was omitted by mistake. I'll make another PR that includes just `<breath>`.

musicEnfanthen:
@pe-ro Thanks for clarification. 

We could include it in this PR, so everything would be in a single place? You would just need to commit once more to the `develop-vgrp` branch. If that's not possible just let me know, and I can add it.

musicEnfanthen (approved):

rettinghaus (approved):

Issue #1451, opened 2024-04-18
schema: exclude directions in extracted parts


rettinghaus:
Currently it is possible to declare if a direction is affects specific parts or all parts with `@part` (and `@partstaff`). However it is not possible to say specifically, that something should not appear in any part, e.g. something like "a2" in the woodwind section. 

I propose to extend the allowed pattern to `(%all|%none|#[\i][\c]+)` to keep something from extracted parts.

Relates to #1450.

NB: Currently we have the same definition for this datatype twice, so it might make sense to extract it into its own datatype.

craigsapp:
I got chatgpt to tell me what `#[\i][\c]+` means after giving it hints that it is related to ODD 😉 

I am wondering about the `+`:  Does that mean that `#A` is an invalid reference ID for the `@part` (or `@partstaff`) value?   I am wondering if it should be `*` instead:

```
 (%all|%none|#[\i][\c]*)
```

And is `#:` be a valid ID reference?

Maybe adding a minus sign before the IDREF could be a way of excluding from a particular extracted part:

```
 (%all|%none|-?#[\i][\c]*)
```

In general `%none` should work for suppressing `a2`. 

There will be text such as `solo` which would be suppressed in only in the non-solo part, which could be `@part="#part1"` or `@part="-#part2".  Not particularly useful for the negation in this case since `@part="#part1"` works well.  The negation would be useful if there were more than two parts to exclude a particular part rather than including multiple parts.

pe-ro:
`#[\i][\c]*` would be fine, too.

I'm less enthusiastic about using '-' to indicate negation. Declarative markup is usually expressed affirmatively, even if that leads to verbosity.  Having an indication that says, "This applies everywhere except ..." relies on an assumption, whereas an indication that says, "Do this here and there" does not.

PR #1473, opened 2024-05-23
schema: add '%none' to `@part` in att.partIdent

lpugin:
* Relates to #1451
* Also change the regex to allow single character part names


rettinghaus (approved):
Looks great, thank you!

source/modules/MEI.shared.xml
```diff
@@ -2501,10 +2501,10 @@
...
         <desc xml:lang="en">Indicates the part in which the current feature should appear. Use '%all' when the
-          feature should occur in every part.</desc>
+          feature should occur in every part, or '%none' in none of the parts.</desc>
         <datatype maxOccurs="unbounded">
           <rng:data type="token">
-            <rng:param name="pattern">(%all|#[\i][\c]+)</rng:param>
+            <rng:param name="pattern">(%all|%none|#[\i][\c]*)</rng:param>
           </rng:data>
         </datatype>
         <constraintSpec ident="check_part_attr_all" scheme="schematron">
```
musicEnfanthen, line 2521: Do we need to update the schematron rule as well? It seems like `'%none'` cannot be mixed with other values, too, right?

musicEnfanthen: Also: As discussed in #1451, does this value need to be updated for partstaff too?

rettinghaus: @musicEnfanthen Having this on `partstaff` doesn't seem to make sense. Why would you say something appears in one part but doesn't show up there on any staff?

musicEnfanthen: Sure. Just wanted to double-check since it was mentioned in #1451.

What do you think about the schematron rule?

pe-ro:
Rather than revising the Schematron rule, I think it would be better to eliminate the rule and modify the definition of `@part` like so
```
<attribute name="part">
  <a:documentation xmlns:a="http://relaxng.org/ns/compatibility/annotations/1.0">Indicates
    the part in which the current feature should appear. Use '%all' when the feature
    should occur in every part or '%none' when the feature should not appear in any 
    part.</a:documentation>
  <choice>
    <value>%all</value>
    <value>%none</value>
    <list>
      <oneOrMore>
        <ref name="mei_data.URI"/>
      </oneOrMore>
    </list>
  </choice>
</attribute>
```
This accomplishes the same goal -- limit `@part` to either "%all", "%none", or a list of IDs attached to part elements.

A new Schematron rule can be written to make sure that `@part` points to a part element within the current instance or points to a part encoded in a separate file.

pe-ro:
Here's a first pass at the new Schematron --
```
<pattern xmlns="http://purl.oclc.org/dsdl/schematron">
    <sch:rule xmlns="http://www.tei-c.org/ns/1.0"
      xmlns:sch="http://purl.oclc.org/dsdl/schematron"
      xmlns:xi="http://www.w3.org/2001/XInclude"
      context="@part[not(matches(., '^%all$|^%none$'))]">
      <sch:assert
        test="every $i in tokenize(., '\s+') satisfies substring($i,2)=//mei:part/@xml:id or 
        matches($i, '^([a-z]+://|\.{1,2}/)')">The values in @part must either
        correspond to the @xml:id of a part element or be an external URL.</sch:assert>
  </sch:rule>
</pattern>
```

pe-ro:
`@partstaff` can/should also be rewritten like so --
```
<attribute name="partstaff">
  <a:documentation xmlns:a="http://relaxng.org/ns/compatibility/annotations/1.0"
    >Signifies the part staff on which a notated feature occurs. Use '%all' when the
    feature should occur on every staff.</a:documentation>
  <choice>
    <value>%all</value>
    <data type="token">
      <param name="pattern">\d+(-\d+)?</param>
    </data>
  </choice>
</attribute>
```
This constrains the value in `@partstaff` to "%all", an integer, or a range of integers.  Currently, there's no way to provide a non-consecutive set of integers.

rettinghaus:
Thanks @pe-ro. The proposed schematron rule will only work if you already have parts encoded. To bring it to use for _automatic part extraction_ I think it should also take `mei:staffGrp`, `mei:staffDef`, and `mei:layerDef` into account.

pe-ro:
@rettinghaus, You're correct.  The rule will only work for already-encoded parts. Rather than make the rule consider automated part extraction, I'm inclined to drop it. I don't see a simple way to capture/validate _the intention to create parts_ with certain ID values.

rettinghaus:
@lpugin would you please be so kind to update this PR as suggested by @pe-ro?

- https://github.com/music-encoding/music-encoding/pull/1473#issuecomment-2135212174
- https://github.com/music-encoding/music-encoding/pull/1473#issuecomment-2135339934
- https://github.com/music-encoding/music-encoding/pull/1473#issuecomment-2135557259

musicEnfanthen:
@lpugin Any updates on this?

musicEnfanthen:
@lpugin Is there a chance this could be updated until next week?

lpugin:
I am not sure how to translate back the suggestion into the source file. I think it will be easier for somebody else to start from scratch. Sorry for the delay.

rettinghaus:
@pe-ro I tried to model it after your suggestions, but I didn't succeed. A choice with values and a list doesn't seem to work.

pe-ro:
@rettinghaus, 
```
<choice>
  <value>%all</value>
  <data type="token">
    <param name="pattern">\d+(-\d+)?</param>
  </data>
</choice>
```
definitely works in RNG.  Do you mean that there's no Pure ODD equivalent?

rettinghaus:
Ok, it _does_ work to build the schema, but our process to build the Guidelines cannot handle this.

Issue #272, opened 2015-10-17
Cannot position and size text in front matter


ahankinson:
I can find no way to position and size text boxes in front matter. `<p>` supports `@x` and `@y`, but there may be multiple paragraphs in a box, and boxes may have width and height, which is not supported. I would think `<div>` would behave as a block-level element (like HTML) but it seems to be missing any positioning or sizing attributes.

![snap](https://cloud.githubusercontent.com/assets/163183/10559557/94924dd2-74c1-11e5-892d-481208d7832a.png)

Currently I have the following markup within `<front>`:

``` xml
<front xml:id="m-40">
  <pb xml:id="m-23" n="1"/>
    <div xml:id="m-28">
      <p xml:id="m-29">Second livre
        <lb xml:id="m-30"/>de chansons à quatre 
        <lb xml:id="m-31"/>Nicolas Duchemin
        <lb xml:id="m-32"/>1549
      </p>
    </div>
  <div xml:id="m-24">
    <p xml:id="m-25">Qu’est il besoing chercher toutes les nuictz</p>
  </div>
  <div xml:id="m-26">
    <p xml:id="m-27">Crequillon</p>
  </div>
</front>
```

pe-ro:
Would allowing `@x` and `@y` on `<div>` be sufficient?  Or does it also need `@x2` and `@y2` in order to define its bounding box?  Would `@ulx, @uly, @lrx, and @lry` better suit this purpose?  Should these bounding box attributes be added to all block-level elements -- those in model.textcomponentLike (lg, model.listLike (biblList, castList, list), p, quote, table)  -- and any `x` and `y` coordinate attributes removed?

Without opening up a discussion just yet of a similar music style attribute/specification), would it be a reasonable alternative for text block elements to allow a `@style` attribute with CSS content?

pe-ro:
Do the same requirements also apply to `<anchoredText>`?

ahankinson:
It might be splitting hairs, but I see `@ulx`, etc. to be defining a box that sits on top of the content (e.g., in the `<zone>` element), while `@x`, `@y`, `@x2` and `@y2` to be defining a box **for** the content. So for this particular case, I think the latter would be a better solution.

I tried looking it up in TEI, but it doesn't seem like they have the same concept of block-level elements.

I think we should wait on `@style` until after this release. That's a much bigger can of worms.

It would be nice to have it on `<anchoredText>` too.

pe-ro:
[...]

PR #1474, opened 2024-05-23
schema: adjust `<p>` and `<div>` for specifying better text positioning

### LINKED ISSUE #1189: Text placement within `div`
I am looking at encoding / rendering `div` for text in-between measures or systems as displayed in the example below. I imagine that text placement can be achieved with a `table`. It seems that using `div` within `div` would be simpler. However, as it stands, the position of a `div` cannot really be encoded.  Looking at existing attributes, having `@halign` and `@valign` on `div` would do the job. Then you can do:
```xml
<div>
  <div halign="left" valign="top">Some text at the top left of the div</div>
  <div halign="center" valign="middle">Text centered</div>
  <div halign="right" valign="bottom">Some text at the bottom right</div>
</div>
```
This would require adding these two attributes to `<div>`. Any objections?

Along the same lines, it would be meaningful to allow `<div>` within `pgHead` and `pgFoot` for positioning to be encoded in the same way.

div-01-1

lpugin:
Relates to #1189


lpugin:
Also addresses #272

rettinghaus:
"A horizontal division" is a bit problematic as it could be read as "horizontally divided", but this is not exactly the point, Is it?

If I understood correctly it is about splitting a `div` into multiple vertically aligned parts right? Why not borrow `cb` (column beginning) from [TEI](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-cb.html)?

lpugin:
As you can see the `@func` is on the `div` and is meant to describe how the division interrupts the flow. When placed between two measures, a `<div func="hdiv"/>` represents a division within the system between the two measures, whereas a `<div func="vdiv"/>` is a vertical division (i.e., replacing a full system).

rettinghaus:
Thanks for the clarification. But I'm still not fully convinced of the benefits. 
You could do the same with 

```xml
<measure />
<div />
<measure />
```

for text between measures, and 

```xml
<measure />
<sb />
<div />
<sb />
<measure />
```

for divisions between systems.

lpugin:
That works if you want to use and rely on encoded layout (i.e., `sb`), but not if you do not want to have encoded `sb`. That is what `div@func` is there for.

rettinghaus:
Still, if you just put it between measures it depends on the content where a renderer would put it, especially if its content also is expected to layout automatically. If you want to say that it fills the space of a full system, then a boolean flag would be enough, something like `systemlike="true"`. But that would also be some sort of "layout" that you need to rely on. 
The best way is to make it explicit by putting it on a higher level between sections or, even better, between mDivs.

lpugin:
I do not see a major difference between `systemlike="true"` and `func="vdiv"`.

> The best way is to make it explicit by putting it on a higher level between sections or, even better, between mDivs

Because the number of levels of nesting of `section` is not constraint and not semantically bound to anything decided in MEI, your approach does not work. Just to give you one example: it would make it impossible to have a `hdiv` between two sections, which depending on what a section is in the piece, might be expected. Unless you introduce then `systemlike="false"` ...

bwbohl:
A bit late to the party, but wouldn’t it be good to add to the description of `att.horizontalAlign` and `att.verticalAlign` something about the scope of alignment or the reference box and that the attributes are about the placement of an element’s contents?

And for better understanding, @lpugin could you provide a "complete" encoding for the example page at the beginning of #1189

source/modules/MEI.shared.xml
```diff
@@ -5524,6 +5524,17 @@
...
+        <desc xml:lang="en">Indicates whether the element acts as a vertical or a horizontal division within the flow.</desc>
+        <valList type="closed">
+          <valItem ident="hdiv">
+            <desc xml:lang="en">A horizontal division.</desc>
+          </valItem>
+          <valItem ident="vdiv">
+            <desc xml:lang="en">A vertical division.</desc>
+          </valItem>
+        </valList>
+      </attDef>
```
lpugin:
```suggestion
      <attDef ident="func" usage="opt">
        <desc xml:lang="en">Indicates whether the element acts as a measure or a system within the flow.</desc>
        <valList type="closed">
          <valItem ident="measure">
            <desc xml:lang="en">A measure division.</desc>
          </valItem>
          <valItem ident="system">
            <desc xml:lang="en">A system division.</desc>
          </valItem>
        </valList>
      </attDef>
```

source/modules/MEI.shared.xml
```diff
@@ -5524,6 +5524,17 @@
       </rng:zeroOrMore>
     </content>
     <attList>
+      <attDef ident="func" usage="opt">
+        <desc xml:lang="en">Indicates whether the element acts as a measure or a system within the flow.</desc>
```
rettinghaus:
```suggestion
        <desc xml:lang="en">Indicates whether the element acts like a measure or a system within the flow.</desc>
```

source/modules/MEI.shared.xml
```diff
@@ -5524,6 +5524,17 @@
...
+        <valList type="closed">
+          <valItem ident="measure">
+            <desc xml:lang="en">A measure division.</desc>
```
rettinghaus:
```suggestion
            <desc xml:lang="en">A measure-like division.</desc>
```

source/modules/MEI.shared.xml
```diff
@@ -5524,6 +5524,17 @@
...
+          </valItem>
+          <valItem ident="system">
+            <desc xml:lang="en">A system division.</desc>
```
rettinghaus:
```suggestion
            <desc xml:lang="en">A system-like division.</desc>
```

rettinghaus (dismissed):

source/modules/MEI.shared.xml
lpugin: Not sure about this one. Maybe an native English speaker can decide.

lpugin: @ahankinson ?

rettinghaus: @ahankinson ?

bwbohl: I think that both are valid alternatives; while `act as` emphasises a role, `act like` emphasises a behaviour. But what about adding a bit more verbosity, e.g.:
```suggestion
        <desc xml:lang="en">Specifies how the element is treated in terms of page flow: as inline element like a measure, or as block element like a system.</desc>
```

rettinghaus (dismissed):

lpugin:
May I ask what is missing for the PR to be approved?

pe-ro:
I don't have any objections to this PR, but it may be helpful to 

1. rename `@func` to `@rend` and
2. allow `@rend` to take the values "inline" and "block".

The name `@rend` is more in keeping with TEI and HTML since it carries info about the _appearance_ of the `<div>`.

The values "inline" and "block" are replacements for "hdiv" and "vdiv", respectively.  These terms are analogous to how objects; e.g., images, are labeled  in flowing text.  `<div func="inline">` indicates that the `<div>` does not interrupt the flow of measures; that is, it fits between measures (essentially taking the place of a measure or measures), while `<div func="block">` means that the `<div>` interrupts the flow of measures; that is, it takes the place of a system.

bwbohl:
MEI ODD meeting: Attribute name should be `@flow`with  values `inline` and `block`.

bwbohl:
@lpugin maybe you could also add the images from the issue to the guidelines for illustrative purposes

bwbohl (changes requested):
MEI ODD meeting January: Attribute name should be `@flow` with values inline and block.

@lpugi,n maybe you could also add the images from the issue to the guidelines for illustrative purposes

bwbohl:
As agreed in today’s ODD meeting, let’s move the documentation section to chapter Text (whatever number) and then it’s good to go

bwbohl (dismissed):

rettinghaus (dismissed):

bwbohl:
GitHub seems to have issues processing my local branch update because the “update branch” button isn’t working… Issuing a new comparison or PR from lpugin:develop-div looks just fine, but this here doesn’t update (see “processing changes” at the top). We can wait just a bit longer or close and create a new PR. @lpugin, what do you think?

bwbohl:
cf. https://github.com/music-encoding/music-encoding/compare/develop...lpugin:music-encoding:develop-div

lpugin:
I updated it locally

bwbohl:
That's what I had done before but then it stopped working, great it worked for you!

bwbohl (approved):

bwbohl:
I merged it bypassing the requirement of two reviews because they had been there before and just got de-approved because of the local rebase

PR #1475, opened 2024-05-23
schema: for consistency, allow `<clef>` within `<scoreDef>`

lpugin:
For consistency, allow `<clef>` within `<scoreDef>` since `@clef.*` are allowed

* Fixes to #1439

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -7892,6 +7892,9 @@
         <rng:ref name="model.symbolTableLike"/>
       </rng:zeroOrMore>
       <rng:optional>
+        <rng:ref name="clef"/>
+      </rng:optional>
+      <rng:optional>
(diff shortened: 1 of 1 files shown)


bwbohl (approved):
I think it is more consistent now.

rettinghaus:
I'm still not convinced that this is an improvement, but if the decision is to move in this direction, it is at least more consistent, yes.

bwbohl:
@musicEnfanthen and I revisited this PR today and looked at the context. We think we should tackle and discuss this in a broader context with more people, e.g., a developer workshop. Thus, postponing to MEI 6.

lpugin:
OK. It did not look like a big deal to me, though.

bwbohl (approved):

martha-thomae (approved):

PR #1477, opened 2024-05-26
Update MEI.visual.xml

### LINKED ISSUE #1379: Allow `@vgrp` on `<harm>` and other `<dir>` derivative elements

pe-ro:
Make att.breath.vis a member of att.verticalGrp.

Similar to #1472

Closes #1379

### DIFF
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -303,6 +303,7 @@
       <memberOf key="att.staffLoc"/>
       <memberOf key="att.staffLoc.pitched"/>
       <memberOf key="att.typography"/>
+      <memberOf key="att.verticalGrp"/>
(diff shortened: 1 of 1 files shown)


musicEnfanthen:
@pe-ro I just added the final change here to #1472, so we have everything in one PR. 

Closing this one.

Issue #1452, opened 2024-04-23
Setup for Customizing the 'music-encoding' Tools


o-sapov:
As announced in the last MEI community forum, we at the Digital Mozart Edition ([dme.mozarteum.at](dme.mozarteum.at)) are about to adopt the current [music-encoding tool](https://github.com/music-encoding/music-encoding) for building our own guidelines,  a process that could also be interesting for other projects as we hope. First step will be generation of the web version of the guidelines. To ensure that our developments have also a return of invest to the MEI community, we would like to discuss the basic integration setup in this issue.

Considering our TEI source and examples, we have aligned the structure with that of the `music-encoding/source`. Additionally, we have copied assets and other artifacts there: 

<img width="150"  src="https://github.com/music-encoding/music-encoding/assets/24490898/40fa5de9-89a6-44fd-a032-236d136c0f4c">

This repository is hosted on our GitLab server, which is currently only for internal use. 

We are currently considering two possible integration setups. In both cases, development is taking place on our [fork](https://github.com/ism-dme/music-encoding) of the music-encoding repository.

**First setup**<br>
The `music-encoding/source` folder is replaced by our source files which are embedded as a submodule: 

<img width="300"  src="https://github.com/music-encoding/music-encoding/assets/24490898/00a55d92-cbde-4932-92e9-e492796a5ac6">

```
[submodule "source"]
	path = source
	url = https://gitlab.mozarteum.at/dme/dime/dime-guidelines.git
```

This setup allows for running the ant target `build-guidelines-html` without any modifications.

**Second setup**<br>
The music-encoding repository itself is embedded as a submodule of our guidelines repository: 

<img width="300"  src="https://github.com/music-encoding/music-encoding/assets/24490898/d18480a2-e8db-4f13-b736-7e3be61bc817">

```
[submodule "music-encoding"]
	path = music-encoding
	url = https://github.com/ism-dme/music-encoding.git
  branch = dime-guidelines-test
```

This setup does not affect the original `music-encoding/source` folder in any way. However, the `odd2html` XSLT-files have to be adopted to allow processing the MEI source located at a different location. 

Currently, after initial discussions with @bwbohl  and @musicEnfanthen , we tend towards adopting the second setup. Any other proposals are welcome!

musicEnfanthen:
@music-encoding/technical-team Any second thoughts?

rettinghaus:
[...]

PR #1491, opened 2024-07-03
utils: separate contributor list from index generation

musicEnfanthen:
This PR is an addition to #1471 and intended to be merged into that PR.

It separates the generation of the contributors list from the generation of indices which prepares an option to not include the full list in the future. A possible use case for that is the reuse of the MEI Utils for project customization specific guidelines (which do not necessarily retrieve any contributors, but would want to include the indices in the back matter) as discussed by @o-sapov in #1452


rettinghaus (approved):
lgtm, thanks!

bwbohl (approved):

PR #1492, opened 2024-07-04
guidelines: clarify use of page/system/line beginnings

### LINKED ISSUE #1489: Guidelines: <lb> elements referenced as line breaks instead of line beginnings
The specs for the `<lb>` element say: 

`(line beginning) – An empty formatting element that forces text to begin on a new line. `

However, the guideline chapters [3.7.1.2](https://music-encoding.org/guidelines/dev/content/metadata.html#textTitlePages) and [8.3](https://music-encoding.org/guidelines/dev/content/lyricsperfdir.html#lyricsInEvents) use and describe these elements in text and examples as "line breaks" (3.7.1.2) or even "line endings" (8.3) which is obviously wrong and a common mistake.

This should be corrected to avoid confusion.

NB: Could be worth to check for `pb` and `sb` as well.

rettinghaus:
This PR changes references to the `lb` element and similar where it could be misleading. 

closes #1489

### DIFF
--- a/source/docs/02-shared.xml
+++ b/source/docs/02-shared.xml
@@ -333,7 +333,7 @@
                         <specDesc key="colLayout"/>
                      </specList>
                   </p>
-                  <p>In order to force a new system in the musical text <gi scheme="MEI">sb</gi> can be used.</p>
+                  <p>In order to force the beginning of a new system in the musical text <gi scheme="MEI">sb</gi> can be used.</p>
                   <p>
                      <specList>
                         <specDesc key="sb"/>
--- a/source/docs/02-shared.xml
+++ b/source/docs/02-shared.xml
@@ -333,7 +333,7 @@
                         <specDesc key="colLayout"/>
                      </specList>
                   </p>
-                  <p>In order to force a system break in the musical text <gi scheme="MEI">sb</gi> can be used.</p>
+                  <p>In order to force a new system in the musical text <gi scheme="MEI">sb</gi> can be used.</p>
                   <p>
                      <specList>
                         <specDesc key="sb"/>
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -1439,7 +1439,7 @@
                            <specDesc key="lg"/>
                         </specList>
                      </p>
-                     <p>The following example shows the encoding of the title page of <name ref="https://en.wikipedia.org/wiki/Ralph_Vaughan_Williams">Vaughan Williams'</name> <hi rend="italic">On Wenlock Edge</hi>. Note the use of the <gi scheme="MEI">lb</gi> element to mark new lines present in the original.</p>
+                     <p>The following example shows the encoding of the title page of <name ref="https://en.wikipedia.org/wiki/Ralph_Vaughan_Williams">Vaughan Williams'</name> <hi rend="italic">On Wenlock Edge</hi>. Note the use of the <gi scheme="MEI">lb</gi> element to mark the beginning of new lines present in the original.</p>
                      <p>
                         <figure>
                            <head/>
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -1439,7 +1439,7 @@
                            <specDesc key="lg"/>
                         </specList>
                      </p>
-                     <p>The following example shows the encoding of the title page of Vaughan Williams' <hi rend="italic">On Wenlock Edge</hi>. Note the use of the <gi scheme="MEI">lb</gi> element to mark the line breaks present in the original.</p>
+                     <p>The following example shows the encoding of the title page of Vaughan Williams' <hi rend="italic">On Wenlock Edge</hi>. Note the use of the <gi scheme="MEI">lb</gi> element to mark the line beginnings present in the original.</p>
                      <p>
                         <figure>
                            <head/>
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -1439,7 +1439,7 @@
                            <specDesc key="lg"/>
                         </specList>
                      </p>
-                     <p>The following example shows the encoding of the title page of Vaughan Williams' <hi rend="italic">On Wenlock Edge</hi>. Note the use of the <gi scheme="MEI">lb</gi> element to mark the line beginnings present in the original.</p>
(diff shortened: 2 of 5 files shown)


bwbohl (dismissed):

bwbohl:
Many thanks!

kepper (dismissed):
I wouldn't want to remove this without further discussion, as I don't think that this whole paragraph is simply wrong. Instead, I believe this is a feature originally planned, but never fully implemented (I checked back until [2010-05](https://music-encoding.org/downloads/MEI_TagLibrary_2010-05.pdf#page=291)). The idea was to support encoding of karaoke applications, or, perhaps more relevant for traditional projects, the ability to encode additional verses not contained in the music, but written as plain text below (or elsewhere). Up until MEI 3, we had an element called `<lyrics>`, which was [removed](https://music-encoding.org/archive/comparison-4.0.html#elementsRemoved) in MEI v4. With https://github.com/music-encoding/music-encoding/commit/088149bdc5b0c9f74464b4a6dd6ae9152390260d, this was removed in favor of `<lg>`. This element is actually allowed inside `<div>`, but the situation isn't really documented well in [8.4 Vocally Performed Text Encoded Separately](https://music-encoding.org/guidelines/dev/content/lyricsperfdir.html#lyricsAfterEvents). So, this paragraph is probably a bit dated, but it was there intentionally, and just removing it doesn't feel right, and it's definitely not an _improvement on the description on breaks_ – I'd say it's rather unrelated to that. Accordingly, I'd rather nto include this commit, and instead open another ticket for this discussion.

bwbohl:
Why close the whole pull request? I’d say let’s just restore the one paragraph, and ok, we can start a discussion on the karaoke paragraph, which – at least from my point of view – is utterly needed; maybe I'm blind, but I can’t grasp the karaoke `verse/lb` model

musicEnfanthen:
@kepper I think the main issue with the karaoke paragraph is not the karaoke application, but the use of `lb` described as "line endings"(!) there (instead of line beginning). So while I agree that the paragraph does not need to be removed necessarily, the intended use of `lb` there is indeed obviously wrong and misleading.

kepper:
I agree there is a problem that needs to be addressed. All I'm saying is that it's a different problem, and removing is not a good solution. No hard feelings.

source/docs/02-shared.xml
```diff
@@ -333,7 +333,7 @@
                         <specDesc key="colLayout"/>
                      </specList>
                   </p>
-                  <p>In order to force a system break in the musical text <gi scheme="MEI">sb</gi> can be used.</p>
+                  <p>In order to force a new system in the musical text <gi scheme="MEI">sb</gi> can be used.</p>
```
musicEnfanthen: I am wondering if this still could confuse people to think of `<sb>` as system "break"? 

Maybe: a new system beginning?

source/docs/03-metadata.xml
```diff
@@ -1439,7 +1439,7 @@
                            <specDesc key="lg"/>
                         </specList>
                      </p>
-                     <p>The following example shows the encoding of the title page of Vaughan Williams' <hi rend="italic">On Wenlock Edge</hi>. Note the use of the <gi scheme="MEI">lb</gi> element to mark the line breaks present in the original.</p>
+                     <p>The following example shows the encoding of the title page of <name ref="https://en.wikipedia.org/wiki/Ralph_Vaughan_Williams">Vaughan Williams'</name> <hi rend="italic">On Wenlock Edge</hi>. Note the use of the <gi scheme="MEI">lb</gi> element to mark new lines present in the original.</p>
```
musicEnfanthen: same here: new line beginnings ?

musicEnfanthen: Suggestion to also adjust the following `text-sample371.txt` where the `lb` elements appear at the end of every line.

musicEnfanthen: Same applies to `text-sample362.txt`

source/docs/08-lyricsperfdir.xml
```diff
@@ -181,7 +181,6 @@
                      <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve" valid="feasible"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/lyricsDesc/lyricsDesc-sample283.txt" parse="text"/></egXML>
                   </figure>
                </p>
-               <p>Optionally, it is possible to include an <gi scheme="MEI">lb</gi> element within <gi scheme="MEI">verse</gi> to explicitly encode line and line group endings. This is specifically meant to facilitate karaoke applications.</p>
```
musicEnfanthen, line 184:
```suggestion
               <p>Optionally, it is possible to include an <gi scheme="MEI">lb</gi> element within <gi scheme="MEI">verse</gi> to explicitly encode line and line group beginnings. This is specifically meant to facilitate karaoke applications.</p>
```

source/docs/02-shared.xml
bwbohl: indeed

riedde:
```suggestion
                  <p>In order to force the beginning of a new system in the musical text <gi scheme="MEI">sb</gi> can be used.</p>
```

source/docs/03-metadata.xml
```suggestion
                     <p>The following example shows the encoding of the title page of <name ref="https://en.wikipedia.org/wiki/Ralph_Vaughan_Williams">Vaughan Williams'</name> <hi rend="italic">On Wenlock Edge</hi>. Note the use of the <gi scheme="MEI">lb</gi> element to mark the beginning of new lines present in the original.</p>
```

riedde: The link is referring not referring to a picture, but to an article where no title plage is shown at the moment. Could we use a permanent url to a stable version or a link to a wiki resource here?

riedde (approved):
Some small comments, but looks fine then!

source/docs/03-metadata.xml
rettinghaus: It links the name of Vaughan Williams to the corresponding article about this composer.

riedde: ah sorry, I was wrong. I just saw a link in a text about a picture and the encoding and expected (maybe also) a link to this. so, just ignore my comment!

musicEnfanthen:
@kepper Statement was adusted without removing the part about karaoke applications. Are you fine with the solution and could withdraw your change requests resp. approve the PR?

musicEnfanthen (approved):

bwbohl (approved):

PR #1497, opened 2024-07-04
guidelines: encode quotes

rettinghaus:
This PR adds a rule for quoted content to the `guidelines.xsl` and properly encodes some quotes.
Additionally it adds a link and marks a foreign word as such.

### DIFF
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -718,11 +718,11 @@
                <p>The IFLA’s FRBR model distinguishes four levels of abstraction, or entities:</p>
                <list type="gloss">
                   <label>Work</label>
-                  <item>FRBR defines a work as a "distinct intellectual or artistic creation", an abstract entity because there is no single material object one can point to as the work.</item>
+                  <item>FRBR defines a work as a <quote>distinct intellectual or artistic creation</quote>, an abstract entity because there is no single material object one can point to as the work.</item>
                   <label>Expression</label>
-                  <item>An expression is defined as "the intellectual or artistic realization of a work in the form of [...] notation, sound, image, object, movement, etc., or any combination of such forms". Expressions are also abstract entities.</item>
+                  <item>An expression is defined as <quote>the intellectual or artistic realization of a work in the form of [...] notation, sound, image, object, movement, etc., or any combination of such forms</quote>. Expressions are also abstract entities.</item>
                   <label>Manifestation</label>
-                  <item>A manifestation is defined as "the physical embodiment of an expression of a work", including, for instance, manuscripts, books, sound recordings, films, video recordings, CD-ROMs, multimedia kits, etc. The manifestation represents all the physical objects that bear the same characteristics, with respect to both intellectual content and physical form.</item>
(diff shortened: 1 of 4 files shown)


bwbohl (dismissed):

utils/guidelines_xslt/odd2html/guidelines.xsl
```diff
@@ -463,6 +462,15 @@
         <span class="mentioned"><xsl:apply-templates select="node()" mode="#current"/></span>
     </xsl:template>
     
+    <xd:doc>
+        <xd:desc>
+            <xd:p>Quoted content</xd:p>
+        </xd:desc>
+    </xd:doc>
+    <xsl:template match="tei:q|tei:quote" mode="guidelines">
+        <xsl:value-of select="'&quot;'"/><xsl:apply-templates select="node()" mode="#current"/><xsl:value-of select="'&quot;'"/>
```
bwbohl, line 471: second thought after approving: we could handle this via CSS and then even decide on hos the quotes actually look like with the CSS `quotes:` property what do you think?

rettinghaus: I thought about this, but I'd say this is actual _content_ and not just _styling_.

source/docs/03-metadata.xml
```diff
@@ -718,11 +718,11 @@
                <p>The IFLA’s FRBR model distinguishes four levels of abstraction, or entities:</p>
                <list type="gloss">
                   <label>Work</label>
-                  <item>FRBR defines a work as a "distinct intellectual or artistic creation", an abstract entity because there is no single material object one can point to as the work.</item>
+                  <item>FRBR defines a work as a <quote>distinct intellectual or artistic creation</quote>, an abstract entity because there is no single material object one can point to as the work.</item>
```
musicEnfanthen, line 721: Since these are inline-quotations, `q` should be used instead of block-level `quote`?

rettinghaus: Are you confusing [MEI](https://music-encoding.org/guidelines/v5/elements/quote.html) with [TEI](https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-quote.html)?

musicEnfanthen: Yes, indeed, just recognized that TEI does not make this assumption.

I guess I got confused by #1498 where it says that `q` should be used... Maybe you can add `quote` there, too, and give some hint when which of them should be used? Could be helpful for those who will work on it as "good first issue" ;)

source/docs/08-lyricsperfdir.xml
```diff
@@ -163,7 +163,7 @@
...
+               <p>As it is common practice in written text, it is assumed that a space separates words. Many vocal texts, however, introduce elisions and connect two syllables into one unit. For example, the vocal text from Mozart’s <hi rend="italic">Don Giovanni</hi> sung by Don Giovanni in Finale II, <quote>Ho fermo il core in petto</quote> introduces an elision between the word fermo and il and between core and in. An elision can be indicated by placing both syllables within the same <gi scheme="MEI">note</gi> and setting the <gi scheme="MEI">syl</gi> element’s <att>con</att> attribute value to 't':</p>
```
musicEnfanthen, line 166: Would it make sense to use foreign xml:lang="it" for the 4 italian words after the quote, too?

rettinghaus: Actually we should use `<foreign xml:lang="it" >` here in addition, but too much for this PR I'd say.

bwbohl (approved):

musicEnfanthen (approved):

PR #1501, opened 2024-07-08
schema: replace outdated tei:macroSpec[type] syntax

rettinghaus:
This PR updates our ODDs to be valid TEI again. 

Background: the `macroSpec` element lost it's `@type` attribute long ago, and introduced the `dataSpec` element instead. These changes adopt the current TEI schema and adjust our XSLTs for generating the Guidelines accordingly. No changes in schema or Guidelines are expected.

### DIFF
--- a/source/validation/mei-source.sch
+++ b/source/validation/mei-source.sch
@@ -82,7 +82,7 @@
             <sch:let name="name" value="string(@name)"/>
             <sch:let name="elements" value="//tei:elementSpec/@ident/string()"/>
             <sch:let name="models" value="//tei:classSpec[@type = 'model']/@ident/string()"/>
-            <sch:let name="macros" value="//tei:macroSpec[@type ='pe']/@ident/string()"/>
+            <sch:let name="macros" value="//tei:macroSpec/@ident/string()"/>
             <sch:let name="datatypes" value="//tei:dataSpec/@ident/string()"/>
             <sch:assert role="error"
                 test="$name = $elements or $name = $models or $name = $macros or $name = $datatypes or $name = ('svg', 'svg_svg')">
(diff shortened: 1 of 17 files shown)


utils/guidelines_xslt/odd2html/specs/macroGroupSpecs.xsl
```diff
@@ -54,7 +54,7 @@
     
     <xd:doc>
         <xd:desc>
-            <xd:p>Creates a section for a single macroSpec[@type='pe']</xd:p>
+            <xd:p>Creates a section for a single macroSpec</xd:p>
```
musicEnfanthen: There is a typo in line 40 above. It should read:

`<xd:p>Creates a section with all macroSpecs</xd:p>`

Can you please fix it, too?

source/validation/mei-source.sch
```diff
@@ -87,7 +87,7 @@
             <sch:let name="elements" value="//tei:elementSpec/@ident/string()"/>
             <sch:let name="models" value="//tei:classSpec[@type = 'model']/@ident/string()"/>
             <sch:let name="macros" value="//tei:macroSpec[@type ='pe']/@ident/string()"/>
```
musicEnfanthen:
```suggestion
            <sch:let name="macros" value="//tei:macroSpec/@ident/string()"/>
```

utils/compare_versions/comparison.xsl
bwbohl, line 1: There are still some templates relying on `@type='dt'` in this file, e.g.:

https://github.com/music-encoding/music-encoding/blob/269bd98612d3d2055a207cc68ae29a4b50e053e5/utils/compare_versions/comparison.xsl#L887-L927

rettinghaus: This change wasn't intended. I reverted it and opened #1510 instead.

PR #1503, opened 2024-07-10
schema: unify remarks

rettinghaus:
This PR combines `remarks` in the schema (by keeping the paragraphs), just to make sure all information in one language is kept together. 

 ✅ No changes in the generated Guidelines.


doerners (approved):
Thanks @rettinghaus

riedde (approved):
thank you @rettinghaus

musicEnfanthen (approved):

bwbohl (approved):
Although the the ODD schema does not require to have just one remarks element per language, I do agree, that changing this aligns our encoding with the encoding praxis of the TEI.

PR #1506, opened 2024-07-16
guidelines: give credit to main editor of chapter 3

rettinghaus:
As @riedde did most of the work on revising Chapter 3 of the Guidelines, we should give him credit.

### DIFF
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -32,8 +32,7 @@
             <title level="a">Metadata in MEI</title>
             <respStmt>
                <resp>Edited by</resp>
-               <name role="edt">Margrethe Bue</name>
-               <name role="edt">Kristina Richts-Matthaei</name>
+               <name role="edt">Clemens Gubsch</name>
                <name role="edt">Dennis Ried</name>
                <note>The editors listed here are responsible for the contents of this very chapter of the MEI Guidelines.</note>
             </respStmt>
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -34,6 +34,7 @@
                <resp>Edited by</resp>
                <name role="edt">Margrethe Bue</name>
                <name role="edt">Kristina Richts-Matthaei</name>
+               <name role="edt">Dennis Ried</name>
                <note>The editors listed here are responsible for the contents of this very chapter of the MEI Guidelines.</note>
             </respStmt>
             <respStmt>


source/docs/03-metadata.xml
```diff
@@ -32,8 +32,7 @@
             <title level="a">Metadata in MEI</title>
             <respStmt>
                <resp>Edited by</resp>
-               <name role="edt">Margrethe Bue</name>
-               <name role="edt">Kristina Richts-Matthaei</name>
+               <name role="edt">Dennis Ried</name>
```
riedde, line 36: I only revised chapter 3.7 not the whole thing, so what about just adding my name here?

riedde (commented):
Thank you @rettinghaus for this nice idea :) I left you a small comment.

source/docs/03-metadata.xml
rettinghaus: As you wish.

musicEnfanthen: Was about to propose the same, thanks! :) 

In general, as I understand it, the persons listed here are meant to be the main responsible editors for the chapters (as "appointed" by the @music-encoding/guidelines-editors-in-chief which is far away from being a "formal appointment" but should at least be coordinated with them). That's a bit different from a comprehensive list of all people that have worked on a certain chapter. 

So I guess the main question is whether the responsibility of the main editors has been transferred to @riedde and @gucl-mu with the change of chairmanship in the @music-encoding/ig-metadata "automatically". However, even if that is the case, it does not answer the question wether we want to keep only the current main editors for the edt role or a list including all former main editors as well (which could become quite a long list over time). 

There are arguments for both approaches.

So this should be discussed or agreed in the Metadata IG and with the former main editors first and also coordinated with @kepper.

musicEnfanthen: In any case, by the initial logic of this PR, @gucl-mu would need to be included here as well.

musicEnfanthen: @riedde @gucl-mu What do you think?

riedde: @musicEnfanthen Before this discussion I was not sure what this responsibility means. When this chapter is in the responsibility of the Chairmen, then @gucl-mu and @riedde are the persons to be namend here.

musicEnfanthen: @riedde It is at least in the responsibility of the "main editors". That don't need to be the IG chairs necessarily but could be delegated as well. But it also makes sense to have this role connected to the IG chairs. 

Maybe you can have a short discussion with everyone involved (@margrethesb , @KristinaRichts and @gucl-mu) if they all would be OK with the change? (Don't expect any objections but think that it would be a matter of fairness.)

@kepper Any additions from your side?

riedde: I talked to @gucl-mu and he is fine with the process.

Summarizing: Responsible for chapter 3 is @music-encoding/ig-metadata represented by the co-chairs or deligates.

@musicEnfanthen If no one, mentioned in this PR, has additions or contra arguments this is the way we do it for MEI 5.1

source/docs/03-metadata.xml
```diff
@@ -34,6 +34,7 @@
                <resp>Edited by</resp>
                <name role="edt">Margrethe Bue</name>
                <name role="edt">Kristina Richts-Matthaei</name>
+               <name role="edt">Dennis Ried</name>
```
```suggestion
               <name role="edt">Clemens Gubsch</name>
               <name role="edt">Dennis Ried</name>
```

riedde (changes requested):
As nobody mentioned reacted to this PR, I suggested the outcome of the discussion. @rettinghaus Please have a short look at it. From my side it can be approved then!

riedde (approved):

bwbohl (approved):

musicEnfanthen (approved):

PR #1513, opened 2024-07-22
schema: align method attribute with TEI

### LINKED ISSUE #1512: schema: revise values allowed by att.regularMethod/`@method`
To be more in line with TEI, the values allowed by `att.regularMethod/@method` should be "silent" and "markup".  Typically, correction that isn't carried out silently is accomplished by using elements, but can potentially involve the use of attributes as well.  Elements + attributes = markup, not tags.

This change breaks backward compatibility, but values of "tags" can easily be converted to "markup" via the next v. 5 to v. 6 XSLT.

rettinghaus:
This PR aligns the values in `att.regularMethod` with those in TEI's [correction](https://github.com/TEIC/TEI/blob/dev/P5/Source/Specs/correction.xml) and [normalization](https://github.com/TEIC/TEI/blob/dev/P5/Source/Specs/normalization.xml). In addition it adds a `gloss`.

Closes #1512.

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -305,7 +305,8 @@
     </remarks>
   </classSpec>
   <classSpec ident="att.regularMethod" module="MEI.header" type="atts">
-    <desc xml:lang="en">Attributes that describe correction and normalization methods.</desc>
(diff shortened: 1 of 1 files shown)


pe-ro (approved):
Looks good!  Thanks.

bwbohl (approved):

PR #1520, opened 2024-08-12
docs: remove misleading regex

rettinghaus:
This PR removes the misleading regex in the datatype description for `data.COURSENUMBER` and `data.STRINGNUMBER`, which left the impression, there is no tenth course. Also it fixes wrong naming.

### DIFF
--- a/source/modules/MEI.stringtab.xml
+++ b/source/modules/MEI.stringtab.xml
@@ -25,7 +25,7 @@
     <desc xml:lang="en">Tablature component declarations.</desc>
   </moduleSpec>
   <macroSpec ident="data.COURSENUMBER" module="MEI.stringtab" type="dt">
-    <desc xml:lang="en" versionDate="2024-08-12">In string tablature, the number of the course to be played.</desc>
+    <desc xml:lang="en">In string tablature, the number of the course to be played.</desc>
     <content>
       <rng:data type="positiveInteger"/>
     </content>
@@ -97,7 +97,7 @@
         </datatype>
       </attDef>
       <attDef ident="tab.string" usage="opt">
-        <desc xml:lang="en">This attribute is deprecated in favour of <att>tab.course</att> and will be removed in a future version. Records which string is to be played.</desc>
+        <desc xml:lang="en">This attribute is deprecated in favor of <att>tab.course</att> and will be removed in a future version. Records which string is to be played.</desc>
         <datatype>
           <rng:ref name="data.STRINGNUMBER"/>
         </datatype>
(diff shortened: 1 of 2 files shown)


musicEnfanthen:
Thanks for catching @rettinghaus !

I wouldn't say the regex is misleading because it gives the impression there is no 10th course (which would be included in "[1-9]+"), but that there could be (theoretically) 8912435678 courses (and many, many more).

Waiting for @musicog, @janjusolja, @DILewis or @reinierdevalk (@music-encoding/ig-tablature) to confirm, maybe it was provided intentionally?

While we may not know what mega string instruments with extraordinary numbers of courses have been built or will be built in the future, there seems to be a physical limit for the possible number of courses. Maybe there is a way to describe this (for the less technical users) in simple words instead of the regex expression? Like "any two-digit number below 29" or similar?

rettinghaus:
@musicEnfanthen It can be dated back to prior [MEI 3](https://music-encoding.org/guidelines/v3/data-types/data.stringnumber.html) and was merely copied over to the new datatype without further revision.

pe-ro:
The restriction on the number of courses/strings was originally planned to cover only modern instruments like the guitar, banjo, mandolin, etc.  But various zither-like instruments, both ancient and modern, may need more than 9 courses.  I suppose if there were some kind of tablature for the harp, then one could treat it like an 88-course instrument.  So, it seems there's no way to predict the number of courses a string instrument can have once we lift the original restriction.  In this case, the easiest thing to do is use the RNG integer data type for `data.COURSENUMBER` and `data.STRINGNUMBER`.

rettinghaus:
Another question for @musicog, @janjusolja, @DILewis or @reinierdevalk (@music-encoding/ig-tablature): 
Why not drop `data.STRINGNUMBER` completely and just replace it with `data.COURSENUMBER`, as it's just a change of name.

musicog:
Agreed that imposing an arbitrary limit on the number of courses should be avoided. The regex is accurate, but probably unnecessarily technical (knowledge of regular expressions shouldn't be a requirement for using MEI). We're already specifying positiveInteger as a data type, that should provide enough clarity and the regex should be removed (thx @rettinghaus).

We would like to retain data.STRINGNUMBER and a few other now deprecated things -- with the deprecation notices -- in 5.1 to not break existing tab encodings. They should be removed in future (e.g., MEI 6).

musicog (approved):
Looks good to me! Possibly also remove the regex from `data.STRINGNUMBER`? But less pressing as it's deprecated and will be removed in future.

musicEnfanthen (approved):

musicog (approved):

bwbohl (approved):

Issue #601, opened 2019-06-17
control events in non-measured notation?


axgeertinger:
The MEI 3 model.controleventLike has changed somewhat into model.controlEventLike.cmn in MEI 4. It seems that its members are no longer available in non-measured notation (within `<layer>`, for instance, where they used to be allowed) except inside editorial markup. Is that intentional? Specifically, my question is: Where can I put a fermata in square (neume) notation?

ahankinson:
I think the short answer is that fermata weren't really considered as part of the neume notation module, and are only really available in CMN. I'm not sure that 'mixed mode' CMN/Neumes were tested.

You could probably create your own customization that re-enables the CMN module with the neumes module. Or are you already using the `mei-all.rng`?

axgeertinger:
Yes, I'm using `mei-all.rng`. The problem is that the members of model.controlEventLike.cmn are not allowed in `<layer>` as they used to be in MEI 3. Would it be a problem to reinstate that option? I'd prefer not having to customize the schema. Below is an example from what we are encoding. Fermatas in lines 2 and 5.
![jesp30](https://user-images.githubusercontent.com/3779074/59676778-9624ed80-91c8-11e9-974d-1edd0bf71109.png)

bwbohl:
For cases like this we should definitely come up with a solution, which might mean to reinstate the control events in layer in mei all and disallowing them in CMN?

davidrizo:
In mensural it happens the same situation. I suppose they were removed because the `tstamp` attribute was difficult to use in non measure notation. However, not it seems we can use this kind of notation that can be used to reinstate the control events:

`<fermata staff="1" startid="#note1" place="above" />`

rettinghaus:
`<fermatas>` are currently part of `model.controlEventLike.cmn` 
First step is to make them part of `model.controlEventLike`
Also it may be in `model.sectionLike` to make it available in a `<section>` 
That way it would be possible to use them in unmeasured music and also to use it *between* two measures without trying to figure out a logical value for `@tstamp` to place it on top of a bar line.

kepper:
might be related to #322. Probably a bigger issue…

PR #1522, opened 2024-08-12
schema: make caesura and breath eventLike

### LINKED ISSUE #1521: schema: make caesura eventLike
A [caesura](https://en.wikipedia.org/wiki/Caesura) indicates a pause or break, and isn't a free-floating sign, but always appears on a staff between two notes/rests. This should be reflected in the schema, where `caesura` currently is member of `model.controlEventLike`. It should also be a member of `model.eventLike` to allow the encoding directly within a layer, which is especially helpful for the encoding of measureless music, where it is yet impossible to use `caesura`. 

NB: A similar case is `breath` which should be treated the same way.

rettinghaus:
This PR makes `caesura` and `breath` member of `model.eventLike` and `model.eventLike.cmn` respectively. 

Closes #1521.

### DIFF
--- a/source/modules/MEI.cmn.xml
+++ b/source/modules/MEI.cmn.xml
@@ -1466,6 +1466,7 @@
       <memberOf key="att.breath.ges"/>
       <memberOf key="att.breath.anl"/>
       <memberOf key="model.controlEventLike.cmn"/>
+      <memberOf key="model.eventLike.cmn"/>
     </classes>
     <content>
       <empty/>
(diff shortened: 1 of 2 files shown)


lpugin:
I would suggest looking at this together with #601. Starting to allow control events within a layer solves only one part on the problem. A more generic solution has been discussed, which I will be happy to put forward for discussion.

martha-thomae:
This should be considered when solving https://github.com/music-encoding/music-encoding/issues/601. We are aiming for a bigger solution (for non-measured music events) after MEI 6.0

PR #1523, opened 2024-08-12
guidelines: add documentation for @beam

### LINKED ISSUE #1517: guidelines: `@beam` mentioned but not explained
In [this bit of the guidelines](https://github.com/music-encoding/music-encoding/blob/ac400fe7820241bd69ccd20f16c71932d1859a14/source/docs/04-cmn.xml#L577), this passage appears (emphasis mine):

> In addition to the explicit encoding of beams accommodated by the [\<beam>](https://music-encoding.org/guidelines/v5/elements/beam.html) and [\<beamSpan>](https://music-encoding.org/guidelines/v5/elements/beamSpan.html) elements ***and the @beam attribute***, MEI allows for specification of default beaming behavior using the following attributes on [\<scoreDef>](https://music-encoding.org/guidelines/v5/elements/scoreDef.html), [\<staffDef>](https://music-encoding.org/guidelines/v5/elements/staffDef.html), and [\<layerDef>](https://music-encoding.org/guidelines/v5/elements/layerDef.html):

It's not clear what is referred to by "the @beam attribute" in this passage. The only attribute with "beam" in its name that appears prior to this point in the document is `@beam.with` (of [att.beamedWith](https://music-encoding.org/guidelines/v5/attribute-classes/att.beamedWith.html)). As far as I could tell, two other attribute classes provide attributes starting with "beam":
- [att.beaming.log](https://music-encoding.org/guidelines/v5/attribute-classes/att.beaming.log.html) provides `@beam.group` and `@beam.rests`, but the section in question is preparing to introduce them and so could not be referring to them.
- [att.beaming.vis](https://music-encoding.org/guidelines/v5/attribute-classes/att.beaming.vis.html) provides `@beam.color`, `rend`, and `slope`. These attributes do not appear anywhere in the document in question so it would not make sense to be referring to them.

Can you clarify which attribute is being referred to by this passage? If it helps, I'm happy to make a pull request updating the guidelines once that question has been settled.

maxrothman:
Resolves #1517 

I took a first pass at writing documentation for `@beam`, which was mentioned in the guidelines as if explained elsewhere, but was not actually explained anywhere. This is very much a first pass to have something to discuss, feel free to rewrite/reject/discuss further as necessary.

### DIFF
--- a/source/docs/04-cmn.xml
+++ b/source/docs/04-cmn.xml
@@ -583,7 +583,7 @@
                      <label>t (terminal)</label>
                      <item>Marks the end of a beam</item>
                   </list>
-                  <p>Unlike <gi scheme="MEI">beam</gi> and <gi scheme="MEI">beamSpan</gi>, <att>beam</att> is purely in the analytical domain, meaning it does not represent the visual rendering of a beam (indeed Verovio does not render it) but that the note in question is <em>under a beam</em>. It may be used in cases where analytical needs require a flatter structure, or when beams might not have well-defined starts or ends, or when it simplifies analysis to have beaming be a property of notes rather than a separate element.</p>
+                  <p>In general, it is recommended to use <gi scheme="MEI">beam</gi> or <gi scheme="MEI">beamSpan</gi> over <att>beam</att> for typical notation use-cases. <att>beam</att> is available for those less typical situations, for example when analytical needs require a flatter structure, or when beams might not have well-defined starts or ends, or when it simplifies analysis to have beaming be a property of notes rather than a separate element.</p>
                   <p>In addition to the explicit encoding of beams accommodated by the <gi scheme="MEI">beam</gi> and <gi scheme="MEI">beamSpan</gi> elements and the <att>beam</att> attribute, MEI allows for specification of default beaming behavior using the following attributes on <gi scheme="MEI">scoreDef</gi>, <gi scheme="MEI">staffDef</gi>, and <gi scheme="MEI">layerDef</gi>:</p>
                   <list type="gloss">
                      <label>beam.group</label>
--- a/source/docs/04-cmn.xml
+++ b/source/docs/04-cmn.xml
@@ -574,6 +574,16 @@
                         <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve" valid="true"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/cmn/cmn-sample-schulhoff.txt" parse="text"/></egXML>
                      </figure>
                   </p>
+                  <p>One more way of notating beams is to use the <att>beam</att> attribute found on <gi scheme="MEI">note</gi>. This attribute allows three values:</p>
+                  <list type="gloss">
+                     <label>i (initial)</label>
+                     <item>Marks the start of a beam</item>
+                     <label>m (medial)</label>
+                     <item>Marks a participant in a beam other than the first or last</item>
+                     <label>t (terminal)</label>
+                     <item>Marks the end of a beam</item>
+                  </list>
+                  <p>Unlike <gi scheme="MEI">beam</gi> and <gi scheme="MEI">beamSpan</gi>, <att>beam</att> is purely in the analytical domain, meaning it does not represent the visual rendering of a beam (indeed Verovio does not render it) but that the note in question is <em>under a beam</em>. It may be used in cases where analytical needs require a flatter structure, or when beams might not have well-defined starts or ends, or when it simplifies analysis to have beaming be a property of notes rather than a separate element.</p>
                   <p>In addition to the explicit encoding of beams accommodated by the <gi scheme="MEI">beam</gi> and <gi scheme="MEI">beamSpan</gi> elements and the <att>beam</att> attribute, MEI allows for specification of default beaming behavior using the following attributes on <gi scheme="MEI">scoreDef</gi>, <gi scheme="MEI">staffDef</gi>, and <gi scheme="MEI">layerDef</gi>:</p>
                   <list type="gloss">
                      <label>beam.group</label>
--- a/source/docs/04-cmn.xml
+++ b/source/docs/04-cmn.xml
@@ -574,7 +574,7 @@
                         <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve" valid="true"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/cmn/cmn-sample-schulhoff.txt" parse="text"/></egXML>
                      </figure>
                   </p>
-                  <p>One more way of notating beams is to use the <att>beam</att> attribute found on <gi scheme="MEI">note</gi>. This attribute allows three values:</p>
+                  <p>One more way of notating beams is to use the <att>beam</att> attribute found on <gi scheme="MEI">note</gi>, <gi scheme="MEI">chord</gi>, <gi scheme="MEI">rest</gi>, and <gi scheme="MEI">space</gi> elements. This attribute allows three values:</p>
                   <list type="gloss">
                      <label>i (initial)</label>
                      <item>Marks the start of a beam</item>
--- a/source/docs/04-cmn.xml
+++ b/source/docs/04-cmn.xml
@@ -576,7 +576,7 @@
                         <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve" valid="true"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/cmn/cmn-sample-schulhoff.txt" parse="text"/></egXML>
                      </figure>
                   </p>
-                  <p>One more way of notating beams is to use the <att>beam</att> attribute found on <gi scheme="MEI">note</gi>, <gi scheme="MEI">chord</gi>, <gi scheme="MEI">rest</gi>, and <gi scheme="MEI">space</gi> elements. This attribute allows three values:</p>
+                  <p>Another way of notating beams is to use the <att>beam</att> attribute found on <gi scheme="MEI">note</gi>, <gi scheme="MEI">chord</gi>, <gi scheme="MEI">rest</gi>, and <gi scheme="MEI">space</gi> elements. This attribute allows three values:</p>
                   <list type="gloss">
                      <label>i (initial)</label>
                      <item>Marks the start of a beam</item>
--- a/source/docs/04-cmn.xml
+++ b/source/docs/04-cmn.xml
@@ -585,7 +585,7 @@
                      <label>t (terminal)</label>
                      <item>Marks the end of a beam</item>
                   </list>
-                  <p>Unlike <gi scheme="MEI">beam</gi> and <gi scheme="MEI">beamSpan</gi>, <att>beam</att> is purely in the analytical domain, meaning it does not represent the visual rendering of a beam (indeed Verovio does not render it) but that the note in question is <em>under a beam</em>. It may be used in cases where analytical needs require a flatter structure, or when beams might not have well-defined starts or ends, or when it simplifies analysis to have beaming be a property of notes rather than a separate element.</p>
+                  <p>In general, it is recommended to use <gi scheme="MEI">beam</gi> or <gi scheme="MEI">beamSpan</gi> over <att>beam</att> for typical notation use-cases. <att>beam</att> is available for those less typical situations, for example when analytical needs require a flatter structure, or when beams might not have well-defined starts or ends, or when it simplifies analysis to have beaming be a property of notes rather than a separate element.</p>
                   <p>In addition to the explicit encoding of beams accommodated by the <gi scheme="MEI">beam</gi> and <gi scheme="MEI">beamSpan</gi> elements and the <att>beam</att> attribute, MEI allows for specification of default beaming behavior using the following attributes on <gi scheme="MEI">scoreDef</gi>, <gi scheme="MEI">staffDef</gi>, and <gi scheme="MEI">layerDef</gi>:</p>
                   <list type="gloss">
                      <label>beam.group</label>
--- a/source/docs/04-cmn.xml
+++ b/source/docs/04-cmn.xml
@@ -585,7 +585,7 @@
                      <label>t (terminal)</label>
                      <item>Marks the end of a beam</item>
                   </list>
-                  <p>In general, it is recommended to use <gi scheme="MEI">beam</gi> or <gi scheme="MEI">beamSpan</gi> over <att>beam</att> for typical notation use-cases. <att>beam</att> is available for those less typical situations, for example when analytical needs require a flatter structure, or when beams might not have well-defined starts or ends, or when it simplifies analysis to have beaming be a property of notes rather than a separate element.</p>
+                  <p>Although <att>beam</att> can be used for rendering beams, <gi scheme="MEI">beam</gi> or <gi scheme="MEI">beamSpan</gi> is preferred over <att>beam</att> in most cases. However, because it makes beaming a <emph>property</emph> of a note, chord, rest, or space rather than an <emph>entity</emph> in its own right, <att>beam</att> can provide a flatter structure that simplifies analytical tasks.</p>
                   <p>In addition to the explicit encoding of beams accommodated by the <gi scheme="MEI">beam</gi> and <gi scheme="MEI">beamSpan</gi> elements and the <att>beam</att> attribute, MEI allows for specification of default beaming behavior using the following attributes on <gi scheme="MEI">scoreDef</gi>, <gi scheme="MEI">staffDef</gi>, and <gi scheme="MEI">layerDef</gi>:</p>
                   <list type="gloss">
                      <label>beam.group</label>
--- a/source/docs/04-cmn.xml
+++ b/source/docs/04-cmn.xml
@@ -576,6 +576,16 @@
                         <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve" valid="true"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/cmn/cmn-sample-schulhoff.txt" parse="text"/></egXML>
                      </figure>
                   </p>
+                  <p>One more way of notating beams is to use the <att>beam</att> attribute found on <gi scheme="MEI">note</gi>. This attribute allows three values:</p>
+                  <list type="gloss">
+                     <label>i (initial)</label>
+                     <item>Marks the start of a beam</item>
+                     <label>m (medial)</label>
+                     <item>Marks a participant in a beam other than the first or last</item>
+                     <label>t (terminal)</label>
+                     <item>Marks the end of a beam</item>
+                  </list>
+                  <p>Unlike <gi scheme="MEI">beam</gi> and <gi scheme="MEI">beamSpan</gi>, <att>beam</att> is purely in the analytical domain, meaning it does not represent the visual rendering of a beam (indeed Verovio does not render it) but that the note in question is <em>under a beam</em>. It may be used in cases where analytical needs require a flatter structure, or when beams might not have well-defined starts or ends, or when it simplifies analysis to have beaming be a property of notes rather than a separate element.</p>
                   <p>In addition to the explicit encoding of beams accommodated by the <gi scheme="MEI">beam</gi> and <gi scheme="MEI">beamSpan</gi> elements and the <att>beam</att> attribute, MEI allows for specification of default beaming behavior using the following attributes on <gi scheme="MEI">scoreDef</gi>, <gi scheme="MEI">staffDef</gi>, and <gi scheme="MEI">layerDef</gi>:</p>
                   <list type="gloss">
                      <label>beam.group</label>
--- a/source/docs/04-cmn.xml
+++ b/source/docs/04-cmn.xml
@@ -576,7 +576,7 @@
                         <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve" valid="true"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/cmn/cmn-sample-schulhoff.txt" parse="text"/></egXML>
                      </figure>
                   </p>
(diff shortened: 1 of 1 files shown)


source/docs/04-cmn.xml
```diff
@@ -574,6 +574,16 @@
...
+                  <p>One more way of notating beams is to use the <att>beam</att> attribute found on <gi scheme="MEI">note</gi>. This attribute allows three values:</p>
+                  <list type="gloss">
+                     <label>i (initial)</label>
+                     <item>Marks the start of a beam</item>
+                     <label>m (medial)</label>
+                     <item>Marks a participant in a beam other than the first or last</item>
+                     <label>t (terminal)</label>
+                     <item>Marks the end of a beam</item>
+                  </list>
+                  <p>Unlike <gi scheme="MEI">beam</gi> and <gi scheme="MEI">beamSpan</gi>, <att>beam</att> is purely in the analytical domain, meaning it does not represent the visual rendering of a beam (indeed Verovio does not render it) but that the note in question is <em>under a beam</em>. It may be used in cases where analytical needs require a flatter structure, or when beams might not have well-defined starts or ends, or when it simplifies analysis to have beaming be a property of notes rather than a separate element.</p>
```
rettinghaus: I see no point in mentioning Verovio here.
```suggestion
                  <p>Unlike <gi scheme="MEI">beam</gi> and <gi scheme="MEI">beamSpan</gi>, <att>beam</att> is purely in the analytical domain, meaning it does not represent the visual rendering of a beam but that the note in question is <em>under a beam</em>. It may be used in cases where analytical needs require a flatter structure, or when beams might not have well-defined starts or ends, or when it simplifies analysis to have beaming be a property of notes rather than a separate element.</p>
```

rettinghaus (commented):
Thanks for this first pass.

source/docs/04-cmn.xml
pe-ro: `@beam` is allowed on chord, rest, and space in addition to note.

pe-ro:
Should've made my comment here -- 
`@beam` is allowed on chord, rest, and space in addition to note.

source/docs/04-cmn.xml
ahankinson: ```<att>beam</att> is purely in the analytical domain, meaning it does not represent the visual rendering of a beam but that the note in question is <em>under a beam</em>```

I think I get what @kepper was trying to say here, but I'm not sure I completely agree with phrasing like "purely in the analytical domain" and "it doesn't represent the visual rendering." There is no reason why an `@beam` attribute can't represent a visual beam, and this seems to be an artificial restriction on its use. It's simply an alternative way of indicating a beam structure. MEI is full of these attribute / element alternates: `<tuplet>` or `@tuplet`; `<tie>` or `@tie`; `<slur>` or `@slur`; `@grace` or `<graceGrp>`. Even `@accid` and `<accid>`, `@artic` and `<artic>`, and `@syl` and `<syl>` are forms of this same pattern.

I'm also with @rettinghaus that there is no need to mention Verovio here.

maxrothman:
> `@beam` is allowed on chord, rest, and space in addition to note.

I've addressed this comment in my latest commit. It seems like there's not a consensus yet on the analytic vs visual bit, and that's a more substantive change, so I'm letting that conversation progress further before suggesting more changes.

craigsapp:
> <em>under a beam</em>

using "under" could be confused with `@place="below"`, so another word or phrase would be better, such as "contained within a beam", "associated with a beam" or "attached to a beam" 

>  It seems like there's not a consensus yet on the analytic vs visual bit

The `<beam>` and `<beamSpan>` systems are sufficient for the visual bit as demonstrated by verovio, so I would say that there is instead no consensus to expand `note@beam` into the visual domain. In particular, the parameter`note@breaksec`, is used instead of `note@beam` for refined notation rendering of beams:

Screenshot 2024-08-14 at 9 54 11 AM


```xml
<beam>
   <note dur="16" oct="4" pname="f" />
   <note dur="16" oct="4" pname="f" />
   <note dur="16" oct="4" pname="f" breaksec="1" />
   <note dur="16" oct="4" pname="f" />
   <note dur="16" oct="4" pname="f" />
   <note dur="16" oct="4" pname="f" />
</beam>  
```

<details>
<summary> Click for full example </summary>

```xml
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<?xml-model href="https://music-encoding.org/schema/5.0/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.0">
 <meiHead>
  <fileDesc>
   <titleStmt>
    <title />
   </titleStmt>
   <pubStmt>
    <unpub>This MEI file was created by Verovio's Humdrum converter. When published, this unpub element should be removed, and the enclosing pubStmt element should be properly filled out.</unpub>
   </pubStmt>
  </fileDesc>
  <encodingDesc>
   <appInfo>
    <application isodate="2024-08-14T09:50:04" version="4.3.0-dev-464b2ce-dirty">
     <name>Verovio</name>
    </application>
   </appInfo>
  </encodingDesc>
  <extMeta>
   <frames xmlns="http://www.humdrum.org/ns/humxml" />
  </extMeta>
 </meiHead>
 <music decls="#work1_encoded">
  <body>
   <mdiv xml:id="m19c00a5">
    <score xml:id="s1ts3xf8">
     <scoreDef xml:id="s1asoosf" tempo.dist="3.0000vu">
      <staffGrp xml:id="s11nczdm">
       <staffDef xml:id="staffdef-L1F1" n="1" lines="5">
        <clef xml:id="ct7hndo" shape="G" line="2" />
        <meterSig xml:id="metersig-L2F1" count="3" unit="8" />
       </staffDef>
      </staffGrp>
     </scoreDef>
      <measure xml:id="measure-L1">
       <staff xml:id="staff-L1F1" n="1">
         <beam xml:id="beam-L3F1-L8F1">
          <note xml:id="note-L3F1" dur="16" oct="4" pname="f" accid.ges="n" />
          <note xml:id="note-L4F1" dur="16" oct="4" pname="f" accid.ges="n" />
          <note xml:id="note-L5F1" breaksec="1" dur="16" oct="4" pname="f" accid.ges="n" />
          <note xml:id="note-L6F1" dur="16" oct="4" pname="f" accid.ges="n" />
          <note xml:id="note-L7F1" dur="16" oct="4" pname="f" accid.ges="n" />
          <note xml:id="note-L8F1" dur="16" oct="4" pname="f" accid.ges="n" />
         </beam>
        </layer>
       </staff>
      </measure>
     </section>
    </score>
   </mdiv>
  </body>
 </music>
</mei>
```

</details>

The meaning of the `note@beam` parameter is somewhat vague (particularly what the numbers mean, although I think I am interpreting them properly further below):
 
Screenshot 2024-08-14 at 9 56 20 AM

I do not know what a nested beam is.  There are nested tuplets.

`note@beam` cannot be used in replacement for `<beam>` or `<beamSpan>` since it would require milestone processing.  Perhaps `note@beam` could function as an alternate to `note@breaksec`:

Screenshot 2024-08-14 at 10 25 06 AM

```xml
<beam>
   <note dur="16" oct="4" pname="f" beam="i1 i2" />
   <note dur="16" oct="4" pname="f" beam="m1 m2" />
   <note dur="16" oct="4" pname="f" beam="m1 t2" />
   <note dur="16" oct="4" pname="f" beam="m1 i2" />
   <note dur="16" oct="4" pname="f" beam="m1 m2" />
   <note dur="16" oct="4" pname="f" beam="m1 t2" />
</beam> 
```

presuming that the numbers refer to sub-beams (I would not call them nested beams), with sub-beam number 1 being the 8th-note level.

The current system using `note@breaksec` cannot render this sort of beaming:

Screenshot 2024-08-14 at 10 10 34 AM

which is non-standard and I have never seen an example of in real music.  In other words, `note@beam` is more generalized than `note@breaksec`, but the generalization is not within the scope of standard music notation, so there would be little benefit other than for very rare non-standard notations.  And the possibility of encoding errors is significant compare to `note@breaksec` which is negligible.

The above notation example could be made explicit with `note@beam`:

```xml
<beam xml:id="beam-L3F1-L8F1">
   <note dur="16" oct="4" pname="f" beam="i1 i2" />
   <note dur="16" oct="4" pname="f" beam="m1 m2" />
   <note dur="16" oct="4" pname="f" beam="t1 m2" />
   <note dur="16" oct="4" pname="f" beam="i1 m2" />
   <note dur="16" oct="4" pname="f" beam="m1 m2" />
   <note dur="16" oct="4" pname="f" beam="t1 t2" />
</beam>
```

-------------------

As a side note, MuseData uses the `note@beam` system.  Here is the first example:

```
F4     1      1 s        [[
F4     1      1 s        ==
F4     1      1 s        =]
F4     1      1 s        =[
F4     1      1 s        ==
F4     1      1 s        ]]
```

The brackets and equals signs are the equivalent to "i", "m", and "t", and the position from left to right are equialent to the numbers:

```
F4     1      1 s        [[          i1   i2
F4     1      1 s        ==          m1   m2
F4     1      1 s        =]          m1   t2
F4     1      1 s        =[          m1   i2
F4     1      1 s        ==          m1   m2
F4     1      1 s        ]]          t1   t2
```

If I try to print the second notation example with the MuseData printing system:

```
F4     1      1 s        [[
F4     1      1 s        ==
F4     1      1 s        ]=
F4     1      1 s        [=
F4     1      1 s        ==
F4     1      1 s        ]]
```

It cannot handle it:

Screenshot 2024-08-14 at 10 19 39 AM

So if `note@beam` is used for visual rendering, there could be non-standard rendering requests for beams as in the second notation example.  The `note@breaksec` cannot have these problems, and less encoding is needed since it is a more semantic description of the realistic visual layout of the beams.

pe-ro:
Sorry for the delayed response.

I have no issue with changing the phrase "under a beam" to "attached to a beam" or the like.

@ahankinson, is correct --`@beam` is an alternative way of representing beams on chords, notes, rests themselves instead of wrapping those elements with a `<beam>` element. There's no reason why it can't be used for rendering beams that don't cross the encoding hierarchy.

However, the values of `@beam` can also represent the (counter) hierarchy created by nested or measure-crossing beams, not secondary, tertiary, etc. beams, which are always calculated by the renderer.  "Nested" beams are those that aren't associated with tuplets, but that cross hierarchies the way tuplets sometimes do.  I don't have any examples at hand, but they can be found in Schenkerian and other so-called "analytical notations", I think.  The existence of `@beam` predates that of `<beamSpan>`, but `<beamSpan>` can be used to encode these "analytical beams" as easily or perhaps better than `@beam`.

`@beam` can be redefined to account for secondary beams. Alternatively, `@beam` can be redefined for analytical purposes by assigning it a Boolean value to indicate that the chord, note, rest, etc. is or is not attached to a beam.  Either way would create backward incompatibility if anyone has actually used `@beam` for hierarchy-crossing beams.

So, I have no issue with redefining `@beam`, but suggest that care be taken to account for "analytical beams" and the need to indicate attachment to a beam at the note level.

maxrothman:
Thanks everyone for contributing to this interesting discussion. Given that the discussion has now grown to include changing the meaning of `@beam` entirely, I wanted to try putting forth a different proposal: what if we simply removed all references to `@beam` from the guidelines? As a reminder, the [original issue](https://github.com/music-encoding/music-encoding/issues/1517) that spawned this PR was that [the guidelines](https://github.com/music-encoding/music-encoding/blob/ac400fe7820241bd69ccd20f16c71932d1859a14/source/docs/04-cmn.xml#L577) mention `@beam` but do not define it:

> In addition to the explicit encoding of beams accommodated by the [\<beam>](https://music-encoding.org/guidelines/v5/elements/beam.html) and [\<beamSpan>](https://music-encoding.org/guidelines/v5/elements/beamSpan.html) elements ***and the @beam attribute***, MEI allows for specification of default beaming behavior using the following attributes on [\<scoreDef>](https://music-encoding.org/guidelines/v5/elements/scoreDef.html), [\<staffDef>](https://music-encoding.org/guidelines/v5/elements/staffDef.html), and [\<layerDef>](https://music-encoding.org/guidelines/v5/elements/layerDef.html):

I'm proposing just removing the italicized text.

rettinghaus:
I'm ok with that approach.

ahankinson:
My personal view is that you can simply remove the final paragraph from your PR ("Unlike ...") . You can describe how `@beam` works without needing to justify it.

source/docs/04-cmn.xml
```diff
@@ -574,6 +574,16 @@
                         <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve" valid="true"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/cmn/cmn-sample-schulhoff.txt" parse="text"/></egXML>
                      </figure>
                   </p>
+                  <p>One more way of notating beams is to use the <att>beam</att> attribute found on <gi scheme="MEI">note</gi>, <gi scheme="MEI">chord</gi>, <gi scheme="MEI">rest</gi>, and <gi scheme="MEI">space</gi> elements. This attribute allows three values:</p>
```
ahankinson:
```suggestion
                  <p>Another way of notating beams is to use the <att>beam</att> attribute found on <gi scheme="MEI">note</gi>, <gi scheme="MEI">chord</gi>, <gi scheme="MEI">rest</gi>, and <gi scheme="MEI">space</gi> elements. This attribute allows three values:</p>
```

maxrothman:
> My personal view is that you can simply remove the final paragraph from your PR ("Unlike ...") . You can describe how `@beam` works without needing to justify it.

While I agree that explaining how `@beam` works is better than mentioning it with no explanation, I think it would also be confusing to explain _how_ it works without including any context about _why_ it might be used, especially since it seems like there's a consensus that `<beam>` and `<beamSpan>` are preferred in situations where they can be used.

What do you think about something like this to replace the "Unlike..." paragraph?
> In general, it is recommended to use `<beam>` or `<beamSpan>` over `@beam` if possible. However, `@beam` is available for unusual situations, for example when analytical needs require a flatter structure, or when beams might not have well-defined starts or ends, or when it simplifies analysis to have beaming be a property of notes rather than a separate element.

maxrothman:
I've pushed a version with my suggested paragraph that hopefully threads the needle here. Thoughts?

Also, I know there are some small wording suggestions that have been posted on this PR. I've held off on those for now while we nail down the broad strokes of the change. Just know I haven't been ignoring them!

maxrothman:
Hi y'all, any thoughts on the suggested change? @ahankinson @rettinghaus @pe-ro @craigsapp

pe-ro (changes requested):
I'm not convinced by the current language.  How about:

```<p>Although <att>beam</att> can be used for rendering beams, <gi scheme="MEI">beam</gi> or <gi scheme="MEI">beamSpan</gi> is preferred over <att>beam</att> in most cases. However, because it makes beaming a _property_ of a note, chord, rest, or space rather than an _entity_ in its own right, <att>beam</att> can provide a flatter structure that simplifies analytical tasks.```

**Comment by craigsapp** [comment]
The problem is that `@beam` is then used for both visual and analytic purposes, which are better to keep separate.   Currently it is only analytical according to the documentation, and using it for visual purposes as well will conflict with any simultaneous `<beam>` and `<beamSpan>`.  In other words, if used for rendering beams, would `@beam` have priority over `<beam>` and `<beamSpan>` or the other way around?  `@beam` is also milestone sort of information which makes is more computationally expensive than using `<beam>` and `<beamSpan>` since essentially `@beam` requires converting into `<beam>` and `<beamSpan>` before processing for visual rendering.

This choice of rendering could be left up to the renderer, which would use `<beam>` and `<beamSpan>` by default, and then have an option to ignore these and use `@beam`, but only one at a time should be considered.   Then a program such as verovio could add an option such as `--use-beam-att` to select using `@beam` instead of `<beam>`/`<beamSpan>`.

**Comment by pe-ro** [comment]
@craigsapp, I think you're right about it being a binary choice **in the case of rendering** -- one should use either `<beam>` and `<beamSpan>` or `@beam`.  So I'm fine with a renderer providing a `--use-beam-att` parameter to switch between these modes of operation.

However, `@beam` can also be useful for analysis -- this note, etc. is associated with a beam -- and for 
translation of representations that use milestone markup, such as MusicXML, into MEI.  `@beam` can be a convenient place to temporarily store beaming info until it can be determined whether the beaming information in the input file can be satisfactorily converted into `<beam>` and `<beamSpan>`; that is, until it can be determined whether the milestone markup agrees or disagrees with the measure/event hierarchy, or if it even makes any sense at all (which milestone-using markup all-too-frequently does not).

So, both options must be valid simultaneously even if the rendering question is resolved.  I think we just haven't found the right language to express this big-tent notion yet.  I'd be very pleased if someone else could wordsmith this better than I.  :-)

**Comment by maxrothman** [comment]
In the spirit of not letting the perfect be the enemy of the good enough, I think it's worthwhile to resolve the ambiguity in the current guidelines one way or another, even if that leaves the purpose/use cases of `@beam` not fully detailed. Then a follow-on PR can always be made to improve that aspect of the guidelines later.

In lieu of coming up with a better wording to describe the recommended uses of `@beam`, how about one of these solutions?
1. Remove all mention of `@beam` from the guidelines (it'd still be in its attr class's docs and those of `<note>`)
2. Reduce the last paragraph to just "In general, it is recommended to use `<beam>` or `<beamSpan>` over `@beam` for typical notation use-cases."

I personally feel like the latter is a decent compromise: it appears that there's consensus among reviewers that the statement is correct and it allows `@beam` to be documented in the guidelines at all compared to option 1.

**Comment by craigsapp** [comment]
> However, `@beam` can also be useful for analysis 

I am wanting `@beam` to remain *only* used for analysis, and not used (directly) for rendering.   Using it as a stepping stone to generate `<beam>` and `<beamSpan>` as an analytical process is fine (note that MusicXML and Humdrum importers do not use this stepping stone and convert directly to `<beam>`/`<beamSpan>`).

Of course if @lpugin wants to implement direct rendering from `@beam` in verovio, then I would let him ;-).   This also relates to `<tuplet>`/`<tupletSpan>`, and it can get quite complicated in the general case to figure out how to go from a milestone system for beams/tuplets into `<beam>`/`<tuplet>` while minimizing use of `<beamSpan>`/`<tupletSpan>`.

> In the spirit of not letting the perfect be the enemy of the good enough, I think it's worthwhile to resolve the ambiguity in the current guidelines one way or another, even if that leaves the purpose/use cases of @beam not fully detailed. Then a follow-on PR can always be made to improve that aspect of the guidelines later.

I would say that leaving out @beam is fine.   It should be described somewhere in the documentation (maybe in a chapter/section describing analytic attributes) as a descriptive attribute (analytic) rather than a proscriptive one to cause a beam to show up in a rendering of the data.  Because otherwise the attribute description is vague:

`beam(optional) Indicates that this event is "under a beam". One or more values from data.BEAM, separated by spaces.`

**Comment by bwbohl** [comment]
Pinging: How does everybody like the current state?
I think it’s a valuable improvement of the guidelines and does indeed express the analytical tendencies of `@beam`.

File: source/docs/04-cmn.xml (hunk 2 above)
Comment (bwbohl) [reply source/docs/04-cmn.xml]: @maxrothman is that ok with you?

**Review by rettinghaus — APPROVED** [review]

**Comment by ahankinson** [comment]
I prefer @pe-ro s suggested text rather than the current version. I’m not sure where the distinction between analytical and visual started but this difference in the two encodings doesn’t really fit with all the other “attribute” vs. “element” methods — beams, tuples, slurs, ties, etc. not to mention accidentals and articulations. Suddenly making the distinction between analytical and visual for the specific case of beam means we will have to sort out this distinction for all the other ones too.

The distinction is really about use cases where you only need to state the presence of something (the attributes) vs needing an object to hang other things off of (color, shape, size, or even independent identity). Analytical vs. visual is a back-formulation at best. So I think Perry’s comment goes far enough in suggesting a use case without being too prescriptive about the use of attributes in general for analytical purposes.

**Comment by maxrothman** [comment]
If we're looking for neutral language to compromise, I had previously suggested reducing the paragraph to just:
> "In general, it is recommended to use `<beam>` or `<beamSpan>` over `@beam` for typical notation use-cases."

I agree that some improvement here is better than nothing, and if a consensus cannot be reached on when it `@beam` should be used, then hopefully the above text is something that we _can_ reach a consensus on.

That said, if a consensus can be reached on another text (e.g. @pe-ro's) then I'm happy to go with that too.

File: source/docs/04-cmn.xml [hunk 3]
```diff
@@ -574,6 +574,16 @@
...
+                  <p>Another way of notating beams is to use the <att>beam</att> attribute found on <gi scheme="MEI">note</gi>, <gi scheme="MEI">chord</gi>, <gi scheme="MEI">rest</gi>, and <gi scheme="MEI">space</gi> elements. This attribute allows three values:</p>
+                  <list type="gloss">
+                     <label>i (initial)</label>
+                     <item>Marks the start of a beam</item>
+                     <label>m (medial)</label>
+                     <item>Marks a participant in a beam other than the first or last</item>
+                     <label>t (terminal)</label>
+                     <item>Marks the end of a beam</item>
+                  </list>
+                  <p>In general, it is recommended to use <gi scheme="MEI">beam</gi> or <gi scheme="MEI">beamSpan</gi> over <att>beam</att> for typical notation use-cases. <att>beam</att> is available for those less typical situations, for example when analytical needs require a flatter structure, or when beams might not have well-defined starts or ends, or when it simplifies analysis to have beaming be a property of notes rather than a separate element.</p>
```
bwbohl: As suggested by @pe-ro 

```suggestion
                  <p>Although <att>beam</att> can be used for rendering beams, <gi scheme="MEI">beam</gi> or <gi scheme="MEI">beamSpan</gi> is preferred over <att>beam</att> in most cases. However, because it makes beaming a <emph>property</emph> of a note, chord, rest, or space rather than an <emph>entity</emph> in its own right, <att>beam</att> can provide a flatter structure that simplifies analytical tasks.</p>
```

bwbohl: as suggested by @maxrothman 

```suggestion
                  <p>In general, it is recommended to use <gi scheme="MEI">beam</gi> or <gi scheme="MEI">beamSpan</gi> over <att>beam</att> for typical notation use-cases.</p>
```

musicEnfanthen (approved):

martha-thomae (approved):

Issue #1525, opened 2024-08-13
schema: indicate the extent of a scribal hand in `<hand>`


ahankinson:
Currently there is no way to specify the scope of a hand on a document. A conversation on Slack highlighted that there is currently no way of saying "hand A is the default hand of this MS, with occasional interventions by hand B."

One possible solution that was proposed by @pe-ro is to borrow `@scope` from TEI: https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.handFeatures.html

There was contradictory remarks on `<hand>` that `@initial` could be used as the main hand (#1524) but this is insufficient as it does not allow one to say "hand A is the initial, but most of the rest of the document is in hand B."

PR #1526, opened 2024-08-13
docs: remove contradictory remarks on `<hand>`

### LINKED ISSUE #1524: Remarks on `<hand>` are contradictory
As noted in a Slack conversation, the Remarks section of `<hand>` and the documentation for `@initial` are not aligned. The remarks say (emphasis mine):

> The `@initial` attribute indicates whether this is the first **or main** hand of the document. 

While the documentation for `@initial` says:

> Marks this hand as the first one of the document.

The discussion on Slack (with @pe-ro and @maxrothman) indicated that using `@initial` for the main hand of the document was insufficient, and an alternate solution should be found, possibly borrowing TEI's `@scope` from `<handNote>` (https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.handFeatures.html)

In any case, the remarks and the description should be aligned, so a PR is incoming to remove the conflict. A separate issue will also be created for `@scope`.

ahankinson:
This PR removes a statement from the remarks of `<hand>` that contradicts the usage statement on the `@initial` attribute.

Fixes #1524
Refs #1525

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -1454,7 +1454,7 @@
       </attDef>
     </attList>
     <remarks xml:lang="en">
-      <p>The <att>initial</att> attribute indicates whether this is the first or main hand of the
+      <p>The <att>initial</att> attribute indicates whether this is the first hand of the
         document. The <att>medium</att> attribute describes the writing medium, <abbr>e.g.</abbr>, 
         <val>pencil</val>, or the tint or type of ink, <abbr>e.g.</abbr>, <val>brown</val>. 
         The <att>resp</att> attribute contains an ID reference to an element containing the name of 


source/modules/MEI.header.xml
```diff
@@ -1454,7 +1454,7 @@
       </attDef>
     </attList>
     <remarks xml:lang="en">
```
rettinghaus, line 1456: Please add a versionDate:
```suggestion
    <remarks xml:lang="en" versionDate="2024-08-13">
```

rettinghaus (approved):
Thanks for this fix.

source/modules/MEI.header.xml
ahankinson: What does this do? I thought the version will be `5.1`?

rettinghaus: Just for keeping track when descriptions/remarks were updated. Super annoying to go through commits trying to find out. (In use by TEI since 2005.)

ahankinson: Uhm... is that the correct usage of `versionDate`? The description seems to indicate something about translation? 

https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.translatable.html

ahankinson: Shouldn't it be "`@change`"? https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.global.change.html

rettinghaus: All languages carry it: https://github.com/TEIC/TEI/blob/dev/P5/Source/Specs/att.translatable.xml
We can discuss it in an ODD meeting.

ahankinson: Yes, I know. But it seems to indicate that it is used to mark a version date for something that is being sent off to translation.

"specifies the date on which the source text was extracted and sent to the translator"

"The versionDate attribute can be used to determine whether a translation might need to be revisited, by comparing the modification date on the containing file with the versionDate value on the translation. If the file has changed, changelogs can be checked to see whether the source text has been modified since the translation was made."

Which I don't think is what we want here?

ahankinson: > We can discuss it in an ODD meeting.

I just looked, and I think the next ODD meeting I can make it to will be 29 November. Surely we can get this merged before then?

rettinghaus: It's just a suggestion, I already approved.

lpugin: I don't think having `@versionDate` is a very helpful practice. Maybe it was in 2005, but now that everything is fully versioned by git it does not make much sense any more. What would you expect to happen to `@versionDate` when the content of the remark is edited? To be updated or to remain the same? It is not satisfactory either way. Looking at the history is much more precise.

pe-ro: I agree with @lpugin.

musicEnfanthen (approved):

Issue #1518, opened 2024-08-10
guidelines: "<", ">", and "@" in referenced elements/attributes are unsearchable/unselectable


maxrothman:
Frequently when scanning through the MEI guidelines, I'm looking for references to a specific element or attribute. To do this, I use my browser's search tool to search for `<tagname` if I'm looking for an element or `@attname` if I'm looking for an attribute. I also might use the guideline's global search or use google to search globally through the guidelines.

Also, when collaborating with others in reference to sections of the guidelines, I often copy passages and paste them in slack, github issues, etc.

In all above cases, these tasks are made more difficult by the fact that the `<`, `>`, and `@` characters decorating element and attribute names in the guidelines cannot be selected, searched, or copied because they're added by content in the `:before`/`:after` pseudo-elements (e.g. [here](https://github.com/music-encoding/music-encoding/blob/ac400fe7820241bd69ccd20f16c71932d1859a14/source/assets/css/screen/mei-website.css#L4252-L4253)). Pseudo-element content [is not in the DOM](https://stackoverflow.com/questions/19914349/how-can-i-make-generated-content-selectable) and therefore can't be selected, searched, indexed by search engines, etc.

I think a straightforward fix for this would be to add those characters instead during the xml transformation process, e.g. [here for elements](https://github.com/music-encoding/music-encoding/blob/ac400fe7820241bd69ccd20f16c71932d1859a14/utils/guidelines_xslt/odd2html/guidelines.xsl#L129-L142). Unfortunately I don't have much experience with xslt so I can't contribute those changes without a little help. If you can provide an example of how one might add those characters in xslt, I'm happy to make a pull request if that's helpful.

musicEnfanthen:
Thanks for reporting! I agree that it would be helpful to be able to select these "markers".

Did a quick research, and at least the `@` symbol issue can definitely be resolved (have it already working on a test branch). Not sure if it is similarily straightforward with the angle brackets, though.

maxrothman:
Thanks for the reply!

I found the code that generates element and attribute names in the xslt while doing research before filing this issue and my (albeit xslt-uninformed) instinct is that they’re both fairly similar; if you can prefix with a character for attributes then it should be straightforward to prefix and suffix for elements? The top post links to the xslt that generates elements of that helps.

musicEnfanthen:
Thanks. 

The issue with those additional signs is that they will need to be included in many more (and much more different) places than one would think....

[...]

PR #1529, opened 2024-08-30
utils: improve searchability of attributes in the guidelines

musicEnfanthen:
This PR improves the searchability of attributes in the guidelines by removing the css-driven pseudo-`@` and includes literal `@`-signs in the utils scripts.

Fixes part of #1518 .

(This is for testing purposes. If we decide to go this route, we can also consider enhancements for elements).

### DIFF
--- a/source/assets/css/print/mei.css
+++ b/source/assets/css/print/mei.css
@@ -737,10 +737,6 @@ div.facet.attributes div.classItem > label {
     display: none;
 }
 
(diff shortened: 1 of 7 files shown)


musicEnfanthen:
Not quite sure if it would be better to use the html-escaped version of the `@`-sign in the scripts, `&#64;`?

musicEnfanthen:
Superseded by #1531

PR #1531, opened 2024-09-21
utils: make attribute and element symbols (`@`, `<`, `>`) searchable & selectable

### LINKED ISSUE #1518: guidelines: "<", ">", and "@" in referenced elements/attributes are unsearchable/unselectable

maxrothman:
Fixes #1518

Previously these symbols were added in css via pseudo-elements, which are not searchable or selectable on the page (see linked issue for details on how this results in poor UX). This change generates these symbols instead through xslt.

### DIFF
--- a/utils/guidelines_xslt/odd2html/generateSearchIndex.xsl
+++ b/utils/guidelines_xslt/odd2html/generateSearchIndex.xsl
@@ -184,7 +184,7 @@
                     let a = document.createElement('a')
                     a.setAttribute('href', pathAdjust + hit.item.url)
                     a.classList.add(hit.item.type)
-                    a.innerHTML = hit.item.ident
+                    a.innerHTML = "&amp;lt;" + hit.item.ident + "&amp;gt;"
                     div.append(a)
                     
                     if(hit.item.type === 'chapter') {
--- a/utils/guidelines_xslt/odd2html/generateSearchIndex.xsl
+++ b/utils/guidelines_xslt/odd2html/generateSearchIndex.xsl
@@ -184,9 +184,12 @@
                     let a = document.createElement('a')
                     a.setAttribute('href', pathAdjust + hit.item.url)
                     a.classList.add(hit.item.type)
-                    a.innerHTML = "&amp;lt;" + hit.item.ident + "&amp;gt;"
+                    a.innerHTML = hit.item.ident
+                    if (hit.item.type === "element") {
+                        a.innerHTML = "&amp;lt;" + a.innerHTML + "&amp;gt;"
+                    }
                     div.append(a)
-                    
+
                     if(hit.item.type === 'chapter') {
                         
                         let remarksHits = hit.matches.filter(m => m.key === 'remarks')
--- a/utils/guidelines_xslt/odd2html/specs/genericSpecFunctions.xsl
+++ b/utils/guidelines_xslt/odd2html/specs/genericSpecFunctions.xsl
@@ -181,7 +181,7 @@
                             <xsl:value-of select="', '"/>
                         </xsl:if>
                         <span class="ident element" title="{$desc}">
-                            <a class="link_odd_elementSpec" href="{tools:linkToElement($current.elem)}"><xsl:value-of select="$current.elem"/></a>
+                            <a class="link_odd_elementSpec" href="{tools:linkToElement($current.elem)}">&lt;<xsl:value-of select="$current.elem"/>&gt;</a>
                         </span>
                     </xsl:for-each>
                 </xsl:variable>-->
@@ -212,7 +212,7 @@
                                 <xsl:sort select="." data-type="text"/>
                                 <xsl:variable name="current.elem" select="." as="xs:string"/>
                                 <item class="element" ident="{$current.elem}" module="{$elements/self::tei:elementSpec[@ident = $current.elem]/@module}">
-                                    <link><a class="{tools:getLinkClasses($current.elem)}" href="#{$current.elem}"><xsl:value-of select="$current.elem"/></a></link>
+                                    <link><a class="{tools:getLinkClasses($current.elem)}" href="#{$current.elem}">&lt;<xsl:value-of select="$current.elem"/>&gt;</a></link>
                                     <desc><xsl:apply-templates select="$elements/self::tei:elementSpec[@ident = $current.elem]/tei:desc" mode="guidelines"/></desc>
                                 </item>
                             </xsl:for-each>
@@ -290,7 +290,7 @@
                     <xsl:value-of select="', '"/>
                 </xsl:if>
                 <span class="ident element" title="{$desc}">
-                    <a class="link_odd_elementSpec" href="{tools:linkToElement($current.elem)}"><xsl:value-of select="$current.elem"/></a>
+                    <a class="link_odd_elementSpec" href="{tools:linkToElement($current.elem)}">&lt;<xsl:value-of select="$current.elem"/><&gt;/a>
                 </span>
             </xsl:for-each>
             
@@ -339,7 +339,7 @@
                     </span>
                 </div>-->
                 <item class="element" ident="{$ident}" module="{$elements/self::tei:elementSpec[@ident = $ident]/@module}">
-                    <link><a class="{tools:getLinkClasses($ident)}" href="#{$ident}"><xsl:value-of select="$ident"/></a></link>
+                    <link><a class="{tools:getLinkClasses($ident)}" href="#{$ident}">&lt;<xsl:value-of select="$ident"/>&gt;</a></link>
                     <desc><xsl:apply-templates select="$elements/self::tei:elementSpec[@ident = $ident]/tei:desc" mode="guidelines"/></desc>
                 </item>
             </xsl:for-each>
@@ -417,7 +417,7 @@
                 <xsl:variable name="desc" select="normalize-space(string-join(tei:desc//text(),' '))" as="xs:string?"/>
                 
                 <item class="element" ident="{$current.element/@ident}" module="{$current.element/@module}">
-                    <link><a class="{tools:getLinkClasses($current.element/@ident)}" href="#{$current.element/@ident}"><xsl:value-of select="$current.element/@ident"/></a></link>
+                    <link><a class="{tools:getLinkClasses($current.element/@ident)}" href="#{$current.element/@ident}">&lt;<xsl:value-of select="$current.element/@ident"/>&gt;</a></link>
                     <desc><xsl:apply-templates select="tei:desc" mode="guidelines"/></desc>
 
                     <xsl:variable name="attributes" select="$current.element//tei:attDef[.//rng:ref[@name = $object/@ident]]" as="node()+"/>
@@ -433,7 +433,7 @@
                 </item>
                 
                 <!--<span class="ident element" data-ident="{$current.element/@ident}" data-module="{$current.element/@module}" title="{$desc}">
-                    <a class="{tools:getLinkClasses($current.element/@ident)}" href="#{$current.element/@ident}"><xsl:value-of select="$current.element/@ident"/></a>
+                    <a class="{tools:getLinkClasses($current.element/@ident)}" href="#{$current.element/@ident}">&lt;<xsl:value-of select="$current.element/@ident"/>&gt;</a>
                 </span>-->
             </xsl:for-each>
         </xsl:variable>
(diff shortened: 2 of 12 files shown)


utils/guidelines_xslt/odd2html/specs/genericSpecFunctions.xsl
```diff
@@ -290,7 +290,7 @@
                     <xsl:value-of select="', '"/>
                 </xsl:if>
                 <span class="ident element" title="{$desc}">
-                    <a class="link_odd_elementSpec" href="{tools:linkToElement($current.elem)}"><xsl:value-of select="$current.elem"/></a>
+                    <a class="link_odd_elementSpec" href="{tools:linkToElement($current.elem)}">&lt;<xsl:value-of select="$current.elem"/><&gt;/a>
```
musicEnfanthen:
```suggestion
                    <a class="link_odd_elementSpec" href="{tools:linkToElement($current.elem)}">&lt;<xsl:value-of select="$current.elem"/>&gt;</a>
```

utils/guidelines_xslt/odd2html/generateSearchIndex.xsl
```diff
@@ -184,7 +184,7 @@
                     let a = document.createElement('a')
                     a.setAttribute('href', pathAdjust + hit.item.url)
                     a.classList.add(hit.item.type)
-                    a.innerHTML = hit.item.ident
+                    a.innerHTML = "&amp;lt;" + hit.item.ident + "&amp;gt;"
```
musicEnfanthen: Unfortunately, this will result in every search result being embraced by brackets:

Screenshot 2024-09-27 114350

musicEnfanthen: Reverting this change for innerHTML should fix the issue.

utils/guidelines_xslt/odd2html/specs/genericSpecFunctions.xsl
```diff
@@ -1405,12 +1405,12 @@
                     <xsl:variable name="current.elem" select="@ident" as="xs:string"/>
                     <xsl:variable name="desc" select="normalize-space(string-join(./tei:desc//text(),' '))" as="xs:string"/>
                     <item class="element" ident="{$current.elem}" module="{$elements/self::tei:elementSpec[@ident = $current.elem]/@module}">
-                        <link><a class="{tools:getLinkClasses($current.elem)}" href="#{$current.elem}"><xsl:value-of select="$current.elem"/></a></link>
+                        <link>&lt;<a class="{tools:getLinkClasses($current.elem)}" href="#{$current.elem}"><xsl:value-of select="$current.elem"/>&gt;</a></link>
```
musicEnfanthen:
```suggestion
                        <link><a class="{tools:getLinkClasses($current.elem)}" href="#{$current.elem}">&lt;<xsl:value-of select="$current.elem"/>&gt;</a></link>
```

musicEnfanthen: Maybe related to the fact that MayContainFacet now is missing the opening bracket:

Screenshot 2024-09-27 115702

musicEnfanthen: To be checked after changes.

musicEnfanthen:
Also, all references to elements in the chapter texts are missing the brackets now:

Screenshot 2024-09-27 120153

musicEnfanthen:
Thank you @maxrothman for this PR. Most of the changes are working, except for some critical things mentioned above. 

It seems that we are still missing the spot where the element links in the guideline chapters are created.

The resulting html for these elements would be now: 

`<a class="link_odd link_odd_elementSpec" href="../elements/meiHead.html">meiHead</a>`, 

so maybe this gives a hint where to find this spot in the xslt scripts?

musicEnfanthen:
PS: If you want to check the rendered output yourself, you can use the `ant build-guidelines-html` command from the build tools. `Building MEI on the commandline` is described here: https://github.com/music-encoding/music-encoding/blob/develop/BUILD_COMMANDLINE.md

rettinghaus:
I thought about this for a while now and I'm still not sure what to make of this or if the proposed solution even is indeed a _solution_. If you really need to search for / copy/paste things _with_ brackets why not take the PDF instead.
I'm wondering if the search itself or the search index could be adjusted to allow searching for element names with brackets. On the other hand, if you search for 'beam' you'll find all places where the word appears, but usually they are in the vicinity of the corresponding element name …

maxrothman:
> I thought about this for a while now and...

Just so we're sure we're on the same page, I came up with some examples on the CMN guideline page:
- Searching for "note" turns up 577 results
- Searching for "measure" turns up 239 results
- Searching for "chord" turns up 103 results
- Searching for "label" (e.g. if you're looking for `<note>`'s `@label`) turns up 30 results, many of which are about `<label>`

So while it may be fine in certain cases (e.g. "beam"), the UX is really bad for others. Sure, if the site search was really good such that it could find references to elements/attributes in guideline pages and take you straight there then this may not be strictly necessary, but it's not and this is a much easier change to make.

To address your PDF suggestion directly, let's say I'm working with a colleague about a piece of notation while I read the guidelines and I want to quote a portion of the guidelines in a message to them. My options are basically a) manually re-add the missing characters or b) pull open the PDF, find the exact same spot as I was already browsing on the web, then copy it. That of course assumes that as a less-technical user I even _know_ that copying out of the PDF will solve my issue.

Just on another note, the general recommendation to not to insert meaningful characters via CSS `content` is fairly well-known, particularly because it breaks user expectations about how webpages work wrt searching/copying. For example, the following note appears on the [MDN page about CSS content](https://developer.mozilla.org/en-US/docs/Web/CSS/content):
> CSS-generated content is not included in the DOM. Because of this, it will not be represented in the accessibility tree and certain assistive technology/browser combinations will not announce it. If the content conveys information that is critical to understanding the page's purpose, it is better to include it in the main document.

So even aside from the search issues, this change brings the MEI guidelines in line with that standard practice and makes the guidelines more accessible.

utils/guidelines_xslt/odd2html/generateSearchIndex.xsl
maxrothman: Unfortunately it seems like making no change makes brackets _never_ appear, even for element search results. I have a change incoming where adding the brackets is behind a conditional, please stand by :)

maxrothman:
Ok, so my latest push includes a fix for the search results page and a fix for element references in the guidelines. Unfortunately, the latter fix also introduces an issue with some html `@title`s on attribute definitions (`.attributeDef .ident.attribute`) on attribute and element descriptions (e.g. https://music-encoding.org/guidelines/v5/attribute-classes/att.accid.log.html#attributes_full_tab). For example, on the aforementioned page, `@when` looks like:
```html
<span class="ident attribute" title="<when&gt; Indicates the point of occurrence of this feature along a time line. Its value must be the ID of a element elsewhere in the document.">@when</span>
```
Interestingly, this issue only seems to occur when the title starts with a reference, and these references don't actually appear in [the attribute's description](https://github.com/maxrothman/music-encoding/blob/7721a63b79c339029f9fbedce621bdf968d7d371/source/modules/MEI.performance.xml#L43)!

There are a few other examples of this behavior, e.g. `@altsym` on `att.altSym` starts with a reference to `<symbolDef>`.

As far as I can tell, attribute titles come from [this function](https://github.com/maxrothman/music-encoding/blob/7721a63b79c339029f9fbedce621bdf968d7d371/utils/guidelines_xslt/odd2html/specs/genericSpecFunctions.xsl#L628), but I can't make heads or tails of why some attributes have identifiers stuck on the front of their titles and most don't. At this point I'm reaching the limit of my capabilities with XSLT. I also don't really understand why my change [over here](https://github.com/maxrothman/music-encoding/blob/7721a63b79c339029f9fbedce621bdf968d7d371/utils/guidelines_xslt/odd2html/guidelines.xsl#L136) caused this issue. Help would be very much appreciated!

musicEnfanthen:
@maxrothman Thanks for digging further into this. 

Regarding the title issue: I think there has already been a bug in the title generation before your changes. When you look at the description for, e.g., tstamp attributes,  there is a `<abbr>i.e.</abbr>` in the middle of that desc in the canonicalized source (the source for the creation of the website content):

```
<attDef ident="tstamp" usage="opt">
        <gloss xml:lang="en">time stamp</gloss>
        <desc xml:lang="en">Encodes the onset time in terms of musical time, <abbr>i.e.</abbr>, beats[.fractional beat part],
          as expressed in the written time signature.</desc>
...
</attDef>
```

However, the generated span than reads:
`<span class="ident attribute" title="i.e. Encodes the onset time in terms of musical time, , beats[.fractional beat part], as expressed in the written time signature.">tstamp</span>`

So it appears, that anything within angled brackets seems to be processed first and gets moved to the beginning of the title. That would explain why there is a "when" at the beginning of the `@when` attribute title, as the inner `<when>` element gets moved to the beginning:
```

<attDef ident="when" usage="opt">
        <desc xml:lang="en">Indicates the point of occurrence of this feature along a time line. Its value must be
          the ID of a <gi scheme="MEI">when</gi> element elsewhere in the document.</desc>
        <datatype>
          <rng:ref name="data.URI"/>
        </datatype>
      </attDef>
```

And the resulting span has the inner when element moved:

`<span class="ident attribute" title="when Indicates the point of occurrence of this feature along a time line. Its value must be the ID of a element elsewhere in the document.">when</span>`

musicEnfanthen:
I will have a closer look where and why this is happening.

In the meantime, all the brackets seem to be generated correctly, except for the blue-ish tiles in the guideline chapters (cf. here: https://github.com/music-encoding/music-encoding/pull/1531#issuecomment-2378915374):

```
<div class="tile-title">
  <span class="specList-elementSpec">
    <a class="link_odd link_odd_elementSpec" href="../elements/music.html">music</a>
  </span>
</div>
```

maxrothman:
I finally had time to bring this PR back from the dead! And I believe I've correctly added the angle brackets to elements in specLists.

It looks like the bug mentioned in https://github.com/music-encoding/music-encoding/pull/1531#issuecomment-2380708315 is still present though. Now that I am slightly more proficient with XSLT I'm happy to take another look at it, any hints or context would be very much appreciated.

maxrothman:
Hey @musicEnfanthen, any chance you've got a sec to take another look at this PR? I _think_ it should be ready to go now, modulo that bug you mentioned in https://github.com/music-encoding/music-encoding/pull/1531#issuecomment-2380704829. I'm happy to dig into it a little more but it'd help to have some hints to get started on it.

musicEnfanthen:
@maxrothman Finally could have a look into it. 

I think it looks very good now! I could not find any missing or unnecessary brackets with the latest changes. 

It also fixes an ugly issue where the XSLTs generated brackets around elements that were referenced in an rng-ref in the declaration section of the rendered specs.

Before:
Screenshot 2025-04-07 142515

After:
Screenshot 2025-04-07 142445

Outsourced the title generation issue into #1662, so this PR would be good to go 🎉

musicEnfanthen (approved):

lpugin (approved):

PR #1532, opened 2024-09-26
docs: update copyright dates

rettinghaus:
This PR updates all copyright notices to the current year to match the recent changes in the tablature chapter.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- 
-  NOTICE: Copyright (c) 2017-2023 by the Music Encoding Initiative (MEI) Board (formerly known as "MEI Council").
+  NOTICE: Copyright (c) 2017-2024 by the Music Encoding Initiative (MEI) Board (formerly known as "MEI Council").
 
   Licensed under the Educational Community License, Version 2.0 (the "License"); you may
   not use this file except in compliance with the License. You may obtain a copy of the License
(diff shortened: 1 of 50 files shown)


lpugin:
Do we really need this date range? Maybe we can drop it so we don't have to update it every year....

rettinghaus:
It's common practice. And we didn't update it every year, only for the last release.

lpugin:
I think this is a useless practice. Having the creation date is fine, but having a date range quite inevitably becomes not updated at some point. That makes things look not maintained, or worse, not under copyright any more. Plus it creates some useless noise in the history of the files, so I would suggest to drop the end date.

ahankinson:
cURL has removed the years from the source code:

https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/

Facebook also dropped the years from the React source code:

https://github.com/facebook/react/commit/b87aabdfe1b7461e7331abb3601d9e6bb27544bc

Here's a blog post about the topic: 

https://hynek.me/til/copyright-years/

I'd be ok with removing the dates altogether.

musicEnfanthen:
I don't think it is useless, there is some value in having an end date. But you are right, the most important date is the first one which claims the start of the copyright (in case of legal disputes).

BTW, you do not loose copyright due to a wrong date. ;) In fact, author's rights are automatically applied under the Berne Convention so things are under copyright even if you wouldn't have a copyright statement (with some exceptions for the US, of course). The statement just makes these claims a bit more visible and easier to find.

This also seems to be some of the reasoning behind the decisions made by cURL et al. to skip the years altogether.

ahankinson:
We could update the LICENSE file to keep the years there? Then there would only be one place. Currently it doesn't have them.

musicEnfanthen:
Yes, I think that's in any case a good idea to add the dates there.

If we want to keep any dates elsewhere, maybe the solution facebook used before they removed the dates could be elegant (and does not require updates)? 

`Copyright (c) 2017-present ...`

musicEnfanthen:
Superseded by #1533

PR #1538, opened 2024-10-06
docs: refine multiRest description

rettinghaus:
A little refinement to clarify that this element combines multiple measures between two bar lines. Furthermore it is not necessarily a single symbol.


lpugin (approved):
Looks good to me.

bwbohl (approved):

PR #1549, opened 2024-11-18
build: extract build properties

musicEnfanthen:
This PR is a follow-up to #1428.

It extracts the version variables into a build properties file to facilitate maintenance. 
It also adds an info target that logs the Ant and Java version used, which is useful for debugging purposes.

Needs #1428.


build.properties
```diff
@@ -0,0 +1,7 @@
+meiversion.latest=5.0
+meiversion.next=5.1-dev
+saxon.edition.version=SaxonHE12-5
+saxon.version=12.5
+schematron.version=8.0.0
+stylesheets.version=7.57.1
+xerces.version=26.1.0.1
```
rettinghaus: Please always add a new line at eof, see https://gist.github.com/camh-/1bebfcff1b0f814e9b191edc60d5206b

rettinghaus (changes requested):
Almost perfect, thanks!

build.properties
musicEnfanthen: Thanks for the hint. Tried to add it, hope it worked?

rettinghaus (approved):
Thanks!

bwbohl (approved):
This is so much more convenient, thanks @musicEnfanthen!

PR #1552, opened 2024-11-20
guidelines: fix tstamp explanation

DILewis:
Guidelines explanation of timestamps ([section 1.3.3](https://music-encoding.org/guidelines/dev/content/introduction.html#timestamps)) states that tstamp=1 is the left bar line. I think it should be 0 (that seems to be how it's interpreted in Verovio).


musicEnfanthen:
@DILewis Thanks for the catch!

musicEnfanthen (approved):

rettinghaus (approved):
Great, thank you!

bwbohl (approved):
in 4.2.6 it sys left barline is 0, which is correct.
@DILewis thanks for spotting this grave issue

Issue #700, opened 2020-07-29
Polymeter


craigsapp:
How should polymetric music be encoded in MEI?

Example:

<img width="573" alt="Screen Shot 2020-07-29 at 3 35 18 AM" src="https://user-images.githubusercontent.com/3487289/88790354-e6e97280-d14c-11ea-8602-2c7212646783.png">

ahankinson:
Very carefully. 😁

craigsapp:
I thought so 😉 

Here is the real-world example that this is based on:

Arthur Honeggar, Pacific 231, Movement symphonique no. 1 (1923), measure 62:

<img width="306" alt="Screen Shot 2020-07-29 at 3 46 40 AM" src="https://user-images.githubusercontent.com/3487289/88791381-793e4600-d14e-11ea-9c7a-e4509b36ce80.png">


In measure 61, all parts are in 4/4, and in m. 62 some switch to 6/4  in the time of 4/4.

PDF of score:

http://conquest.imslp.info/files/imglnks/usimg/5/56/IMSLP12979-Honegger_-_Pacific_231_(orch_score).pdf

rettinghaus:
We need a way to signal if beats (or tstamps) in polymeric music match between different staves, like in this example form Mozart's Don Giovanni:
![Mozart-DG](https://user-images.githubusercontent.com/7693447/88792702-efb56500-d19b-11ea-831f-72ca572e5c76.png)

craigsapp:
Probably some attribute in `<staffDef>`?

craigsapp:
Here is how I encode such music in Humdrum:

```
**kern  **kern
*clefG2 *clefG2
*M6/4   *M4/4
=1      =1
*rscale:3/2     *
6e      4cc
6f      .
.       4b
6g      .
6e      4a
6f      .
.       4g
6g      .
=2      =2
6g      4f
6f      .
.       4e
6e      .
6g      4d
6f      .
.       4c
6e      .
=       =
*-      *-
```

I encode all parts in a normalize time system (encoding the 6/4 quarter notes as triplet quarter notes in 4/4 in this case), and then I have a rhythm scaling factor of 3/2 to increase the duration of the notes for visual display (converting the triplet-sounding quarter notes into visual quarter notes).

[...]

PR #1554, opened 2024-11-21
schema: make additional attribute classes member of att.duration.ratio

lpugin:
This PR makes the following att.* classes member of att.duration.ratio:
* `att.layerDef.log`
* `att.scoreDef.log`
* `att.staffDef.log`

This addresses the simple cases of polymeter as discussed and agreed in #700.

Dropping `@num.default` and `@numbase.default` might be done at a later stage if desired.


rettinghaus:
If this is to address durational ratios between staves, does it make sense to also have it on `scoreDef`?

lpugin:
> If this is to address durational ratios between staves, does it make sense to also have it on scordeDef?

The most common use is quite likely to be on `staffDef`. However, having it on `scoreDef` might make sense for successive `socreDef`s, or for `scoreDef` in `parts/part`.

rettinghaus (approved):

bwbohl:
I like the idea and am not opposed to it, but given some open discussions and the lack of compound examples for everything concerning `att.duration.default` and `att.duration.ratio` in the guidelines, I would like to push this to a later release. Could we tackle this during the developer meeting?

cf. #1071 and https://music-encoding.org/guidelines/v5/attribute-classes/att.duration.ratio.html

lpugin:
I am glad you like the idea. However, I think the `*.default` question is essentially unrelated, so I'm not sure what else needs to be discussed.

bwbohl:
I think

- [ ] we would need a corresponding example for encoding “simple polymeter” in the guidelines.
- [ ] Moreover, the relation of `num` and `numbase` on one staff vs their absence on another `staffDef` should be clarified
- [ ] and the relation to `default.*`attributes (especially when co-occurring).

musicEnfanthen:
ODD meeting 11/2024: This should definitely be accompanied by instructive examples and guideline documentation about the meaning and usage of the corresponding attributes.

Therefore, moving to v6 (and dev workshop 2025).

lpugin:
Sounds good. I can prepare the examples and describe the relation to other staves - which I think will be made explicit by the examples.

However, I am not sure what to do regarding the relation to `default.num` and `default.numbase` since I do not really understand the use-case for these attributes. If somebody else can add some explanations in the guidelines about what they should currently be used for, that would be helpful and I should be able to take it from there.

pe-ro:
@lpugin, I agree that `num.default`, and `numbase.default` don't have anything to do with making `<staffDef>` and `<scoreDef>` members of att.duration.ratio.  But, keeping these attributes may cause confusion and removing them may lead to another issue.

The history of these attributes (`@dur.default`, `num.default`, and `numbase.default`) is long and complex, but they do have a purpose.  The need for them arises because `@dur` (as well as `@num` and `@numbase`) on `<note>` is optional.  When `@dur` isn't provided, the note's duration must be determined in some other fashion.  A common case is when `@dur` isn't given because the note is a member of a chord, where the duration is specified.  An uncommon, but legal case is when all the notes of a measure have the same duration, `@dur` may only be provided on the first one.  Going further, if all the notes on a staff or even in the entire piece have the same duration, then `@dur` need not be given on any of the notes, but provided via `@dur.default` on `<staffDef>` or `<scoreDef>` instead, as appropriate.  

`@num.default` and `@numbase.default` support similar functionality for mensural notation.  Like, `@num` and `@numbase` on `<note>`, they provide a default duration (given on `<staffDef>` or `<scoreDef>`), not a ratio between staves.

To avoid the confusion created by the presence of `@num` and `@numbase` as well as  `@num.default` and `@numbase.default` on `<staffDef>` and `<scoreDef>` (already exhibited amongst ourselves), we can simply remove the *.default attributes, but then another mechanism must be created to deal with the fact that duration is optional.  Making `@dur` required is off the table due to the problem it creates with chords and because the change wouldn't be backward-compatible.  Renaming the *.default attributes, creating a Schematron rule that warns against the lack of duration info, and/or addressing this situation in the documentation are reasonable alternatives.  I'm not in favor of a simple  "Don't do that" or "That never happens" dismissal of the problem.  If something is valid, then someone, somewhere will eventually do it.

lpugin:
@bwbohl would you be able to clarify what you would like to see explained regarding `default.num` and `default.numbase`?

bwbohl (approved):

bwbohl:
Maybe we should just add another issue for adding to the Guidelines a small section:

"Encoding Polymeter"
* what kind of polymeter is supported -> simple polymeter
* what we understand under the term "simple polymeter"
* how to encode it
* An example could be those from the original issue
* how to encode it
  * Does every staffDec need num numbase, or what’s the implied meaning if one doesn’t?
  * is there any implication from the *default attributes

Thoughts?

pe-ro:
The general problem of representing polymeter is still unresolved.  It seems to me there are 2 approaches:

1.  use attributes on scoreDef to specify a ratio between staves, or
2.  explicitly encode the polymeter using something like I proposed in #700.

I don't think we need to pick one. Both solutions are good for different uses.  I'm not convinced these solutions have been thoroughly vetted, though.

bwbohl:
@pe-ro I think the issue and the outcome of its discussion give reason enough to add something to MEI once again that others might prove wrong in the future ;-) We want to tackle that simple case of measures aligning, so we have something to build upon for more complicated cases in the future.

Essential to getting this vetted more is a section in the Guidelines advertising "How to deal with Polymeter in MEI"

martha-thomae:
@lpugin, do you think you can have the examples in the guidelines by the next ODD Meeting (March 27th)? Then we can just proceed to directly merge this PR at the meeting.

@pe-ro, would you mind helping Laurent with this? Laurent could focus on the first two items of the list provided by Benni and you, Perry, on the last one(?) https://github.com/music-encoding/music-encoding/pull/1554#issuecomment-2503396892

> * [ ]  We would need a corresponding example for encoding “simple polymeter” in the guidelines.
> * [ ]  Moreover, the relation of `num` and `numbase` on one staff vs their absence on another `staffDef` should be clarified
> * [ ]  and the relation to `default.*`attributes (especially when co-occurring).

I will be happy to merge this during or before the next ODD Meeting, if you have the examples ready by then :)

pe-ro:
There are several examples of polymetric notation at https://github.com/music-encoding/music-encoding/issues/700.  My hope is that @lpugin can adapt these.

martha-thomae:
Thanks for pointing that out, Perry!

PR #1555, opened 2024-11-23
schema: remove defaultVal from meiversion attributes

### LINKED ISSUE #1347: Use of a:defaultValue causes problems with schematron for <tempo>

musicEnfanthen:
This PR removes the defaultVal from the meiversion attributes in the customizations.

Fixes #1347

TODOs

- [x] fix schematron rule for tempo about allowed attributes when not in score or part in anyStart customisation
- [x] add schematron rule for meiversion by @pe-ro from #1347 
- [x] Fix rule schematron rule for tempo https://github.com/music-encoding/music-encoding/issues/1347#issuecomment-1751286288
- [x] make meiversion easily accessible, e.g., encode meiversion in editionStmt
- [x] get rid of meiversion defaultVal
- [x] fix usage of defaultVal in comparison.xsl
  https://github.com/music-encoding/music-encoding/blob/0c10d2a53ffa5212dd17636cf497a786cfb40551/utils/compare_versions/comparison.xsl#L18

   https://github.com/music-encoding/music-encoding/blob/0c10d2a53ffa5212dd17636cf497a786cfb40551/utils/compare_versions/comparison.xsl#L34

- [x] fix usage of defaultVal in odd2html.xsl
  https://github.com/music-encoding/music-encoding/blob/0c10d2a53ffa5212dd17636cf497a786cfb40551/utils/guidelines_xslt/odd2html.xsl#L110)

### DIFF
--- a/customizations/mei-CMN.xml
+++ b/customizations/mei-CMN.xml
@@ -105,8 +105,8 @@
         <classSpec ident="att.meiVersion" module="MEI.shared" type="atts" mode="change">
           <attList>
             <attDef ident="meiversion" usage="rec" mode="change">
-              <defaultVal>5.1-dev+CMN</defaultVal>
               <valList type="closed" mode="change">
+                <valItem ident="5.1-dev+CMN"/>
                 <valItem ident="5.1-dev+anyStart" mode="delete"/>
                 <valItem ident="5.1-dev+basic" mode="delete"/>
                 <valItem ident="5.1-dev+Mensural" mode="delete"/>
--- a/customizations/mei-CMN.xml
+++ b/customizations/mei-CMN.xml
@@ -106,6 +106,7 @@
           <attList>
             <attDef ident="meiversion" usage="rec" mode="change">
               <valList type="closed" mode="change">
+                <valItem ident="5.1-dev"/>
                 <valItem ident="5.1-dev+CMN"/>
                 <valItem ident="5.1-dev+anyStart" mode="delete"/>
                 <valItem ident="5.1-dev+basic" mode="delete"/>
--- a/customizations/mei-Mensural.xml
+++ b/customizations/mei-Mensural.xml
@@ -106,6 +106,7 @@
           <attList>
             <attDef ident="meiversion" usage="rec" mode="change">
               <valList type="closed" mode="change">
+                <valItem ident="5.1-dev"/>
                 <valItem ident="5.1-dev+Mensural"/>
                 <valItem ident="5.1-dev+anyStart" mode="delete"/>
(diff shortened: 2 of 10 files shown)


musicEnfanthen:
@bwbohl Could you double-check if this would do the trick for #1347?

Left the defaultVal in shared, so 5.1-dev will be propagated to all customizations as additional option.

Also not totally happy about the descriptions for the custom versions in shared, maybe better: "MEI version"/"MEI version specific to basic customization." (etc.)? Or "Version of MEI"/"Version of MEI basic customization" (ect.)? 
[In any case, I would remove "development", so we wouldn't have to remove/reintroduce the "development" during releases...]

bwbohl:
As the `defaultValue` gets propagated to all customisation schemas we should completely remove it and add it as just another value in `MEI.shared.xml`.

rettinghaus (approved):
lgtm

bwbohl (approved):

PR #1556, opened 2024-11-25
schema: add usage to attribute definitions

rettinghaus:
This is a small PR adding a `usage` attributes to all attribute definitions, as most of them had one. 
It only adds the default value in all instances, so effective change to the schema.


musicEnfanthen (approved):

bwbohl (approved):
Thanks, this makes it more consistent and explicit encoding instead of relying on a default behaviour is always preferable.

Issue #1065, opened 2023-01-06
Analytical bracketSpan


rettinghaus:
The current set of values for `bracketSpan@func` is very limited. 
For _Hauptstimme_ and _Nebenstimme_ we can set the symbol with `@lstartsym`, but that reflects only the visual appearance. 

![](https://cloud.githubusercontent.com/assets/3487289/22934805/25e0cd58-f285-11e6-9e90-3f8e7a96e735.png)

It wood be good to have also the values "hauptstimme" and "nebenstimme" and/or the generic term "analytical" available. 

See #403

ahankinson:
Would you have an `@lstartsym` of "H" or "N" where it wasn't Hauptstimme or Nebenstimme?

IOW, why can't `@lstartsym` be both the symbol and the function? There's not really anything there that says it *must* "reflect only the visual appearance."

rettinghaus:
`@func` is required.

ahankinson:
Ah, good point. 

Maybe we should try to not have `@func` be required? Given that a lot of the symbols in start/end represent *not* one of ligature / coloration / cross-rhythm, and that sometimes a bracket is just a bracket, making `@func` optional would be another option?

rettinghaus:
That basically works for me. But even if a bracket is just a bracket, you should be able to point out that it serves an "analytical" purpose rather than an _editorial_ one (which the currently available values represent).

craigsapp:
> It wood be good to have also the values "hauptstimme" and "nebenstimme" and/or the generic term "analytical" available.

Those are also the names of the brackets in English, by the way (but often not capitalized).

> Would you have an @lstartsym of "H" or "N" where it wasn't Hauptstimme or Nebenstimme?

Those symbols are hardwired to have a single meaning, so an H-bracket always means Hauptstimme and N-bracket always means Nebenstimme (developed/used by Arnold Schoenberg).

His student, Alban Berg, invented (or at least used) additional analytic brackets, which are not common-knowledge in music like Hauptstimme and Nebenstimme:

https://imslp.eu/files/imglnks/euimg/2/28/IMSLP755449-PMLP9697-Berg_-_Violin_Concerto_CE.pdf
https://imslp.org/wiki/Violin_Concerto_(Berg,_Alban)

[...]

PR #1557, opened 2024-11-25
schema: add unspecified as value for bracket spans

rettinghaus:
This adds another value to the `func` attribute for `brackSpan` as proposed by @pe-ro in #1065.


musicEnfanthen:
@lpugin does that work for you?

lpugin (approved):

bwbohl (approved):

PR #1558, opened 2024-11-27
schema: simplify meterSigGrp model

### LINKED ISSUE #970: Nested meterSigGrps

rettinghaus:
This PR remodels `meterSigGrp` to use `model.meterSigLike` as proposed by @kepper 
closes #970

### DIFF
--- a/source/modules/MEI.cmn.xml
+++ b/source/modules/MEI.cmn.xml
@@ -1863,15 +1863,9 @@
       <memberOf key="model.meterSigLike"/>
     </classes>
     <content>
-      <rng:choice>
-        <rng:ref name="meterSig"/>
-        <rng:ref name="meterSigGrp"/>
-      </rng:choice>
+      <rng:ref name="model.meterSigLike"/>
       <rng:oneOrMore>
-        <rng:choice>
-          <rng:ref name="meterSig"/>
-          <rng:ref name="meterSigGrp"/>
-        </rng:choice>
+        <rng:ref name="model.meterSigLike"/>
       </rng:oneOrMore>
     </content>
   </elementSpec>


bwbohl:
Still an empty `<meterSigGrp/>` validates… Somehow an `<empty/>` element makes it to the parsed RNG:

```xml
<define name="mei_meterSigGrp">
  <element name="meterSigGrp">
     <a:documentation xmlns:a="http://relaxng.org/ns/compatibility/annotations/1.0">(meter signature group) Used to capture alternating, interchanging, mixed or other non-standard meter signatures.</a:documentation>
     <ref name="mei_model.meterSigLike"/>
     <oneOrMore>
        <ref name="mei_model.meterSigLike"/>
     </oneOrMore>
     <ref name="mei_att.facsimile.attributes"/>
     <ref name="mei_att.meterSigGrp.anl.attributes"/>
     <ref name="mei_att.meterSigGrp.ges.attributes"/>
     <ref name="mei_att.meterSigGrp.log.attributes"/>
     <ref name="mei_att.meterSigGrp.vis.attributes"/>
     <empty/>
  </element>
</define>
```

lpugin:
I am not sure I have a full understanding of the discussion here, but allowing an empty `meterSigGrp` might be needed in some cases. One typical case encountered elsewhere (beams) is when `@sameas` is given. This should be done by Schematron, but it might actually be simpler just to allow empty meterSigGrp. This might in fact also be useful when your encoding it just a high level one and used to represent what is on a page without detailed information about it.

rettinghaus:
@bwbohl You find this empty element at the end of every definition in the RNG. It has nothing to do with the allowed content.
How are you validating this?

bwbohl:
@rettinghaus I had a sample file with varying configurations of `meterSigGrp` from empty to nested several times. All of those did validate. But I got curious when I deleted the `fileDesc`, and still everything did validate.

I have to apologise: after a couple of follow-up tests, I am sure there is something astray with RNG validation in my @oxygenxml… still have to figure it out!

Now I did validate with jing and everything is being reported as we want it!

bwbohl (approved):

bwbohl:
As stated [here](https://github.com/music-encoding/music-encoding/pull/1558#issuecomment-2516624060) there is some bug in my validation scenario with oXygen XML. Jing did the job.

rettinghaus:
@lpugin Could you please provide an example with `@sameas`? I'm trying to understand in which context this could happen.

lpugin:
I am thinking the case with a mid-measure `meterSigGrp` with two layers. You would encode the one in the second layer with `meterSigGrp@sameas`.

musicEnfanthen (approved):

PR #1569, opened 2024-12-14
schema: extract data class for harp pedal position

rettinghaus:
This PR introduces `data.HARPPEDALPOSITION` for the pitch-named attributes in `att.harpPedal.log`, because they were all the same. 

Background: This doesn't change the schema itself, but helps in application built on it like LibMEI. Currently in LibMEI there are special data types for every single pitch, which means that you need a spearate converter for all of them.

### DIFF
--- a/source/modules/MEI.cmn.xml
+++ b/source/modules/MEI.cmn.xml
@@ -73,6 +73,22 @@
       </valList>
     </content>
   </macroSpec>
+  <macroSpec ident="data.HARPPEDALPOSITION" module="MEI.cmn" type="dt">
+    <desc xml:lang="en">Indicates the pedal setting for a harp strings.</desc>
+    <content>
+      <valList type="closed">
+        <valItem ident="f">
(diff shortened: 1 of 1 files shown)


lpugin (dismissed):

bwbohl (dismissed):

source/modules/MEI.cmn.xml
```diff
@@ -581,110 +597,54 @@
       <memberOf key="att.startEndId"/>
     </classes>
     <attList>
+      <attDef ident="d" usage="opt">
```
musicEnfanthen: @rettinghaus The order of the string attributes is now d, c, b, e, f ... correct?

bwbohl (approved):
I reinstated the original order

musicEnfanthen (approved):

musicEnfanthen:
Thank you @rettinghaus

rettinghaus:
My order was the actual order of the pedals, which seems to make more sense …

PR #1586, opened 2025-01-22
schema: replace version 5.1 with 6.0-dev

musicEnfanthen:
This PR udpates version 5.1 to the latest dev version.


bwbohl (dismissed):
thanks

lpugin (dismissed):

bwbohl (dismissed):

source/mei-source.xml
```diff
@@ -80,7 +80,7 @@
             </respStmt>
          </titleStmt>
          <editionStmt>
-            <edition>MEI 5.1</edition>
+            <edition>MEI 6</edition>
```
musicEnfanthen:
```suggestion
            <edition>MEI 6.0</edition>
```

musicEnfanthen: ping @bwbohl another one ;)

bwbohl:
I just stumbled over something else: the `/TEI/@version` in `mei-source.xml` and in all `docs` is set to `5.0`, which doesn’t make sense because the newest version of TEI actually is `4.9.0` and I thus have the impression that it is the MEI version that has been encoded here…

bwbohl:
> I just stumbled over something else: the `/TEI/@version` in `mei-source.xml` and in all `docs` is set to `5.0`, which doesn’t make sense because the newest version of TEI actually is `4.9.0` and I thus have the impression that it is the MEI version that has been encoded here…

Let's fix this in another pull request. (cf. https://github.com/music-encoding/music-encoding/pull/1608)

bwbohl (approved):

bwbohl:
ping @lpugin

lpugin (approved):

PR #1591, opened 2025-02-01
build: add validation on PR

### LINKED ISSUE #790: build: add odd validation to GH Actions

musicEnfanthen:
This PR adds a GH action to run source validation on every PR.

It also adds default permissions to the deploy workflow, to be consistent.

Fixes #790

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -1,4 +1,4 @@
-name: Validate Source
+name: PR Source Validation
 
 on:
   pull_request:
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-validation.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+    - name: Checkout head ref of repository (PR)
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      with:
+        repository: ${{github.event.pull_request.head.repo.full_name}}
+        ref: ${{github.event.pull_request.head.ref}}
+        fetch-depth: 0
+
+    - name: Run validation with Docker image on PR
+      run: docker run --rm -v $(pwd):/opt/docker-mei/music-encoding ghcr.io/music-encoding/docker-mei:latest ant validate-source -noinput -buildfile music-encoding/build.xml
```
rettinghaus, line 27: Why run Docker instead running `ant validate-source` directly?

rettinghaus (changes requested):

.github/workflows/pr-validation.yml
musicEnfanthen: To run run the ant target directly, you would need to setup java and ant with the correct/controlled versions inside the workflow resp. in all workflows that use the build script (like deploy). 
Using the docker image makes sure, we use the same controlled environment for everything. 
In addition, invoking the ant target via docker makes sure, that not only the ant target is working correctly, but that it can be correctly triggered from the docker image.

rettinghaus: Isn't that the whole idea of actions-runner? That you have a controlled environment by choosing e.g. [ubuntu-24.04](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md) that brings everything out of the box that is needed. My impression is that starting the docker container adds just overhead here.

rettinghaus: Ok, I see the point in not relying on the check _within_ the PR.

rettinghaus (approved):

bwbohl (approved):

PR #1596, opened 2025-02-05
schema: revise handling of agent-like elements (`model.respLikePart`)

### LINKED ISSUE #555: publisher of a work

### LINKED ISSUE #576: make composer, author etc. member of att.name?
According to the guidelines, `<composer>`, `<author>` and similar elements are supposed to contain the _name_ of a contributor. Shouldn't these person-related elements be members of att.name in order to allow reference to authority files, then? Or are they meant to contain `<persName>` elements marking the actual names in their contents?

### LINKED ISSUE #622: dedicatee is senseless
`<dedicatee>` isn't allowed anywhere it would make sense (like `<item>` or `<manifestation>`). Instead it is part of `model.textPhraseLike.limited`, which allows it in all those non-sense places like `<date>`, `<dynam>`, and `<tempo>`. 

The IG Metadata recommends to drop this element completely and use `@role="dedicatee"` whenever and wherever it is needed.

### LINKED ISSUE #1316: repurpose members of respLike.part
Creating a separate issue for the suggestion I made in the discussion of issue #555 where I said that the members of respLike.part (`<arranger>`, `<author>`, `<composer>`, `<contributor>`, `<editor>`, `<funder>`, `<librettist>`, `<lyricist>`, and `<sponsor>`) or their replacement (`<agent>`) can be made more biographical; that is, they can be used as containers for more detailed information that just that normally used in bibliographic description, such as those defined by TEI -- affiliation, age education, faith, dates of activity, gender, language knowledge, nationality, occupation, persona, residence, sex, socio-economic status, states (time-based characteristics), and traits (unalterable characteristics).  This list can also be extended to include other characteristics like place of birth and death.

riedde:
From the MEI Developer Workshop 2025 (Amsterdam)

Task: getting rid of the resp-like semantic wrappers, e.g., `<composer>` or `<lyricist>`

Idea: using `<respStmt>` containing `<name>`, `<persName>` and `<corpName>` with `@role`

- deleting the elements `<arranger>`, `<contributer>`,  `<composer>`, `<librettist>`, `<lyricist>`
- the element `<author>` implies the necessarity of an element `<composer>`. Therefore `<author>` will be renamed to `<creator>` (@pe-ro)
- updating guidelines (review is necessary)

**Side-effect: MEI-basic is including `<arranger>`, `<composer>`, `<lyricist>` manually. This PR deactivates this to avoid compilation failure.**

Closes #1435 
Closes #1434 
Closes #799 
Closes #1370
Closes #555 
Closes #1316 
Closes #576 
Closes #622


pe-ro (changes requested):
We need to keep `<contributor>`.  The members of respLikePart should be: creator (renamed from author), contributor, editor, funder, and sponsor.  The elements that should be removed are: arranger, librettist, and lyricist.

riedde:
@pe-ro I restored `<contributor>`.  What about `<dedicatee>`?  Is this another case because it can be used for transcription? I ask because a `<persName role="dte">` could also work. (see also #622 )

pe-ro:
`<dedicatee>` shouldn't be part of the respLikePart class because a dedicatee is not a "responsible party" included in the `<titleStmt>` or `<respStmt>`.  It belongs elsewhere -- certainly within `<work>`, maybe other places as well.

riedde:
@pe-ro I noticed that `<dedicatee>` is not documented in the GL at the moment what I like to change. I'm just thinking about the way. If we get rid of the semantic respLike-Elements (in some sence) it feels logical to me also to get rid of `<dedicatee>` because the mor generic `persName/@role` will do that job also. And this is already possible in `<respStmt>` even if it might be "wrong".

So why should we keep `<dedicatee>`?

If we keep it–and I think we should not–it must become part of the content of `<work>` also, or?

riedde:
@pe-ro Sorry, but once again. What about getting rid of `<dedicatee>` for the benefit of a `<dedication>`?

Or where is the current place to keep an information like this:
`Dedicated with honour to the <corpName role="dedicatee">“Harmonie Philadelphia” choral society </corpName> and its conductor <persName role="dedicatee">Mr Right</persName>`

pe-ro (dismissed):
Looks good to me!

customizations/mei-all.xml
```diff
@@ -119,7 +119,7 @@
             </attDef>
           </attList>
         </classSpec>
-
+
```
musicEnfanthen:
```suggestion
```

customizations/mei-basic.xml
```diff
@@ -95,8 +95,9 @@
                 <!--<moduleRef key="MEI.msDesc"/>-->
                 <moduleRef key="MEI.midi" include="instrDef" />
                 <moduleRef key="MEI.namesdates" include="persName" />
+                <!-- the following elements are not available anymore in MEI.shared: arranger, composer, lyricist -->
                 <moduleRef key="MEI.shared" 
-                    include="accid arranger artic body caesura chord clef clefGrp composer date dir dynam ending label labelAbbr layer lb lyricist mdiv mei music note ornam pb pgFoot pgHead pubPlace rend respStmt rest sb score scoreDef section space staff staffDef staffGrp symbol syl tempo title "/>
+                    include="accid artic body caesura chord clef clefGrp date dir dynam ending label labelAbbr layer lb mdiv mei music note ornam pb pgFoot pgHead pubPlace rend respStmt rest sb score scoreDef section space staff staffDef staffGrp symbol syl tempo title "/>
```
musicEnfanthen: @lpugin Can you have a look if there are any side effects for MEI Basic (MuseScore Import etc.) when removing these elements from shared. Do we need to further adjust the customization?

source/modules/MEI.shared.xml
```diff
@@ -7495,14 +7534,18 @@
...
+        <sch:rule context="mei:respStmt[not(ancestor::mei:change) and not(ancestor::mei:work)]">
           <sch:assert
             test="(mei:resp and (mei:name or mei:corpName or mei:persName)) or
             count(mei:*[@role]) = count(mei:*) and count(mei:*) &gt; 0"
             role="warning">At least one element pair (a resp element and a name-like element) is
             recommended. Alternatively, each name-like element may have a @role
             attribute.</sch:assert>
         </sch:rule>
+        <sch:rule context="mei:respStmt[ancestor::mei:work]">
+          <sch:report test="count(mei:*[@role]) &lt; count(mei:*) or count(mei:*) = 0" role="warning">Name-like elements with a @role are recommented here.</sch:report>
```
```suggestion
          <sch:report test="count(mei:*[@role]) &lt; count(mei:*) or count(mei:*) = 0" role="warning">Name-like elements with a @role are recommended here.</sch:report>
```

source/modules/MEI.shared.xml
```diff
@@ -7495,14 +7534,18 @@
...
           <sch:assert
             test="(mei:resp and (mei:name or mei:corpName or mei:persName)) or
             count(mei:*[@role]) = count(mei:*) and count(mei:*) &gt; 0"
             role="warning">At least one element pair (a resp element and a name-like element) is
             recommended. Alternatively, each name-like element may have a @role
             attribute.</sch:assert>
         </sch:rule>
+        <sch:rule context="mei:respStmt[ancestor::mei:work]">
+          <sch:report test="count(mei:*[@role]) &lt; count(mei:*) or count(mei:*) = 0" role="warning">Name-like elements with a @role are recommended here.</sch:report>
+          <sch:report test="mei:resp" role="warning">Name-like elements with a @role are recommented here (instead of resp).</sch:report>
```
```suggestion
          <sch:report test="mei:resp" role="warning">Name-like elements with a @role are recommended here (instead of resp).</sch:report>
```

ahankinson:
A couple questions arise: 

1. Can a single `<persName>` have multiple entries for `@role`?
2. If I enter a `<persName>` inside a `<dedication>` should I assume that that person is a dedicatee? If I have a dedication like "To my darling husband John who spent countless hours raising our children Robert and Mary while I composed this work", then I would assume only "John" gets `@role=dte`?

riedde:
> A couple questions arise:
> 
>     1. Can a single `<persName>` have multiple entries for `@role`?
> 
>     2. If I enter a `<persName>` inside a `<dedication>` should I assume that that person is a dedicatee? If I have a dedication like "To my darling husband John who spent countless hours raising our children Robert and Mary while I composed this work", then I would assume only "John" gets `@role=dte`?

to 1. interesting question. Since the valuelist is semi-open it should be possible to enter lists.
to 2. I'm not sure if I got the point. If you use `<persName>` it is just a markup that a string references a person. If you like to markup a persName as dedicatee a role like `dte` or `dedicatee` is necessary or let me say: more explicit. But that does not exlude using also persName markup for other named persons.

ahankinson:
Looking at: https://music-encoding.org/guidelines/dev/attribute-classes/att.name.html

It seems the value of `@role` is text? In that case, I don't think it allows multiple values? I would have to brush up on my datatypes, but I would expect something like this:

```xml
<datatype maxOccurs="unbounded">
     <rng:data type="NMTOKEN"/>
</datatype>
```

for a list of values, but instead I see this:

```xml
<datatype>
    <rng:text/>
</datatype>
```

While I think `rng:text` allows spaces, it's because it can be a string ("The Quick Brown Fox...") and not a list ("itm1 itm2 itm3")

For my second question, I didn't actually have a clear idea when I asked, but something was knawing at me about it. Thinking about it more, I guess I was just trying to probe the question of whether using `<dedication>` was really a replacement for dedicatee, in response to your (@riedde) comment above:

> What about getting rid of `<dedicatee>` for the benefit of a `<dedication>`?

That is, since you can already do:

```
<dedication>Blah blah blah <persName role="dte">John</persName></dedication>
```

Then the removal of `<dedicatee>` has nothing really to do with `<dedication>` -- it's simply being removed in favour of only using `<persName@role=dte>`. Is that right?

musicEnfanthen:
> It seems the value of `@role` is text? <snip>
> 
> for a list of values, but instead I see this:
> 
> ```
> <datatype>
>     <rng:text/>
> </datatype>
> ```

When you look at the changes in this PR, this has been adjusted to a new  semi-open list: `data.RELATOR`.

https://github.com/music-encoding/music-encoding/pull/1596/files#diff-88fd1aced3baa36ffacb0e712159e911cf0e90962ea8675fa252c097f8def3e4R2273-R2275 (it is in the MEI.shared.xml file that may need a manual loading trigger)

ahankinson:
@musicEnfanthen I actually don't see that change.... Cmd-F for `data.RELATOR` on this PR, and on the page you linked to, don't show it?

ahankinson:
Ah! Ok I see it now.

source/modules/MEI.shared.xml
```diff
@@ -171,6 +171,153 @@
...
     </content>
   </macroSpec>
+  <macroSpec ident="data.RELATORS" module="MEI.shared" type="dt">
+    <desc versionDate="2025-02-05" xml:lang="en">List of relators. The values and the definitions are taken from the <ref target="http://www.loc.gov/marc/relators/relaterm.html">MARC Code List for Relators</ref>.</desc>
+    <content>
+      <alternate>
+        <valList type="semi">
+          <valItem ident="act">
+            <gloss versionDate="2025-02-05" xml:lang="en">Actor</gloss>
+            <desc versionDate="2025-02-05" xml:lang="en">A performer contributing to an expression of a work by acting as a cast member or player in a musical or dramatic presentation, etc.</desc>
```
ahankinson: Is `expression` in this case a FRBR expression? I didn't think a specific individual could contribute to an "expression"? I know that this comes directly from the marc21 page, but maybe it's something we need to consider?

pe-ro: @ahankinson, an individual or group of people can certainly contribute to an expression as defined by FRBR. Performers, translators, arrangers are examples.  However, in the MARC relators definitions, the words "work" and "expression" are not used in their FRBR sense.

ahankinson: But can an actor? At best I think they can contribute **as an actor** at the manifestation level.

pe-ro: A performance, musical or theatrical, is an expression, not a manifestation.

ahankinson: My FRBR may be a bit rusty, but I thought it went something like:

Work: Orfeo
Expression: Setting of L'Orfeo by Monteverdi and Striggio
Manifestation: Recording of Orfeo from the 2012 Boston Early Music Festival
Item: The particular recording of that concert that I am holding in my hand.

In this case, "Actor: Aaron Sheehan in the role of Orfeo" only makes sense at the manifestation level, since there have been many, many actors that have played Orfeo over the years.

I may have it a bit mixed up at the work and expression level, but I'm struggling to see how a single actor can be listed in the expression of "Orfeo." 

Except, maybe in very rare cases... "John Malkovich" in "Being John Malkovich" comes to mind as a possible exception, but those are the exception rather than the rule.

Do I have that wrong?

(Incidentally, this is why I absolutely *hate* FRBR. No clear definition of what the levels are, and no authority to appeal to to get it right. My word against yours. *sigh*)

pe-ro: As I understand FRBR, each performance is also an expression.  Expanding your example --
```
Work: Orfeo
Expression: Setting of L'Orfeo by Monteverdi and Striggio
Expression: Monteverdi and Striggio's setting as performed by the Boston Early Music 
  Festival Players
Expression: Monteverdi and Striggio's setting as performed by The Not Ready for 
  Primetime Players
Manifestation: Recording of Orfeo from the 2012 Boston Early Music Festival
Item: The particular recording of that concert that I am holding in my hand.
```
The performances are based on Monteverdi and Striggio's "rendering" of the story of Orfeo into Italian, for singers of a certain range, and so on, and probably (but not necessarily) also on a given manifestation -- that is, the text of the opera had to have a physical form from which the performers performed. So, already the idea of a simple hierarchy is blown.

A recording, being a physical instantiation, is a manifestation of a particular expression; that is, the one by the Boston Early Music Festival Players _at a particular time and place_.  Of course, this leads to the natural conclusion that each time the opera is performed, a new expression is created, even if it's the same opera performed by the same people on the same instruments in the same hall, etc.  However, in most cases, we can compress this down to a single expression.

FRBR can be confusing.  But I don't think FRBR is any more or less confusing than other bibliographic description methods.  With FRBR, I do find it best to think in "bottom up" terms rather than "top down"; that is, starting from the item and going "up" to the work, creating as many manifestations and expressions as are necessary to adequately describe the situation. Complex situations end up being graphs rather than strict hierarchies.  So --
```
Item: A physical copy of a recording of a concert/performance
Manifestation: All the copies of this recording
Expression: Monteverdi and Striggio's setting as performed by the Boston Early Music 
  Festival Players
Work: Orfeo

Expression: Monteverdi and Striggio's setting as performed by The Not Ready for 
  Primetime Players
Expression: Setting of L'Orfeo by Monteverdi and Striggio
```
The choice of expression here (Monteverdi and Striggio's setting as performed by the Boston Early Music Festival Players and not Setting of L'Orfeo by Monteverdi and Striggio) is a personal choice and depends on whether one is emphasizing the performance aspect, which seems logical in this case since we started with a performance recording. 

The trailing expressions above don't fit the primary Item/Manifestation/Expression/Work hierarchy, but are part of a graph of "everything you might want to know about Orfeo". The web of relationships between the two of them and between them and the work can be expressed with stand-off markup.  Of course, another approach would be not to have a hierarchy at all, but only a set of relationships.

In short, FRBR is a complex tool for capturing the nuances of complex situations.  It doesn't have to be used in all cases.

pe-ro:
I believe @ahankinson is correct. As currently defined, `@role` will only allow a single value.  I believe 
```
<attDef ident="role" usage="opt">
  <desc xml:lang="en">Used to specify further information about the entity referenced by
    this name, for example, the occupation of a person or the status of a place.</desc>
  <datatype>
    <rng:ref name="data.RELATORS"/>
  </datatype>
  <remarks xml:lang="en">
    <p>When applicable, values from the <abbr>MARC</abbr> relator term list (<ref
        target="http://www.loc.gov/marc/relators/relaterm.html"
        >http://www.loc.gov/marc/relators/relaterm.html</ref>) or code list (<ref
        target="http://www.loc.gov/marc/relators/relacode.html"
        >http://www.loc.gov/marc/relators/relacode.html</ref>) are recommended for
        <att>role</att>.</p>
  </remarks>
</attDef>
```
should be 
```
<attDef ident="role" usage="opt">
  <desc xml:lang="en">Used to specify further information about the entity referenced by
    this name, for example, the occupation of a person or the status of a place.</desc>
  <list>
    <oneOrMore>
      <rng:ref name="data.RELATORS"/>
     </oneOrMore>
  </list>
  <remarks xml:lang="en">
    <p>When applicable, values from the <abbr>MARC</abbr> relator term list (<ref
        target="http://www.loc.gov/marc/relators/relaterm.html"
        >http://www.loc.gov/marc/relators/relaterm.html</ref>) or code list (<ref
        target="http://www.loc.gov/marc/relators/relacode.html"
        >http://www.loc.gov/marc/relators/relacode.html</ref>) are recommended for
        <att>role</att>.</p>
  </remarks>
</attDef>
```
I don't know how/if this can be translated to Pure ODD, though.

source/modules/MEI.shared.xml
ahankinson: ... aaanyway, this is secondary to the gist of the comment in that I'm not entirely sure that simply copying the MARC21 relators descriptions wholesale is entirely problem-free, particularly where there may be terminology differences that can cause confusion.

riedde:
> Then the removal of <dedicatee> has nothing really to do with <dedication> -- it's simply being removed in favour of only using <persName@role=dte>. Is that right?

Yes, `dedicatee` will simply removed (as `lyricist` and others). Making possible ro record a `dedication` should be the benefit, but while developing @pe-ro and me found out that `<dedication>` already exists.

source/modules/MEI.shared.xml
riedde: One issue that I see is that the list is quite big. (By the way it is not the entire list. I reduced it to those values that might be senseful in some way from an MEI perspective.) In the PR I added this macro as datatype to `@role` which is allowed, e.g., in `<persName>`. Since `<persName>` is allowed in multiple places (e.g., `<event>`) this list shouldn't provide a set of relators. If you see to many complications here, I'm open for discusion and to revise the list or divide it or assign it in another way.

(We are very often talking about standardization, best practice and making MEI more suitebale for beginners. That's why I'm proposing this.)


> Is expression in this case a FRBR expression? I didn't think a specific individual could contribute to an "expression"? I know that this comes directly from the marc21 page, but maybe it's something we need to consider?

This is the definition given in MARC21, thats whay I refer that this list of definititions is taken from that concept (l. 175). Not everything in MEI is FRBR ;-) But I see the confusion in terminology...

riedde:
> I believe @ahankinson is correct. As currently defined, `@role` will only allow a single value. I believe
> 
> ```
> <attDef ident="role" usage="opt">
>   <desc xml:lang="en">Used to specify further information about the entity referenced by
>     this name, for example, the occupation of a person or the status of a place.</desc>
>   <datatype>
>     <rng:ref name="data.RELATORS"/>
>   </datatype>
>   <remarks xml:lang="en">
>     <p>When applicable, values from the <abbr>MARC</abbr> relator term list (<ref
>         target="http://www.loc.gov/marc/relators/relaterm.html"
>         >http://www.loc.gov/marc/relators/relaterm.html</ref>) or code list (<ref
>         target="http://www.loc.gov/marc/relators/relacode.html"
>         >http://www.loc.gov/marc/relators/relacode.html</ref>) are recommended for
>         <att>role</att>.</p>
>   </remarks>
> </attDef>
> ```
> 
> should be
> 
> ```
> <attDef ident="role" usage="opt">
>   <desc xml:lang="en">Used to specify further information about the entity referenced by
>     this name, for example, the occupation of a person or the status of a place.</desc>
>   <list>
>     <oneOrMore>
>       <rng:ref name="data.RELATORS"/>
>      </oneOrMore>
>   </list>
>   <remarks xml:lang="en">
>     <p>When applicable, values from the <abbr>MARC</abbr> relator term list (<ref
>         target="http://www.loc.gov/marc/relators/relaterm.html"
>         >http://www.loc.gov/marc/relators/relaterm.html</ref>) or code list (<ref
>         target="http://www.loc.gov/marc/relators/relacode.html"
>         >http://www.loc.gov/marc/relators/relacode.html</ref>) are recommended for
>         <att>role</att>.</p>
>   </remarks>
> </attDef>
> ```
> 
> I don't know how/if this can be translated to Pure ODD, though.

I'll have an eye on that soon and figure out if there are other ways of integrating this.

source/modules/MEI.shared.xml
ahankinson: The size of the list isn't necessarily the problem, it's the inclusion of the descriptions wholesale, without modifying it for the purposes of MEI, that I am questioning. 

I'm also OK with suggesting or even recommending the use of relator codes. That's also absolutely fine.

Are we also committing to keeping this list up-to-date with changes to the relator codes and their descriptions? I see a date of February 6th 2025 on the relator code page, so I think it changes frequently.

source/validation/mei_odds.rng
bwbohl, line 1: I think these changes to mei_odds.rng were unintentional and probably should be reverted.

riedde: If I remember right, the test has failed before.

riedde:
> [...] As currently defined, `@role` will only allow a single value. [...]

The pureODD solution for alowing multiple values is
```
<datatype minOccurs="1" maxOccurs="unbounded">
    <rng:ref name="data.RELATORS"/>
</datatype>
```
The rng-fragment `rng:ref/@name` should be switched to `dataRef/@key` some day, but the pureODD way produces a lot of warnings.

(To allow also custom text values I had to split the data type analog to `data.RELATIONSHIP`.)

bwbohl (dismissed):

riedde:
> > [...] As currently defined, `@role` will only allow a single value. [...]
> 
> The pureODD solution for alowing multiple values is
> 
> ```
> <datatype minOccurs="1" maxOccurs="unbounded">
>     <rng:ref name="data.RELATORS"/>
> </datatype>
> ```
> 
> The rng-fragment `rng:ref/@name` should be switched to `dataRef/@key` some day, but the pureODD way produces a lot of warnings.
> 
> (To allow also custom text values I had to split the data type analog to `data.RELATIONSHIP`.)

I finally got it. in PureODD it can be:

```
<content>
      <alternate>
        <macroRef key="data.MARCRELATORS"/>
        <macroRef key="data.NMTOKEN"/>
      </alternate>
    </content>
```

Effect: Allowing a value list and custom values mixed as whitespace separated list. Thank you @bwbohl for the coincidental hint ;-)

bwbohl (dismissed):

bwbohl:
@musicEnfanthen and I just gave it some more thought:
As the list is also published at https://id.loc.gov/vocabulary/relators.html, we would prefer to point there, instead of https://www.loc.gov/marc/relators/relacode.html in general and add a link to the respective page for each term to their description to make clear that both value and description come from there.

bwbohl:
I like the idea with the "basic" relators

riedde:
The "basic" group is more or less the roles the former semantic container elements have represented. The division was also made to support customizations with a smaller setup, e.g., MEI basic

riedde:
> @musicEnfanthen and I just gave it some more thought: As the list is also published at https://id.loc.gov/vocabulary/relators.html, we would prefer to point there, instead of https://www.loc.gov/marc/relators/relacode.html in general and add a link to the respective page for each term to their description to make clear that both value and description come from there.

@musicEnfanthen @bwbohl I changed the general source link, added the specific link to each value and updated the descriptions. The update is done using a XQuery that calls the source link and reads the label and description automatically. I can share the script, if you're interested.

source/modules/MEI.shared.xml
```diff
@@ -242,6 +242,167 @@
...
     </content>
   </macroSpec>
+  <macroSpec ident="data.MARCRELATORS.basic" module="MEI.shared" type="dt">
+    <desc versionDate="2025-05-25" xml:lang="en">Relators [MARC]. Values and definitions taken from <ref target="https://id.loc.gov/vocabulary/relators.html">http://id.loc.gov/vocabulary/relators</ref>.</desc>
+    <content>
+        <valList type="semi">
+          <valItem ident="arr">
+            <gloss versionDate="2025-05-25" xml:lang="en">Arranger</gloss>
+            <desc versionDate="2025-05-25" xml:lang="en">
+              A person, family, or organization contributing to a musical work by rewriting the composition for a medium of performance different from that for which the work was originally intended, or modifying the work for the same medium of performance, etc., such that the musical substance of the original composition remains essentially unchanged. For extensive modification that effectively results in the creation of a new musical work, see composer. [<ref target="https://id.loc.gov/vocabulary/relators/arr.html">MARC</ref>]</desc>
```
```suggestion
            <desc versionDate="2025-05-25" xml:lang="en">A person, family, or organization contributing to a musical work by rewriting the composition for a medium of performance different from that for which the work was originally intended, or modifying the work for the same medium of performance, etc., such that the musical substance of the original composition remains essentially unchanged. For extensive modification that effectively results in the creation of a new musical work, see composer. [<ref target="https://id.loc.gov/vocabulary/relators/arr.html">MARC</ref>]</desc>
```

source/modules/MEI.shared.xml
```diff
@@ -242,6 +242,167 @@
       </valList>
     </content>
   </macroSpec>
+  <macroSpec ident="data.MARCRELATORS.basic" module="MEI.shared" type="dt">
+    <desc versionDate="2025-05-25" xml:lang="en">Relators [MARC]. Values and definitions taken from <ref target="https://id.loc.gov/vocabulary/relators.html">http://id.loc.gov/vocabulary/relators</ref>.</desc>
+    <content>
+        <valList type="semi">
+          <valItem ident="arr">
+            <gloss versionDate="2025-05-25" xml:lang="en">Arranger</gloss>
```
```suggestion
            <gloss versionDate="2025-05-25" xml:lang="en">arranger</gloss>
```

source/modules/MEI.shared.xml
```diff
@@ -242,6 +242,166 @@
...
+          </valItem>
+          <valItem ident="lyr">
+            <gloss versionDate="2025-05-25" xml:lang="en">lyricist</gloss>
+            <desc versionDate="2025-05-25" xml:lang="en">An author of the words of a non-dramatic musical work (e.g. the text of a song), except for oratorios. [<ref target="https://id.loc.gov/vocabulary/relators/lyr.html">MARC</ref>]</desc>
+          </valItem>
+        </valList>
+    </content>
+  </macroSpec>
+  <macroSpec ident="data.MARCRELATORS.extended" module="MEI.shared" type="dt">
+    <desc versionDate="2025-02-05" xml:lang="en">MARC list of relators. Values and definitions taken from the <ref target="http://www.loc.gov/marc/relators/relaterm.html">MARC Code List for Relators</ref>.</desc>
```
musicEnfanthen: @riedde This description needs to be updated, too?

riedde:
```suggestion
    <desc versionDate="2025-02-05" xml:lang="en">Relators [MARC]. Values and definitions taken from <ref target="https://id.loc.gov/vocabulary/relators.html">http://id.loc.gov/vocabulary/relators</ref>.</desc>
```

bwbohl (dismissed):

bwbohl:
@riedde, many thanks! Sharing the script is a good idea so we can reproduce or rerun it. I could also imagine creating a test that runs regularly to check if the Library of Congress has changed its definitions.

bwbohl (dismissed):

musicEnfanthen (dismissed):

source/modules/MEI.shared.xml
bwbohl: @ahankinson several changes have happened, trying to include your caveats, any chance you re-review the current state and resolve this conversation?

pe-ro (dismissed):
The commits appear to address everything, so I'm inclined to approve these changes, but I need to have an RNG schema to perform testing.

source/docs/09-textencoding.xml
```diff
@@ -166,7 +166,7 @@
                         <label>'frontispiece'</label>
                         <item>A pictorial frontispiece, possibly including some text.</item>
                      </list>
-                     <p>The following extended example demonstrates how various parts of the front matter of a text may be encoded. The front part begins with a title page, which is presented in section <ptr target="#textTitlePages"/>, below. This is followed by a dedication and a preface, each of which is encoded as a distinct div:</p>
+                     <p>The following extended example demonstrates how various parts of the front matter of a text may be encoded. The front part begins with a title page, which is presented in section <ptr target="#textTitlePages"/>, below. This is followed by a dedication and a preface, each of which is encoded as a distinct div. Note that if the dedication is placed on the title page, it can also be marked up there using <gi scheme="MEI">dedication</gi>.</p>
```
ahankinson: This makes it sound like `<dedication>` can only be used on the front page?

source/docs/03-metadata.xml
```diff
@@ -1423,10 +1419,9 @@
...
-                     <p>Detailed analysis of the title page and other preliminaries of older printed books and manuscripts is of major importance in descriptive bibliography and the cataloging of printed books. The following elements are suggested as a means of encoding the major features of most title pages for faithful rendition:</p>
+                     <p>Detailed analysis of the title page and other preliminaries of older printed books and manuscripts is of major importance in descriptive bibliography and the cataloging of printed books. The following elements are suggested as a means of encoding the structural features of most title pages for faithful rendition:</p>
```
```suggestion
                     <p>Detailed analysis of the title page and other preliminaries of older printed books and manuscripts is important in descriptive bibliography. The following elements are suggested as a means of encoding the structural features of most title pages for faithful rendition:</p>
```

No need to be redundant and hyperbolic.

source/docs/09-textencoding.xml
riedde: @ahankinson I'm with you. My goal was to note here, that `<dedication>` exists. Before there was only `div/@type` or something like that. Do you have an idea how to reword this? Or shall we leave it out (in this PR)?
However, I think a revision of this chapter might be necessary (in the nearer future).

riedde: Resolved in 2386a31f0fcd55dfadc516087abab9e02bad57af by removing the phrase. Reset state to before commit.

riedde:
@bwbohl @musicEnfanthen I resolved the last conversation.

pe-ro (dismissed):
Looks good

customizations/mei-basic.xml
```diff
@@ -148,8 +148,9 @@
                 <!--<moduleRef key="MEI.msDesc"/>-->
                 <moduleRef key="MEI.midi" include="instrDef" />
                 <moduleRef key="MEI.namesdates" include="persName" />
-                <moduleRef key="MEI.shared"
-                    include="accid arranger artic body caesura chord clef clefGrp composer date dir dynam ending label labelAbbr layer lb lyricist mdiv mei music note ornam pb pgFoot pgHead pubPlace rend respStmt rest sb score scoreDef section space staff staffDef staffGrp symbol syl tempo title "/>
+                <!-- the following elements are not available anymore in MEI.shared: arranger, composer, lyricist -->
```
ahankinson: Is this comment necessary?

source/docs/03-metadata.xml
```diff
@@ -1423,10 +1419,9 @@
...
                      </p>
                      <p>The <gi scheme="MEI">titlePage</gi> element, modelled after a similar element in the Encoded Archival Description (EAD), may occur within the textual matter preceding or following the musical content of the encoding. Since a diplomatic transcription of the title page is often necessary to accurately identify musical material contained within a source, the <gi scheme="MEI">titlePage</gi> element may also be used within the metadata header as a child of the <gi scheme="MEI">physDesc</gi> element.</p>
-                     <p>Detailed analysis of the title page and other preliminaries of older printed books and manuscripts is of major importance in descriptive bibliography and the cataloging of printed books. The following elements are suggested as a means of encoding the major features of most title pages for faithful rendition:</p>
+                     <p>Detailed analysis of the title page and other preliminaries of older printed books and manuscripts is important in descriptive bibliography. The following elements are suggested as a means of encoding the structural features of most title pages for faithful rendition:</p>
```
ahankinson: I’m not sure about “faithful rendition “ here. What does this mean? There is not enough information to capture the exact look and layout of the title page, which is what I would expect if it’s been “rendered.”

I would put a period after “title pages” and remove the last three words.

source/docs/09-textencoding.xml
```diff
@@ -997,7 +993,7 @@
...
+                  <p>The elements <gi scheme="MEI">title</gi>, <gi scheme="MEI">edition</gi>, <gi scheme="MEI">series</gi>, and <gi scheme="MEI">identifier</gi> fall into the first category as do the elements <gi scheme="MEI">creator</gi>, <gi scheme="MEI">funder</gi>, <gi scheme="MEI">sponsor</gi>, and <gi scheme="MEI">respStmt</gi>. The respStmt element is provided for marking responsibility roles (<abbr>e.g.</abbr>, composer, arranger, lyricist). The <gi scheme="MEI">biblScope</gi> element also carries information of an identifying nature.</p>
```
ahankinson:
```suggestion
                  <p>The elements <gi scheme="MEI">title</gi>, <gi scheme="MEI">edition</gi>, <gi scheme="MEI">series</gi>, and <gi scheme="MEI">identifier</gi> fall into the first category as do the elements <gi scheme="MEI">creator</gi>, <gi scheme="MEI">funder</gi>, <gi scheme="MEI">sponsor</gi>, and <gi scheme="MEI">respStmt</gi>. The  <gi scheme=“MEI”>respStmt</gi> element is provided for marking responsibility roles (<abbr>e.g.</abbr>, composer, arranger, lyricist). The <gi scheme="MEI">biblScope</gi> element also carries information of an identifying nature.</p>
```

source/validation/mei_odds.rng
```diff
@@ -4934,8 +4934,8 @@ Elements]</a:documentation>
          <empty/>
       </element>
    </define>
-   <define name="author">
-      <element name="author">
+   <define name="creator">
+      <element name="creator">
          <a:documentation xmlns:a="http://relaxng.org/ns/compatibility/annotations/1.0">in a bibliographic reference, contains the name(s) of an author, personal or corporate, of a work; for example in the same form as that provided by a recognized bibliographic name authority. [3.11.2.2. Titles, Authors, and Editors 2.2.1. The Title Statement]</a:documentation>
```
ahankinson:
```suggestion
         <a:documentation xmlns:a="http://relaxng.org/ns/compatibility/annotations/1.0">in a bibliographic reference, contains the name(s) of a creator, personal or corporate, of a work; for example in the same form as that provided by a recognized bibliographic name authority. [3.11.2.2. Titles, Authors, and Editors 2.2.1. The Title Statement]</a:documentation>
```

ahankinson (commented):
Some suggestions

customizations/mei-basic.xml
riedde: Yes.

source/docs/03-metadata.xml
riedde: ****
```suggestion
                     <p>Detailed analysis of the title page and other preliminaries of older printed books and manuscripts is important in descriptive bibliography. The following elements are suggested as a means of encoding the structural features of most title pages:</p>
```

customizations/mei-basic.xml
ahankinson: Why? If they’ve been removed then there is no more action that needs to be taken.

riedde: This is a reminder because it will could have side-effects to the musescore plug-in. I talked to @lpugin and he will remove it, when this is checked.

ahankinson: I just checked with @lpugin and the MuseScore plugin uses `<persName>` with `@role`.

riedde: okay, then the comment is not necessary anymore.
```suggestion
```

ahankinson:
Sorry, I keep having more questions.

The `<author>` element is now being replaced by `<creator>`, and the other creator-like elements are being removed. 

But as far as I can see in this PR, `@role` isn't added to `<creator>`, right? So how do I tell what kind of creator they are? Or how do I otherwise specify what kind of creator they are?

I understand that `@role` is on `<persName>`, but this would mean that you would always have to do this:

```xml
<creator>
   <persName role="composer">John Smith</persName>
</creator>
<creator>
   <persName role="arranger">John Smith</persName>
</creator>
<creator>
    <persName role="librettist">Mary Jones</persName>
</creator>
```

Doesn't the role type belong on the `<creator>` tag (or indeed, some sort of typification of the relationship), instead of the `<persName>` tag?

(I'm not even sure `<persName@role>` is appropriate here, since it seems to suggest that the relator codes give a person's occupation, and not necessarily their relationship to a specific piece of music.)

riedde:
> I understand that `@role` is on `<persName>`

exactly, `<creator>` is a container for `<persName>` and has no role.
 
 ```
 <creator>
    <persName role="composer">John Smith</persName>
    <persName role="librettist">Mary Jones</persName>
 </creator>
 ```

But I see that in case of an author it is a bit strange, e.g.,

 ```
 <creator>
    <persName role="author">James Joyce</persName>
 </creator>
 ```

> (I'm not even sure `<persName@role>` is appropriate here, since it seems to suggest that the relator codes give a person's occupation, and not necessarily their relationship to a specific piece of music.)

If I think about all the work and discussions done on this during the last three+ years, this solution might be acceptable but not perfect. I'm open to discuss this again in the future.

pe-ro:
The `@role` attribute exists on `<name>` and its variants, `<persName>`, `<corpName>`, etc. for capturing the role of the person, etc. when these elements are used outside the `<creator>` context; that is, elsewhere in the header or body.  In those cases, the `<creator>` element won't be present to carry the name's role.  I see the logic in Andrew's suggestion:
```
<creator role="composer">
   <persName>John Smith</persName>
</creator>
<creator role="arranger">
   <persName>John Smith</persName>
</creator>
<creator role="librettist">
    <persName>Mary Jones</persName>
</creator>
```
but adding `@role` to `<creator>` doesn't reduce the verbosity of the encoding:
```
<creator>
   <persName role="composer">John Smith</persName>
</creator>
<creator>
   <persName role="arranger">John Smith</persName>
</creator>
<creator>
    <persName role="librettist">Mary Jones</persName>
</creator>
```
and would result in a different use of `<persName>`, etc. in the header than elsewhere:
```
<history xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.1+anyStart">
  <p>
    <persName role="composer">John Smith</persName> composed this work in <date>2025</date>.
  </p>
</history>
```
By the way, multiple creators, as in @riedde's John Smith/Mary Jones example above, should be marked up using multiple `<creator>` elements:
```
<creator>
  <persName role="composer">John Smith</persName>
</creator>
<creator>
  <persName role="librettist">Mary Jones</persName>
</creator>
```
Multiple names may occur in `<creator>` when there are multiple forms of the creator's name, such as real/pseudonym names like Mark Twain and Samuel Clemens, or like Tchaikovsky and its myriad transliterations.

pe-ro:
Regarding
> (I'm not even sure <persName@role> is appropriate here, since it seems to suggest that the relator codes give a person's occupation, and not necessarily their relationship to a specific piece of music.)

`@role` captures the relationship between the person and the intellectual entity being described.  The MARC documentation for bibliographic records describes the $e subfield as "[d]esignation of function that describes the relationship between a name and a work."  Here's a portion of the bibliographic record for a work by Thomas Jefferson:
```
100 1 Jefferson, Thomas,|d1743-1826,|eauthor.
245 10Order to Militia officers to enforce the Embargo Act, 22 January 1809.
264 _0[Place of production not identified :|bproducer not identified,|c1809]
300 _ _1 online resource.336 _ _text|2rdacontent
337 _ _computer|2rdamedia
338 _ _online resource|2rdacarrier
etc.
```
However, in authority headings, there's a separate designator for occupation; that is, datafield 374. see [41866059.marc21.txt](https://github.com/user-attachments/files/21303004/41866059.marc21.txt)

MEI doesn't support "in-line" authority records yet.  It's expected that the `<name>` element will point to an external file, e.g.
```
<creator>
  <persName role="composer" auth.uri="http://id.loc.gov/authorities/names/n79089957">Jefferson, Thomas</persName>
</creator>
```
So, `@role` records Jefferson's relationship to the executive order and `@auth.uri` describes Jefferson's occupation(s).

ahankinson:
Yeah, I'm sorry for being so picky about this. But I think the current formulation changes some fundamental semantics and I want to make sure it's right.

I wasn't trying to make the encoding less verbose, I was trying to make the relationships more explicit.

If we take the MEI document as the context, then the hierarchy represents these relationships. Given these two alternate encodings:

```
<creator>
    <persName role="composer">John Smith</persName>
</creator>
```

and 

```
<creator role="composer">
    <persName>John Smith</persName>
</creator>
```

The second one says the relationship of "John Smith" to the document is a "creator" and, specifically, a "composer". "John Smith" exists as a person on their own.

But encoding the role on the `<persName>` element, and not on the `<creator>` element seems to change the semantics to say that "John Smith is a composer" and not that "John Smith is the creator of the music, with the role of composer". I think the latter is what is meant, but the former is currently what is being promoted.

Further, this is currently allowed:

```
<creator>
    Anonymous
</creator>
```

"Anonymous" isn't a person name, or even a name at all, but that anonymous person is the creator. Without `@role` on `<creator>` I'm not sure how we would state this? And how do we specify a role for things with just textual content?

Looking at the upgrade path, how does `<composer>` get translated to either `<creator>` or `<contributor>`? Will the upgrade from 5 to 6 require moving textual content into a `<persName>` element in order to capture their role? I'm guessing this can't really be automated? 

Which brings me to the second issue: the implicit value judgement being made. I think we need to at least acknowledge that a "contributor" is a lesser role in the work than a "creator". In the previous versions we could simply state that someone was a `<lyricist>` and this would make no judgement on whether their contribution was a core part of the work or not -- think of a libretto that stands as its own work prior to being set to music. 

Now we will have to make the judgement whether the librettist is a "creator" or merely a "contributor" to the work. And we will also have to determine whether the `<composer>` is always the `<creator>`, or if they simply "contributed" the music to the piece.

Again, this will be tricky to manage in any automated upgrade, and makes it harder to communicate what tags should be used in any given situation.

---

To sum up, if `@role` is added to `<creator>` and `<contributor>` I think this, while not perfect, would serve to at least bring back some of the semantics that I think are lost, and help the upgrade path. I don't like the implied value judgements being introduced by removing the _prima facie_ tags, but if all the work and discussions haven't flagged that yet, then I'm probably in the minority.

That said, I've tried my best to explain (in pretty significant detail) what I think the problems with this change are. I won't block it, but I probably won't approve it as it stands. You can merge it if you want as-is if that's what most people want.

riedde:
> That said, I've tried my best to explain (in pretty significant detail) what I think the problems with this change are.

Thank you!

> I won't block it, but I probably won't approve it as it stands. You can merge it if you want as-is if that's what most people want.

I know, but I'm getting tired of the endless discussion on this topic. Sorry for the harsh tone of my last comments!

> To sum up, if `@role` is added to `<creator>` and `<contributor>` I think this, while not perfect, would serve to at least bring back some of the semantics that I think are lost, and help the upgrade path. I don't like the implied value judgements being introduced by removing the prima facie tags, but if all the work and discussions haven't flagged that yet, then I'm probably in the minority.

It doesn't came up in the discussions, so it is worth to have a look at this. While scanning the schema I noticed that `<contributor>` already has a `@role` attribute definded as a direct child. And since `<creator>` will become also a generic container why shouldn't `@role` also be possible here. I am not sure if I really got your point, however, but we may create an inconsistency here if we keep this PR as it is. From this point of view I would argue keeping `@role` at `<contributor>` should be preferred and in consequence, adding it to `<creator>` because this is less invasive (especially for the upgrade path).

pe-ro:
I don't remember how/why `<contributor>` ended up with a `@role` attribute, but seeing as it does, my argument for not allowing `<creator>` to have `@role` isn't strong.

Adding `@role` to `<creator>` is easy, but it may cause confusion.  For example, `@role` has to be kept for cases like --
```
<respStmt xmlns="http://www.music-encoding.org/ns/mei">
  <persName role="editor">Joe</persName>
  <persName role="encoder">Jane</persName>
</respStmt>
```
but this admits the possibility of
```
<creator role="composer">
  <persName role="arranger">Mr. Confusion</persName>
</creator>
```
So, if `@role` is added to `<creator>` (and it continues to stand on `<contributor>`), then a context-dependent solution is needed.  One way of handling this is to use Schematron to disallow `@role` when one of the name elements is a descendant of `<creator>` or `<contributor>`.  The documentation for this scenario will need to make it clear that while `@role` is permitted on the name elements, it will be disallowed by the Schematron rule in certain circumstances.  It's a little weird -- it's allowed by the schema, but disallowed by the Schematron rule.  And if the Schematron rules are not applied, then all bets are off.

Another option is to allow `<creator>` and `<contributor>` to have different content models based on where they appear. This will be easier for the user because it will be easier for the parser.  It also won't rely on Schematron being turned on.  But creator models 1 and 2, and contributor models 1 and 2 will present a "trip hazard" for current and future maintainers of MEI.

We have shied away from the latter solution until now, but I think in this case it's the better path.

riedde:
> but this admits the possibility of
> 
> ```
> <creator role="composer">
>   <persName role="arranger">Mr. Confusion</persName>
> </creator>
> ```

there are multiple places in MEI where you can do thinks like that or similar. But I don't think that's a problem as long as the encoders think about their work. And in my opinion, we can certainly expect that.

If we can't expect that (would be sad), then we can think about schematron rules that restrict the possible combinations of values here. (just my opinion)

bwbohl (approved):

musicEnfanthen (approved):

PR #1608, opened 2025-02-05
docs: use latest TEI version

rettinghaus:
* fix version
* remove unused namespaces
* <s>use "official" rng</s>

Revalidated all files.

### DIFF
--- a/source/docs/01-introduction.xml
+++ b/source/docs/01-introduction.xml
@@ -19,9 +19,7 @@
 -->
 <?xml-model href="../validation/mei_odds.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
 <?xml-model href="../validation/mei_odds.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
-<TEI xmlns:rng="http://relaxng.org/ns/structure/1.0"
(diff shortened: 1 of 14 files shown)


rettinghaus:
see https://github.com/music-encoding/music-encoding/pull/1586#issuecomment-2636977290

bwbohl:
thanks @rettinghaus !!

bwbohl:
I would like to postpone this merge until after updating to pureODD because otherwise it would report errors that will not be there after converting.

rettinghaus:
How exactly is this connected to pureODD?

rettinghaus:
@bwbohl?

bwbohl (dismissed):

bwbohl:
I think we should validate `mei-source.xml` the `modules` and the `docs` against the same schema. I tweaked the `compare-versions` ant target a bit and ran it with mei_odds vs tei_all compiled odds and I don’t really see where that benefits us (lots of new elements that we don’t actually use) except from making sure we’re still compatible with the latest version of TEI.
I’d rather update the `mei_odds` customization and drill it down to what we actually use in order to harmonize the current encoding.

bwbohl (approved):

musicEnfanthen (approved):
Thank you. Finally back from future :D

PR #1617, opened 2025-02-06
schema: fix version number in edition

musicEnfanthen:
This PR fixes the version number in source's edition to include dev. That was overlooked in #1586


bwbohl (approved):
thx

lpugin (approved):

PR #1620, opened 2025-02-06
utils: add schematron to precheck current source for "purification"

bwbohl:
prepare conversion to pureODD


source/validation/mei-purify.sch
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<sch:schema queryBinding="xslt2" xmlns:sch="http://purl.oclc.org/dsdl/schematron">
+    <sch:ns prefix="tei" uri="http://www.tei-c.org/ns/1.0"/>
+    <sch:ns prefix="rng" uri="http://relaxng.org/ns/structure/1.0"/>
+
+    <sch:pattern id="check_rngChoice">
+        <sch:rule context="rng:choice">
+            <sch:assert role="error" test="count(*) > 1">An &lt;rng:choice&gt; element hast to contain more than one option to choose from.</sch:assert>
```
rettinghaus:
```suggestion
            <sch:assert role="error" test="count(*) > 1">An &lt;rng:choice&gt; element has to contain more than one option to choose from.</sch:assert>
```

source/validation/mei-purify.sch
```diff
@@ -0,0 +1,24 @@
...
+        </sch:rule>
+    </sch:pattern>
+
+    <sch:pattern id="check_teiConstraintSpec_ident">
+        <sch:rule context="tei:constraintSpec">
+            <sch:assert role="error" test="@ident != preceding::tei:constraintSpec/@ident">The @ident on constraintSpec has to be unique across all of mei-source.xml.</sch:assert>
```
rettinghaus: `mei-source.xml` actually doesn't contain constraints, better to use something like "across all MEI source files".

rettinghaus (changes requested):

source/validation/mei-purify.sch
bwbohl: Hmm… right will work in a context where xi:includes get resolved automatically, e.g. with XInclude processing activated in oXygen XML or when canonicalising first and then validating the canonicalised file like in the newly pushed ant target ;-)

bwbohl: I’d be ok with having this as ant-target exclusively; what do you guys think @rettinghaus & @musicEnfanthen?

rettinghaus: Sounds okay with me.
As it will be of use only for a limited time, maybe add a comment for future us?

bwbohl: added a xml comment

build.xml
```diff
@@ -303,6 +303,18 @@
         <property name="validate-source.done" value="true"/>
         <echo>canonicalized source validated: ${validate-source.done}</echo>
     </target>
+    
+    <target name="validate-purification" depends="canonicalize-source" description="Validates the canonicalized source with mei-purify.sch. This target will be triggered before all `build-...` targets.">
+        <echo unless:set="validate-source.done">validating source...</echo>
```
```suggestion
        <echo unless:set="validate-purification.done">validating purification of source...</echo>
```

build.xml
```diff
@@ -303,6 +303,18 @@
...
+        <taskdef name="schematron" classname="com.helger.schematron.ant.Schematron" classpathref="mei.classpath" loaderref="mei.classpath.loader" />
+        <schematron unless:set="validate-source.done" schematronFile="${dir.validation}/mei-purify.sch" expectSuccess="true" schematronProcessingEngine="schematron" languageCode="en" failOnError="true" failOnValidationError="true">
```
```suggestion
        <schematron unless:set="validate-purification.done" schematronFile="${dir.validation}/mei-purify.sch" expectSuccess="true" schematronProcessingEngine="schematron" languageCode="en" failOnError="true" failOnValidationError="true">
```

build.xml
```diff
@@ -303,6 +303,18 @@
...
+            <fileset dir="${dir.build}">    
+                <include name="${canonicalized.file.next}" ></include>
+            </fileset>
+        </schematron>
+        <property name="validate-source.done" value="true"/>
```
```suggestion
        <property name="validate-purification.done" value="true"/>
```

build.xml
```diff
@@ -303,6 +303,18 @@
...
+        <echo>canonicalized source validated: ${validate-source.done}</echo>
```
```suggestion
        <echo>purification of canonicalized source validated: ${validate-purification.done}</echo>
```

rettinghaus (dismissed):

musicEnfanthen (dismissed):

bwbohl:
@musicEnfanthen fixed the syntax and the check runs successfully. The purify-check only issues warnings, no errors @musicEnfanthen how do we get those reported in the check result?

musicEnfanthen:
@bwbohl Would not open a new issue for every check run. We can simply look into the action logs whenever we need it, right?

bwbohl:
thx @rettinghaus for the hint here: https://github.com/musescore/MuseScore/pull/27050/files

bwbohl:
> @bwbohl Would not open a new issue for every check run. We can simply look into the action logs whenever we need it, right?

I didn‘t mean issues but a status in the check that is something else than green/passed but a warning

rettinghaus (dismissed):
lgtm

bwbohl:
tried different things:
1. add a problem matcher but that did not report anything… why? I don’t know --> help appreciated
2. made the schematron check with com.helger.schematron.ant.Schematron failOnValidationWarn

@rettinghaus & @musicEnfanthen which way to go? I tend to (2.) because we want that stuff fixed and digging into the problem matcher for this intermediate task seems (at least to me) like an unnecessary overhead.

What’s your take?

bwbohl:
Probably the purification check should also go into a separate action to make it clear that the general validation did not fail

musicEnfanthen:
I would not failOnValidationWarn since then this might block the merging of an otherwise fine PR.

musicEnfanthen:
@bwbohl Moved the purification job into a separate job, so it is reported in the PR separately from the source validation.

.github/problem_matchers/schematron.json
```diff
@@ -0,0 +1,20 @@
+{
+    "problemMatcher": [
+        {
+            "name": "schematron",
+            "owner": "schematron",
```
musicEnfanthen:
```suggestion
            "owner": "music-encoding",
```

.github/problem_matchers/schematron.json
musicEnfanthen, line 1: would rename this file: schematron-problem-matcher.json

.github/workflows/pr-validation.yml
```diff
@@ -22,6 +26,27 @@ jobs:
         repository: ${{github.event.pull_request.head.repo.full_name}}
         ref: ${{github.event.pull_request.head.ref}}
         fetch-depth: 0
-
+
```
musicEnfanthen:
```suggestion

```

source/modules/MEI.header.xml
musicEnfanthen, line 1: @riedde @gucl-mu Please double check

.github/problem_matchers/schematron.json
musicEnfanthen: @bwbohl

musicEnfanthen:
@rettinghaus How do you test the matcher?

rettinghaus:
@musicEnfanthen I don't 😉  
It's a regex. Take a line from the output of the CI and build a regex. You can then check it here: https://regex101.com/

bwbohl:
> @bwbohl Moved the purification job into a separate job, so it is reported in the PR separately from the source validation.

fantastic!

source/modules/MEI.header.xml
bwbohl: yes please, @riedde and @gucl-mu check if the resulting rng matches your original design ;-)

source/modules/MEI.genetic.xml
musicEnfanthen, line 1: @bwbohl Can we move the fixes for the warnings into a separate PR? Maybe it's a bit overcautious, but seems to be better in terms of separation of concerns?

bwbohl: Wanted to try the results but, was thinking about that, too!
Of course, we can move them ;-)

musicEnfanthen: @bwbohl Perfect, I think the PR is ready to be merged as soon as the fixes are outsourced.

musicEnfanthen: Thank you!

bwbohl:
I rewrote the history of this PR’s branch to drop the schema fixes and moved them to #1665

musicEnfanthen (approved):

musicEnfanthen:
Perfect, thanks!

PR #1627, opened 2025-02-07
patch: develop-profileDescPart

riedde:
developing further

### DIFF
--- a/source/examples/header/header-sample080.txt
+++ b/source/examples/header/header-sample080.txt
@@ -1,0 +1,33 @@
+<fileDesc>
+    <titleStmt>
+        <title>La clé du caveau</title>
+    </titleStmt>
+    <pubStmt>
+        <publisher>Capelle &amp; Renand</publisher>
+        <pubPlace>Paris</pubPlace>
+    </pubStmt>
+    <sourceDesc>
+        <source>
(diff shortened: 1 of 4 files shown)


riedde:
fixing bugs and adding documentation started in #1122 (modified with updated example)

pe-ro (changes requested):
Looks good to me, although I think in the example there should be a value in the source/biblFull/titleStmt/title element.

riedde:
Yes, that might be good. In the base of this example adaption there was no titleStmt, but I thought it might be helpful here. We can add a value!

source/examples/header/header-sample080.txt
```diff
@@ -0,0 +1,33 @@
...
+    </titleStmt>
+    <pubStmt>
+        <publisher>Capelle &amp; Renand</publisher>
+        <pubPlace>Paris</pubPlace>
+    </pubStmt>
+    <sourceDesc>
+        <source>
+            <biblFull>
+                <titleStmt>
+                    <title/>
```
riedde:
```suggestion
                    <title>La clé du caveau</title>
```

pe-ro (approved):
Perfect! Thanks.

PR #1628, opened 2025-02-07
docs: clarify use of dur.ges

### LINKED ISSUE #1605: Clarify use of @dur.ges
data.DURATION.gestural and data.DURATION both refer to data.DURATION.cmn (and data.DURATION.mensural). That restricts the values of `@dur.ges` to the same as those for `@dur`, which doesn't make sense, does it?

krHERO:
Closes #1605

### DIFF
--- a/source/modules/MEI.gestural.xml
+++ b/source/modules/MEI.gestural.xml
@@ -222,7 +222,7 @@
       duration.</desc>
     <attList>
       <attDef ident="dur.ges" usage="opt">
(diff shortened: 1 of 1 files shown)


kepper (approved):
fits our discussions here in Amsterdam – many thanks to @krHERO for taking this on. This doesn't address the overarching question of clarifying the relation between different domains, but that should not stop this one…

bwbohl (approved):
many thanks!

PR #1629, opened 2025-02-07
patch: develop-relax-frbr

riedde:
updating documentation and fixing bugs


source/docs/03-metadata.xml
```diff
@@ -900,15 +900,16 @@
...
                <list rend="numbered">
                   <item>
                      <gi scheme="MEI">identifier</gi>
                   </item>
                   <item>
                      <gi scheme="MEI">title</gi>
                   </item>
-                  <item>Responsibility-like elements including <gi scheme="MEI">composer</gi>, <gi scheme="MEI">lyricist</gi>, <gi scheme="MEI">contributor</gi> and others</item>
+                  <item>
+                     <gi scheme="MEI">respStmt</gi> (before MEI 6.0: Responsibility-like elements (e.g.,  composer, lyricist and others)</item>
```
musicEnfanthen: @riedde There are two opening parentheses "( ... (", but only one is closing ")"

source/docs/03-metadata.xml
```diff
@@ -900,15 +900,16 @@
                      <specDesc key="work"/>
                   </specList>
                </p>
-               <p>All the components of <gi scheme="MEI">work</gi> are optional, but they must occur in the following order:</p>
+               <p>All the components of <gi scheme="MEI">work</gi> are optional. Before MEI 6.0 they had to occur in the following order:</p>
```
musicEnfanthen: @riedde @pe-ro If I get it right, there is no strict order any more for MEI 6.0, right? Would be good to have short indication of that

source/docs/03-metadata.xml
```diff
@@ -1338,7 +1339,9 @@
...
                </p>
                <p>Many of these elements are already described in chapter <ptr target="#headerstructure"/>, especially in <ptr target="#headerWorkDescription"/>.</p>
-               <p>The <gi scheme="MEI">manifestationList</gi> is available to create lists of physical sources representing a work, for instance for use in a thematic catalog or a critical edition. The <gi scheme="MEI">manifestation</gi> child element corresponds to the <ptr target="#FRBR"/> level of the same name, that is, it describes embodiments of certain expressions of a work. The list below reflects the order in which the optional components of manifestation must occur. <specList>
+               <p>The <gi scheme="MEI">manifestationList</gi> is available to create lists of physical sources representing a work, for instance for use in a thematic catalog or a critical edition. The <gi scheme="MEI">manifestation</gi> child element corresponds to the <ptr target="#FRBR"/> level of the same name, that is, it describes embodiments of certain expressions of a work. The list below reflects the order in which the optional components of manifestation had to occur before MEI 6.0.
```
musicEnfanthen: @riedde @pe-ro Same here. Short indication of how it should/can be used now (with MEI 6.0 and later) would be helpful and reduce confusion.

source/docs/03-metadata.xml
riedde:
```suggestion
                     <gi scheme="MEI">respStmt</gi> [before MEI 6.0: Responsibility-like elements (e.g.,  composer, lyricist and others)]</item>
```

Oh, thats a typo!

source/docs/03-metadata.xml
riedde: Is this not clear by the change I propose?

source/docs/03-metadata.xml
riedde: I thought about cutting this text, but then we will lost the documentation for pre-MEI6. What about something like: "The following elements can be used in any order. Before MEI6 they had to be ordered as shown below"?

source/docs/03-metadata.xml
ahankinson: Since this is the documentation *for* MEI 6.0, then there's no need to specify that it's "Before MEI 6.0..." -- it's sufficient just to state it outright. "They may occur in any order".

If people want the pre-6 documentation where this is different, they can use that.

source/docs/03-metadata.xml
ahankinson: Again, this is the documentation *for* 6.0, not the documentation for earlier versions.

source/docs/03-metadata.xml
musicEnfanthen: Yes, good point @ahankinson. I thought about this after I had already made the comment above, too, and I agree, since we follow the same approach in other chapters as well. 
The documentation for pre-MEI6 can be found in the documentation for MEI5 which reads: `... are optional, but they must occur in the following order:`

Changing it to `... are optional, and may occur in any order:` sounds like a good solution.

riedde: Okay, I just didn't want wo loose this information. But I'm absolutely with you @ahankinson @musicEnfanthen !

riedde:
```suggestion
               <p>All the components of <gi scheme="MEI">work</gi> are optional. They may occur in any order.</p>
```

source/docs/03-metadata.xml
riedde:
```suggestion
               <p>The <gi scheme="MEI">manifestationList</gi> is available to create lists of physical sources representing a work, for instance for use in a thematic catalog or a critical edition. The <gi scheme="MEI">manifestation</gi> child element corresponds to the <ptr target="#FRBR"/> level of the same name, that is, it describes embodiments of certain expressions of a work. The components of manifestation may occur in any order.
```

source/docs/03-metadata.xml
```diff
@@ -900,15 +900,16 @@
...
                <list rend="numbered">
                   <item>
                      <gi scheme="MEI">identifier</gi>
                   </item>
                   <item>
                      <gi scheme="MEI">title</gi>
                   </item>
-                  <item>Responsibility-like elements including <gi scheme="MEI">composer</gi>, <gi scheme="MEI">lyricist</gi>, <gi scheme="MEI">contributor</gi> and others</item>
+                  <item>
+                     <gi scheme="MEI">respStmt</gi> [before MEI 6.0: Responsibility-like elements (e.g.,  composer, lyricist and others)]</item>
```
```suggestion
                     <gi scheme="MEI">respStmt</gi></item>
```

riedde: @ahankinson Sorry, I missed this one

ahankinson: That’s ok!

source/docs/03-metadata.xml
```diff
@@ -900,15 +900,16 @@
                      <specDesc key="work"/>
                   </specList>
                </p>
-               <p>All the components of <gi scheme="MEI">work</gi> are optional, but they must occur in the following order:</p>
+               <p>All the components of <gi scheme="MEI">work</gi> are optional. They may occur in any order.</p>
                <list rend="numbered">
```
musicEnfanthen:
```suggestion
               <list rend="bulleted">
```

musicEnfanthen: Sorry, just recognized this one.

At the moment, there is no difference in the rendering between numbered or bulleted lists, but maybe at some point there will be. Since the components here are not strictly ordered anymore, we should also "relax" the listing.

rettinghaus: There is, but it must "ordered", not "numbered".

riedde: After investigation through the mei guidelines there is `list/@rend` with the following values: `simple` (one time), `numbered` and `bulleted`. What I like to see here is `bulleted`.

rettinghaus (approved):
MEI has simple/marked/ordered as values for the form attribute. So I guess you mean TEI Guidelines. 😉

riedde:
> MEI has simple/marked/ordered as values for the form attribute. So I guess you mean TEI Guidelines. 😉

Try `//list[@rend]` on `/source/docs` ;-)

rettinghaus:
Now I get what you meant.

PR #1634, opened 2025-02-12
utils: fix example images being created with the wrong suffix

maxrothman:
When generate_examples.py was migrated to use Python's pathlib in 4ee9761d, the suffixes of generated example images (in assets/images/GeneratedImages) went from being ".mei.svg" to just ".svg" because Path.with_suffix() replaces, rather than appends to, the existing suffix. This caused references to those files on the site to be broken.

This change reverts to the previous behavior, restoring those links.

Refs #1458

### DIFF
--- a/utils/guidelines_xslt/odd2html/guidelines.xsl
+++ b/utils/guidelines_xslt/odd2html/guidelines.xsl
@@ -567,7 +567,7 @@
         <xsl:variable name="verovioClass" select="if($renderedLive) then(' verovio') else('')" as="xs:string"/>
         <xsl:variable name="id" select="generate-id(.)"/>
         <xsl:if test="$renderedLive">
-            <xsl:variable name="imageUrl" select="$assets.folder.generated.images.rel || $id || '.mei.svg'"/>
+            <xsl:variable name="imageUrl" select="$assets.folder.generated.images.rel || $id || '.svg'"/>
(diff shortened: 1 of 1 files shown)


rettinghaus:
Thanks for spotting this and proposing the fix. However, I find double suffixes (like `.mei.svg`) extremely ugly. 

Maybe better to change this line: 
https://github.com/music-encoding/music-encoding/blob/a4d4e3ec40beda34e97c24275f1bcb814a0fac72/utils/guidelines_xslt/odd2html/guidelines.xsl#L570

maxrothman:
@rettinghaus done, ready for rereview

rettinghaus (approved):

musicEnfanthen:
Thanks for catching this, @maxrothman ! And thanks for the adjusted fix @rettinghaus, seems to be indeed the cleanest solution! What do you think @bwbohl ?

> [!CAUTION]
> BEFORE merging this PR, this will also need an update in the docker-mei repo, since the node script appends, not replaces, the svg suffix there, too.

musicEnfanthen:
Needs music-encoding/docker-mei#75

bwbohl (approved):

rettinghaus (approved):

musicEnfanthen:
Thank you!

musicEnfanthen:
@maxrothman Could you please double-check that it is now working as expected?

PR #1638, opened 2025-02-13
guidelines: document `<repeatMark>`, expand `<expansion>` docs

maxrothman:
When `<repeatMark>` was added, no documentation to the guidelines was added to describe it, and portions of the guidelines as well as the docs on `<dir>` were not updated to reflect that `<repeatMark>` was taking over use cases previously solved with `<dir>`. This change remedies that, and expands the documentation of `<expansion>`.

A few things that came up while I was working on this:
- The graphic in the `<repeatMark>` docs renders very large in the browser. Is there a way to set the max size on images?
- It appears that the documentation of `<score>` and `<parts>` is duplicated: once in [2.1.2.2](https://music-encoding.org/guidelines/v5/content/shared.html#sharedMdivContent) and again in [2.2.1](https://music-encoding.org/guidelines/v5/content/shared.html#sharedScoreAndParts)
- I noticed that `<repeatMark>` has the capability to represent Da Capo and Dal Segno, but not D.<S | C>. al <fine | coda>. I guess right now you'd have to use `<dir>`? Was that an oversight, and if so should they be added? Happy to make an issue if it needs discussion.


ahankinson:
NB: Edited the text to encode the tags in backticks.

.gitignore
musicEnfanthen, line 1: @maxrothman Can you remove this unrelated file from the PR please?

maxrothman: Would you be open to keeping this change in a separate commit in the same PR? I don't feel strongly enough about this change to derail the entire PR discussion about it, but I think it's worthwhile to include:
- `.ant-targets-build.xml` is generated by the build, so like `dist/` and `build/` it makes sense to ignore
- It's fairly typical to ignore editor-generated files in gitignores for the convenience of maintainers

pe-ro:
> I noticed that `<repeatMark>` has the capability to represent Da Capo and Dal Segno, but not D.<S | C>. al <fine | coda>. I guess right now you'd have to use `<dir>`? Was that an oversight, and if so should they be added? Happy to make an issue if it needs discussion.
```
<repeatMark>D.S. al coda</repeatMark>
<repeatMark>D.C. al fine</repeatMark>
```
are legitimate uses of `<repeatMark>`.

`@func` is intended to be used as an alternative to textual content in `<repeatMark>`.  For instance,
```
<repeatMark func="coda"/>
```
will result in the display of a coda sign.  

`@func` serves another purpose as well. It allows one to assign a "repetition function" to any symbol or text.  For instance,
```
<repeatMark func="coda">::</repeatMark>
```
indicates that the text "::" represents a coda. In other words, `@func` serves a normalizing purpose.  The various permutations of "D.C.", "D.S", "coda", and "fine" could be added to the value list of `@func`, but they would duplicate the text values usually found in `<repeatMark>`.

This is explained (perhaps too briefly) in the remarks section at https://music-encoding.org/guidelines/dev/elements/repeatMark.html.  Perhaps a longer explanation should be added to the Guidelines as well.

pe-ro:
> It appears that the documentation of `<score>` and `<parts>` is duplicated: once in [2.1.2.2](https://music-encoding.org/guidelines/v5/content/shared.html#sharedMdivContent) and again in [2.2.1](https://music-encoding.org/guidelines/v5/content/shared.html#sharedScoreAndParts)

While `<score>` and `<parts>` are mentioned in both places, their use is explained only in section 2.1.2.2.  They're just mentioned in passing in 2.2.1. The focus in this section is on the items discussed in the other subsections of 2.2.

Of course, suggestions for improvement are always welcome.

.gitignore
musicEnfanthen: @maxrothman Totally agree that it is good to have things excluded via the gitignore. But would not introduce it via this "content-related" PR, but make a separate PR for this.

I am wondering about the `.ant-targets-build.xml`, though. Never got this one when building with ant. Is this specific to your OS?

maxrothman: No idea, maybe it's OS- or editor-specific? I've split this out into https://github.com/music-encoding/music-encoding/pull/1642.

maxrothman:
@pe-ro 

> D.<S | C>. al <fine | coda>

This seems like it's going to grow into a larger discussion, I've split that discussion out into https://github.com/music-encoding/music-encoding/issues/1643.

maxrothman:
> While `<score>` and `<parts>` are mentioned in both places, their use is explained only in section 2.1.2.2. They're just mentioned in passing in 2.2.1. The focus in this section is on the items discussed in the other subsections of 2.2.
> 
> Of course, suggestions for improvement are always welcome.

That makes sense to me. Perhaps my confusion came more from the section titles than anything else, would it be fair to say that 2.2.1 is more about what goes _inside_ `<score>` & `<parts>` than the elements themselves? If so, how about "Content of Scores and Parts", "Qualities of Scores and Parts", or similar?

And perhaps we could add a crosslink from 2.2.1 to 2.1.2.2. I'm happy to add that to this PR if you're on board.

pe-ro:
I agree that section 2.2.1 is about the content of the `<score>` and `<part>` elements.  So, I suggest --

1. Change the heading of this section to "Score and Part Content";
2. Modify the first sentence from "The following elements are provided for the capture of scores and parts:" to "The following elements are provided for capturing the content of a score or part:";
3.  Eliminate `<score>`, `<parts>`, and `<part>` from the list of elements at the top of this section.

The content of section 2.1.2.2 can stand unchanged.

Adding a link from 2.1.2.2 to 2.2.1 is fine with me.

pe-ro (changes requested):
Change:  

```
<p>The following elements are provided for capturing the content of scores 
and parts. The capturing of scores and parts themselves is discussed in 
section <ptr target="#sharedMdivContent"/>.</p>
```

to

```
<p>The following elements are provided for capturing the content of scores 
and parts. The score and parts elements themselves are discussed in 
section <ptr target="#sharedMdivContent"/>.</p>
```

pe-ro (dismissed):
Looks good to me

maxrothman:
It looks like this PR is still waiting on a second reviewer to sign off. Was there something I'm supposed to be doing to get that going, or is it just a matter of waiting for another maintainer to have the time for it?

source/docs/02-shared.xml
```diff
@@ -233,72 +233,90 @@
...
+                        <p>Rather than using a single monolithic <gi scheme="MEI">expansion</gi>, <gi scheme="MEI">expansion</gi> elements can be placed throughout the document, indicating repetition of their sibling/descendent elements:</p>
+                        <p>
+                           <figure>
+                              <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/shared/shared-expansion-many.txt" parse="text"/></egXML>
+                           </figure>
+                        </p>
+                        <p>In cases where different sources indicate different expansions of the music, multiple <gi scheme="MEI">expansion</gi> elements can be specified, each with a different <att>source</att> attribute. When multiple sources agree, <att>source</att> can contain multiple values. This indicates semantically similar information to the critical apparatus features described in <ptr target="critApp"/> and is provided for cases where the nested markup required to use those features would be clumsy.</p>
+                        <p></p>
```
musicEnfanthen: Empty line needed?

source/examples/shared/shared-expansion-many.txt
```diff
@@ -0,0 +1,26 @@
+<music>
+  <body>
+    <mdiv>
+      <score>
+        <section>
+          <expansion plist="#A #B #A" />
+          <section xml:id="A">
+            <!-- "A" section -->
+          </section>
+          <section xml:id="B">
```
musicEnfanthen, line 10:
```suggestion
          <section xml:id="B">
            <!-- "B" section -->
```

source/examples/shared/shared-expansion-many.txt
```diff
@@ -0,0 +1,26 @@
...
+            <expansion plist="#B0 #B1 #B0 #B2" />
+            <section xml:id="B0">
+              <!-- "B" section -->
```
musicEnfanthen: Actually, the full B section is the parent section, right? Including the endings. So this one would be the B0 section according to the ids used?

source/examples/shared/shared-expansion-many.txt
```diff
@@ -0,0 +1,26 @@
+<music>
+  <body>
+    <mdiv>
+      <score>
+        <section>
+          <expansion plist="#A #B #A" />
```
musicEnfanthen, line 6: Maybe we can go with the same style of ids as in the sample-009 file: `shared.A` `shared.B` `shared.B0` etc. This would make the examples easier to compare.

source/docs/02-shared.xml
```diff
@@ -233,72 +233,90 @@
...
+                        <p>
+                           <figure>
+                              <head/>
+                              <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/shared/shared-sample009.txt" parse="text"/></egXML>
+                           </figure>
+                        </p>
+                        <p>If present, all of the <gi scheme="MEI">expansion</gi> elements must be the initial children of their parent element.</p>
+                        <p>Rather than using a single monolithic <gi scheme="MEI">expansion</gi>, <gi scheme="MEI">expansion</gi> elements can be placed throughout the document, indicating repetition of their sibling/descendent elements:</p>
```
musicEnfanthen: placed --> nested ?

source/docs/02-shared.xml
```diff
@@ -366,13 +384,10 @@
                <head>General Music Elements</head>
                <p>This section lists the elements defined in the shared module that are available within the music element.</p>
                <div xml:id="sharedScoreAndParts" type="div3">
-                  <head>Score and Parts</head>
-                  <p>The following elements are provided for the capture of scores and parts:</p>
+                  <head>Score and Part Content</head>
+                  <p>The following elements are provided for capturing the content of scores and parts. The score and parts elements themselves are discussed in section <ptr target="#sharedMdivContent"/>.</p>
```
musicEnfanthen:
```suggestion
                  <p>The following elements are provided for capturing the content of scores and parts. The score and parts elements themselves are discussed in section <ptr target="#sharedScoreAndPartsStruc"/>.</p>
```

musicEnfanthen: Or did you refer to the parent chapter for a specific reason?

source/docs/04-cmn.xml
```diff
@@ -484,6 +484,23 @@
             <div xml:id="cmnAdditionalMeasureContent" type="div2">
                <head>Advanced CMN Features</head>
                <p>Over time, in addition to the basic features of note, chord, and rest, many other symbols have been added to CMN. The following section describes some of these symbols and introduces their handling in MEI.</p>
+               <div xml:id="cmnBarlines" type="div3">
+                  <head>Bar lines</head>
+                  <p>The bar lines bounding measures can take a variety of different forms to indicate various musical features.</p>
+                  <p>
+                     <specList>
+                        <specDesc key="att.measure.log" atts="left right"/>
```
```suggestion
                        <specDesc key="att.measure.log" atts="right left"/>
```

musicEnfanthen: `right` would be usual use case, so it should be listed first.

source/docs/04-cmn.xml
```diff
@@ -484,6 +484,23 @@
             <div xml:id="cmnAdditionalMeasureContent" type="div2">
                <head>Advanced CMN Features</head>
                <p>Over time, in addition to the basic features of note, chord, and rest, many other symbols have been added to CMN. The following section describes some of these symbols and introduces their handling in MEI.</p>
+               <div xml:id="cmnBarlines" type="div3">
+                  <head>Bar lines</head>
+                  <p>The bar lines bounding measures can take a variety of different forms to indicate various musical features.</p>
```
musicEnfanthen: Can we add a sentence that the different forms can be controlled in MEI via the following attributes on the measure element ....? 

Otherwise the attributes are listed here without any context.

source/docs/04-cmn.xml
```diff
@@ -1152,7 +1169,7 @@
...
+                  <p>In CMN scores, there is often a large number of natural language instructions. Some of them concern the loudness and the speed of the performance, in which case MEI offers the elements <gi scheme="MEI">dynam</gi> (described at <ptr target="#cmnDynam"/>) and <gi scheme="MEI">tempo</gi>. Others are structural repetition instructions, for which MEI offers the <gi scheme="MEI">repeatMark</gi> element. In other cases, however, they provide other instructions for the performer. Instead of providing separate elements for all possible types of such directions, MEI offers the generic <gi scheme="MEI">dir</gi> element. Although this element is not CMN specific (it is defined in <ptr target="#shared"/>), it is frequently used in this repertoire.</p>
```
musicEnfanthen: I am not sure about the benefit of this addition. `repeatMark` is not discussed in this section, but the following. So this could be potentially confused with `rehearsalMark` discussed in this section?

source/docs/02-shared.xml
```diff
@@ -233,72 +233,90 @@
...
+                        <p>However, staff-by-staff organization is more appropriate for music without measures and is provided when either the MEI.mensural or MEI.neumes module is employed. Coverage of mensural notation is provided in chapter <ptr target="#mensural"/>, while <ptr target="#neumes"/> describes neumatic notation.</p>
+                        <p>It must be noted that, when both the MEI.cmn and MEI.mensural modules are available, it is possible to encode CMN notation without using <gi scheme="MEI">measure</gi> elements; that is, staff-by-staff organization may be used and the ends of measures marked using <gi scheme="MEI">barLine</gi> elements.</p>
+                        <p>In certain circumstances, this approach may be preferable for reproduction of the visual layout of the music. However, the simultaneous use of the <gi scheme="MEI">measure</gi> and <gi scheme="MEI">barLine</gi> elements may lead to confusion and should be avoided.</p>
+                     </div>
+                     <div xml:id="sharedStrucRep" type="div5">
+                        <head>Structural Repetition</head>
```
musicEnfanthen, line 277: Maybe rename this section to "Large-Scale Structural Repetition"? (to distinguish it from those strucReps in chapter 4)

source/docs/04-cmn.xml
```diff
@@ -1193,7 +1210,29 @@
                   <p>Repetition is a characteristic feature of music. Many musical forms rely on repetition (sometimes with modification) of distinct sections of the music. Repetition in this sense can be thought of as ‘structural’. At the same time, composers and engravers of music often use local symbols for repeating smaller portions of music instead of writing them in full more than once. In this case, the repetition is better defined as a species of abbreviation.</p>
                   <div xml:id="cmnRepStruct" type="div4">
                      <head>Structural Repetition</head>
```
musicEnfanthen, line 1238: Maybe rename to "Repetition Marks"? (Similar to "Rehearsal Marks" above and in disctinction from the large-scale structural repetions discussed in chapter 2? And since the section here only refers to the element repeatMark, renaming it would be helpful?)

musicEnfanthen: Also the id could be then `cmnRepeatMark`? (would need to be checked for references)

source/docs/04-cmn.xml
```diff
@@ -1193,7 +1210,29 @@
...
+                        <specList>
+                           <specDesc key="repeatMark"/>
+                        </specList>
+                     </p>
+                     <p>
+                        Its <att>func</att> attribute indicates the function of the mark, usually
+                        implying a corresponding form; allowed values are described in <ident
+                        type="class">att.repeatMark.log</ident>. Textual content or glyph
+                        information (via <ident type="class">att.extSym.names</ident>) can also be
+                        provided, and will override the symbol implied by <att>func</att>.
```
```suggestion
                        Its <att>func</att> attribute indicates the function of the mark, usually implying a corresponding form; allowed values are described in <ident type="class">att.repeatMark.log</ident>. Textual content or glyph information (via <ident type="class">att.extSym.names</ident>) can also be provided, and will override the symbol implied by <att>func</att>.
```

musicEnfanthen: Also: described -> listed ?

source/docs/04-cmn.xml
```diff
@@ -1193,7 +1210,29 @@
...
+                     </p>
+                     <p>The placement the repeat mark is indicated using a <att>tstamp</att>, <att>tstamp.ges</att>, <att>tstamp.real</att>, or <att>startId</att> attribute. To place the mark above a bar line, a <att>tstamp</att> value of one more than the number of beats in the measure can be used:</p>
+                     <p>
+                        <figure>
+                           <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code verovio"
+                           xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/verovio/repeatMark.mei" parse="text"/></egXML>
+                        </figure>
+                     </p>
+                     <p>The semantic meaning of the repetition in both cases can be encoded using the <gi scheme="MEI">expansion</gi> element, discussed in section <ptr target="#sharedStrucRep"/>.</p>
```
musicEnfanthen: what is "both cases" referring to?

source/docs/04-cmn.xml
```diff
@@ -1193,7 +1210,29 @@
...
+                        </specList>
+                     </p>
+                     <p>
+                        Its <att>func</att> attribute indicates the function of the mark, usually
+                        implying a corresponding form; allowed values are described in <ident
+                        type="class">att.repeatMark.log</ident>. Textual content or glyph
+                        information (via <ident type="class">att.extSym.names</ident>) can also be
+                        provided, and will override the symbol implied by <att>func</att>.
+                     </p>
+                     <p>The placement the repeat mark is indicated using a <att>tstamp</att>, <att>tstamp.ges</att>, <att>tstamp.real</att>, or <att>startId</att> attribute. To place the mark above a bar line, a <att>tstamp</att> value of one more than the number of beats in the measure can be used:</p>
```
```suggestion
                     <p>The placement of the repeat mark is indicated using a <att>tstamp</att>, <att>tstamp.ges</att>, <att>tstamp.real</att>, or <att>startId</att> attribute. To place the mark above a bar line, a <att>tstamp</att> value of one more than the number of beats in the measure can be used (e.g., <val>5</val> in 4/4 time):</p>
```

source/docs/04-cmn.xml
```diff
@@ -1193,7 +1210,29 @@
...
                   <div xml:id="cmnRepStruct" type="div4">
                      <head>Structural Repetition</head>
-                     <p>Large-scale structural repetition, utilizing <gi scheme="MEI">section</gi> and <gi scheme="MEI">expansion</gi> elements, is discussed in section <ptr target="#sharedMdivContent"/>. This section will focus on repetition within sections.</p>
+                     <p>The most common variety of structural repetition is the use of repeat bar lines and endings. The encoding of bar lines is discussed in section <ptr target="cmnBarlines"/>, and endings in <ptr target="#sharedStrucRep"/>.</p>
+                     
+                     <p>For larger-scale repetition using da capo, segnos, codas, and etc., MEI provides the <gi scheme="MEI">repeatMark</gi> element:</p>
```
musicEnfanthen, line 1240:
```suggestion
                     <p>The most common variety of structural repetition is the use of repeat bar lines and endings. The encoding of bar lines is discussed in section <ptr target="cmnBarlines"/>, and that of large-scale structural repetition, utilizing <gi scheme="MEI">section</gi>, <gi scheme="MEI">ending</gi> and <gi scheme="MEI">expansion</gi> elements, is discussed in section <ptr target="#sharedStrucRep"/>.</p>
                     <p>This section will focus on repetition within sections using repetition marks like da capo, segnos, codas, and etc. For this puprose, MEI provides the <gi scheme="MEI">repeatMark</gi> element:</p>
```

musicEnfanthen: Above is meant as a proposal to keep the reference to the large-scale struReps in ch. 2

source/modules/MEI.shared.xml
```diff
@@ -5456,8 +5456,7 @@
       </constraint>
     </constraintSpec>
     <remarks xml:lang="en">
-      <p>Examples include text strings, such as 'affettuoso', and music symbols, such as segno and
-        coda symbols, fermatas over a bar line, etc. Directives can be control elements. That is,
+      <p>Examples include text strings, such as 'affettuoso' and music symbols. Directives can be control elements. That is,
```
```suggestion
      <p>Examples include text strings, such as 'affettuoso', and music symbols. Directives can be control elements. That is,
```

musicEnfanthen:
@maxrothman Thank you! Some suggestions above.

source/examples/shared/shared-expansion-many.txt
maxrothman: That sounds right to me. I was using these comments primarily as a sort of "your content goes here" mark. If it's unclear as-is, perhaps I could change this comment to "B section repeated content" or something like that?

source/examples/shared/shared-expansion-many.txt
maxrothman: I agree it makes sense to both examples the same way. Do you have a preference for which one is changed? I feel like the "#A #B #A" style is easier to read, so my first instinct would be to alter sample-009.

source/docs/02-shared.xml
maxrothman: How about "placed in sections and endings"? To me, "nested" sounds like you're putting expansions inside expansions.

source/docs/02-shared.xml
maxrothman: It was a typo, fixed!

maxrothman:
Thanks for the review! I took care of the low-hanging fruit, I'll come back around to the more involved comments sometime this week hopefully.

source/examples/shared/shared-expansion-many.txt
musicEnfanthen: Sounds good to me.

source/docs/04-cmn.xml
maxrothman: My intent was to, like with `<tempo>` and `<dynam>`, indicate to readers that `<repeatMark>` should be preferred to `<dir>` when it would be appropriate to use. If I changed it to read something like "...offers the repeatMark element, discussed in section x.x.x" would that address the concern?

source/docs/02-shared.xml
maxrothman: I'd think it'd make sense to apply that suggestion to chapter 4, i.e. rename 4.3.11.1 to "Large-Scale Structural Repetition" and leave 2.1.2.2.3 as "Structural Repetition", since the one in ch4 is specifically about `<repeatMark>` at the one in ch2 is more general to all types of structural repetition. Are you alright with that change?

source/docs/04-cmn.xml
maxrothman: Looking holistically at this comment and the previous one, it seems like this chapter could be titled "Repeat Marks" or "Large-Scale Structural Repetition". I don't have a super strong opinion about those choices, though it does occur to me that in common musician lingo, a "repeat mark" usually refers to a repeat barline, which could be confusing to readers.

Thoughts?

source/docs/04-cmn.xml
maxrothman: Left over from a previous iteration of the text, removed

source/docs/04-cmn.xml
maxrothman: I'm not sure I follow the intent of this change? My general understanding is that repeat bars and endings are typically smaller-scale than the marks represented by `<repeatMark>`, so I find the reference to endings being "large-scale" confusing. Also, `<section>` and `<expansion>` do not encode any marks found in music encoding directly, so I think readers may find their mention here confusing among references to encoded marks like repeat barlines and endings.

maxrothman:
@musicEnfanthen I think I've addressed all the comments you left in your last review in either comments or changes. Let me know if I missed anything!

maxrothman:
Hi @musicEnfanthen, it looks like this branch has ended up conflicting with develop. I'd prefer to only have to do conflict resolution once, so would you mind giving me a heads up when you're ready to come back around to giving this PR another look? Thanks!

source/docs/04-cmn.xml
musicEnfanthen: Personally, I would prefer "Repeat Marks", but would be interested what other native speakers think about it? @pe-ro @ahankinson ?

source/docs/04-cmn.xml
musicEnfanthen: I see. Yes, a pointer to the other section would be helpful here.

musicEnfanthen:
@maxrothman Resolved the conflict which in fact was GH marking your newly added changes to shared.xml as conflict for whatever reason.

maxrothman:
Thanks for your help resolving that conflict. I've addressed the two comments you made recently, if I'm not mistaken then the last remaining issue is about the naming of the sections?

musicEnfanthen:
@pe-ro @ahankinson Any second thoughts on the outstanding comments? Cf. https://github.com/music-encoding/music-encoding/pull/1638#discussion_r1966745982 and following.

maxrothman:
Do you mean https://github.com/music-encoding/music-encoding/pull/1638#discussion_r1966747370? As far as I can tell, https://github.com/music-encoding/music-encoding/pull/1638#discussion_r1966745982 is resolved.

musicEnfanthen:
Yes, of course, you're correct. Getting confused by my own comments 🙈

martha-thomae:
> Thanks for your help resolving that conflict. I've addressed the two comments you made recently, if I'm not mistaken then the last remaining issue is about the naming of the sections?

@ahankinson and @pe-ro, any thoughts about the state of this PR? Could you check the last remaining three comments about renaming a few sections? These ones:
- https://github.com/music-encoding/music-encoding/pull/1638#discussion_r1975706019
- https://github.com/music-encoding/music-encoding/pull/1638#discussion_r2035729972
- https://github.com/music-encoding/music-encoding/pull/1638#discussion_r1975730807

This will be very helpful to resolve the issue with the lack of documentation for `<repeatMark>`. Thank you :)

ahankinson:
I think my only involvement in this ticket was to format the original one with code ticks. I don't see any places where I've commented, and don't have anything to add to @musicEnfanthen 's comments.

pe-ro:
@martha-thomae, I'd like to suggest the following wording for chapter 4, which I hope will make clear the use of `<repeatMark>` as well as the distinction between "large-scale", "sectional", and "intra-measure" repetition .

Chapter 2 can stand as-is.

I leave it to someone else (@maxrothman ?) to integrate this into the Guidelines. Coordination/integration of this text and Max's would be welcome.

```
4.3.10 Repetition in CMN

Repetition is a characteristic feature of music. Many musical forms rely on repetition (sometimes
with modification) of distinct sections of the music. Repetition in this sense can be thought of
as ‘structural’. At the same time, composers and engravers of music often use local symbols for
repeating smaller portions of music instead of writing them in full more than once. In this case,
the repetition is better defined as a species of abbreviation.

4.3.10.1 Large-Scale Structural Repetition

Large-scale structural repetition; that is, repetition of multiple sections, can be encoded 
using one or more &lt;expansion> elements or via the &lt;repeatMark> element.

Use of &lt;section> and &lt;expansion> elements is discussed in section 2.1.2.2 Content of
Musical Divisions. This approach enables efficient encoding of complex musical structures
(repeats, da capos, multiple endings, etc.) separate from their performance implications. Multiple
performed versions, even conflicting ones, can be captured without changing the encoded structures
themselves.

The &lt;repeatMark> element, a specialized form of the &lt;dir> (performance directive) element,
takes a more visually-oriented approach. The textual content of the &lt;repeatMark> element holds
the text and/or symbols that visually indicate repetition. For example, the text may use a
commonly-accepted phrase or a combination of text and symbols:

<repeatMark>D.C. al fine</repeatMark>
<repeatMark>D.S.</repeatMark>
<repeatMark>D.S. al <symbol glyph.num="#x1D10C"/></repeatMark>

The @func attribute serves multiple purposes. First, it can provide an alternative to the textual 
content inside the &lt;repeatMark> element. For instance,

<repeatMark func="coda"/>

can be used by a renderer to display a coda sign at the point indicated by the @tstamp attribute.

In addition, it allows one to assign a "repetition function" to any symbol or text. For instance,

<repeatMark func="coda">::</repeatMark>

indicates that the text "::" represents a coda. In other words, @func can be used to "normalize"
historical or individual practice. It may also be used to capture editorial suggestions not
present in the music notation itself; e.g.,

<repeatMark func="rptLeft"/>

4.3.10.2 Section-Level Repetition

Repetition of the contents of a section can be encoded using the @left and @right attributes on
the first and last measures of the repeated material:

<section>
  <measure left="rptstart">
    <!-- content of m. 1 -->
  </measure>
  <measure right="rptend">
    <!-- content of m. 2 -->
  </measure>
</section>

When a measure ends one repeated section and begins another, both the @left and @right 
attributes can be used:

<section>
  <measure left="rptstart" right="rptboth">
    <!-- content of m. 1 -->
  </measure>
  <measure right="rptend">
    <!-- content of m. 2 -->
  </measure>
</section>

4.3.10.3 Mid-measure Repetition

In addition to repetition at the section level, CMN includes a number of different symbols for
intra-measure repetitions. Many of these symbols are found in manuscripts and may be regarded as
personal conventions of their respective authors. Some signs, however, have been widely adopted.
For example, it is common to indicate the repetition of a single beat or an entire measure with
one or more diagonal lines, sometimes with dots at the upper left and lower right, much like a
percent sign. The illustration below contains the most common signs:

etc.
```

martha-thomae:
Thank you, @pe-ro! 
@maxrothman, if you agree with these changes, please proceed to include them and we will happily approve and merge this PR soon. Thanks everyone!

maxrothman:
I'm working on integrating those suggestions, I had a few questions that came up as I'm working through it:

@pe-ro your suggested text indicates that `@left` and `@right` may be used for encoding repeat barlines, but the guidelines seem to discourage the use of `@left`:
> - @left(optional)Indicates the visual rendition of the left bar line. It is present here only for facilitation of translation from legacy encodings which use it. Usually, it can be safely ignored. Value conforms to [data.BARRENDITION](https://github.com/music-encoding/music-encoding/data-types/data.BARRENDITION.html).
> - @right(optional)Indicates the function of the right bar line and is structurally important. Value conforms to [data.BARRENDITION](https://github.com/music-encoding/music-encoding/data-types/data.BARRENDITION.html).

That would also seem to imply that `<measure n="1" right="rptstart">...</measure><measure n="2" right="rptend">...</measure>` is preferred to `<measure n="1">...</measure><measure n="2" left="rptstart" right="rptend">...</measure>`, which is contrary to the suggested text above. Is this interpretation of the guidelines correct, and if not, perhaps the documentation for `@left` should be updated?

On another note, the renderings of `examples/verovio/repeatMark.mei` in 4.3.11.1 and `images/ExampleImages/barlines.png` in 4.3.1 are showing up very large in the browser rendering. Is there a way to set a max-width on those images?

pe-ro:
Hi, @maxrothman, it seems you're confused about the purpose of `<repeatMark>`.  It is essentially a specialized form of `<dir>` (directive) that can be used to capture repeat marks, like "D.C. al Coda", etc., and left | right | both repeat signs that occur **_above/below the staff_**.  It should not be used to encode repeat marks when they occur **_on the staff_**.  That's what the `@left` and `@right` attributes on `<measure>` are for. The comments you're referring to concerning `@left` apply only there.

@lpugin or @musicEnfanthen can advise about the appearance of images in the Guidelines.

maxrothman:
@pe-ro to clarify, I understand the intended purpose of `<repeatMark>`. I was referring to the content of the section in your suggestion entitled "4.3.10.2 Section-Level Repetition" which contains this snippet for instance:

```xml
<section>
  <measure left="rptstart">
    <!-- content of m. 1 -->
  </measure>
  <measure right="rptend">
    <!-- content of m. 2 -->
  </measure>
</section>
```

If the use of `@left` is discouraged, then I would instead expect the `left="rptstart"` to be on the prior measure.

The section of the guidelines I pasted above discouraging the use of `@left` was drawn from [att.measure.log](https://music-encoding.org/guidelines/v5/attribute-classes/att.measure.log.html), so for use on `<measure>`s, not on `<repeatMark>`s.

pe-ro:
Sorry, looks like I didn't understand. :-)

You're correct, given the recommendation to avoid `@left`, the previous measure should have `@right="rptStart"`.  However, if the example is changed to 
```
<measure right="rptstart"/>
<section>
  <measure>
    <!-- content of m. 1 -->
  </measure>
  <measure right="rptend">
    <!-- content of m. 2 -->
  </measure>
</section>
```
it won't validate.

I'm not sure I completely agree with the admonition to avoid `@left`.  What's supposed to be the markup if the first measure begins with a repeat sign?  In this case, there's no previous measure to attach the repeat sign to.

@lpugin, care to join in?

maxrothman:
Throwing my 2c in, I always found the recommendation against `@left` to be odd. I would've thought that I'd want to always do `@left="rptstart"` and `@right="rptend"`, but never the reverse. Then at least the repeat barlines are on the measures that they "belong" to.

martha-thomae:
@lpugin, I added you as one of the reviewers as I see @pe-ro is asking for your thoughts here: https://github.com/music-encoding/music-encoding/pull/1638#issuecomment-4014099032

lpugin:
```xml
<section>
  <measure left="rptstart">
    <!-- content of m. 1 -->
  </measure>
  <measure right="rptend">
    <!-- content of m. 2 -->
  </measure>
</section>
```
Looks OK to me because this measure is the first in its section. So instead of "discouraging" the use of `@left`, maybe we can instead "recommend" its use only when either at the beginning of a piece or a section. Or course, it does not make much of a difference...

martha-thomae:
@maxrothman, now that things have been clarified, would you mind please adding the text suggested by Perry (https://github.com/music-encoding/music-encoding/pull/1638#issuecomment-4007698438)? 

We could also add the suggestion by Laurent about softening the language about the `@left` attribute:

> So instead of "discouraging" the use of `@left`, maybe we can instead "recommend" its use only when either at the beginning of a piece or a section.

This would be the last change, and we can accept and merge this PR. We have a deadline for July 31st to merge pull requests that are ready. Thanks for all your work on this!

---

Regarding the image, at this moment, the only viable solution is to manually resize your image.

maxrothman:
Sorry, I haven't had time to get to this recently. I don't think it's going to make it over the line by July 31, but hopefully I should have the time shortly thereafter.

bwbohl:
@maxrothman now after the summer break, we’re still working on some technical stuff, and if you’d find the time to update the PR, we’d be happy to include it.

PR #1650, opened 2025-02-28
schema: add additional options to repeatMark's `@func`

### LINKED ISSUE #1643: Options available in <repeatMark>'s @func & documentation thereof
Split from https://github.com/music-encoding/music-encoding/pull/1638#issuecomment-2659638544

Prior context:

@maxrothman:
> I noticed that `<repeatMark>` has the capability to represent Da Capo and Dal Segno, but not D.<S | C>. al <fine | coda>. I guess right now you'd have to use `<dir>`? Was that an oversight, and if so should they be added? Happy to make an issue if it needs discussion.

@pe-ro:
> ```
> <repeatMark>D.S. al coda</repeatMark>
> <repeatMark>D.C. al fine</repeatMark>
> ```
> are legitimate uses of `<repeatMark>`.
> 
> `@func` is intended to be used as an alternative to textual content in <repeatMark>. For instance,
> 
> ```
> <repeatMark func="coda"/>
> ```
> will result in the display of a coda sign.
> 
> `@func` serves another purpose as well. It allows one to assign a "repetition function" to any symbol or text. For instance,
> 
> ```
> <repeatMark func="coda">::</repeatMark>
> ```
> indicates that the text "::" represents a coda. In other words, `@func` serves a normalizing purpose. The various permutations of "D.C.", "D.S", "coda", and "fine" could be added to the value list of `@func`, but they would duplicate the text values usually found in <repeatMark>.
> 
> This is explained (perhaps too briefly) in the remarks section at https://music-encoding.org/guidelines/dev/elements/repeatMark.html. Perhaps a longer explanation should be added to the Guidelines as well.

maxrothman:
Fixes #1643

Some small notes on this change:
- As discussed in the issue, repeatLeft and repeatRight should render their SMuFL small versions by default, so I used those codes in the docs (E04C leftRepeatSmall and E04D rightRepeatSmall) rather than the characters whose name matches the respective `@func` value (E040 repeatLeft and E041 repeatRight).
- However, SMuFL does not have a small variant of repeatRightLeft, so I used the full-size code for that one.
- Unicode has no repeatRightLeft character at all, so I left the Unicode character out for that one.

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -2890,6 +2890,27 @@
           <valItem ident="fine">
             <desc xml:lang="en">Fine. (text)</desc>
           </valItem>
+          <valItem ident="daCapoAlFine">
+            <desc xml:lang="en">D.C. al Fine. (text)</desc>
+          </valItem>
+          <valItem ident="dalSegnoAlFine">
+            <desc xml:lang="en">D.S. al Fine. (text)</desc>
+          </valItem>
+          <valItem ident="daCapoAlCoda">
+            <desc xml:lang="en">D.C. al Coda. (text)</desc>
+          </valItem>
+          <valItem ident="dalSegnoAlCoda">
+            <desc xml:lang="en">D.S. al Coda. (text)</desc>
+          </valItem>
+          <valItem ident="repeatLeft">
+            <desc xml:lang="en">Left repeat barline as a separate symbol disconnected from staff lines (SMuFL E04C or Unicode 1D106)</desc>
+          </valItem>
+          <valItem ident="repeatRight">
+            <desc xml:lang="en">Right repeat barline as a separate symbol disconnected from staff lines (SMuFL E04D or Unicode 1D107)</desc>
+          </valItem>
+          <valItem ident="repeatRightLeft">
+            <desc xml:lang="en">Right and left repeat barline as a separate symbol disconnected from staff lines (SMuFL E042)</desc>
+          </valItem>
         </valList>
       </attDef>
     </attList>


bwbohl (dismissed):

lpugin (dismissed):

bwbohl:
@musicEnfanthen any idea why the validate-source check is stuck?

musicEnfanthen:
@bwbohl because it is not in develop yet, so there is no pureoddd validation action

musicEnfanthen:
We can ignore that check for now

source/modules/MEI.shared.xml
```diff
@@ -2890,6 +2890,27 @@
...
+            <desc xml:lang="en">D.S. al Fine. (text)</desc>
+          </valItem>
+          <valItem ident="daCapoAlCoda">
+            <desc xml:lang="en">D.C. al Coda. (text)</desc>
+          </valItem>
+          <valItem ident="dalSegnoAlCoda">
+            <desc xml:lang="en">D.S. al Coda. (text)</desc>
+          </valItem>
+          <valItem ident="repeatLeft">
+            <desc xml:lang="en">Left repeat barline not appearing in the staff (SMuFL E04C or Unicode 1D106)</desc>
```
```suggestion
            <desc xml:lang="en">Left repeat barline (SMuFL E04C or Unicode 1D106)</desc>
```

source/modules/MEI.shared.xml
```diff
@@ -2890,6 +2890,27 @@
...
+          </valItem>
+          <valItem ident="repeatRight">
+            <desc xml:lang="en">Right repeat barline not appearing in the staff (SMuFL E04D or Unicode 1D107)</desc>
```
```suggestion
            <desc xml:lang="en">Right repeat barline (SMuFL E04D or Unicode 1D107)</desc>
```

source/modules/MEI.shared.xml
```diff
@@ -2890,6 +2890,27 @@
...
+          </valItem>
+          <valItem ident="repeatRightLeft">
+            <desc xml:lang="en">Right and left repeat barline not appearing in the staff (SMuFL E042)</desc>
```
```suggestion
            <desc xml:lang="en">Right and left repeat barline (SMuFL E042)</desc>
```

source/modules/MEI.shared.xml
maxrothman: When I wrote this, I thought it would be helpful to clarify since it's pretty unusual to have a repeat barline that isn't in the staff. Maybe there's a better phrasing for this? how about "Left repeat barline as a separate symbol (outside the staff)"?

Ditto for the comments below

musicEnfanthen: We were talking about this today during the ODD meeting. "outside the staff" was one option we considered, but since you can have `place="within"` on `<repeatMark>`, we thought it might be contrainductive then. But "Left repeat barline as a separate symbol" would be clarifying enough for me...

@lpugin?

maxrothman: Ah, that makes sense. Maybe "as a separate symbol disconnected from staff lines"?

musicEnfanthen: Sounds good to me

musicEnfanthen (approved):

musicEnfanthen:
@lpugin Wording improved, can you have a look, please, if that works for you, too?

lpugin:
I am not sure why we need such as specific explanation since it it about a **control event mark** and that seems by itself explains that it is "a separate symbol disconnected from staff lines", but I am OK with it.

lpugin (approved):

bwbohl (approved):

musicEnfanthen:
Thank you @maxrothman

PR #1651, opened 2025-03-12
schema: revise `incipCode/@form` attribute documentation

pe-ro:
Currently, `incipCode/@form` uses the value "parsons" to indicate that the incipit contains an encoding of melodic contour. A better value here would be "melodicContour", defined as 

> "Representation of 1] diatonic interval direction and quality (e.g., "P0 +P5 +P0 +M2 +P0 -M2"), 2] precise interval direction with size in half steps (e.g., "0 +7 +0 +2 +0 -2"), or 3] general interval direction only (e.g., "sususddsdsdsd", often called "Parsons code")."

This change will better align the `@form` attribute values with the intended values of the `<incipCode>` element.

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -1557,15 +1557,21 @@
           </valItem>
           <valItem ident="parsons">
             <desc xml:lang="en">Parsons code containing intervallic direction only 
-              (e.g., "sususddsdsdsd").</desc>
+              (e.g., "sususddsdsdsd"). Permitted values are composed of a series of tokens, which may 
+              be "u" (Up), "d" (Down) or "s" (Same). Each token should not be separated by a space.</desc>
           </valItem>
-          <valItem ident="diatonicMelodicContour">
-            <desc xml:lang="en">Representation of intervallic direction and diatonic quality
-              (e.g., "P0 +P5 +P0 +M2 +P0 -M2").</desc>
+          <valItem ident="intervals">
+            <desc xml:lang="en">Representation of interval quality and direction 
+              (e.g., "P1 +P5 P1 +M2 P1 -M2"). Values must be composed of a series of space-separated tokens, where each token
+              is comprised of three components: a direction indicator ("+" for "up" and "-" for down); an interval quality, being one of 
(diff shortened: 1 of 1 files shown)


ahankinson:
I think we should keep "Parsons" as-is, since it's a known and [defined specification](https://archive.org/details/directoryoftunes0000pars/page/14/mode/2up) (p. 16), and it has the advantage of remaining backwards-compatible.

Then we could either have "melodicContour" take both other representations, or split the two representations into their own attribute values so that parsers would know what kind of code to expect when reading it.

pe-ro:
I started off doing as you suggest, but figured you'd complain that there would be 3 different values for recording melodic contour.  Seems I can't win.

ahankinson:
I'm sorry imaginary me had a different opinion than real me. 😁

ahankinson:
I just proposed a PR to alter the code types. I hope it doesn't come across as an "aggressive" move -- I meant it as a form of "working, not just talking".

rettinghaus:
Instead of calling it "diatonicMelodicContour" why not use the name "[humdrumMint](https://www.humdrum.org/rep/mint/index.html)" for `-P4 -m2 +M3` which would go nicely along "humdrumKern"?

ahankinson:
Looking here: 

https://www.humdrum.org/Humdrum/commands/mint.html

It looks like `mint` produces output like this:

```
**mint
=1
[c]
+P5
+m3
.
-d5
=2
+m2
8r
[C]
-P5
=3
*-
```

Which means that a format of `humdrumMint` would need more than just the intervals, right? And it probably should be formatted vertically? That sounds like a bit more than is necessary or desired?

rettinghaus:
Maybe @craigsapp could clarify?

ahankinson:
I mean, the documentation seems pretty clear to me already? 

Also, isn't `**mint` just another spine type, already covered by the `humdrumKern` incipit type? I don't remember any restrictions being put on the specific spine types that can be included in the incipit codes.

musicEnfanthen:
Any action needed for this to continue?

bwbohl (approved):

musicEnfanthen (approved):

PR #1652, opened 2025-03-13
schema: proposed alteration to incipit code types

ahankinson:
This is a proposal for altering the acceptable incipit code types in #1651

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -1557,15 +1557,21 @@
           </valItem>
           <valItem ident="parsons">
             <desc xml:lang="en">Parsons code containing intervallic direction only 
-              (e.g., "sususddsdsdsd").</desc>
+              (e.g., "sususddsdsdsd"). Permitted values are composed of a series of tokens, which may 
+              be "u" (Up), "d" (Down) or "s" (Same). Each token should not be separated by a space.</desc>
           </valItem>
-          <valItem ident="diatonicMelodicContour">
-            <desc xml:lang="en">Representation of intervallic direction and diatonic quality
-              (e.g., "P0 +P5 +P0 +M2 +P0 -M2").</desc>
+          <valItem ident="intervals">
+            <desc xml:lang="en">Representation of interval quality and direction 
+              (e.g., "P1 +P5 P1 +M2 P1 -M2"). Values must be composed of a series of space-separated tokens, where each token
+              is comprised of three components: a direction indicator ("+" for "up" and "-" for down); an interval quality, being one of 
+              "P" (Perfect), "M" (Major), "m" (Minor), "A" (Augmented), or "d" (Diminished); and finally an integer indicating 
+              interval size, where "1" is a Unison. (For a Perfect Unison, P1, the direction indicator may be omitted.). 
+            </desc>
           </valItem>
-          <valItem ident="steppedMelodicContour">
-            <desc xml:lang="en">Representation of intervallic direction and size in half steps
-              (e.g., "0 +7 +0 +2 +0 -2").</desc>
+          <valItem ident="semitones">
+            <desc xml:lang="en">Representation of intervallic direction and size in half steps (semitones)
+              (e.g., "0 +7 +0 +2 +0 -2"). Values must be composed of a series of space-separated positive or negative 
+              integers indicating the number of semitones to ascend or descend, respectively.</desc>
           </valItem>
         </valList>
       </attDef>
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -1556,12 +1556,12 @@
             <desc xml:lang="en">**kern representation of the Humdrum format.</desc>
           </valItem>
           <valItem ident="parsons">
-            <desc xml:lang="en">Parsons code containing intervallic direction only 
+            <desc xml:lang="en">Parsons code containing intervallic direction only
               (e.g., "sususddsdsdsd"). Permitted values are composed of a series of tokens, which may 
               be "u" (Up), "d" (Down) or "s" (Same). Tokens should not be separated by spaces.</desc>
           </valItem>
           <valItem ident="intervals">
-            <desc xml:lang="en">Representation of interval quality and direction 
+            <desc xml:lang="en">Representation of interval quality and direction
               (e.g., "P1 +P5 P1 +M2 P1 -M2"). Values must be composed of a series of space-separated tokens, where each token
               is comprised of three components: a direction indicator ("+" for "up" and "-" for down); an interval quality, being one of 
               "P" (Perfect), "M" (Major), "m" (Minor), "A" (Augmented), or "d" (Diminished); and finally an integer indicating 
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -1555,11 +1555,17 @@
           <valItem ident="humdrumKern">
             <desc xml:lang="en">**kern representation of the Humdrum format.</desc>
           </valItem>
-          <valItem ident="melodicContour">
-            <desc xml:lang="en">Representation of 1] intervallic direction and diatonic quality
-              (e.g., "P0 +P5 +P0 +M2 +P0 -M2"), 2] intervallic direction and size in half steps
-              (e.g., "0 +7 +0 +2 +0 -2"), or 3] general intervallic direction only (e.g.,
-              "sususddsdsdsd", often called "Parsons code").</desc>
+          <valItem ident="parsons">
+            <desc xml:lang="en">Parsons code containing intervallic direction only 
+              (e.g., "sususddsdsdsd").</desc>
+          </valItem>
+          <valItem ident="diatonicMelodicContour">
+            <desc xml:lang="en">Representation of intervallic direction and diatonic quality
+              (e.g., "P0 +P5 +P0 +M2 +P0 -M2").</desc>
+          </valItem>
+          <valItem ident="steppedMelodicContour">
+            <desc xml:lang="en">Representation of intervallic direction and size in half steps
+              (e.g., "0 +7 +0 +2 +0 -2").</desc>
           </valItem>
         </valList>
       </attDef>
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -1563,9 +1563,9 @@
           <valItem ident="intervals">
             <desc xml:lang="en">Representation of interval quality and direction
               (e.g., "P1 +P5 P1 +M2 P1 -M2"). Values must be composed of a series of space-separated tokens, where each token
-              is comprised of three components: a direction indicator ("+" for "up" and "-" for down); an interval quality, being one of 
+              is comprised of three components: a direction indicator ("+" for "up" and "-" for "down"); an interval quality, being one of 
               "P" (Perfect), "M" (Major), "m" (Minor), "A" (Augmented), or "d" (Diminished); and finally an integer indicating 
-              interval size, where "1" is a Unison. (For a Perfect Unison, P1, the direction indicator may be omitted.). 
+              interval size, where "1" is a Unison. (For a Perfect Unison, P1, the direction indicator may be omitted). 
             </desc>
           </valItem>
           <valItem ident="semitones">
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -1558,7 +1558,7 @@
           <valItem ident="parsons">
             <desc xml:lang="en">Parsons code containing intervallic direction only 
               (e.g., "sususddsdsdsd"). Permitted values are composed of a series of tokens, which may 
-              be "u" (Up), "d" (Down) or "s" (Same). Each token should not be separated by a space.</desc>
+              be "u" (Up), "d" (Down) or "s" (Same). Tokens should not be separated by spaces.</desc>
           </valItem>
           <valItem ident="intervals">
             <desc xml:lang="en">Representation of interval quality and direction 
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -1570,7 +1570,7 @@
           </valItem>
           <valItem ident="semitones">
             <desc xml:lang="en">Representation of intervallic direction and size in half steps (semitones)
-              (e.g., "0 +7 +0 +2 +0 -2"). Values must be composed of a series of space-separated positive or negative 
+              (e.g., "0 +7 0 +2 0 -2"). Values must be composed of a series of space-separated positive or negative 
               integers indicating the number of semitones to ascend or descend, respectively.</desc>
           </valItem>
         </valList>


bwbohl:
I like the idea of dividing the different possibilities into different values!

pe-ro (changes requested):
I'm a little unsure about the value "steppedMelodicContour".  Can we replace it with "numericMelodicContour"?  I think this term is a better match with "diatonicMelodicContour".  Also, it might be worth considering dropping "Melodic" from both terms, resulting in "diatonicContour" and "numericContour".

bwbohl:
> I'm a little unsure about the value "steppedMelodicContour". Can we replace it with "numericMelodicContour"? I think this term is a better match with "diatonicMelodicContour". Also, it might be worth considering dropping "Melodic" from both terms, resulting in "diatonicContour" and "numericContour".

I can see the problem… step like in diatonic step; what about `halftoneMelodicContour`? (Because both is numeric, at least to some degree ;-)

ahankinson:
What about `intervalDirection` to replace `steppedMelodicContour`? I'm also ok with `diatonicContour`.

bwbohl:
I gave this some more thought:
I  think the name should express the content. This is perfectly true for **parsons**, because it’s known enough to be considered a terminus technicus. As for the other two, I think **contour** is good and melodic can be dropped.

When I read _diatonicContour_ I’d rather think of scale steps instead of (im-)perfect or altered intervals, as the numeric indicator in chord symbols, e.g., G**7** so the above example would be "1 5 5 6 6 5" or "I V V VI VI V" but that’s not what we want or mean. Actually, the _perfect fifth_ (P0) or major second_ (M2) in the example are intervals, so  **intervallicContour** would be a good candidate.

Looking at the steppedContour, for me, the problem is just the "step", because we want to express it in half steps, but the value doesn’t contain "half" and could be mistaken for the double. _numericContour_ is too abstract for me, not connected to the half steps anymore. Now for @ahankinson proposal _ intervalDirection_, for me would rather be "P0 +P5 +P0 +M2 +P0 -M2", but I like the "direction" in it, so sth. like _halftoneDirection_?

All in all, I like the direction most and would propose:

* **parsons**
* **directedInterval(s)**
* **directedHalfstep(s)**

What do you think?

craigsapp:
`P0` should be `P1` for a perfect unison (no zeros in musical mathematics, since it was not widely accepted until the 16th/17th centuries). (Perfect fifth is `P5`, where `P0` is a typo, but in the example data you seem to be using P0 for a unison). 

I will bring your attention to my 2004 ISMIR paper:

https://zenodo.org/records/1417913 (paper)
https://bit.ly/ismir2004-melodic-searching (slides)

Of particular interest is slide 3:

Screenshot 2025-04-09 at 05 36 52

In other words there are many descriptors for pitch contour (7 different ones examined in the paper), as well as for rhythm.  The one called *musical interval* has the regular expression form: `[-+]?([PMm]|A+|d+)?\d+`.  Examples: `2` = a second (either up or down, any quality (major, minor, augmented, diminished).  `+A2` = up an augmented second, `-dd3` = down a doubly diminished third), `P1` = perfect unision (repeated note), `A1` = augmented unison (such as `C` to `C#`), `d8` = diminished octave (such as `C` to `C-flat`).

For enharmonic searching (such as treating `C#` and `D-flat` as the same pitch), twelve-tone intervals are more useful.  In this case `0` equals a repeated note, `1` = either an augmented unison or a minor second, or `1 semitone` (not *halftone*, which is likely the directly translation of the German term, which makes sense, but only after thinking about it :-).

Parsons code is equivalent to "pitch gross contour", and "pitch refined contour" is an extension of this to differentiating between steps/leaps.

So there are four main interval descriptions if pitch contour, in order from most explicit to least explicit:

(1) Intervals based on traditional music theory (which derive from Pythagorean tuning).
(2) Intervals based on twelve-tone music theory (semitones) (derived from equal temperament)
(3) Reducing the intervals to 5 categories of leap up/down, step up/down, and repeated notes.
(4) Reducing further (Parsons code) to 3 categories, up, down, repeated.

The other three categories in the slide are the pitches that are used to calculate the intervals:

(A) Pitch names with octave information (used to calculate no. 1, 3)
(B) MIDI key numbers (used to calculate  no. 2, 4)
(C) Scale degrees which are a variant of (A) but involve an interval to the tonic of the reference key, so it is a hybrid descriptor.

Likewise for the rhythmic features, I created 7 features to test, with "duration" being analogous to pitch, and "metric"/"beat" being somewhat analogous to "interval".

Slide 22:

Screenshot 2025-04-09 at 06 13 50

Illustrates the relative effectiveness of searching for pitch gross contour (Parsons code) and rhythm gross contour compared to twelve-tone interval.   PGC has a small number of bits for each token in the search query (and RGC, not displayed, has even less).  But when using both PGC and RGC together, they match the bits per query token (so using two low-information features is as effective as one higher-information feature).  This may be of use to avoid errors/noise (variants) when searching.

ahankinson:
Thanks Craig! So... what would you suggest we call them?

craigsapp:
> what would you suggest we call them?

Good luck on that 😉 

> I'm a little unsure about the value "steppedMelodicContour". Can we replace it with "numericMelodicContour"? I think this term is a better match with "diatonicMelodicContour". Also, it might be worth considering dropping "Melodic" from both terms, resulting in "diatonicContour" and "numericContour".

I don't like "stepped" as well, but it is not overly objectionable, other than "step" is understandable, and "stepped" is not understandable without thinking about it.   And removing "melodic" seems good since it is pretty much the same as "contour". "Interval" is a more musical term for "contour", but both seem good to me. ("melodic" in contrast to "harmonic" intervals).

> When I read diatonicContour I’d rather think of scale steps

This seems to indicate that contour is a bit ambiguous in a musical context, and can cause confusion between pitch and interval (with scale degrees being more pitchlike, but they have a hidden interval built into them).  "Interval" cannot have the same confusion as "contour", as it is the "slope" between the pitches (or scale degrees when including octave information).

"twelvetone (pitch|interval)" might be a good descriptor for that avoids complications related to "numeric (pitch|interval)".  The standard music-theory names are essentially numeric (numbers for steps), so calling twelvetone intervals numeric is somewhat ambiguous. "Semitones" is a good substitution for twelvetone (numeric) intervals, and I prefer more than "halftones".

For musical intervals, which are based on `@pname`/`@accid*`/`@oct`, maybe just call them "pitch intervals".   This would contrast with twelve-tone intervals, maybe using the term "semitone intervals", or "twelve-tone intervals" (but that is a bit long and has a hypen to figure out how to condense).

Using the word "direction" in the descriptions is confusing: all intervals posses direction and it seems to be a proxy for differentiation between scale degrees and "pitch intervals" (and the two are not equivalent since one is absolute (with a relative component) and the other is relative.  In mathematical terms, an "interval" is the slope of a function and a function is equivalent to the pitch feature (with scale degrees + octave information being more like pitch but have a constant interval reference to the tonic).

----------------------

One possibility is to have both a `form` and an `authority`.  For example `parsons` is a "pitch gross contour" in Humdrum, which is a more generic term.  But for any feature there has to be a technical description of it for implementing searches against it.  Not that this distinction is necessary in this situation.

For example parsons uses "udr" for the three states in pitch gross contour, while Themefinder uses "uds" for up/down/same (reserving "r" for potential inclusion of rests).   This could also be described with "<>=".  In Themefinder "UuDds" are used for "refined pitch contour" with the upper/lower cases distinguishing between steps and leaps.

If parsons is to be kept as the feature form, then as @rettinghaus mentioned in the other thread `humdrumMint` (or perhaps better for brevity `mint`) would probably be a good `form` for things like `+P5` since it is also a formalized description of intervals suitalbe for searching.

So I would propose:

(1) `@form="parsons"` = gross pitch contour/interval and already exists as a form.
(2) `@form="mint"` = musical intervals using the same syntax as the `mint` tool in Humdrum where the regular expression for each item is: `[-+]?([PMm]|A+|d+)?\d+`.  If intervals always include the direction, then spaces betwen intervals would be optional like in "parsons" (but without spaces they would be harder to read).
(3) `@form="semitones"` could be the numeric interval form (that is derived from the numeric pitch as in MIDI key numbers).

An important feature to be able to search would be scale degrees, which is what I would use in tonal context (but specifics of scale degrees get complicated such as whether or not to include octave information, accidental information (altered scale degrees), and some complications related to mode.

For further extension to all 7 features that I used for analysis in the above paper, the "pitch name" would be a useful form to have available: If you have an anonymous manuscript, then searching by pitches in RISM is most efficient, and if nothing found you would switch to "mint" and then continue to "parsons" (that is what I would typically do in Themefinder, starting with the most specific feature and then lowering the resolution if I do not have any matches, although the reverse could be done to limit match counts if too numerous: starting with parsons, and then going to refined contour of steps/leaps, then scale degrees than pitches).

Also related to this discussion, `verovio` could be used to automate creation and management of such incipit descriptors (I presume you guys have been adding parsons manually?).   If there is more than "parsons", then that will cause problems when searching across different datasets of MEI files, since one might have "parsons" and another might have "mint", for example.   For small databases, parsons is good, but for large databases (RISM incipits), "mint" would be more useful to minimize unwanted matches, or "degree" is what I typically prefer.   Adding all features is interesting, although that is a bit verbose.   When implementing a search interface on the web for MEI files, you would not literally search the incipit features within the files, but would extract the features only into another file for searching (as is done in Themefinder), or have some dedicated server that loads the MEI files and extracts the incipit features (but this could be automated with an interface to verovio to extract features).  With a full score it is also useful to be able to search the entire score for melodic patterns, which would be outside of the incipit features and be too verbose to practically encode within the score.

In https://polishscores.org, I only have pitch search available on the homepage, but sometime I will add the "mint" search to find transposed matches.  And having gross pitch contour, "parsons", would be good.  For the pitch searches, they are actually done within the web browser, from a pre-calculated index:

https://humdrum.nifc.pl/popc2-pitch-index.txt

And I would create other indices for searching other types of pitch features.

For monophonic music/incipits extracting them automatically is easy.  For piano music it can be difficult (there may be a boring introduction before the opening melody, for example, or the melody may be in the bass staff).  It might be useful to add incipit start/stop markers in the MEI score to assist in automatic extraction of incipits so that they do not need to be created by hand.

Also note: starting from "mint", then "semitones" and "parsons" features can be created automatically (but not in the other direction).   

Related to comment https://github.com/music-encoding/music-encoding/pull/1651#issuecomment-2720874567, the Humdrum file structure would be removed to create a one-dimensional string as in @bwbohl's example `"P1 +P5 +P1 +M2 +P1 -M2`, since those are the actual `mint` intervals (and other things are not necessary, such as `=` for barlines, and `[c]` for indicating the starting note/octave).

pe-ro:
"parsons", "mint", and "semitones" sound good to me.

ahankinson:
I'm not so keen on "mint" because of the expectation that it would need to support *all* of the mint code, which is effectively also **kern, and that's already covered. (that is, If you really want it, you can use the "humdrumKern" option that already exists.)

pe-ro:
@ahankinson, when the "humdrumKern" value is used, the expectation is that there will be a complete encoding of the incipit.  "mint" will use the same values as a Humdrum `**mint` spine, but will only encode the melody.

pe-ro:
Schematron rules for the melodic encodings can be based on the existing mei_data.INTERVAL.MELODIC datatype.

ahankinson:
`**mint` seems overkill for what is needed, though. From the [docs](https://github.com/humdrum-tools/humdrum-documentation/blob/master/guide/ch11/index.md)

```
**mint
*M4/4
=1-
.
+m2 -P5
-m2 +M2
+P4 +M2
-M3 r
=2
+P4
-m3 (+M2) (-P5) (+A4) (-m3) +m3
-M2 (-m3) (-P5) (P1) (-M3) (+P4) (+m2) -m2
*-
```

```
**mint	**mint
*M4/4	*M4/4
=1-	=1-
.	.
+m2	.
-m2	-P4
.	+M2
+m2	r
=2	=2
P1	.
-m3 +M2	-m3 +m3
-M2 (+M2) (-m3) P1	+m2 -m2
*-	*-
```

As @craigsapp notes, you can also generate this from an existing `**kern` representation with the same tools you would use to make sense of the kern encoding. 

It seems like a big step to go from trying to find a name for a fairly small and contained thing (interval quality, direction, amount) to needing to bring in support for the full `**mint` format.

Can't we just call "P1 +P5 +P0 +M2 +P0 -M2" "qualified" or "quality" and leave it at that?

`parsons`, `quality`, `semitones`?

pe-ro:
I'd say that the intervallic encoding is "mint-like", so I don't have a problem calling it "mint" even though it's not exactly Humdrum mint.  Along those lines, we already have a `@intm` attribute (that name's just an anagram of 'mint') that allows the same values.

Still, I'd be ok with using "intervals" instead of "mint".  I think this pairs well with "semitones".

craigsapp:
> Still, I'd be ok with using "intervals" instead of "mint". I think this pairs well with "semitones".

That sound good to me.   "humdrumMint" is over-specific, because the "mint" description of intervals is fairly standard in music theory, so there is nothing particularly Humdrum-specific about it:

https://ycmusictheory.commons.gc.cuny.edu/mus-110-fundamentals-of-music-theory/standard-4-intervals

Calling it "quality" is problematic since the "A", "d", "M", "m", "P" are the quality component of the intervals, which doesn't include the interval direction part or the interval size.

ahankinson:
`parsons`, `intervals`, `semitones` sounds good to me.

I like the _idea_ of aligning with `@intm`, but I'm not sure how good a fit it is? `parsons` would come "out-of-the-box" with `u/d/s` but can't support `sd` or `su` since it's ambiguous -- is `sus` "same-up-same", or "same or higher-same"? I don't think parsons supports spaces between tokens?

Also not sure if we want to allow intervals specified with no direction or amount, which `@intm` allows. "M2" -- up or down? `+M`?

pe-ro:
@ahankinson, I didn't mean that the values allowed in `@intm` would be used literally in `<incipCode>`, only that they could provide a starting point.  For example, we can create new datatypes that go along with the `@form` values of "parsons", "intervals", and "semitones", and allow those to be alternatives in the content declaration of `<incipCode>`.  Alternatively, we can leave the current content declaration of `<incipCode>`; that is, "text", in place but create Schematron rules that restrict that text when `@form` matches "parsons", "intervals", or "semitones".

ahankinson:
Regexes for each of the three types would be fairly easy to write, so I like the idea of having a form of validation for them. I just updated this PR to change the form values, as well as expand the description to specify the expected content.

pe-ro:
Looks good to me.

craigsapp:
> but can't support sd or su since it's ambiguous -- is `sus` "same-up-same", or "same or higher-same"? I don't think parsons supports spaces between tokens?

* `s` = same as previous pitch (or `r` for repeated note in original specification).
* `u` = higher than previous pitch
* `d` = lower than previous pitch

`sd` is valid: it means that the pattern starts with a repeated note, and then goes down, such as `C4 (s) C4 (d) B3`

`sus` is always relative to the previous token, so this represents two repeated notes in an upward pattern, such as `C4 (s) C4 (u) F4 (s) F4`

craigsapp:
> Also not sure if we want to allow intervals specified with no direction or amount, which `@intm` allows. "M2" -- up or down? `+M`?

You should always want to encode the complete feature (direction/quality/size), but when searching such features, you can make them selective as illustrated below for example on the JRP website repertory search interface:
     https://josquin.ccarh.org/search

Searching for the interval pattern `2 3 2 3 2 3 2 3 2` which means an alteration of seconds and thirds in any direction with any quality (quality is somewhat ambiguous in Renaissance era music due to performers avoiding melodic tritones and creating leading tones at cadences):

Screenshot 2025-04-15 at 03 38 15

First match in the list:

Screenshot 2025-04-15 at 03 41 47

You can also search for `232323232`, and I adjust the query to add spaces to show the segmentation of the non-spaced query.  Here is searching for a 12th, which occurs 3 times in two pieces:

Screenshot 2025-04-15 at 03 48 14

And here is searching for a unison (repeated note) followed by a step (up or down): 

Screenshot 2025-04-15 at 03 49 19

Five major intervals in a row:

Screenshot 2025-04-15 at 03 50 12

Example (All major seconds):

Screenshot 2025-04-15 at 04 08 44

Searching for a diminished interval followed by an augmented interval:

Screenshot 2025-04-15 at 03 52 18

only one, probably a encoding error:

Screenshot 2025-04-15 at 03 52 09

Using interval direction as a proxy for Parson's code.

Screenshot 2025-04-15 at 04 03 38

Screenshot 2025-04-15 at 03 58 54

(P1 would be used as the token for "same").

ahankinson:
> `sd` is valid: it means that the pattern starts with a repeated note, and then goes down, such as C4 (s) C4 (d) B3

@craigsapp take a look at the [docs for `data.INTERVAL.MELODIC`](https://music-encoding.org/guidelines/v5/data-types/data.INTERVAL.MELODIC.html), which is what I was responding to:

Interval direction only: 
 - u = up/higher, 
 - d = down/lower, 
 - s = same, 
 - n = neutral/unknown, 
 - sd = same or lower (but not higher), 
 - su = same or higher (but not lower)

The last two values is what I was referring to. Parsons code does not have this, and in a stream of `sdsdsdsdsdsdsdsusus` you can't tell the difference between the "same or higher / lower" and the other options.

ahankinson:
> but when searching such features, you can make them selective as illustrated below for example on the JRP website repertory search interface

Sure; However, I think in the context of this issue, though, we're concerned with capturing the contour of the incipit, not defining a query format.

ahankinson:
To move this PR forward, now that (I think?) we have agreement, someone else should merge it. The target is *not* the develop branch, but the initial PR proposed by @pe-ro -- thus I don't think it needs the two approvals (it will need that when it goes into `develop`). 

Perhaps @musicEnfanthen or @bwbohl ?

craigsapp:
> https://music-encoding.org/guidelines/v5/data-types/data.INTERVAL.MELODIC.html

A ha.  That is a mixture of "parsons" (extended), "intervals" and "semitones" interval types:

```xml
<content>
   <rng:choice>
      <rng:data type="token">
         <rng:param name="pattern">u|d|s|n|sd|su</rng:param>
      </rng:data>
      <rng:data type="token">
         <rng:param name="pattern">(\+|\-)?([AdMmP])?[0-9]+</rng:param>
      </rng:data>
      <rng:data type="token">
         <rng:param name="pattern">(\+|\-)?\d+(\.\d+)?hs</rng:param>
      </rng:data>
   </rng:choice>
</content>
```

The pattern `u|d|s|n|sd|su` is Parson's code but adding `n`, `sd` and `su`, so extended Parson's code.  Why were these added @pe-ro?

`sd` and `su` are more for querying rather than for describing&mdash;in querying `sd` can be expressed as `(s|d)` when searching normal parsons.  `n` purpose is unknown to me&mdash;used to encode rests in melodic sequences I am presuming or intervals between unpitched/semipitched notes? Since these are all "tokens", that would imply that they need space separations when stored in an incipit string.

The interval pattern `(\+|\-)?([AdMmP])?[0-9]+` allows for `0`, which could represent a rest?   Otherwise `[1-9]\d*` is better, since this interval representation does not otherwise have a `0` in its system.  `AA` can represent doubly augmented and `dd` can represent doubly diminished. so an improvement for the quality could be: `(A+|d+|M|m|P)?` instead of `([AdMmP])?`. The characters allowed are contextual to the following number, but that is not easy to control with regular expressions (for example `M4` is not allowed). 

I am wondering about the validity of using `\-` in the regular expression `(\+|\-)`, since `-` is not a meta character unless it is inside of `[]`.   The extended regular expression `[+-]?` is a more concise form for `(\+|\-)?`, or more correctly: `(\+|-)?`.

So a better regular expression description of "intervals" is `[+-]?(A+|d+|M|m|P)?[1-9]\d*` (still not perfect since it does not check for valid combinations of quality and interval size).


> Sure; However, I think in the context of this issue, though, we're concerned with capturing the contour of the incipit, not defining a query format.

Yes, I understand that.  You need to create a feature description that can predictably be searched, so encoding some incipits using "intervals" with and without quality information would generate unpredictable results.

So when encoded as an incipt, if "intervals" format is used, the full direction/quality/number should be given for the incipit, and a query can be adjusted, assuming all components of the "intervals" form are present.   For example querying `M M M M M` for 5 major intervals in a row would be expanded to the regular expression: `([-+]M\d+) ([-+]M\d+) ([-+]M\d+) ([-+]M\d+) ([-+]M\d+)`.  If the "intervals" description does not include the quality then you could not search the quality of the interval.

pe-ro:
@craigsapp, 

- as I recall, the so-called "extended" Parsons values were added at your recommendation. 😊  Removing them is fine with me.  I doubt anyone has used them.
- the '0' in the pattern `(\+|\-)?([AdMmP])?[0-9]+` allows one to capture intervals beyond a 9th, for example "M10"; however,
- your suggestions for improvement of the regEx patterns are fine with me.

pe-ro:
I would also suggest that we remove the trailing 'hs' from the pattern `(\+|\-)?\d+(\.\d+)?hs`, at least in the incipit context.

craigsapp:
> I would also suggest that we remove the trailing 'hs' from the pattern `(\+|\-)?\d+(\.\d+)?hs`, at least in the incipit context.

I was also thinking about that, but the problem is that the "interval" data type can be numbers only, so there would be an ambiguity.   For example,  "8" could mean an octave in the "interval" data type, or 8 "semitones" (= minor 6th, or "6" without direction or quality).

One possibility is to have separate attributes dedicated for each data type such as `@intm="+P5"`, `@intm.parsons="u"` and `@intm.semi="+7"`.   And similar for incipits.

> the '0' in the pattern `(\+|\-)?([AdMmP])?[0-9]+` allows one to capture intervals beyond a 9th, for example "M10".

Yes, going beyond a 9th is important.  The enhancement avoids allowing for `0` by itself, which does not exist (unless used to represent something like a rest).  Intervals above a 9th are still allowed by the pattern `[1-9]\d*` (or `[1-9][0-9]*`).

> as I recall, the so-called "extended" Parsons values were added at your recommendation. 😊

😜 That is too long ago for me to remember...  I think getting rid of the multi-character parsons tokens would be good.  This would allow the incipit values to make spaces optional.

musicEnfanthen:
@ahankinson Happy to merge this PR, but there seems to be some ongoing discussion? Not sure if it is related to the very changes of this PR, though? How do you want to proceed?

source/modules/MEI.header.xml
```diff
@@ -1555,11 +1555,23 @@
...
+            <desc xml:lang="en">Representation of interval quality and direction 
+              (e.g., "P1 +P5 P1 +M2 P1 -M2"). Values must be composed of a series of space-separated tokens, where each token
+              is comprised of three components: a direction indicator ("+" for "up" and "-" for down); an interval quality, being one of 
+              "P" (Perfect), "M" (Major), "m" (Minor), "A" (Augmented), or "d" (Diminished); and finally an integer indicating 
+              interval size, where "1" is a Unison. (For a Perfect Unison, P1, the direction indicator may be omitted.). 
+            </desc>
+          </valItem>
+          <valItem ident="semitones">
+            <desc xml:lang="en">Representation of intervallic direction and size in half steps (semitones)
+              (e.g., "0 +7 +0 +2 +0 -2"). Values must be composed of a series of space-separated positive or negative
```
musicEnfanthen: @ahankinson Can the direction be omitted for unisons (value 0) with semitones, too? Otherwise, the first `0` needs a `+` sign like the following `0`s?

ahankinson:
Yeah I need to find a bit of time to change it. Please don’t merge yet

lpugin:
> One possibility is to have separate attributes dedicated for each data type.

I think that would be a good idea. I makes typing and processing much easier. I also like the proposed naming with `@intm.*`

source/modules/MEI.header.xml
craigsapp: I would suggest that `0` semitones not include a sign; otherwise searching via regular expressions could become complicated if not careful.   When searching semitones, `[+-]?0` could handle any of the equivalent `0 -0 +0` if signs are allowed on `0`.

Likewise for P1 is should not have a direction since it is going neither up nor down.   Exceptions would be related to a non-perfect quality: `+A1` for an augmented unison (`C` to `C#`, for example), or `-d1` for a diminished  unison (`C` to `C-flat`).

Note for full generalization of regular intervals, `d` and `A` can repeat as often as needed (`m`, `M`, and `P` can only occur once).  Example `+A5` = `C` &rarr; `G#`, `+AA5` = `C` &rarr; `G##`, `+AAA5` = `C` &rarr; `G###`.  Where `AA` is "doubly augmented" and `ddd` is "triply diminished".

ahankinson: Do you think needing to specify a "triply diminished" interval when _notating an incipit_ is useful? I would suggest using the `humdrumKern` or `pae` formats if that's the level of analysis needed.

craigsapp: It is theoretically possible.  "triply" probably does not exist in realistic music, but there are a few "doubly" augmented/diminished intervals in real music.

pe-ro:
` @intm.*` makes sense in the `<note>` context, but it doesn't make sense for `<incipit>`.  This discussion is about incipits.  Let's leave the modification of `@intm` to another issue please.

craigsapp:
>  `@intm.*` makes sense in the `<note>` context, but it doesn't make sense for `<incipit>`. This discussion is about incipits. Let's leave the modification of `@intm` to another issue please.

Not much documentation for incipits yet:
https://music-encoding.org/guidelines/v5/content/metadata.html#headerWorkIncipit

In the `<incipit>` context `@intm.*` relates to the `incipitCode@form`, which I am presuming would be encoded like this:

```
<incipCode form="parsons">sududuuudd</incipCode>
<incipCode form="intervals">P1 +M2 -M2 +M2 -M2 +M2 +m2 +M3 -M2 -m2</incipCode>
<incipCode form="semitones">0 +2 -2 +2 -2 +2 +1 +3 -2 -1</inciptCode>
```

And presumably more than one `incipCode` would be allowed, and people could search in whichever `@form` they prefer. 

What would the default `@form` be?   Allowing a mixture of forms at the same time for the intervals in `incipCode` would not be very useful.

pe-ro:
> What would the default `@form` be?

There are no defaults per se in MEI, so there is no default value for `@form`. However, either `@form` or `@mimetype` is required by a Schematron rule.  As far as I know, no mimetypes are available for music notation formats at present, so users will likely choose `@form`.  When they do, the enumerated values for `@form` and the associated regexes come into play.

pe-ro:
> presumably more than one incipCode would be allowed, and people could search in whichever `@form` they prefer.

Yes, zero or more `<incipCode>` elements are already allowed in `<incip>`.

bwbohl:
Is this ready to go into the develop-incipCode-form branch?

I like the current state. Would it be possible to add some more verbosity to the description of `plaineAndEasie` and `humdrumKern` in the same run?

musicEnfanthen:
pining @ahankinson @pe-ro

pe-ro:
The label "parsons" is ok since that's how many people refer to "melodic contour", but to be more precise I'd change 
```
<valItem ident="parsons">
  <desc xml:lang="en">Parsons code containing intervallic direction only 
    (e.g., "sususddsdsdsd"). Permitted values are composed of a series of tokens, which may 
   be "u" (Up), "d" (Down) or "s" (Same). Each token should not be separated by a space.</desc>
</valItem>
```
 to 
```
<valItem ident="contour">
  <desc xml:lang="en">Representation containing intervallic direction only 
    (e.g., "sususddsdsdsd"). Permitted values are composed of a series of tokens, which may 
   be "u" (Up), "d" (Down) or "s" (Same). Each token should not be separated by a space. 
    Sometimes referred to as "Parsons" code.</desc>
</valItem>
```
Other than that, I think it's good to go.

pe-ro:
> Would it be possible to add some more verbosity to the description of plaineAndEasie and humdrumKern in the same run?

The descriptions in the schema should be short aides-de-memoire.  Longer explanations should go in the text of the Guidelines.

bwbohl:
> > Would it be possible to add some more verbosity to the description of plaineAndEasie and humdrumKern in the same run?
> 
> The descriptions in the schema should be short aides-de-memoire. Longer explanations should go in the text of the Guidelines.

Compared to the long explanations of the others, I think we should at least add a short example and maybe provide a link to some official documentation.

Also, I think the element description in line 1526 should be adjusted to list more than Plane &amp; Easie

And Guidelines is a good catch!
The examples provided in the value descriptions should also be provided in the Guidelines. Maybe it would be a good idea to have one music notation example and its representation in all of the alternative formats?

source/modules/MEI.header.xml
bwbohl:
```suggestion
              (e.g., "0 +7 0 +2 0 -2"). Values must be composed of a series of space-separated positive or negative 
```

bwbohl (approved):

bwbohl (approved):

Issue #483, opened 2017-11-16
Revise encoding of ossia


pe-ro:
The resolution of issue #482 is connected to the encoding of ossia -- those that are displayed on additional staves and those that are rendered in the same layer as "original" material.  Also, I think the current method of handling ossia is not explicit enough; that is, its processing requires differentiating between staves/layers that carry `@n` and those that don't.  It would be better to use a different element for these staves/layers, which would make for a cleaner resolution of #482.  

Ossia material displayed on a separate staff could be encoded as --
```
<measure>
  <staff n="1">
    <!-- first staff, without ossia -->
  </staff>
  <ossia>
    <oStaff>
      <!-- alternative content on reduced-size staff -->
    </oStaff>
    <staff n="2">
      <!-- original content on regular staff -->
    </staff>
  </ossia>
  <staff n="3">
    <!-- third staff, without ossia -->
  </staff>
</measure>
```
while an in-measure ossia could be encoded as
```
<measure>
  <staff n="1">
    <!-- content omitted -->
  </staff>
  <staff n="2">
    <ossia>
      <layer n="1">
        <!-- original content in regular layer -->
      </layer>
      <oLayer>
        <!-- alternative content in separate layer -->
      </oLayer>
    </ossia>
  </staff>
  <staff n="3">
    <!-- content omitted -->
  </staff>
</measure>
```
So-called "regular" staves would continue to carry the `@n` attribute to link them to the appropriate staff definitions.  The `<oStaff>` and `<oLayer>` elements would not use `@n`, instead taking their key, clef, etc. from the `<staff>` or `<layer>` element with which they're paired.  The children of `<ossia>` can be in any order; therefore, `<oStaff>` and `<oLayer>` can have an `@place` attribute ("above"/"below") to explicitly say where they should be rendered with respect to the "regular" staff.  The encoding order of multiple `<oStaff>` and `<oLayer>` elements can be used to render them with respect to each other or another attribute (`@order` perhaps or even `@n` now that these have a different name than `<staff>`) can be added to indicate the order of their appearance.

Since this change isn't backwards compatible, we should make it in version 4.0.0 -- If you're going to break things, do it sooner rather than later.  :-)

pe-ro:
Here's the ODD --
```
<!-- Add oStaff element with same content model as staff -->
<elementSpec ident="oStaff" module="MEI.shared" mode="add">
  <classes>
    <memberOf key="att.basic"/>
    <memberOf key="att.facsimile"/>
[...]

PR #1654, opened 2025-03-17
guidelines: fix ossia section

rettinghaus:
After introducing `oStaff` and `oLayer` elements (#483) nobody took care of adjusting the relevant section in the Guidelines. 
This PR updates the given examples and the explanation.


pe-ro (approved):
The new language is fine with me.  The larger question is whether we need `<ossia>` at all.  I think we should think about whether `<choice>` is a reasonable replacement for `<ossia>`.

craigsapp:
There are also cue staves which I came across for a discussion in the verovio issues:

Screenshot 2025-03-19 at 07 34 33

https://s9.imslp.org/files/imglnks/usimg/3/3a/IMSLP53008-PMLP27823-Mendelssohn-Op070.Bassoon.pdf#page=5

These would function like ossia staves, but cannot be considered a `<choice>`.

pe-ro:
They may rendered in a similar fashion to ossia staves, but they're definitely not ossia, so I don't think one should use either `<choice>` or `<ossia>` for these staves.

craigsapp:
How would it currently be done in MEI? 😜 In verovio it would be handled exacty as an ossa staff regardless of its categorization/function.  Probably as `cueStaff`?  Then it could programmatically be removed if creating a score from parts containing cues.

craigsapp:
MuseData printing system can handle this case:

Screenshot 2025-03-19 at 08 41 41

http://scores.ccarh.org/mendelssohn/elijah/parts/current/mendelssohn-elijah-bsn1.pdf
https://wiki.ccarh.org/wiki/Elijah#Woodwinds

But I would have to review the source to see how it is done.  There is not an encoded full score, so for example when creating a full score from the already generated parts, the cue notes would have to be removed, although in MuseData, there would typically be a more data-oriented part file generated first, and then the cues are added to create part-specific files (and the data-oriented parts would also be used in a separate process to generate the score).

bwbohl:
@rettinghaus thanks for the fix! If there’s need for further discussion of `ossia` vs. `cueStaff` vs `choice` please feel free to open an issue ;-)

bwbohl (approved):

PR #1655, opened 2025-03-17
schema: allow only one staff within measure ossias

rettinghaus:
If I understand the concept correctly there should only be one "normal" staff allowed within an ossia within measure. @pe-ro ?


pe-ro:
No, multiple staves must be allowed to handle multi-staff "original" notation accompanied by multi-staff "ossia" notation, for example in notation for piano.

rettinghaus:
True, didn't think of that.

PR #1660, opened 2025-03-31
ci: fix regex again

musicEnfanthen:
As a follow-up to #1659, this PR moves back to the original regex pattern but with line delimiter "$".  As it turns out, missing the delimiter and/or having a severity option in the pattern object, seems to result in the warnings not showing up in the action output.

It also renames the matcher file for better clarity.


rettinghaus (commented):
Could you please explain where the benefit is? The output looks the same to me.

bwbohl (approved):

musicEnfanthen:
> Could you please explain where the benefit is? The output looks the same to me.

Yes, hopefully it is the same ;) New (old) regex is more specific regarding the structure of the Schematron log messages. Makes it a bit easier for our future (s)elves to get what it does.

PR #1665, opened 2025-04-24
schema: fix pureODD prechecks

bwbohl:
This PR fixes the issues raised by the pureOdd-precheck.

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -2902,20 +2902,18 @@
       <memberOf key="model.physDescPart"/>
     </classes>
     <content>
-      <rng:choice>
-        <rng:group>
-          <rng:zeroOrMore>
-            <rng:ref name="head"/>
-          </rng:zeroOrMore>
-          <rng:zeroOrMore>
-            <rng:choice>
-              <rng:ref name="watermarkList"/>
-              <rng:ref name="watermark"/>
-              <rng:ref name="p"/>
-            </rng:choice>
-          </rng:zeroOrMore>
-        </rng:group>
-      </rng:choice>
+      <rng:group>
+        <rng:zeroOrMore>
+          <rng:ref name="head"/>
+        </rng:zeroOrMore>
+        <rng:zeroOrMore>
+          <rng:choice>
+            <rng:ref name="watermarkList"/>
+            <rng:ref name="watermark"/>
+            <rng:ref name="p"/>
+          </rng:choice>
+        </rng:zeroOrMore>
+      </rng:group>
     </content>
   </elementSpec>
   <elementSpec ident="watermarkList" module="MEI.header">
--- a/source/modules/MEI.genetic.xml
+++ b/source/modules/MEI.genetic.xml
@@ -44,6 +44,7 @@
           ad-hoc modification; that is, one which does not interrupt the writing process.</desc>
         <datatype>
           <rng:ref name="data.BOOLEAN.unknown"/>
+        </datatype>
       </attDef>
       <attDef ident="state" usage="opt">
         <desc xml:lang="en">Points to the genetic state that results from this modification.</desc>
--- a/source/modules/MEI.genetic.xml
+++ b/source/modules/MEI.genetic.xml
@@ -43,11 +43,7 @@
         <desc xml:lang="en">The @instant attribute is syntactic sugar for classifying a scribal intervention as an
           ad-hoc modification; that is, one which does not interrupt the writing process.</desc>
         <datatype>
-          <rng:choice>
-            <rng:ref name="data.BOOLEAN"/>
-            <rng:value>unknown</rng:value>
-          </rng:choice>
-        </datatype>
+          <rng:ref name="data.BOOLEAN.unknown"/>
       </attDef>
       <attDef ident="state" usage="opt">
         <desc xml:lang="en">Points to the genetic state that results from this modification.</desc>
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -548,6 +548,22 @@
       </valList>
     </content>
   </macroSpec>
+  <macroSpec ident="data.BOOLEAN.unknown" module="MEI" type="dt">
+    <desc xml:lang="en">Boolean attribute values and "unknown".</desc>
+    <content>
+      <valList type="closed">
+        <valItem ident="true">
+          <desc xml:lang="en">True.</desc>
+        </valItem>
+        <valItem ident="false">
+          <desc xml:lang="en">False.</desc>
+        </valItem>
+        <valItem ident="unknown">
+          <desc xml:lang="en">Unknown.</desc>
+        </valItem>
+      </valList>
+    </content>
+  </macroSpec>
   <macroSpec ident="data.CANCELACCID" module="MEI" type="dt">
     <desc xml:lang="en">Indicates where cancellation accidentals are shown in a key signature.</desc>
     <content>


source/modules/MEI.header.xml
```diff
@@ -2889,20 +2889,18 @@
...
+        <rng:zeroOrMore>
+          <rng:choice>
+            <rng:ref name="watermarkList"/>
+            <rng:ref name="watermark"/>
+            <rng:ref name="p"/>
+          </rng:choice>
+        </rng:zeroOrMore>
+      </rng:group>
     </content>
   </elementSpec>
```
musicEnfanthen, line 2918: @riedde @gucl-mu Can you have a look please if that still meets your needs?

gucl-mu: Looks fine to me. I was just wondering about the `p`, I wouldn't need it on my side, but it's probably a best practice to have it there, right?

bwbohl: We could try to go without it, it was there before my proposed changes, though ;-)

pe-ro:
I'm not sure what the issue is with ODD purification in this case, but I'd like to take this opportunity to say that I'm currently working on a proposal to change the content model of watermarkDesc.  In RNG-speak, I'm thinking of the following --
```
<define name="mei_watermarkDesc">
  <element name="watermarkDesc">
    <a:documentation xmlns:a="http://relaxng.org/ns/compatibility/annotations/1.0">(watermark
      description) Contains a description of the watermark(s) found in the base material of a
      resource.</a:documentation>
    <choice>
      <zeroOrMore>
        <group>
          <optional>
            <ref name="mei_label"/>
          </optional>
          <ref name="mei_watermark"/>
          <zeroOrMore>
            <ref name="mei_annot"/>
          </zeroOrMore>
        </group>
      </zeroOrMore>
      <zeroOrMore>
        <choice>
          <text/>
          <ref name="mei_model.paracontentPart.restricted"/>
          <ref name="mei_model.msInline"/>
          <ref name="mei_condition"/>
        </choice>
      </zeroOrMore>
    </choice>
    <ref name="mei_att.bibl.attributes"/>
    <empty/>
  </element>
</define>
```
This model allows `<watermarkDesc>` to contain either 1) a narrative text that describes the watermark(s) found in the resource as a group, e.g.,
```
<watermarkDesc>texty texty text</watermarkDesc>
```
or 2) a structured list of watermark sub-elements with an optional label and optional annotations for each watermark , e.g.,
```
<watermarkDesc>
    <label>watermark 1</label>
    <watermark>description of watermark 1</watermark>
    <annot>note pertaining to watermark 1</annot>
    <label>watermark 2</label>
    <watermark>description of watermark 2</watermark>
    <annot>note pertaining to watermark 2</annot>
  </watermarkDesc>
```
Use of the members of model.paracontentPart.restricted (yet to be added to the schema) -- model.lgLike, model.listLike, model.quoteLike, model.tableLike, and model.textPhraseLike.basic -- without intervening text makes it possible to create database-like descriptions of each watermark, e.g., 
```
<watermark>
  <name role="inventor">Lucas, van Leyden, 1494-1533</name>
  <identifier>DE-HAUMB-VA-3373</identifier>
  <date cert="high">1510</date>
  <dimensions type="watermark" unit="mm">
    <height>287</height>
    <width>452</width>
  </dimensions>
  <term type="school">Dutch</term>
  <term type="style">Northern Renaissance</term>
  <term type="subject">Ecce Homo (Large)</term>
  <term type="genre">religious composition</term>
  <term type="IPH classification">T31 (Mason's or merchant's marks without Hermes cross)</term>
  <ref type="equivalent">DE-HAUMB-VA-3372</ref>
  <name role="paperMaker">Robert Le Bé</name>
  <dimensions type="paper" unit="in">folio (<height>15</height>" x <width>22</width>")</dimensions>
  <geogName type="paperProductionPlace">Troyes (France)</geogName>
</watermark>
```
but, of course, mapping from any particular database to `<watermark>` sub-elements will be up to the individual encoder/project.

gucl-mu:
Thanks for your proposal @pe-ro! It seems to me that your proposal would benefit from addressing several important aspects of watermark description. For example, how can one distinguish between paper mould structure and watermarks? How can research data itself be coded (and not just metadata linked)? How should fragmented watermarks be described? Essential dimensions are also worth considering: although the height and width of a watermark are important, they are not necessarily the most significant factors, especially since measuring individual marks depends heavily on the shape of the marks. What I find particularly important is that the entities watermark, paper mould (structure), and papermaker/metadata should be coded separately from each other to ensure clarity and precision in the data model. Perhaps it would be useful to discuss your suggestions further at the next IG meeting? (By the way, Dennis and I have already presented a two-stage plan for this module in the ODD-meeting. What you see here is stage 1 of it.)

pe-ro:
@gucl-mu, thanks for your comments.

>how can one distinguish between paper mould structure and watermarks?
> What I find particularly important is that the entities watermark, paper mould (structure), and papermaker/metadata should be coded separately from each other to ensure clarity and precision in the data model.

Description of the paper and the paper-making process can be described separately in `<physDesc>/<baseMat>` (base material). I should not have included them in my example.  Mea culpa.

>How can research data itself be coded (and not just metadata linked)?

I'm not sure I understand exactly what you're asking here, but MEI is not the "research markup language".  😄 There's only so far one can go in terms of capturing "secondary" data.  This is why `<annot>` and `<extMeta>` exist.

The example I provided wasn't intended to be exhaustive. For clarity's sake, when `<watermarkDesc>` contains text, it may use these elements -- abbr, address, annot, bibl, biblFull, biblStruct, condition , date, dimensions, expan, extent, fig, identifier, lb, lg, list, locus, locusGrp, name, num, ptr, q, quote, ref, rend, seg, stack, symbol, table, term, and title. 

Similarly, `<watermark>` may contain: abbr, address, and heraldry, annot, bibl, biblFull, biblStruct, bloc, condition, country, date, depth, dim, dimensions, district, expan, extent, fig, geogFeat, geogName, height, identifier, lb, lg, list, locus, locusGrp, name, num, ptr, q, quote, ref, region, relation, rend, seg, settlement, stack, symbol, table, term, title, and width. 

These content models can be merged so that either element's text may contain the same sub-elements. They can also be expanded/contracted now as part of this revision or by user customization later. 

>How should fragmented watermarks be described? Essential dimensions are also worth considering: although the height and width of a watermark are important, they are not necessarily the most significant factors, especially since measuring individual marks depends heavily on the shape of the marks.

Please forgive my ignorance, but I don't know what "fragmented watermarks" are.  Perhaps my comment above about secondary info can be helpful in this case.

Simple dimensions other than height and width can be encoded using `<dim>`, where the `@form` attribute captures the kind of dimension being measured. Some complex measurements, such as calculation of a bounding box for the watermark or its position relative to the margins of the paper, can be handled using `<surface>/<zone>`.

bwbohl:
@pe-ro & @gucl-mu: I didn’t want to start a discussion about the content model of `watermarkDesc` (please continue the discussion in other places). This pull request is about getting the current RNG modelling aligned to an automatic transformation to pureODD, and the question above was to check whether the changes in this branch still produce the same results that previous changes to develop did – just wrapping it in group which should be transparent, requeste @gucl-mu because he’s the watermark guy ;_)

martha-thomae (approved):

riedde (approved):
Concerning `watermarkDesc`: This PR contains some changes in the rng, but the result is the same. I compiled the schema on this branch and the possibilities of encoding using the compiled schema are fine.

PR #1669, opened 2025-05-25
schema: clean up attributes on figured bass

### LINKED ISSUE #1668: Too many attributes on f element
I noticed something strange on encoding of figured bass: 

```xml
<harm staff="4" startid="#noteA" place="above">
   <fb>
      <f staff="1" place="below">7b</f>
      <f startid="#noteB">5</f>
      <f staff="2">#</f>
   </fb>
</harm>
```

The `f` element allows attributes that are already specified on its ancestor `harm`. That doesn't seem to make sense, as every `fb` is understood as (and supposed to be) a single unit.
One could argue, that these would be _nice to have_ on `fb`, if you'd want to encode multiple within a single `harm`, but there they're not available.

I suspect that this implementation was intentional, but I couldn't wrap my head around it.

rettinghaus:
This PR moves the `remarks` on `f@tstamp2` to the respective attribute in the logical domain.

Also it removes the memberships on `att.controlEvent` and `att.placementRelStaff`that are conflicting with the information encoded in its ancestor `harm` element.

Closes #1668

### DIFF
--- a/source/modules/MEI.harmony.xml
+++ b/source/modules/MEI.harmony.xml
@@ -41,7 +41,6 @@
   <classSpec ident="att.f.log" module="MEI.harmony" type="atts">
     <desc xml:lang="en">Logical domain attributes.</desc>
     <classes>
-      <memberOf key="att.controlEvent"/>
       <memberOf key="att.duration.additive"/>
       <memberOf key="att.startEndId"/>
       <memberOf key="att.timestamp2.log"/>
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -544,7 +544,6 @@
       <memberOf key="att.altSym"/>
       <memberOf key="att.color"/>
       <memberOf key="att.extender"/>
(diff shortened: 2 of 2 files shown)


kepper (dismissed):
As the remark is about `@tstamp2`, which is coming through `att.f.log`, I believe this remark should go into MEI.harmony.xml instead, no?

rettinghaus:
Good catch! I confused `att.timestamp2.log` with `att.timestamp2.ges`.

rettinghaus:
Looking at it again, the remark actually is about the `@extender` and should stay in the visual domain.

bwbohl (approved):

musicEnfanthen:
Pinging @kepper

rettinghaus:
@kepper please check again (and approve)

bwbohl:
@kepper

riedde (approved):
Thank you for your work!

bwbohl:
merged with approval of @riedde

PR #1670, opened 2025-05-27
schema: allow mspace in MEI Basic

### LINKED ISSUE #1539: customizations: mSpace missing in MEI Basic
I just noticed that `mSpace` is not allowed in MEI Basic, while `space` is. Is this on purpose? Should one use `space` with a specific duration over `mSpace` here?

rettinghaus:
closes #1539

### DIFF
--- a/customizations/mei-basic.xml
+++ b/customizations/mei-basic.xml
@@ -82,7 +82,7 @@
                 <moduleRef key="MEI"/>
                 <moduleRef key="MEI.analytical" />
                 <moduleRef key="MEI.cmn"
-                    except="attacca beamSpan bend bracketSpan meterSig meterSigGrp mRpt2 mSpace oLayer ossia oStaff tupletSpan"/>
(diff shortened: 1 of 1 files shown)


kepper:
I am not so sure if mSpace is really necessary for mei basic. Obviously, mRest is a no-brainer, but I would like to see many requests from the wider community that this is really what they need for their daily work with mei basic – otherwise, I see a certain risk of feature-creeping all of mei-all into basic over time. So, my question would be – what are the use cases for this, and where is the user base requesting this? If there is such a need (perfectly possible!), I'm happy to support this addition, but until then, I think it's better to keep pretending music notation would be nice and clean…

kepper (changes requested):
can we get more input on the need for this, beyond what is said in #1539?

martha-thomae:
@rettinghaus, anything to answer @kepper's question (https://github.com/music-encoding/music-encoding/pull/1670#pullrequestreview-2871565987)?

We need a justification that `mSpace` is absolutely necessary in MEI-Basic to accept this PR. Thank you!

PR #1658, opened 2025-03-21
build: guidelines for customizations

bwbohl:
* fixes: restructures  ODD compilation to include both guidelines and specs
* modifies: odd2html to support standalone execution and ant based execution
* adds: ant targets for building customization guidelines HTML and PDF

# TODOs:

## make it work

- [x] unify filenames and `schemaSpec/@ident` of customizations to lowercase (except anyStart)
- [x] remove `$compiledOddUri`
- [x] Make it work on Windows in Docker
- [x] Make it work natively on Windows

## refactoring

~~- [ ] Code cleanup and todo comments, e.g. `$isCompiledOdd` vs `$source.is.canonicalized`~~
~~- [ ] unify variable values for input XMLs, `sequence /` vs sequence `tei:TEI`~~
~~- [ ] closer typing of functions and parameters, e.g. `element(section)` instead  of `node()*`~~
~~- [ ] adjust documentation~~[ moved to #1676 ]

## improvements

~~- [ ] Introduce customization specific ant targets~~ [moved to new issue #1675 ]
~~- [ ] Fixes to run on uncompiled customizations~~[ moved to #1676 ]
~~- [ ] build: fix hardcoded prince.input -> e.g. by adding ant-contrib to the project and use `<foreach>` task~~[ moved to #1677  ]
~~- [ ] Fixes to run on mei-source.xml~~[ moved to #1676 ]

Fixes #1483

PR #1674, opened 2025-06-02
build: ensure cross-platform compatibility for index page generation

musicEnfanthen:
This PR updates the index page generation logic to ensure cross-platform compatibility. 

It normalizes the customizationIndexPages path to a semicolon-separated list with forward slashes of only the folder names and their index.html. This way, the full path does not need to be processed by the XSLT (leading to some issues on non-unix systems).

Goes to customizationDocs branch, so #1658

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -575,7 +575,6 @@
     </target>
     
     <target name="generate-guidelines-index-page" depends="init-mei-classpath">
-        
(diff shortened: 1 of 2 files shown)


musicEnfanthen:
@bwbohl Can you test on Mac, please?

bwbohl (approved):
many thanks for the improvements

Issue #1548, opened 2024-11-17
Store application information in MEI Basic


rettinghaus:
In order for MEI Basic to serve as an exchange format, it would be useful to store information about the application that created an MEI file.

Maybe the `application` element could be utilized for that without the necessity to use [appInfo](https://music-encoding.org/guidelines/dev/elements/appInfo.html) and [encodingDesc](https://music-encoding.org/guidelines/dev/elements/encodingDesc.html).

pe-ro:
Hi @rettinghaus, how do you imagine that would work?  I wouldn't recommend making `<application>` available somewhere else in the header as that would break conformance with the "normal" header.

I'm more inclined to add `<encodingDesc>` back to MEI Basic.  All of the children of `<encodingDesc>` are optional, so doing this wouldn't add any extra burden to encoding in Basic.  If all of the children of `<encodingDesc>` aren't desirable, then they can be eliminated in the Basic customization.  I can see that it may be helpful to remove all the children of `<encodingDesc>` save `<appInfo>` in Basic.

bwbohl:
I’m with @pe-ro; allowing `encodingDesc` in mei-basic seems more plausible because the general MEI model would be used consistently.

musicEnfanthen:
Pinging @musicog and @wergo to this discussion, since they have a need for appInfo in mei-friend as well.

wergo:
We had an issue with mei-friend trying to write an `appInfo` also to the header of MEI basic files, but we fixed it. 

I agree with the others that this would make sense to also have in MEI basic. Thanks @rettinghaus for raising this issue.

PR #1683, opened 2025-06-05
schema: make encodingDesc available in MEI Basic

rettinghaus:
solves #1548

### DIFF
--- a/customizations/mei-basic.xml
+++ b/customizations/mei-basic.xml
@@ -149,7 +149,7 @@
                 <moduleRef key="MEI.midi" include="instrDef" />
                 <moduleRef key="MEI.namesdates" include="persName" />
                 <moduleRef key="MEI.shared" 
-                    include="accid artic body caesura chord clef clefGrp date dir dynam ending label labelAbbr layer lb mdiv mei music note ornam pb pgFoot pgHead pubPlace rend respStmt rest sb score scoreDef section space staff staffDef staffGrp symbol syl tempo title "/>
+                    include="accid artic body caesura chord clef clefGrp date dir dynam ending label labelAbbr layer lb mdiv mei music name note ornam pb pgFoot pgHead pubPlace rend respStmt rest sb score scoreDef section space staff staffDef staffGrp symbol syl tempo title "/>
                 <moduleRef key="MEI.stringtab"/>
                 <!--<moduleRef key="MEI.text"/>-->
                 <moduleRef key="MEI.visual"/>
--- a/customizations/mei-basic.xml
+++ b/customizations/mei-basic.xml
@@ -148,8 +148,7 @@
                 <!--<moduleRef key="MEI.msDesc"/>-->
                 <moduleRef key="MEI.midi" include="instrDef" />
                 <moduleRef key="MEI.namesdates" include="persName" />
-                <moduleRef key="MEI.shared"
-                    include="accid arranger artic body caesura chord clef clefGrp composer date dir dynam ending label labelAbbr layer lb lyricist mdiv mei music note ornam pb pgFoot pgHead pubPlace rend respStmt rest sb score scoreDef section space staff staffDef staffGrp symbol syl tempo title "/>
+                <moduleRef key="MEI.shared" include="accid arranger artic body caesura chord clef clefGrp composer date dir dynam ending label labelAbbr layer lb lyricist mdiv mei music name note ornam pb pgFoot pgHead pubPlace rend respStmt rest sb score scoreDef section space staff staffDef staffGrp symbol syl tempo title "/>
                 <moduleRef key="MEI.stringtab"/>
                 <!--<moduleRef key="MEI.text"/>-->
                 <moduleRef key="MEI.visual"/>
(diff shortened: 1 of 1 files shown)


musicog (commented):
To support the `<appInfo>` blocks inserted by Verovio and mei-friend, `<name>` and `<p>` would also be needed. Of course they could be modified in this context if this is not desired...

rettinghaus:
Then we need to include `p` and `name` from the shared module. Do we want the complexity in Basic?

pe-ro (dismissed):
`<name>` and `<p>` should be added to facilitate up-conversion of mei-basic to other formats. They should be modified, however, to contain only textual content, dropping all of their currently-allowed sub-elements.

bwbohl:
I second @pe-ro’s suggestion to keep it as close to the MEI all model but reduce it. Small caveat:  Adding `p`and `name` might need disallowing them in other places.

rettinghaus:
So, I added name back to MEI Basic, so it is allowed as child of `application`. I tested also allowing `p` but that brings a whole lot of problems (or unwanted content) in other locations. 
I would propose to keep `p` out of MEI Basic, otherwise we had to refactor many models just for the Basic customization. At least I didn't found an easy way to accomplish this.

bwbohl:
MAybe you could try to remove `model.pLike` from `model.textComponentLike` and add `p` directly to `appInfo`?

rettinghaus:
@bwbohl That would fix it in `application`, but then `p` would also still be available in all those unwanted places like `pgHead`.

bwbohl (approved):

lpugin (approved):

musicEnfanthen:
`<name>` was added, `<p>` will not be added. Before MEI 6, `<persName>` needs to be removed in favor or `<name>` in MEI Basic, cf. #1702

musicEnfanthen:
@pe-ro Can you please double-check if requested changes are adressed and then give apporval. Thank you.

pe-ro:
The `<p>` element is disallowed as it's not included in the spec for MEI.shared in this customization --
```
<moduleRef key="MEI.shared" 
  include="accid artic body caesura chord clef clefGrp date dir dynam ending label labelAbbr layer lb 
  mdiv mei music name note  ornam pb pgFoot pgHead pubPlace rend respStmt rest sb score scoreDef 
  section space staff staffDef staffGrp symbol syl tempo title "/>
```
But I think it can be added back by:
1. adding a new spec for p (note the use of "add" instead of "change") -- 
```
<elementSpec ident="p" module="MEI.shared" mode="add">
  <gloss xml:lang="en">paragraph</gloss>
  <desc xml:lang="en">One or more text phrases that form a logical prose passage.</desc>
  <classes>
    <memberOf key="att.common"/>
    <memberOf key="att.lang"/>
    <!-- These attribute classes aren't needed in MEI Basic -->
    <!--<memberOf key="att.facsimile"/>-->
    <!--<memberOf key="att.metadataPointing"/>-->
    <!--<memberOf key="att.xy"/>-->
    <!--<memberOf key="model.pLike"/>-->
  </classes>
  <!-- The content of the p element is restricted to text content only, no children! -->
  <content>
    <rng:text/>
  </content>
  <remarks xml:lang="en">
    <p>A paragraph is usually typographically distinct: The text usually begins on a new
      line and the first letter of the content is often indented, enlarged, or both.</p>
  </remarks>
  <remarks xml:lang="en">
    <p>The model of this element is based on the <ref
        target="https://www.loc.gov/ead/EAD3taglib/EAD3-TL-eng.html#elem-p">p</ref>
      element of the Encoded Archival Description, the <ref
        target="https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-p.html">p</ref>
      element of the Text Encoding Initiative (TEI), and the <ref
        target="https://html.spec.whatwg.org/#the-p-element">p</ref> element of
        <abbr>HTML</abbr> .</p>
  </remarks>
</elementSpec>
```
2. making reference to the new p element in a revision of the spec for `<application>` --
```
<elementSpec ident="application" module="MEI.header" mode="change">
  <desc xml:lang="en">Provides information about an application which has acted upon the
    current document.</desc>
  <classes>
    <memberOf key="att.common"/>
    <memberOf key="att.datable"/>
  </classes>
  <content>
    <ref name="name"/>
    <zeroOrMore>
      <ref name="p"/>
    </zeroOrMore>
  </content>
  <attList>
    <attDef ident="version" usage="opt">
      <desc xml:lang="en">Supplies a version number for an application, independent of its
        identifier or display name.</desc>
      <datatype>
        <rng:data type="NMTOKEN"/>
      </datatype>
    </attDef>
  </attList>
  <remarks xml:lang="en">
    <p>The model of this element is based on the <ref
        target="https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-application.html"
        >application</ref> element of the Text Encoding Initiative (TEI).</p>
  </remarks>
</elementSpec>
```
This is untested, so please take it with a grain of salt.

bwbohl:
@lpugin, @musicog, and & @rettinghaus: would you like to have `p` avaiilable and follow @pe-ro’s suggestion or go without `p`?

rettinghaus:
As said before, I still think it's better to keep `p` out of MEI Basic. Rewriting the specs here is not really a solution.

musicog:
No huge requirement for the addition of `p` from my perspective. The proposal as stands is sufficient for our needs without it.

bwbohl:
@pe-ro, thanks for the potential additions regarding `p`, but as everybody seems content without it, I'll dismiss your requested changes to make this mergeable.

pe-ro:
Works for me

PR #1685, opened 2025-06-11
schema: remove `<rest[@line]` schematron rule

mwkohout:
Per a conversation with @bwbohl on [slack](https://music-encoding.slack.com/archives/C5HNVG41E/p1749649921472739), I've altered this rule to validate using loc.

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -7629,12 +7629,12 @@
         </rng:choice>
       </rng:zeroOrMore>
     </content>
-    <constraintSpec ident="Check_restline" scheme="schematron">
+    <constraintSpec ident="Check_restloc" scheme="schematron">
       <constraint>
-        <sch:rule context="mei:rest[@line]">
+        <sch:rule context="mei:rest[@loc]">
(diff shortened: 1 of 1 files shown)


rettinghaus:
Well spotted, but as rests can appear above or below the staff I would recommend to drop this rule entirely.

mwkohout:
thanks, @rettinghaus.  I'm happy altering this rule however folks want it since I'm just a visitor here.

Should I alter the pr to remove it now, or wait for more feedback?  
As an alternative would it make sense creating a rule that takes the different locations of `rest` in a document, similar to the rules around the `n` attribute in `staff` and `staffdef`?

ahankinson:
I agree with @rettinghaus -- `@loc` can be a number different from the number of lines on a staff. `@loc="5"` is a rest on the first ledger line above the staff, which is a valid location. 

So this rule can be removed entirely IMO.

mwkohout:
I've removed the rule, @rettinghaus, @ahankinson and @bwbohl.

rettinghaus (approved):
Thanks!

musicEnfanthen (approved):

bwbohl (approved):
many thanks @mwkohout !

PR #1687, opened 2025-06-23
schema: allow mid-measure changes to instrument and staff layout

th-we:
This allows changing the staff layout and instruments mid-measure.  Here is an example:

```xml
<score>
  <scoreDef>
    <staffGrp>
      <staffDef n="1" lines="1" clef.shape="perc" clef.line="1">
        <label>Tambourin</label>
        <instrDef midi.instrname="Tambourine"/>
      </staffDef>
    </staffGrp>
  </scoreDef>
  <section>
    <measure n="1">
      <staff n="1">
        <layer n="1">
          <note dur="4"/>
          <rest dur="4"/>
          <rest dur="4"/>
          <staffDef n="1" lines="5" clef.shape="G" clef.line="2" trans.semi="12" xml:id="instrChange">
            <label>Glockenspiel</label>
            <instrDef midi.instrname="Glockenspiel"/>
          </staffDef>
          <note pname="e" oct="4" dur="8"/>
          <note pname="f" oct="4" dur="8"/>
        </layer>
        <layer n="2">
          <space dur="2" dots="1"/>
          <staffDef sameas="#instrChange"/>
          <note pname="c" oct="4" dur="4"/>
        </layer>
      </staff>
    </measure>
  </section>
</score>
```

This encodes:

* A change in the number of staff lines
* A change of the instrument and its sound
* A change of transposition

... all of which is not possible mid-measure, unless `<staffDef>` is allowed inside `<layer>`.

`<staffDef>` is similar to `<clef>` and `<keySig>`. All of them have the issue that when there are multiple `<layer>`s in a `<staff>`, they should probably be repeated on every `<layer>` with `@sameas`. I made this a draft pull request as the Schematron rules enforcing this are still untested and I think there is a need to discuss whether this makes sense at all.

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -8050,6 +8050,7 @@
       <memberOf key="att.staffDef.ges"/>
       <memberOf key="att.staffDef.log"/>
       <memberOf key="att.staffDef.vis"/>
+      <memberOf key="model.eventLike"/>
       <memberOf key="model.staffDefLike"/>
     </classes>
     <content>
@@ -8179,6 +8180,25 @@
             <sch:let name="scorePPQ" value="preceding::mei:scoreDef[@ppq][1]/@ppq"/>
             <sch:assert test="($scorePPQ mod $staffPPQ) = 0">The value of ppq must be a factor of
               the value of ppq on a preceding scoreDef.</sch:assert>
+          </sch:rule>
+        </sch:pattern>
+      </constraint>
+    </constraintSpec>
+    <constraintSpec ident="Check_staffDef_in_layer" scheme="schematron">
+      <constraint>
+        <sch:pattern>
+          <sch:rule context="mei:staffDef[ancestor::mei:layer or ancestor::mei:oLayer]">
+            <sch:let name="staff" value="ancestor::*[self:mei:staff or self::mei:oStaff][1]"/>
+            <sch:let name="layers" value="$staff//mei:layer|$staff//mei:oLayer]"/>
+            <sch:let name="mainLayer" value="$layers[.//mei:staffDef[not(@sameas)]]"/>
+            <sch:let name="mainStaffDefs" value="$mainLayer//mei:staffDef[not(@sameas)]"/>
+            <sch:let name="secondaryStaffDefs" value="$layers//mei:staffDef[@sameas]"/>
+            <sch:assert
+              test="count($mainLayer) = 1
+                and not($layers[count(.//mei:staffDef) != count($mainStaffDefs)])
+                and count($secondaryStaffDefs) = count($mainStaffDefs) * (count($layers) - 1)"
+              >One layer must be used for main staffDefs. All other sibling layers must have
+              secondary staffDefs pointing to main staffDefs with @sameas.</sch:assert>
           </sch:rule>
         </sch:pattern>
       </constraint>


ahankinson:
Can you give a concrete example of notation that uses this?

I'm pretty skeptical that this is an appropriate solution. Changing `<staffDef>` to an event is a small change in code, but a really big change in the conceptual model of it. Allowing `<staffDef>` to occur anywhere brings with it a ton of attributes that are not suitable to appear within individual voices, like default octaves, number of staff lines, beam slope, etc.

There's a reason why individual elements (like `<clef>` and `<keySig>`) are used as milestones, even though you can define them on `<staffDef>` and `<scoreDef>`. They constrain the scope of possibilities to just the change that is needed. If this is really needed, then it would probably be worth exploring introducing a dedicated element, or changing the model of an existing specific element (like `<instrDef>`?).

th-we:
Here is a real-life example of a mid-measure instrument change that affects the sound and the number of staff lines:

image

rettinghaus:
@ahankinson Using specialized elements wouldn't be satisfactory, because there are too many parameters that can change, e.g. the abbreviated instrument name and the transposition. 

See [this corresponding example](https://www.w3.org/2021/06/musicxml40/musicxml-reference/examples/instrument-change-element/) for a way to encode this in MusicXML. 

`<staffDef>` is allowed within staff, so the question is, if it makes sense to encode partial layers in sequence. 

Usually full layers are encoded in parallel: 

```xml
<staff n="1">
    <layer n="1"></layer>
    <layer n="2"></layer>
    <layer n="3"></layer>
</staff>
```

Could we interrupt the layers at a given point and change staff properties like the following?

```xml
<staff n="1">
    <layer n="1"></layer>
    <layer n="2"></layer>
    <staffDef label="instrument change" />
    <!-- continuation of layers -->
    <layer n="1"></layer>
    <layer n="2"></layer>
</staff>
```

Which is valid MEI (but probably never encountered in the wild).

pe-ro:
I don't support allowing `<staffDef>` within `<layer>`.  In @th-we's encoding example, the existence of `<staffDef>` in both layers _prima facia_ signals a hierarchy problem with the model.  Allowing `layer/staffDef` would allow too much encoder freedom, resulting in significant reduction in interchangeability, not to mention severe limitation of interoperability.  

I do, however, support @rettinghaus' suggestion.  It may also bring in some misunderstandings, but because an attribute that was specified on an earlier `<staffDef>` but omitted on the most recent one continues to be in effect, and because there is no requirement that `<staffDef>` occur or take effect, at the start of a measure, actually `<staffDef>` is already treated as a measure-level event.  I think any new confusion can be ameliorated through better documentation.

That being said, _encoding_ a mid-measure staff definition/change and _rendering_ it are entirely different animals, so the effects of this new way of handling `<staffDef>` elements would have to be carefully examined.  For instance, how would the alignment of the staff lines in @th-we's visual example be specified?  What's supposed to happen if no alignment hints are provided?  Can mid-measure staff (re-)definitions be constrained via Schematron to reduce encoder abuse of the element and its attributes?  And so on ...

I eagerly await @lpugin's comments.

ahankinson:
> I do, however, support @rettinghaus' suggestion

`<layer>` elements are understood to be vertically aligned 'voices', so I would see this: 

```xml
<staff n="1">
    <layer n="1"></layer>
    <layer n="2"></layer>
    <staffDef label="instrument change" />
    <!-- continuation of layers -->
    <layer n="1"></layer>
    <layer n="2"></layer>
</staff>
```

As four stacked voices. I don't think that works?

"The term 'layer' is used instead of 'voice' in order to avoid confusion between 'voice' and 'voice leading' and 'voicing'. "

https://music-encoding.org/guidelines/v5/elements/layer.html

th-we:
What could be a better solution? Adding `@tstamp` to `<staffDef>`?

pe-ro:
Technically, multiple layers are just a sequence.  So far, we've treated that sequence as vertically-aligned voices/streams.  This interpretation is reinforced by presence of the `@n` attribute.  But, multiple `<layer>` elements with the same value of `@n` could indicate partial/incomplete layers.  Adding a `@join` attribute would help clarify this, and make it possible to link the incomplete layer elements into a single, logical entity  --
```
<staff n="1">
    <layer n="1" xml:id="layer1part1" join="#layer1part1 #layer1part2"></layer>
    <layer n="2" xml:id="layer2part1" join="#layer2part1 #layer2part2"></layer>
    <staffDef label="instrument change" />
    <!-- continuation of layers -->
    <layer n="1" xml:id="layer1part2" join="#layer1part1 #layer1part2"></layer>
    <layer n="2" xml:id="layer2part2" join="#layer2part1 #layer2part2"></layer>
</staff>
```
Schematron can be used to further clarify/constrain this situation.  For instance, `@join` could be required when there's more than one layer with the same `@n` value or when the sequence layer+, staffDef, layer+ is present.

Issue #1633, opened 2025-02-12
Rename neume@type to neume@neumetype


lpugin:
Overall, in MEI, the `@type` attribute is a generic attribute left for application-specific use, with no predefined content. However, for `neume`, there is a predefined list, which goes against the general practice.

I would suggest renaming it to `@neumetype` (following the `staffDef@notationtype` example) and keep `@type` as a generic attribute.

Tagging the @music-encoding/ig-neumes

rettinghaus:
Other possibilites could be `neume@name` or `neume@sign`.

pe-ro:
`neume/@type` is a semi-closed (or is it semi-open?) list. That is, the predefined values are simply suggestions to the encoder.  Other values are permitted.  So,  `<neume type="foobar">` is fine.  I don't see the problem.

ahankinson:
`@type` is mostly generic for application specific use in the rest of MEI, so having a specific function for that attribute here precludes the generic use of `@type`.

pe-ro:
@ahankinson, not true! Both suggested and user-supplied values are accommodated. `<neume type="oriscus"/>`, `<neume type="foobar"/>`, and `<neume type="oriscus foobar"/>` are all good.

ahankinson:
Sure, but what if foobar is a hint for some specific processing system? Surely it would be better to separate the concerns so that the list of acceptable names (open ended) are separate from application or project values?

pe-ro:
`@type` may carry multiple values, some of which a given processor may not be concerned about. We don't separate application or project values in `@type` anywhere else. Why do so with `<neume>`?  For example,

We allow
`<dir type="type1 type2">foo</dir>`,
so why not
`<neume type="type1 type2"></neume>`?

IMO, forcing `<neume>` to be different, e.g.,
`<neume name="oriscus" type="myType"></neume>`
will increase the number of attributes unnecessarily and result in more confusion.

ahankinson:
`<dir>` is a member of `att.typed`, which is the generic attribute class for `@type` and `@class`.  There is no standardization, nor none implied, in this attribute class. Most, if not all, other elements in MEI are a member of this class. (I tried to find examples of elements that are not a member of this class and I could not. If they do exist, caveat my next paragraph, in any case they are rare). 

`<neume>` belongs to `att.neumeType`, of which only two elements, `<nc>` and `<neume>` are a member. (Notably, these are *not* a member of `att.typed`, so are the only two exceptions). 

[...]

PR #1692, opened 2025-07-13
schema: add a couple of neume types

martha-thomae:
I am adding two extra values for the neume@type attribute semi-open list: https://music-encoding.org/guidelines/v5/attribute-classes/att.neumeType.html. 

These two neume types are very well known (for chant scholars): 

- cephalicus
  smufl_chant-cephalicus

- epiphonus
  smufl_chant-epiphonus

I am aware that there is a discussion about changing this `neume@type` attribute into `@neumetype` (or `@neumename` or something similar): https://github.com/music-encoding/music-encoding/issues/1633. But until this is not decided, I prefer to not include this change in this PR, and just include the extra values. I can make another PR in the future to account for the new attribute.

---
This would be related to the issue added in Verovio to render these two shapes (already available in SMuFL): https://github.com/rism-digital/verovio/issues/4119

### DIFF
--- a/source/modules/MEI.neumes.xml
+++ b/source/modules/MEI.neumes.xml
@@ -206,8 +206,10 @@
         <valList type="semi">
           <valItem ident="apostropha"/>
           <valItem ident="bistropha"/>
+          <valItem ident="cephalicus"/>
           <valItem ident="climacus"/>
           <valItem ident="clivis"/>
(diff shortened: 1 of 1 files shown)


lpugin (approved):

bwbohl (approved):

musicEnfanthen (approved):
Thank you!

PR #1693, opened 2025-07-15
guidelines: fix title attribute for attributes

### LINKED ISSUE #1663: Utils: html title attributes are not generated correctly
> Ok, so my latest push includes a fix for the search results page and a fix for element references in the guidelines. Unfortunately, the latter fix also introduces an issue with some html `@title`s on attribute definitions (`.attributeDef .ident.attribute`) on attribute and element descriptions (e.g. https://music-encoding.org/guidelines/v5/attribute-classes/att.accid.log.html#attributes_full_tab). For example, on the aforementioned page, `@when` looks like:
> ```html
> <span class="ident attribute" title="<when&gt; Indicates the point of occurrence of this feature along a time line. Its value must be the ID of a element elsewhere in the document.">@when</span>
> ```
> Interestingly, this issue only seems to occur when the title starts with a reference, and these references don't actually appear in [the attribute's description](https://github.com/maxrothman/music-encoding/blob/7721a63b79c339029f9fbedce621bdf968d7d371/source/modules/MEI.performance.xml#L43)!
> 
> There are a few other examples of this behavior, e.g. `@altsym` on `att.altSym` starts with a reference to `<symbolDef>`.
> 
> As far as I can tell, attribute titles come from [this function](https://github.com/maxrothman/music-encoding/blob/7721a63b79c339029f9fbedce621bdf968d7d371/utils/guidelines_xslt/odd2html/specs/genericSpecFunctions.xsl#L628), but I can't make heads or tails of why some attributes have identifiers stuck on the front of their titles and most don't. At this point I'm reaching the limit of my capabilities with XSLT. I also don't really understand why my change [over here](https://github.com/maxrothman/music-encoding/blob/7721a63b79c339029f9fbedce621bdf968d7d371/utils/guidelines_xslt/odd2html/guidelines.xsl#L136) caused this issue. Help would be very much appreciated!

_Originally posted by @maxrothman in https://github.com/music-encoding/music-encoding/issues/1531#issuecomment-2380232549_

rettinghaus:
With this change proper `@title` content is generated:

Bildschirmfoto 2025-07-15 um 18 17 46

closes #1663

### DIFF
--- a/utils/guidelines_xslt/odd2html/specs/genericSpecFunctions.xsl
+++ b/utils/guidelines_xslt/odd2html/specs/genericSpecFunctions.xsl
@@ -632,7 +632,7 @@
         <item class="attribute" ident="{$current.att/@ident}" module="{$module}">
             <link><xsl:value-of select="$current.att/@ident"/></link>
             <desc>
-                <span class="ident attribute" title="{normalize-space(string-join($desc/descendant-or-self::text(),' '))}">@<xsl:value-of select="$current.att/@ident"/></span>
+                <span class="ident attribute" title="{replace(normalize-space(string-join($desc,' ')), ' ([,.])', '$1')}">@<xsl:value-of select="$current.att/@ident"/></span>
                 <xsl:if test="$usage">
                     <span class="attributeUsage">(<xsl:value-of select="$usage"/>)</span>
                 </xsl:if>
--- a/utils/guidelines_xslt/odd2html/specs/genericSpecFunctions.xsl
+++ b/utils/guidelines_xslt/odd2html/specs/genericSpecFunctions.xsl
@@ -632,7 +632,7 @@
         <item class="attribute" ident="{$current.att/@ident}" module="{$module}">
             <link><xsl:value-of select="$current.att/@ident"/></link>
             <desc>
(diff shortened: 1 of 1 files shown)


utils/guidelines_xslt/odd2html/specs/genericSpecFunctions.xsl
```diff
@@ -632,7 +632,7 @@
         <item class="attribute" ident="{$current.att/@ident}" module="{$module}">
             <link><xsl:value-of select="$current.att/@ident"/></link>
             <desc>
-                <span class="ident attribute" title="{normalize-space(string-join($desc/descendant-or-self::text(),' '))}">@<xsl:value-of select="$current.att/@ident"/></span>
+                <span class="ident attribute" title="{replace(normalize-space(string-join($desc,' ')), ' ([,.])', '$1')}">@<xsl:value-of select="$current.att/@ident"/></span>
```
bwbohl: Both the original and the new version seem overly complicated would’t it work to just `normalize-space()` of `$current.att/tei:desc`?

rettinghaus: Yes.

bwbohl (approved):

musicEnfanthen (approved):
Very nice :)

martha-thomae (approved):

Issue #1478, opened 2024-05-30
Why is @cutout not a boolean?


rettinghaus:
The `cutout` attribute only allows the value "cutout". Is there a reason for this not to be a boolean? As far as I can see this is the only attribute with only a single value allowed.

rettinghaus:
Another thing that is not totally clear: Have `mRest cutout="cutout"` and `mSpace cutout="cutout"` the same visual expression, i.e. is the whole measure "invisible" or stayes the rest glyph for the former? 

NB: The Guidelines clearly state that "the complete staff including the staff lines will not be rendered".

rettinghaus:
Is there is a use case for a "per measure" encoding? Why is it not possible to define it as a general rule (refers to #1479). 
And why is `<space>` a member of `att.cutout`, while `<rest>` is not?

rettinghaus:
How would @cutout be expected to work with multiple layers?

pe-ro:
@rettinghaus, I'll try to answer your questions:

> The cutout attribute only allows the value "cutout". Is there a reason for this not to be a boolean?

Using "cutout" for `@cutout` was done to allow for the possibility of different types of cutouts or different renditions of a cutout. For example, I can imagine a kind of cutout where each cutout measure is replaced by its bounding box.  This could be indicated using `@cutout="boxed"`.  Switching to a Boolean value takes away this possibility.  

> Another thing that is not totally clear: Have `mRest cutout="cutout"` and `mSpace cutout="cutout"` the same visual expression, i.e. is the whole measure "invisible" or stayes the rest glyph for the former?

The same visual rendition is expected for both; that is, the entire measure -- staff lines, rests, spaces, etc., should not be rendered.  It may still be useful, however, to mark the measure as containing a rest or a space to facilitate a quick change to a non-cutout score by dropping the `@cutout` attribute.

> Is there is a use case for a "per measure" encoding? Why is it not possible to define it as a general rule (refers to https://github.com/music-encoding/music-encoding/issues/1479).

A "per measure" encoding is more explicit.  However, there's no reason why `@cutout` can't be added to `<scoreDef>`.  In that case, it would apply wherever a `<mRest>` or `<mSpace>` is encountered.  I don't consider this a species of optimization as you suggest in #1479.

> And why is `<space>` a member of` att.cutout`, while `<rest>` is not?

[...]

PR #1694, opened 2025-07-18
Clarify the implications of `@cutout="cutout"`

th-we:
Clarification based on discussion at #1478.

### DIFF
--- a/source/modules/MEI.cmn.xml
+++ b/source/modules/MEI.cmn.xml
@@ -450,11 +450,12 @@
         <desc xml:lang="en">"Cut-out" style.</desc>
         <valList type="closed">
           <valItem ident="cutout">
-            <desc xml:lang="en">The staff lines should not be drawn for the duration of the element.
-              It is sufficient to set <att scheme="MEI">cutout</att> on a single layer. Directly
-              adjacent barlines are only shown if on the other side of the barline, staff lines are
-              drawn (within the same system). The visibility of other elements is not
-              affected.</desc>
+            <desc xml:lang="en">Means that the staff lines "behind" the element are hidden for the
+              entire duration of the element, based on <att scheme="MEI">dur</att> and <att
+                scheme="MEI">dots</att>. <att scheme="MEI">cutout</att> does not imply that any
+              other elements are hidden. Only barlines directly adjacent to the cut-out area may be
+              hidden or shortened. When rendering, barline handling is up to the rendering
+              application.</desc>
           </valItem>
         </valList>
       </attDef>
--- a/source/modules/MEI.cmn.xml
+++ b/source/modules/MEI.cmn.xml
@@ -450,7 +450,11 @@
         <desc xml:lang="en">"Cut-out" style.</desc>
         <valList type="closed">
           <valItem ident="cutout">
-            <desc xml:lang="en">The staff lines should not be drawn.</desc>
+            <desc xml:lang="en">The staff lines should not be drawn for the duration of the element.
+              It is sufficient to set <att scheme="MEI">cutout</att> on a single layer. Directly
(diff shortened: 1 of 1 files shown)


bwbohl (dismissed):

pe-ro (changes requested):
I'm not sure what is meant by "It is sufficient to set `<att scheme="MEI">cutout</att>` on a single layer."  Remove this sentence?

bwbohl:
> I'm not sure what is meant by "It is sufficient to set `<att scheme="MEI">cutout</att>` on a single layer." Remove this sentence?

As I read it, this means that if any layer of a staff has an element with `@cutout=cutout`, the staff lines should not be drawn – one would not have to add it for all events at the specific position (which I think is fairly impossible or at least become quite challenging).

pe-ro:
@bwbohl, that's not my understanding of how cutout on `<rest>` or `<space>` is supposed to work.  If `@cutout` is specified on an element, the lines "behind" that element should not be drawn.  That's different than "turning off" the lines for the entire layer by providing `@cutout` on a single event-like element.  At least in theory, if not in practice.  :-)

th-we:
> If `@cutout` is specified on an element, the lines "behind" that element should not be drawn.

Right. That's what I tried to express with the "for the duration of the element".

> That's different than "turning off" the lines for the entire layer by providing `@cutout` on a single event-like element.

That's not what I intended.  We could maybe omit the sentence.  My intention was to eliminate any uncertainty what an encoding like this means:

```xml
<staff n="1">
  <layer n="1">
    <space dur="2" cutout="cutout"/>
    <chord>
      <note pname="g" oct="4" dur="4"/>
      <note pname="c" oct="4" dur="4"/>
    </chord>
    <note pname="f" oct="4" dur="4"/>
  </layer>
  <layer n="2">
    <space dur="2" dots="1"/>
    <note pname="d" oct="4" dur="8"/>
    <note pname="c" oct="4" dur="8"/>
  </layer>
</staff>
```

I wanted to make clear that this encoding means: For the first half of the measure, no staff lines are visible. No `@cutout` is required on layer 2, but it's of course O.K. to encode it redundantly. In this example, this would require splitting the space element in layer 2 because it's partly in the cutout part of the bar and partly in the part with ~~barlines~~ staff lines.

th-we:
I thought about how to make the wording clearer. Before I potentially add multiple revision commits to this pull request, here is a suggestion:

`Means that the staff lines "behind" the element are hidden for the entire duration of the element, based on <att scheme="MEI">dur</att> and <att scheme="MEI">dots</att>. <att scheme="MEI">cutout</att> does not imply that any other elements are hidden. Only barlines directly adjacent to the cut-out area may be shorter or hidden. When rendering, barline handling is up to the rendering application.`

In the examples posted [here](https://github.com/music-encoding/music-encoding/issues/1478#issuecomment-3070432759) we can see that barlines at cutouts may be shortened or omitted on specific staves. We can't encode this right now. Should we introduce features for encoding it, I'm sure they won't be mandatory. This means that in the the absence of such information, a rendering application still needs to decide what to do with the barlines.

rettinghaus:
I like this.

pe-ro:
A couple of slight modifications -- 

`Means that the staff lines "behind" the element are hidden for the entire duration of the element, based on <att scheme="MEI">dur</att> and <att scheme="MEI">dots</att>. <att scheme="MEI">cutout</att> does not imply that any other elements are hidden. Only barlines directly adjacent to the cut-out area may be hidden or shortened. When rendering, barline handling is up to the rendering application.`

th-we:
Thank you for the feedback, @pe-ro, I added a commit with the changed wording.

bwbohl (approved):

pe-ro (approved):
I think this is good to go.

PR #1695, opened 2025-07-21
schema: allow name in application only once

rettinghaus:
I couldn't think of any case where the following would make sense:

```xml
<encodingDesc>
   <appInfo>
      <application>
         <name>This name</name>
         <name>That name</name>
         <name>Yet antother name</name>
         <name>Your mother's name</name>
         <name>No name</name>
         <name></name>
      </application>
   </appInfo>
</encodingDesc>
```

This PR proposes to allow `name` only once as child of `application`.

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -471,9 +471,7 @@
       <memberOf key="att.datable"/>
     </classes>
     <content>
-      <rng:oneOrMore>
-        <rng:ref name="name"/>
-      </rng:oneOrMore>
+      <rng:ref name="name"/>
       <rng:choice>
         <rng:zeroOrMore>
           <rng:ref name="model.locrefLike"/>


bwbohl:
agree

bwbohl (approved):

pe-ro:
It is unlikely but possible that software can change its name.

ahankinson:
MuseScore -> MuseScore Studio?

rettinghaus:
I'm not sure what you are trying to say @pe-ro. Do you mean collecting all names of an application (past/present/future) in a single entry?

```xml
 <encodingDesc>
   <appInfo>
      <application>
         <name>Logic</name>
         <name>Logic Gold</name>
         <name>Logic Pro</name>
      </application>
   </appInfo>
</encodingDesc>
```

Would you then also encode all the versions of this application?

```xml
 <encodingDesc>
   <appInfo>
      <application version="4.0.1:4.0.2:4.1.0:5.0.0:17.6.1">
         <name>Logic</name>
         <name>Logic Gold</name>
         <name>Logic Pro</name>
      </application>
   </appInfo>
</encodingDesc>
```

pe-ro:
@rettinghaus,  I'm not sure about @ahankinson's example/question, but if it means that the same software is known by multiple names, then allowing `<name>` to occur more than once is desirable.  It could also be useful if the same software is called different things in different operating systems/countries/languages.

Of course, since `@version` isn't intended to allow multiple values, if the version number changes then a new `<application>` element is required.

Just my $.02.

martha-thomae:
So, in the end, we do want to allow for more than one `<name>` in `<application>`, given that an app can have multiple names, right? @pe-ro, @rettinghaus, and @ahankinson ?

If we do, I think it is ok to just close this PR.

pe-ro:
I'm fine with reducing the occurrence of `<name>` in this case to 1.  We can always increase it later if necessary.

ahankinson:
I think the thing to do is to leave it as-is, until there is a compelling use case to change it. 

You could have an application name in multiple languages, for example. Or an official and a common name: 

```
<name>Microsoft Word</name>
<name type="aka">Word</name>
```

bwbohl:
multiple languages convinves me not to touch this, thanks @ahankinson

PR #1696, opened 2025-07-21
tests: update schema path

rettinghaus:
Since we have not yet decided what to do with the test files, I thought it would at least be a good idea to update the `href` attributes in the PIs.


bwbohl (dismissed):

musicEnfanthen (approved):
THanks!

martha-thomae (approved):

PR #1704, opened 2025-07-26
schema: revise part and partstaff attributes

### LINKED ISSUE #1451: schema: exclude directions in extracted parts

rettinghaus:
closes #1451

NB: Schema builds fine but the generation of the Guidelines fails with this approach.

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -3162,8 +3162,8 @@
     <desc xml:lang="en">Reference to parts.</desc>
     <content>
       <rng:choice>
-        <rng:value>%all</rng:value>
-        <rng:value>%none</rng:value>
+        <rng:value>all</rng:value>
+        <rng:value>none</rng:value>
         <rng:list>
           <rng:oneOrMore>
             <rng:ref name="data.URI"/>
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -3158,6 +3158,31 @@
       </rng:data>
     </content>
   </macroSpec>
+  <macroSpec ident="data.PART" module="MEI" type="dt">
+    <desc xml:lang="en">Reference to parts.</desc>
+    <content>
+      <rng:choice>
+        <rng:value>%all</rng:value>
+        <rng:value>%none</rng:value>
+        <rng:list>
+          <rng:oneOrMore>
+            <rng:ref name="data.URI"/>
+          </rng:oneOrMore>
+        </rng:list>
+      </rng:choice>
+    </content>
+  </macroSpec>
+  <macroSpec ident="data.PARTSTAFF" module="MEI" type="dt">
+    <desc xml:lang="en">Reference to part staves.</desc>
+    <content>
+      <rng:choice>
+        <rng:value>%all</rng:value>
+        <rng:data type="token">
+          <rng:param name="pattern">\d+(-\d+)?</rng:param>
+        </rng:data>
+      </rng:choice>
+    </content>
+  </macroSpec>
   <macroSpec ident="data.PEDALSTYLE" module="MEI" type="dt">
     <desc xml:lang="en">Styling of piano pedal marks.</desc>
     <content>
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -3176,7 +3176,7 @@
     <desc xml:lang="en">Reference to part staves.</desc>
     <content>
       <rng:choice>
-        <rng:value>%all</rng:value>
+        <rng:value>all</rng:value>
         <rng:data type="token">
           <rng:param name="pattern">\d+(-\d+)?</rng:param>
         </rng:data>
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -2858,14 +2858,14 @@
     <desc xml:lang="en">Attributes for identifying the part in which the current feature appears.</desc>
     <attList>
       <attDef ident="part" usage="opt">
-        <desc xml:lang="en">Indicates the part in which the current feature should appear. Use '%all' when the feature
-          should occur in every part or '%none' when the feature should not appear in any part.</desc>
+        <desc xml:lang="en">Indicates the part in which the current feature should appear. Use 'all' when the feature
+          should occur in every part or 'none' when the feature should not appear in any part.</desc>
         <datatype>
           <rng:ref name="data.PART"/>
         </datatype>
       </attDef>
       <attDef ident="partstaff" usage="opt">
-        <desc xml:lang="en">Signifies the part staff on which a notated feature occurs. Use '%all' when the
+        <desc xml:lang="en">Signifies the part staff on which a notated feature occurs. Use 'all' when the
           feature should occur on every staff.</desc>
         <datatype maxOccurs="unbounded">
           <rng:ref name="data.PARTSTAFF"/>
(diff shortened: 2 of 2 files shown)


bwbohl:
@rettinghaus is there any other reason than the failing build for this being a draft pull request? It seems plausible and a valid addition. If it’s just about the build, we can try fix it

bwbohl:
a local test we ran revealed that moving the datatype definition of partstaff and part attributes outside to their own macroSpec definitions would make the guidelines build work

rettinghaus:
@bwbohl I added two new datatypes, Guidlines build and test file validation works as expected.

bwbohl (dismissed):

bwbohl:
many thanks @rettinghaus !

pe-ro:
The line `<rng:param name="pattern">(%all|#[\i][\c]+)</rng:param>` was intended to place a limitation on the use of a URI in the `@part` attribute; that is, rather than allowing any URI, including targets elsewhere, it permitted only those *inside* the current document.  In other words, ID refs.  We can loosen this restriction with data.URI as @rettinghaus suggests or we can keep it  (which I prefer) by creating a new `data.ID` datatype:
```
<macroSpec ident="data.ID" module="MEI" type="dt">
  <desc xml:lang="en">An XML ID.</desc>
  <content>
    <rng:data type="token">
      <rng:param name="pattern">#[\i][\c]*</rng:param>
    </rng:data>
  </content>
</macroSpec>
```
which is invoked in `data.PART`:
```
<macroSpec ident="data.PART" module="MEI" type="dt">
  <desc xml:lang="en">Reference to parts.</desc>
  <content>
    <rng:choice>
      <rng:value>%all</rng:value>
      <rng:value>%none</rng:value>
      <rng:list>
        <rng:oneOrMore>
          <rng:ref name="data.ID"/>
        </rng:oneOrMore>
      </rng:list>
    </rng:choice>
  </content>
</macroSpec>
```

bwbohl:
I think `data.ID` would be a valuable addition to the specs, although the name might be misleading because it’s actually an ID prefixed with `#`, i.e., a (local) fragment identifier. A benefit of `data.URI`, which basically is `xs:anyURI`, is that editing environments with enough intellisense would propose local fragment identifiers. So an alternative would be to define it as `data.URI`or `anyURI` with your pattern data facet (maybe with head and tail: `^#[\i][\c]*$` – don’t know if that’s actually needed in validation contexts).

pe-ro:
>it’s actually an ID prefixed with #, i.e., a (local) fragment identifier

You're right. A better name would be `data.IDREF`.  You're also correct that Oxygen and similar editors will likely suggest fragment identifiers anyway, so the whole idea might be overkill.  One might actually want to reference a `<part>` element encoded in a location outside the current document, so opening up `data.PART` is fine.

ahankinson:
Wouldn't MEI `data.IDREF` be confusable with `xs:IDREF`? https://www.w3.org/TR/xmlschema-2/#IDREF

xs:IDREF (as far as I know) doesn't require the fragment identifier to be valid. You could keep the same idea without using the same name if you called it `data.IDReference`.

pe-ro:
data.IDREF could be confused with xs:IDREF, but no more so than data.NMTOKEN (which we already use) is confuse-able with xs:NMTOKEN or data.URI with xs:anyURI. Directly invoking xs:IDREF would be fine with me; I was just trying to be consistent with data.NMTOKEN.

Of course, ultimately data.NMTOKEN just resolves to the RNG NMTOKEN datatype, but it wouldn't necessarily have to do that.  And neither would data.IDREF.  Indirection can help with future-proofing.

Whether id references are validated in RNG; that is, checked to make sure they actually exist in the current document, is controlled by whether one turns on DTD compatibility.  Currently, we use a different approach; that is, the W3C datatypes:
```
<grammar xmlns="http://relaxng.org/ns/structure/1.0" xmlns:tei="http://www.tei-c.org/ns/1.0"
  xmlns:teix="http://www.tei-c.org/ns/Examples" xmlns:xlink="http://www.w3.org/1999/xlink"
  datatypeLibrary="http://www.w3.org/2001/XMLSchema-datatypes"
  ns="http://www.music-encoding.org/ns/mei">
```
(note the **datatypeLibrary** line) plus Schematron like:
```
<sch:assert role="warning"
        test="every $i in tokenize(., '\s+') satisfies substring($i,2)=//mei:*/@xml:id">The value in
        @startid should correspond to the @xml:id attribute of an element.</sch:assert>
```
which checks whether a referenced entity exists. I think we did this because using the RNG compatibility datatypes created problems, although I don't remember exactly what the issues were.

We can use the data.IDREF datatype for data.PART, avoid checking the id reference with a regex "^#[\i][\c]*$", and add a Schematron rule that makes sure `@part` contains a reference to an existing ID.  Or we can follow @rettinghaus's lead and use data.URI and rely on the combination of editor suggestions of likely values (as @bwbohl proposes) and encoder preference (when the referenced part actually is outside the document).  Either way is fine with me.

ahankinson:
But data.NMTOKEN and xs:NMTOKEN have the same definition, don’t they? data.IDREF would require the leading #, while xs:IDREF does not, as far as I can see?

Im not saying we shouldn’t shadow the names, but we should at least ensure those that share names also share definitions.

pe-ro:
>data.IDREF would require the leading #, while xs:IDREF does not, as far as I can see?

That's correct. So, let's just use data.URI for `@part` (in RNG syntax) --
```
<attribute name="part">
  <a:documentation xmlns:a="http://relaxng.org/ns/compatibility/annotations/1.0">Indicates the part
    in which the current feature should appear. Use '%all' when the feature should occur in every
    part, '%none' when the feature should not appear in any part, or one or more URIs, including
    fragment identifiers, to identify one or more parts to which the feature
    applies.</a:documentation>
  <choice>
    <value>%all</value>
    <value>%none</value>
    <list>
      <oneOrMore>
        <ref name="mei_data.URI"/>
      </oneOrMore>
    </list>
  </choice>
</attribute>
```
The `<choice>` can be abstracted into data.PART if desired.  I think we can also drop the "%" signs in the values "all" and "none".

pe-ro (commented):
Using data.URI in data.PART allows referencing of parts not defined in the current file. This contravenes the original intention of the regEx.  If `@part` references should continue to only point to part elements in the current file, a new Schematron rule is required.

I suggest removing the % signs in the part and partstaff datatype values for "all" and "none".

bwbohl:
Developer Meeting:
* let’s go with `data.URI`, remove the `%`’s and the regeEx.

source/modules/MEI.xml
```diff
@@ -3182,6 +3182,31 @@
       </rng:data>
     </content>
   </macroSpec>
+  <macroSpec ident="data.PART" module="MEI" type="dt">
+    <desc xml:lang="en">Reference to parts.</desc>
+    <content>
+      <rng:choice>
+        <rng:value>%all</rng:value>
+        <rng:value>%none</rng:value>
```
```suggestion
        <rng:value>all</rng:value>
        <rng:value>none</rng:value>
```

source/modules/MEI.xml
```diff
@@ -3182,6 +3182,31 @@
...
+          </rng:oneOrMore>
+        </rng:list>
+      </rng:choice>
+    </content>
+  </macroSpec>
+  <macroSpec ident="data.PARTSTAFF" module="MEI" type="dt">
+    <desc xml:lang="en">Reference to part staves.</desc>
+    <content>
+      <rng:choice>
+        <rng:value>%all</rng:value>
```
```suggestion
        <rng:value>all</rng:value>
```

martha-thomae (dismissed):

martha-thomae (approved):

bwbohl (approved):
thanks for updating this so fast ;-)

PR #1712, opened 2025-09-03
schema: add new notationtype value `tab.staff-like`

janjusolja:
adding a new `@notationtype` value `tab.cmn` to enable rendering of CMN notes within `<tabGrp>`

This PR is proposed in connection with [Verovio issue #4121](https://github.com/rism-digital/verovio/issues/4121) regarding the rendering of CMN notes within `<tabGrp>`, and suggests adding a new value to the `@notationtype`. The documentation is consequently updated and an encoding example is provided. This notation type has been used in multiple tablature editions dealing with diplomatic ("transcription brute") or automated transcriptions, such as those by Charnassé, Meylan and Ducasse (_Hans Gerle, Tabulatur auff die Laudten (Nuremberg, 1533), Transcriptions automatiques_, 1975), and El-Mallah (_Ein Tanzzyklus des 16. Jahrhunderts für Laute von Jacomo Gorzanis_, 1979). It is also currently being used in the E-LAUTE project. Co-contributors: @DILewis , @musicog and @reinierdevalk .

refs rism-digital/verovio#4121


ahankinson:
“Cmn” would imply all of common western notation. Is this correct? Or is it more like “tab.noteheads” or something along those lines?

reinierdevalk:
@ahankinson Probably more the latter. We just used the term 'CMN' to mean 'not tablature', which is of course an oversimplification. If this is about the naming only, then we would be happy to change that.

ahankinson:
`tab.staff-notation` perhaps? 

The examples in the Verovio issue would suggest that it’s not actually CMN, but some form of stylized staff notation.

pe-ro:
Riffing on @ahankinson's suggestion, how about "tab.staff-notation-like"? Or expanding on @reinierdevalk's original name, "tab.cmnLike"? I prefer the 2nd suggestion better because it conveys the combination of tablature and CMN notation in a shorter form.

reinierdevalk:
I am open to all three suggestions, but I like `tab.cmn-like` the best -- for the exact same reason that @pe-ro mentions.

ahankinson:
My suggestion to avoid the terminology of “CMN” came from looking at the linked Verovio issue. It’s pretty clear that it’s not actually “CMN” but instead a stylized form of staff notation. Other than the use of a staff for placing the pitches, it doesn’t really bear much resemblance to CMN. Even the clefs are formed differently. 

So I would avoid “CMN” terminology all together. Instead it seems to be a form of tab notation written on a staff.  By calling it differently that actually gives you more freedom to break from the strictures of “CMN-like” behaviours and do things that are only found in this type of notation — like drawing the beams above the staff unconnected from the note heads.

pe-ro:
Saying something is "bird-like" doesn't mean it _IS_ a "bird".  It can just mean it has wings without any of the other features of a bird.

As far as I know, there's no existing label for this kind of notation; that is, tab with the pitches rendered on a staff.  So, 2 options come to mind -- name it according to its features; e.g., "tab-on-a-staff", or name it for something that it resembles; e.g., "CMN-like". 

"tab.staff-notation" is fine.  But so is "tab.cmn-like".

ahankinson:
Sure, but saying something is “bird-like” at a conference of ornithologists probably conjures a bunch of unintended assumptions when what you are talking about is an airplane.

We already have a defined system of notation in MEI that is named CMN, and this aligns with very little of that system.We can certainly take pieces of it, but better to not conjure up expectations of what it must have, and instead let it be defined by what it actually is.

reinierdevalk:
We thought about something like `tab.notehead` ourselves when we came up with this idea, but after some discussion we didn't go with it. How about that name?

ahankinson:
`tab.staff-like`?

The most defining characteristic of this notation doesn’t seem to be the note heads, but the fact that it’s on stafflines.

reinierdevalk:
`tab.staff-like` seems like a good compromise.

source/docs/07-tablature.xml
```diff
@@ -181,6 +181,18 @@
                   </figure>
                </p>
                </div>
+               <div xml:id="tab-cmn" type="div3">
```
musicEnfanthen:
```suggestion
               <div xml:id="tab-staff-like" type="div3">
```

source/docs/07-tablature.xml
```diff
@@ -181,6 +181,18 @@
...
+               <head>Special Case: 'Notehead' Notation</head>
+                  <p>A special kind of <att>notationtype</att> is <val>tab.staff-like</val>, also referred to as 'notehead notation'. This notationtype is a hybrid between CMN and tablature that uses solid, stemless noteheads on a CMN staff, with tablature rhythm flags placed above the staff. It provides only pitch and onset information, and gives no interpretation of note duration or voice leading; as such, it faithfully transcribes tablature into a CMN-like form.</p>
```
musicEnfanthen:
```suggestion
                  <p>A special kind of <att>notationtype</att> is <val>tab.staff-like</val>, also referred to as 'notehead notation'. This notationtype is a hybrid between CMN and tablature that uses solid, stemless noteheads on a CMN staff, with tablature rhythm flags placed above the staff. It provides only pitch and onset information, and gives no interpretation of note duration or voice leading; as such, it faithfully transcribes tablature into a staff-like form.</p>
```

source/modules/MEI.xml
```diff
@@ -3037,6 +3037,9 @@
         <valItem ident="tab">
           <desc xml:lang="en">Tablature notation.</desc>
         </valItem>
+        <valItem ident="tab.staff-like">
+          <desc xml:lang="en">"Notehead" notation: a hybrid between CMN and tablature notation consisting of solid, stemless noteheads on a CMN staff with tablature rhythm flags placed above the staff.</desc>
```
```suggestion
          <desc xml:lang="en">"Notehead" notation: a hybrid between staff and tablature notation consisting of solid, stemless noteheads on a CMN staff with tablature rhythm flags placed above the staff.</desc>
```

musicEnfanthen (approved):

martha-thomae (approved):

PR #1716, opened 2025-09-19
schema: make mdiv member of att.lang

rettinghaus:
I noticed, that it not possible to have multiple versions of a pice in different languages in multiple mdivs by specifying the language of the mdiv. This PR adds `@xml:lang` to `mdiv` to allow structures like

```xml
<music>
    <mdiv xml:lang="de">
        <!-- German version -->
    </mdiv>
    <mdiv xml:lang="en">
        <!-- English version -->
    </mdiv>
    <mdiv xml:lang="fr">
        <!-- French version -->
    </mdiv>
</music>
```

### DIFF
--- a/customizations/mei-basic.xml
+++ b/customizations/mei-basic.xml
@@ -1236,6 +1236,7 @@
                     <classes mode="replace">
                         <memberOf key="att.common"/>
                         <!--<memberOf key="att.facsimile"/>-->
+                        <memberOf key="att.lang"/>
                         <!--<memberOf key="att.metadataPointing"/>-->
                         <!--<memberOf key="att.mdiv.anl"/>-->
                         <memberOf key="att.mdiv.ges"/>
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -6699,6 +6699,7 @@
     <classes>
       <memberOf key="att.common"/>
       <memberOf key="att.facsimile"/>
+      <memberOf key="att.lang"/>
       <memberOf key="att.metadataPointing"/>
       <memberOf key="att.mdiv.anl"/>
       <memberOf key="att.mdiv.ges"/>


ahankinson:
Wouldn't you specify the language on the textual parts, not on the music itself?

rettinghaus:
Sure, but you have to do some pretty complex xpath juggling to figure out what you want, because the lyrics may be (mostly) German, the directives Italian and even more weird mixtures. 
But mdiv is the container you would specify the **main** language, just like in HTML or TEI.

ahankinson:
Noting that this PR also makes the change to the MEI Basic schema.

customizations/mei-basic.xml
```diff
@@ -1236,6 +1236,7 @@
                     <classes mode="replace">
                         <memberOf key="att.common"/>
                         <!--<memberOf key="att.facsimile"/>-->
+                        <memberOf key="att.lang"/>
```
bwbohl, line 1239: @kepper & @lpugin could you please assess this change, does it make sense to you, too?

bwbohl (approved):

customizations/mei-basic.xml
bwbohl, line 1239: @kepper & @lpugin could you please assess this change, does it make sense to you, too?

bwbohl (approved):

customizations/mei-basic.xml
lpugin: I am not fully convinced we need this in MEI-Basic, but I am not against it. If would still be useful to know more in which context this came up.

rettinghaus: I added `att.lang` to `mdiv` also to mei-basic, because it is generelly allowed in Basic. Otherwise it would be a strange inconsistency, and we should better completely remove it from MEI Basic.

lpugin: I am not against adding it to `mdiv`, but I would be against removing it completely if it is not added there. I don't think MEI-Basic is that broken as it is.

lpugin: Also, maybe @pe-ro could comment on why is has been omitted in the first place, and why is it not on `section` either, since I would expect a similar argument. Namely, that a piece could have different sections in different languages, and that is not possible to encode in MEI. (I am not saying we need.)

pe-ro: Of course `@xml:lang` *"cascades"* down the hierarchy.  However, putting aside the fact that `<mdiv>` elements can't exist directly within `<music>`, it's generally better to put the language designation as close to the linguistic content as possible.
```
<music>
  <mdiv xml:lang="de">
    <!-- German version -->
  </mdiv>
  <mdiv xml:lang="en">
    <!-- English version -->
  </mdiv>
  <mdiv xml:lang="fr">
    <!-- French version -->
  </mdiv>
</music>
```
says that everything in the mdiv (including the music notation itself) is in German, English, or French!  On the other hand,
```
<music>
  <mdiv>
    <score>
      . . .
      <div xml:lang="de"><!-- lyrics in German--></div>
      <div xml:lang="en"><!-- lyrics in English --></div>
      <div xml:lang="fr"><!-- lyrics in French --></div>
    </score>
</music>
```
makes it clear that only the additional lyrics are in different languages.

`<mdiv>` should be used to mark large-scale; i.e., movement-level, structure within a score. It shouldn't be used for different "versions" of some material.  One should use critical apparatus markup (e.g., `<app>` and `<rdg>`) or editorial markup (e.g., `<orig>` and `<reg>`, `<sic>` and `<corr>`, etc.) for both notational and textual variation.

bwbohl:
Just to give this a second thought:
The XML specs define: `@xml:lang` (and `@xml:space`,`@xml:base`, and `@xml:id`) in a way that they may be schema-controlled, but also that they may be inserted/used (where applicable). So I’m wondering whether we _really_ should try to foresee where someone encoding MEI might want to use them, or _instead_ just allow them on any element?

rettinghaus:
@bwbohl thanks for the suggestion! I like it.

lpugin:
FWIW, `xml:lang` has been [deprecated](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/xml:lang) from SVG

pe-ro:
I think this is a bad idea. 

- Every element in MEI already allows the xml:id attribute.  
- All elements except the mei element itself already allows the xml:base attribute.  
- Elements where whitespace could be considered to be significant (extData, extMeta, incipCode, and rend) already allow xml:space.
- Elements with potential linguistic content already allow the xml:lang attribute.

Making xml:space and xml:lang universal invites the kind of abuse I described above.  I don't see how allowing `<note xml:lang="fr">` is an improvement. It's the `<syl>` or `<verse>` descendants of note that need `@xml:lang`, not the note.  xml:space marks whether whitespace in the element is significant or not.  Empty elements, like `<bracketSpan>`,  or elements that don't allow text, like '<note>', don't need it and can't meaningfully use it.  Allowing it will only invite confusion.

ahankinson:
I would agree with @pe-ro here -- non-linguistic content shouldn't be marked with a language tag. 

The only exception *might* be (and it's not a very strong might) allowing `xml:lang` on `<mei>` itself, as a possible mirror to having a language tag for an HTML document. (`<html lang="en">` <-> `<mei xml:lang="en">`)

I'm not prepared to argue for that, though. Just a thought.

pe-ro:
@ahankinson, thanks for commenting.

I'll add that allowing xml:lang on `<mei>` is essentially specifying a default language; that is, it says where a language isn't specified further down in the document, use the value from `mei/@xml:lang`.

Putting aside the question of whether music notation (as opposed to the text sometimes found in music notation) is any particular language, default values may save a few keystrokes or bits in memory or on disk , but they create another (I'd say bigger) problem because they introduce _**assumptions**_ into the markup. Explicit markup is better because it reduces the number of assumptions.  In the case where some text is in German and some is in Italian, it's better to explicitly mark each element's language than to rely on a distant default value.  Similar to attribute values, when they're not in a specific namespace they're not in the default namespace but in the "no namespace" namespace, when no language is specified for textual content they should be thought of as in the "no language" language.

PR #1718, opened 2025-09-29
build: move build directories to properties

rettinghaus:
This moves the designation of paths out of the `build.xml` into `build.properties` switching to the more common naming scheme. 

NB: It is supposed to work on all platforms, but should be tested on windows.


bwbohl:
reposting here beacuse I unintentionally posted it to another issue:
Just asking me, which apporach is better, this one here or what I did in https://github.com/music-encoding/music-encoding/tree/refactor%2Fbuildxml

musicEnfanthen:
@bwbohl Your link leads to a 404

bwbohl:
just moved the branch to my fork https://github.com/bwbohl/music-encoding/tree/refactor%2Fbuildxml

nontheless I think the best option would be a combination of both: moving parameters to a separate file but keeping it an ant file and importing it back to build.xml

PR #1723, opened 2025-11-27
schema: add attribute class and attribute for recording the status of a file

riedde:
This PR introduces an attribute class for attributes that describe or define a status. The idea came up in https://github.com/Edirom/MerMEId/issues/283 and has been discussed in the @music-encoding/ig-metadata meetings and the [working meeting](https://github.com/music-encoding/metadata-ig/wiki/2025#2025-11-13-10am-utc1) in November 2025.

The attribute `@status` will be allowed in `<change>` to make possible documentation of the change of a documents status. Further it will be allowed in all elements that can be root in mei-all (`<mei>`, `<meiHead>`, `<meiCorpus>`, `<music>`).

At the moment it is also allowed in `<revisionDesc>`, analogously to TEI, but this just an option to be discussed.

The value list for `@status` is also a reduced setup of the definition in TEI. Descriptions for the single values have been added. @pe-ro offered to revise the wording.
The value list is proposed as a basic setup, but adjustable if not sufficient, since it is semi-open!

ToDo:

- [x] Adding documentation to the Guidelines

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -1697,6 +1697,7 @@
       <memberOf key="att.meiVersion"/>
       <memberOf key="att.responsibility"/>
       <memberOf key="model.startLike.header"/>
+      <memberOf key="att.docStatus"/>
     </classes>
     <content>
       <rng:zeroOrMore>
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -2311,7 +2311,6 @@
     <classes>
       <memberOf key="att.common"/>
       <memberOf key="att.bibl"/>
-      <memberOf key="att.docStatus"/>
     </classes>
     <content>
       <rng:zeroOrMore>
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -725,6 +725,7 @@
       <memberOf key="att.common"/>
       <memberOf key="att.bibl"/>
       <memberOf key="att.datable"/>
+      <memberOf key="att.docStatus"/>
     </classes>
     <content>
       <rng:optional>
@@ -2309,6 +2310,7 @@
     <classes>
       <memberOf key="att.common"/>
       <memberOf key="att.bibl"/>
+      <memberOf key="att.docStatus"/>
     </classes>
     <content>
       <rng:zeroOrMore>
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -6729,6 +6729,7 @@
       <memberOf key="att.id"/>
       <memberOf key="att.meiVersion"/>
       <memberOf key="att.responsibility"/>
+      <memberOf key="att.docStatus"/>
     </classes>
     <content>
       <rng:ref name="meiHead"/>
@@ -6835,6 +6836,7 @@
       <memberOf key="att.common"/>
       <memberOf key="att.meiVersion"/>
       <memberOf key="att.metadataPointing"/>
+      <memberOf key="att.docStatus"/>
     </classes>
     <content>
       <rng:zeroOrMore>
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -1110,7 +1110,6 @@
         <datatype>
             <rng:data type="NMTOKEN"/>
         </datatype>
-        <defaultVal>draft</defaultVal>
         <valList type="semi">
           <valItem ident="draft">
             <desc>Initial or preliminary version of the document.</desc>
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -1146,17 +1146,17 @@
           </change>
           <change isodate="2025-10-16" resp="#CG" status="approved">
             <changeDesc>
-              <p>Proofreading of the document.</p>
+              <p>The document has been proofread. It's ready for publishing</p>
             </changeDesc>
           </change>
           <change isodate="2025-10-05" resp="#DR" status="candidate">
             <changeDesc>
-              <p>Elaborating the content of the document.</p>
+              <p>The content of the document is elaborated and the document is now a candidate for publishing. It can be proofread.</p>
             </changeDesc>
           </change>
           <change isodate="2025-10-01" resp="#DR" status="draft">
             <changeDesc>
-              <p>Creating the document.</p>
+              <p>The document was created.</p>
             </changeDesc>
           </change>
         </revisionDesc>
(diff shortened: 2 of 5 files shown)


rettinghaus (commented):
Looks good. But I think, this is the first case where we actually make use of `tei:exemplum`, which is not tendered in the Guidlines, correct? Maybe for the time being add these examples also as single files to show up in the Guidlines.

riedde:
> Looks good. But I think, this is the first case where we actually make use of `tei:exemplum`, which is not tendered in the Guidlines, correct? Maybe for the time being add these examples also as single files to show up in the Guidlines.

The documentation in the guidelines is missing (sorry, I missed it.). The `<exemplum>` is not rendered in the guidelines, but it is shown in the attribnute definition, so I would add an example to the guidelines, but also would like to keep it here, if it doesn't cause problems.

source/modules/MEI.shared.xml
```diff
@@ -1102,6 +1102,70 @@
...
   </classSpec>
+  <classSpec ident="att.docStatus" module="MEI.shared" type="atts">
+    <desc versionDate="2025-11-27" xml:lang="en">Attributes for describing the status of a document.</desc>
+    <attList>
+      <attDef ident="status" usage="opt">
+        <desc versionDate="2025-11-27" xml:lang="en">Used to describe the status of a document (currently or at the time indicated by a date).</desc>
+        <datatype>
+            <rng:data type="NMTOKEN"/>
+        </datatype>
+        <defaultVal>draft</defaultVal>
```
riedde: I wondered, why we don't use default values. I'm going to use them more often in my projects. But here it is optional for me.

source/modules/MEI.shared.xml
```diff
@@ -1102,6 +1102,70 @@
...
+        <valList type="semi">
+          <valItem ident="draft">
+            <desc>Document is created or the content is a draft.</desc>
```
riedde:
```suggestion
            <desc>Initial or preliminary version of the document.</desc>
```

source/modules/MEI.shared.xml
```diff
@@ -1102,6 +1102,70 @@
...
+          </valItem>
+          <valItem ident="proposed">
+            <desc>Content of the document is proposed for review or still in progress (if the status "candidate" is used).</desc>
+          </valItem>
+          <valItem ident="candidate">
+            <desc>Content of the document is ready for reviewing.</desc>
+          </valItem>
```
pe-ro:
```
<valItem ident="candidate">
  <desc>Content of the document is ready for review.</desc>
</valItem>
```

source/modules/MEI.shared.xml
pe-ro, line 1: Suggested revision:
1. Remove `<defaultValue>`.
2. Change "reviewed" to "in-process".
3. Revise descriptions.
4. Modify example markup.
5. I also suggest removing `@status` from `<revisionDesc>` since the latest status of the document is captured by the chronologically last `<change>`.
```
<classSpec ident="att.docStatus" module="MEI.shared" type="atts"
  xmlns:rng="http://relaxng.org/ns/structure/1.0">
  <desc versionDate="2025-11-27" xml:lang="en">Attributes for describing the status of a
    document.</desc>
  <attList>
    <attDef ident="status" usage="opt">
      <desc versionDate="2025-11-27" xml:lang="en">Used to characterize the MEI document
        with respect to its electronic file life cycle.</desc>
      <datatype>
        <rng:data type="NMTOKEN"/>
      </datatype>
      <valList type="semi">
        <valItem ident="draft">
          <desc>Initial state of the document.</desc>
        </valItem>
        <valItem ident="in-process">
          <desc>The content of the document is under development.</desc>
        </valItem>
        <valItem ident="candidate">
          <desc>The content of the document is ready for review.</desc>
        </valItem>
        <valItem ident="approved">
          <desc>The content of the document has been reviewed and is ready for
            publication.</desc>
        </valItem>
        <valItem ident="published">
          <desc>Document is published.</desc>
        </valItem>
        <valItem ident="withdrawn">
          <desc>Publication of the document has been reversed.</desc>
        </valItem>
        <valItem ident="embargoed">
          <desc>Document can't be published (in full or in part) because of an
            embargo.</desc>
        </valItem>
      </valList>
    </attDef>
  </attList>
  <exemplum>
    <egXML xmlns="http://www.tei-c.org/ns/Examples" xml:space="preserve" valid="true">
        <revisionDesc>
          <change isodate="2025-11-27" resp="#DR" status="embargoed">
            <changeDesc>
              <p>Not accessible until 2027-01-01.</p>
            </changeDesc>
          </change>
          <change isodate="2025-10-16" resp="#CG" status="approved">
            <changeDesc>
              <p>Candidate approved.</p>
            </changeDesc>
          </change>
          <change isodate="2025-10-16" resp="#CG" status="candidate">
            <changeDesc>
              <p>Proofreading complete.</p>
            </changeDesc>
          </change>
          <change isodate="2025-10-05" resp="#DR" status="in-process">
            <changeDesc>
              <p>Elaborating the content of the document.</p>
            </changeDesc>
          </change>
          <change isodate="2025-10-01" resp="#DR" status="draft">
            <changeDesc>
              <p>Creating the document.</p>
            </changeDesc>
          </change>
        </revisionDesc>
      </egXML>
  </exemplum>
  <remarks>
    <p>The model of this attribute class is based on the <ref
        target="https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-att.docStatus.html"
        >att.docStatus</ref> class of the Text Encoding Initiative (TEI).</p>
  </remarks>
</classSpec>
```

source/modules/MEI.header.xml
```diff
@@ -2309,6 +2311,7 @@
     <classes>
       <memberOf key="att.common"/>
       <memberOf key="att.bibl"/>
+      <memberOf key="att.docStatus"/>
```
pe-ro: Remove `<revisiondesc>` membership in `att.docStatus` since the current status of a document can be determined from the chronologically last `<change>` element.

pe-ro (changes requested):
See comments

source/modules/MEI.header.xml
riedde: Thank you @pe-ro , we discussed this at the ODD-Meeting last week. There, we also pointed out to remove it here.

pe-ro (approved):
Thanks for following up on this.  Looks good to me.

riedde:
@rettinghaus Could you have a final eye on this? Thanks!

rettinghaus (approved):

source/modules/MEI.shared.xml
pe-ro: Looks like my comment didn't actually show up.

Default values are practical in local, project-specific customizations because they can speed up markup creation. However, they're not used in MEI because they create a kind of "hidden knowledge" that's not apparent to the user.  One can be caught off guard when a value shows up when a document is processed that was not specified in the original markup.  Markup added by a parser is generally a bad idea.

PR #1728, opened 2025-11-28
schema: add exemplums for `@intm`

rettinghaus:
Trying to extend the exemplums we currently have in the schema. First adding three examples back from the Guidelines to `@intm`. 
Additionally use markup for the citation and change the HTML generation for `tei:bibl`.

### DIFF
--- a/source/modules/MEI.analytical.xml
+++ b/source/modules/MEI.analytical.xml
@@ -268,9 +268,9 @@
               <staff>
                 <layer>
                   <note dur="4" oct="4" pname="c" />
-                  <note dur="4" oct="4" pname="d" intm="1.1" />
-                  <note dur="4" oct="5" pname="d" intm="7.9" />
-                  <note dur="4" oct="5" pname="c" intm="-2.334" />
+                  <note dur="4" oct="4" pname="d" intm="2hs" />
+                  <note dur="4" oct="5" pname="d" intm="12hs" />
+                  <note dur="4" oct="5" pname="c" intm="-2.334hs" />
                 </layer>
               </staff>
             </measure>
--- a/source/modules/MEI.analytical.xml
+++ b/source/modules/MEI.analytical.xml
@@ -229,6 +229,53 @@
         <datatype>
           <rng:ref name="data.INTERVAL.MELODIC"/>
         </datatype>
+        <exemplum xml:lang="en">
+          <p>The following example shows interval relationships indicated by the Parsons Code:</p>
+          <egXML xmlns="http://www.tei-c.org/ns/Examples">
+            <measure n="1">
+              <staff n="1">
+                <layer>
+                  <note dur="4" oct="4" pname="c" />
+                  <note dur="4" oct="4" pname="d" intm="u" />
+                  <note dur="4" oct="4" pname="e" intm="u" />
+                  <note dur="4" oct="4" pname="f" intm="u" />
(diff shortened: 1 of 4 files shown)


martha-thomae:
Hi @rettinghaus! Thank you for the examples. At first glance, it looks alright to me. Would you mind building the guidelines and sharing screenshots of how these changes would look in the website? Both for the changes in the guidelines (`<bibl>` in the analysis chapter) and the changes in the specs (the examples included). 

This will make it easier to approve the PR by next meeting (March 27th) and it saves reviewers a bit of time (not having them building things for each PR). Thank you, Klaus :)

bwbohl (dismissed):

pe-ro (changes requested):
To be valid, the example using half steps should include a trailing "hs"  -- 
```
<exemplum xml:lang="en">
  <egXML xmlns="http://www.tei-c.org/ns/Examples">
    <measure>
      <staff>
        <layer>
          <note dur="4" oct="4" pname="c"/>
          <note dur="4" oct="4" pname="d" intm="1hs"/>
          <note dur="4" oct="5" pname="d" intm="7hs"/>
          <note dur="4" oct="5" pname="c" intm="-2.334hs"/>
        </layer>
      </staff>
    </measure>
  </egXML>
</exemplum>
```
Since integer values are more common than decimal ones, I changed the first two values in this example.  There should also be an explanation in the text that decimal values are permitted to accommodate microtonal or otherwise "detuned" music.

rettinghaus:
Bildschirmfoto 2026-03-07 um 22 30 08

bwbohl (commented):
Do we really need the `xml:lang` attributes on the exemplum elements? And what about the `<bibl>`?

rettinghaus:
In the exemplum is an English explanation, so I guess it's better to have it.

What about the bibl?

bwbohl:
> What about the bibl?

 I totally agree that we should improve on semantic markup in our modules and docs. Nevertheless, this commit is about `exemplums` in the schema, aka modules. And at the moment, I’d be happy with #1626 getting out of the way, first. Moreover, I think we should systematically trace missing `bibl` elements – there just are none at the moment!

bwbohl:
> Since integer values are more common than decimal ones, I changed the first two values in this example. There should also be an explanation in the text that decimal values are permitted to accommodate microtonal or otherwise "detuned" music.

@pe-ro which text ar you referring to?

pe-ro:
> which text ar[e] you referring to?

The Guidelines, of course. 😁

bwbohl:
@rettinghaus same as in #1626, please let’s do another iteration of the sources for catching all bibliographic references ;-) We might even cherry-pick c1f5e7c84837e2efa5665fc61aeef2805f398954 to a new branch

bwbohl:
> @rettinghaus same as in #1626, please let’s do another iteration of the sources for catching all bibliographic references ;-) We might even cherry-pick [c1f5e7c](https://github.com/music-encoding/music-encoding/commit/c1f5e7c84837e2efa5665fc61aeef2805f398954) to a new branch

@rettinghaus, are you ok with proceeding as described above?

rettinghaus:
@bwbohl sure

bwbohl (approved):
I moved the commits concerning bibls to #1785

pe-ro (approved):

PR #1734, opened 2025-12-03
schema: develop text phrase like

pe-ro:
This request lays the groundwork for future requests by creating new textPhraseLike classes that are more basic/elemental than textPhraseLike.limited; i.e. textPhraseLike.bare and textPhraseLike.basic, and by modifying the textPhraseLike.limited class.

The textPhraseLike.bare class only allows the following elements: lb, ptr, ref, rend, stack, seg, and symbol.

The textPhraseLike.basic class allows: abbr, address, annot, bibl, biblStruct, date, dimensions, expan, extent, fig, identifier, name, num, term, and title, plus the members of textPhraseLike.bare.

The textPhraseLike.limited class allows: bloc, corpName, country, depth, dim, dimensions, district, extent, geogFeat, geogName, height, periodName, persName, postBox, postCode, q, region, relation, relationList, repository, settlement,  street, styleName, and width, plus the members of textPhraseLike.basic.

Just to refresh the collective memory, the textPhraseLike class allows all the members of textPhraseLike.limited, plus the `<pb>` element. It has not been modified.

The new class hierarchy can be used to specify what's allowed in text with more detail than before, for example, in metadata elements.

PR #1736, opened 2025-12-05
guidelines: add missing documentation of physMedium (issue 1222)

### LINKED ISSUE #29: Missing documentation: <physMedium>
I quickly created a new issue so that we don't get confused and the discussion runs separately. This is @riedde s first draft: 

> Within physMedium the material of which an object is composed (physical materials, such as ink, paper, etc.) can be recorded. This can be done by using a single physMedium element containing the whole description or by using multiple physMedium elements for recording different materials separately. This container element may contain textual content and semantic markup. Because of the rich model class model.physDescPart the container can also contain highly structured code.

`<physDesc xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.1-dev+anyStart">
    <physMedium>Thick yellow paper.</physMedium>
    <physMedium>Extra paper <height unit="mm">35</height>x<width unit="mm">150</width> glued on the source.</physMedium>
    <physMedium>Black ink is used for the notes, red ink for the dynamics</physMedium>
</physDesc>`

**by using multiple physMedium elements for recording different materials separately.** --> I would not recommend this, on the one hand I wonder whether a source can have more than one physMedium. Is it one object carrier or several (more of a philosophical question perhaps). More importantly, it seems to me that the same effect you suggest can be achieved within a physmedium through paragraphs. On the other hand, physMedium as an element with its history (no equivalent in TEI and introduced because the object description was not as mature earlier in MEI 3) is a remnant and perhaps one should rather suggest using the numerous other elements such as `<supportDesc>`, `<extent>`, `<dimensions>` etc. For me it would be important to clarify how physMedium relates to other elements such as `<supportDesc>`, `<handList>` or `<bindingDesc>`. And then whether these elements might be linked to each other

### LINKED ISSUE #1222: Add description of `physMedium` in GL Chapter 3.7
`physMedium` missing here? To be added in next release @riedde ?

_Originally posted by @musicEnfanthen in https://github.com/music-encoding/music-encoding/pull/1095#discussion_r1229371627_

riedde:
Fixes #1222

This PR introduces documentation for `<physMedium>`. It has been adressed in https://github.com/music-encoding/metadata-ig/issues/29. The text has been revised by @pe-ro

closes https://github.com/music-encoding/metadata-ig/issues/29

### DIFF
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -1543,6 +1543,17 @@
                         </figure>
                      </p>
                      <p><specList>
+                           <specDesc key="physMedium"/>
+                     </specList></p>
+                     <p><gi>physMedium</gi> is intended to capture technical metadata in a narrative form. One should think of this element as a container for notes on or a synopsis of the physical characteristics of an item rather than a place to capture the characteristics themselves. It can used instead of or in addition to the structured sub-elements of <gi>physDesc</gi>. For example,</p>
+                     <p>
+                        <figure>
+                           <head>Example of <gi scheme="MEI">physMedium</gi></head>
+                           <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/header/header-sample097.txt" parse="text"/></egXML>
+                        </figure>
+                     </p>
+                     <p>It is possible to use more than one <gi>physMedium</gi> elements.</p>
+                     <p><specList>
                         <specDesc key="scoreFormat"/>
                      </specList></p>
                      <p>The <gi scheme="MEI">scoreFormat</gi> element is a form of classification. This element is part of <gi scheme="MEI">physDesc</gi> because within the MARC21 standard, the format of the music (score, piano score, etc.) is defined as a physical property.</p>
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -1545,14 +1545,14 @@
                      <p><specList>
                            <specDesc key="physMedium"/>
                      </specList></p>
-                     <p><gi>physMedium</gi> is intended to capture technical metadata in a narrative form. One should think of this element as a container for notes on or a synopsis of the physical characteristics of an item rather than a place to capture the characteristics themselves. It can used instead of or in addition to the structured sub-elements of <gi>physDesc</gi>. For example,</p>
+                     <p><gi scheme="MEI">physMedium</gi> is intended to capture technical metadata in a narrative form. One should think of this element as a container for notes on or a synopsis of the physical characteristics of an item rather than a place to capture the characteristics themselves. It can used instead of or in addition to the structured sub-elements of <gi scheme="MEI">physDesc</gi>. For example,</p>
                      <p>
                         <figure>
                            <head>Example of <gi scheme="MEI">physMedium</gi></head>
                            <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/header/header-sample097.txt" parse="text"/></egXML>
                         </figure>
                      </p>
-                     <p>It is possible to use more than one <gi>physMedium</gi> elements.</p>
+                     <p>It is possible to use more than one <gi scheme="MEI">physMedium</gi> elements.</p>
                      <p><specList>
                         <specDesc key="scoreFormat"/>
                      </specList></p>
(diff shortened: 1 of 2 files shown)


source/docs/03-metadata.xml
```diff
@@ -1542,6 +1542,17 @@
                            <egXML xmlns="http://www.tei-c.org/ns/Examples" rend="code" xml:space="preserve"><xi:include xmlns:xi="http://www.w3.org/2001/XInclude" href="../examples/header/header-sample084.txt" parse="text"/></egXML>
                         </figure>
                      </p>
+                     <p><specList>
+                           <specDesc key="physMedium"/>
+                     </specList></p>
+                     <p><gi scheme="MEI">physMedium</gi> is intended to capture technical metadata in a narrative form. One should think of this element as a container for notes on or a synopsis of the physical characteristics of an item rather than a place to capture the characteristics themselves. It can used instead of or in addition to the structured sub-elements of <gi scheme="MEI">physDesc</gi>. For example,</p>
```
riedde, line 1548: @pe-ro Isn't this a bit far from the element definition, which is "Records the physical materials used in the source, such as ink and paper."?

pe-ro: `<physMedium>` can be used to summarize the physical materials used by the resource as outlined by the text above. Alternatively, it can be used to provide (more-or-less) plain text values that describe the materials --
```
<physMedium>ink</physMedium>
<physMedium>paper</physMedium>
```
These values can be "controlled" by using `@auth` and/or `@auth.uri` --
```
<physMedium auth.uri="http://id.loc.gov/vocabulary/mmaterial/ink">iron gall ink</physMedium>
<physMedium auth.uri="http://id.loc.gov/vocabulary/mmaterial/paper">paper</physMedium>
```
If one wants to describe the support materials independently, then the following can be used --
```
<physMedium auth.uri="http://id.loc.gov/vocabulary/mmaterial/ink">ink</physMedium>
<!-- Or perhaps -->
<physMedium auth.uri="http://id.loc.gov/vocabulary/mmaterial/ink">ink and paper</physMedium>
<supportDesc>
    <support auth.uri="http://id.loc.gov/vocabulary/mmaterial/paper">paper</physMedium>
  </supportDesc>
```
I recognize your uncertainty, so we should rethink the text regarding `<physMedium>`. I suppose we could rename `<physMedium>` to `<material>` or `<surfaceMaterial>` to make the distinction between "layers" clearer, but then there'd be no place to record "notes" on the materials used by the resource in a generic way.   I know it's not very good at the moment, but I still think there's a way to document the dual nature of `<physMedium>` without changing the schema.

pe-ro:
The problem I described above is caused by try to merge traditional MARC-like practice with TEI's `<supportDesc>`.  So, I propose to remove `<supportDesc>`  and revise `<physMedium>` (in RNG speak) --
```
<define name="mei_physMedium">
  <element name="physMedium">
    <a:documentation xmlns:a="http://relaxng.org/ns/compatibility/annotations/1.0"
      >(physical medium) Records information about the physical materials used by a
      resource.</a:documentation>
    <zeroOrMore>
      <ref name="mei_head"/>
    </zeroOrMore>
    <zeroOrMore>
      <choice>
        <ref name="mei_baseMat"/>
        <ref name="mei_collation"/>
        <ref name="mei_condition"/>
        <ref name="mei_foliation"/>
        <ref name="mei_p"/>
        <ref name="mei_surfaceMat"/>
      </choice>
    </zeroOrMore>
    <ref name="mei_att.bibl.attributes"/>
    <empty/>
  </element>
</define>
```
One can choose a generic, narrative description using the (head*, p*) part of the model or a more structured one incorporating the other available elements like `<baseMat>`, `<collation>`, `<condition>`, `<foliation>`, and `<surfaceMat>`, some of which were formerly available in `<supportDesc>`. Each of the sub-elements uses the struc-unstrucContent macro that (with the merging of #1734) allows either the (head*, p*) or the textPhraseLike.basic model.  The `<baseMat>` (base material, called support in TEI) and `<surfaceMat>` elements also have an open list of suggested values.

Example markup:
```
<physMedium xmlns="http://www.music-encoding.org/ns/mei" meiversion="5.1+anyStart">
  <head>Physical Media</head>
  <p>Blah, blah, blah.</p>
  <foliation><p>The whole set of <name>textPhraseLike.limited</name> elements can be used in a
      paragraph.</p></foliation>
  <foliation>Alternatively, if the element contains text, then only the
    textPhraseLike.basic elements may <rend rend="bold">appear</rend>.</foliation>
  <baseMat auth.uri="">vellum</baseMat>
  <surfaceMat auth.uri="">tempura</surfaceMat>
  <condition/>
</physMedium>
```
In Oxygen, suggested values for `<baseMat>` and `<surfaceMat>` are presented when the trailing `>` of the opening tag is typed:

image

riedde:
In today's meeting of @music-encoding/ig-metadata we discussed this PR. After thinking about for a while we decided to close this PR to come up with a new proposal.

PR #1737, opened 2025-12-05
schema: restructure the struc-unstrucContent macro

pe-ro:
Restricting the struc-unstrucContent macro will allow fewer sub-elements in those places where the macro is used; that is, mostly in metadata elements in the header.  If more markup granularity is desired, one can use the `<p>` element that's part of the structured part of the macro.

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -80,7 +80,7 @@
     </content>
   </macroSpec>
   <macroSpec ident="macro.struc-unstrucContent" module="MEI.shared" type="pe">
-    <desc xml:lang="en">Provides a choice between structured and unstructured/mixed content.</desc>
+    <desc xml:lang="en">Provides a choice between structured and unstructured text content.</desc>
     <content>
       <rng:choice>
         <rng:group>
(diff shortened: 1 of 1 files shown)


pe-ro:
Does this PR fail because #1734, which creates creates model.textPhraseLike.basic, hasn't been merged yet? If so, #1734 will need to be merged first.

riedde:
@pe-ro yes, I guess it fails because the referenced model class doesn't exist and could not be validated. If the target of this PR would be the branch `develop-textPhraseLike`instead of `develop` it should work. So #1734 bust be merged before this PR or the branch has to be changed.

riedde (dismissed):
This is an addition to #1734 which has to be merged before this one.

pe-ro:
Needs work, converted to draft

pe-ro:
@riedde , I changed the target to develop-textPhraseLike. Is the next step to merge this PR into #1734? 

Since I'm a git newby, I'm reluctant to hit the "Merge pull request" button without being sure of an un-do path.

bwbohl (approved):

riedde (approved):

pe-ro:
Thanks, @riedde , for approving.  But I don't think this macro is useful anymore, so revising it isn't helpful.  I'm closing this pull request.  Apologies for the thrashing.

bwbohl:
@pe-ro should we also delete the corresponding branch [develop-macro.struc-unstrucContent](https://github.com/music-encoding/music-encoding/tree/develop-macro.struc-unstrucContent) or will you put it to another use?

pe-ro:
@bwbohl, just delete it please.

PR #1743, opened 2026-01-09
schema: update `<annot>` schema and guidelines 

DILewis:
Following on from the Annote project (Oxford & RISM Digital) and discussions with the community at MEC and DLfM, here's the first part of a proposal to update `<annot>`, primarily to make annotations of musical notation easier to make and view in engraving tools such as Verovio (which has a prototype implementation of this already).

Changes are:

- Add `@func` to the `<annot>` element and specify `display` as a possible value. This is intended to allow authors to indicate that an annotation can be visualised by engraving software.
- Changes to Guidelines § 9.2.13:
    * Clarify and tighten language, including adding cross-reference for other uses in the guidelines
    * Add explanation of `@func`
    * Add examples targetting of `annot`, particularly with control element attributes

A second part (to come) is a section within chapter 14 of the Guidelines outlining the use of Web Annotations for MEI. This is an independent part, and will not affect the schema.

Note: The proposal to use `@func` came from discussion with the community at MEC 2025 (Verovio currently uses `@type="score"`, since this requires no schema adjustments). We note that MEI currently has both `@func` and `@function`, and possibly several other attributes that could be used in a similar context. 

Some context is given in Kevin R Page, David Lewis, and Laurent Pugin. 2025. Annotation of digital music notation documents: Surveying needs for a generalised implementation. In Proceedings of the 12th International Conference on Digital Libraries for Musicology (DLfM '25). Association for Computing Machinery, New York, NY, USA, 81–89. [https://doi.org/10.1145/3748336.3748346](https://doi.org/10.1145/3748336.3748346)


pe-ro:
Looks good to me.

lpugin (approved):

bwbohl (approved):

PR #1760, opened 2026-03-26
guidelines: add example for `@data` and `@decls`

### LINKED ISSUE #1412: Chapter 03 (Metadata): Add examples for Metadata and Data association
Dear Metadata IG (@margrethesb , @KristinaRichts , @axgeertinger et al.),

would it be possible (and feasible) to add some small examples to the section `3.1.1.7.1. Associating Metadata and Data` in the Metadata chapter (#82) for the use of `@data` and `@decls` to see the mechanism working?

Background: As far as I understand, `decls` is also a pointing/linking technique, that is available on `<ptr>` and `<ref>`. The attribute will be mentioned in chapter 13 (Linking, music-encoding/guidelines#92) but I'd like to avoid bigger duplications. 

(Of course, if you think it would be better to have a more detailed explanation and examples in the Linking chapter where you can link to from the MD chapter, or if you don't mind duplication, please let me know.)

Thanks a lot in advance.

riedde:
I found this long outstanding issue. Here just an example is added. If there are better examples just let me know to adjust this.

Fixes #1412

### DIFF
--- a/source/docs/03-metadata.xml
+++ b/source/docs/03-metadata.xml
@@ -260,7 +260,23 @@
                <div xml:id="headerAssociatingMetadataAndData" type="div3">
                   <head>Associating Metadata and Data</head>
                   <p>In the MEI header, the <att>data</att> attribute may be used to associate metadata with related notational elements.</p>
+                  <p>
+                     <specList>
+                        <specDesc key="att.pointing" atts="data"/>
+                     </specList>
+                  </p>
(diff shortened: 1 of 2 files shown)


musicEnfanthen (approved):

bwbohl (commented):
Many thanks for adding it, I think we might come up with some more examples later.

bwbohl (approved):

PR #1761, opened 2026-03-26
guidelines: add basic documentation for `<physMedium>`

### LINKED ISSUE #29: Missing documentation: <physMedium>

### LINKED ISSUE #1222: Add description of `physMedium` in GL Chapter 3.7

riedde:
I revised the work done in previous steps consulting the discussion on #1222. This may not be perfect, but there is an ondoing task by @music-encoding/ig-metadata on a revision of the `<physMedium>` element. However, since this task will not be completed by the release of MEI 6, it is important to provide a simple documentation for MEI 6.
I'm open to constructive criticism. My stance on this issue: It is better to have simple documentation in Verison 6 than none at all.

fixes #1222
fixes https://github.com/music-encoding/metadata-ig/issues/29

### DIFF
--- a/source/examples/header/header-sample100.txt
+++ b/source/examples/header/header-sample100.txt
@@ -5,7 +5,7 @@
   <perfDuration isodur="PT12H30M5S">1:33:02</perfDuration>
   <soundChan num="2">stereo</soundChan>
   <supportDesc>polycarbonate plastic</supportDesc>
-  <physMedium>1.2-millimetre (0.047 in) thick, polycarbonate plastic, and weighs 14–33
+  <physMedium>1.2-millimeter (0.047 in) thick, polycarbonate plastic, and weighs 14–33
     grams. A thin layer of aluminum or, more rarely, gold is applied to the surface,
     making it reflective. The metal is protected by a film of lacquer normally spin coated
     directly on the reflective layer. The label is printed on the lacquer layer, usually
--- a/source/examples/header/header-sample100.txt
+++ b/source/examples/header/header-sample100.txt
@@ -1,0 +1,13 @@
+<physDesc>
+  <carrierForm>sound disc</carrierForm>
+  <captureMode>analog</captureMode>
+  <dimensions unit="in" type="diameter">4.7</dimensions>
+  <perfDuration isodur="PT12H30M5S">1:33:02</perfDuration>
+  <soundChan num="2">stereo</soundChan>
+  <supportDesc>polycarbonate plastic</supportDesc>
+  <physMedium>1.2-millimetre (0.047 in) thick, polycarbonate plastic, and weighs 14–33
+    grams. A thin layer of aluminum or, more rarely, gold is applied to the surface,
+    making it reflective. The metal is protected by a film of lacquer normally spin coated
+    directly on the reflective layer. The label is printed on the lacquer layer, usually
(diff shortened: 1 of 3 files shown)


source/examples/header/header-sample100.txt
```diff
@@ -0,0 +1,13 @@
+<physDesc>
+  <carrierForm>sound disc</carrierForm>
+  <captureMode>analog</captureMode>
+  <dimensions unit="in" type="diameter">4.7</dimensions>
+  <perfDuration isodur="PT12H30M5S">1:33:02</perfDuration>
```
musicEnfanthen: isodur and displayed duration seem out of sync?

source/examples/header/header-sample100.txt
```diff
@@ -0,0 +1,13 @@
...
+  <soundChan num="2">stereo</soundChan>
+  <supportDesc>polycarbonate plastic</supportDesc>
+  <physMedium>1.2-millimetre (0.047 in) thick, polycarbonate plastic, and weighs 14–33
```
musicEnfanthen:
```suggestion
  <physMedium>1.2-millimeter (0.047 in) thick, polycarbonate plastic, and weighs 14–33
```

source/examples/header/header-sample100.txt
riedde:
```suggestion
  <perfDuration isodur="PT1H33M2S">1:33:02</perfDuration>
```
@musicEnfanthen Yes, I'm not sure where this example is coming from, but with this suggestion it should be right, souldn't it?

musicEnfanthen (approved):

bwbohl (approved):
let’s please wait for #1626 to be merged before this one

PR #1764, opened 2026-03-27
MEI.shared: fix – add `rng`-prefix to rng elements remove `mei_` prefix from references

bwbohl:
The build failed due to a malformed syntax this PR fixes that


riedde (approved):
Thank you @bwbohl for this fix. The rng-schema compiles again!

PR #1766, opened 2026-03-27
utils: make data types linkable

### LINKED ISSUE #1765: guidelines: links to definiton and chapter reference for datatype missing
In the Guidelines we can link from the text to definitions of element, attribute classes and so on, but not to datatypes. In the other way round, the datatype definitions have no chapter references. Even if we don't use this functionallity very often, it's will be a nice feature we can benefit from.

riedde:
This PR makes data types linkable in the html-guidelines
- adding chapter references to the data type definitions
- adding links to the text (pointing to the definitions)
- adding markup to the guidelines text where missing

Note: The process of generating the PDF-Guidelines has not been adjusted yet.

Fixes #1765


source/docs/14-integration.xml
```diff
@@ -205,8 +205,8 @@
                   <p>The <gi scheme="MEI">instrDef</gi> element can be used to record MIDI
                      instrument names or numbers using the <att>midi.instrname</att> and
                         <att>midi.instrnum</att> attributes. The <att>midi.instrname</att> attribute
-                     must contain an instrument name from the list provided by the data.MIDINAMES
-                     data type. By default, data.MIDINAMES contains General MIDI Instrument
+                     must contain an instrument name from the list provided by the <ident type="macro">data.MIDINAMES</ident>
+                        data type. By default, <ident type="macro">data.MIDINAMES</ident> contains General MIDI Instrument
```
musicEnfanthen: whitespace needed here?

riedde: Is there to much whitespace?

musicEnfanthen: Yes, it looks like added whitespace in front of "data type." ?

riedde: You're right. I edited Line 208 and Oxygen as added an indent to the following line. I'll fix that.

utils/guidelines_xslt/odd2html/guidelines.xsl
```diff
@@ -161,6 +161,25 @@
...
+        </xd:desc>
+    </xd:doc>
+    <xsl:template match="tei:ident[@type = 'macro']" mode="guidelines">
+        <xsl:variable name="text" select="string(text())" as="xs:string"/>
+        <xsl:choose>
+            <xsl:when test="$text = //tei:macroSpec/@ident">
+                <a class="{tools:getLinkClasses($text)}" href="#{$text}"><xsl:value-of select="$text"/></a>
+            </xsl:when>
+            <xsl:otherwise>
+                <xsl:message terminate="no" select="'ERROR: Unable to identify class ' || $text || ' from tei:ident element. No link created.'"/>
```
musicEnfanthen:
```suggestion
                <xsl:message terminate="no" select="'ERROR: Unable to identify datatype' || $text || ' from tei:ident element. No link created.'"/>
```

utils/guidelines_xslt/odd2html/guidelines.xsl
```diff
@@ -161,6 +161,25 @@
...
     
+    <xd:doc>
+        <xd:desc>
+            <xd:p>Datatype references</xd:p>
+        </xd:desc>
+    </xd:doc>
+    <xsl:template match="tei:ident[@type = 'macro']" mode="guidelines">
+        <xsl:variable name="text" select="string(text())" as="xs:string"/>
+        <xsl:choose>
+            <xsl:when test="$text = //tei:macroSpec/@ident">
```
musicEnfanthen, line 172: For now, this captures both datatypes and macros, i.e., macroSpec type="dt" and type="pe", right? 
Since this syntax will probably undergo revision with the switch to pureOdd and be changed to dataSpec and macroSpec: would we need different idents for actual macros vs. datatypes? (cf. #1501)

utils/guidelines_xslt/odd2html/specs/dataTypeSpecs.xsl
```diff
@@ -60,6 +60,7 @@
     <xsl:function name="tools:getDataTypeSpecPage" as="node()">
         <xsl:param name="data.type" as="node()"/>
         
+        <xsl:variable name="refs" select="tools:getReferencesFacet(tools:getReferencingChapters($data.type/@ident, 'modelClass'))" as="node()?"/>
```
musicEnfanthen:
```suggestion
        <xsl:variable name="refs" select="tools:getReferencesFacet(tools:getReferencingChapters($data.type/@ident, 'dataType'))" as="node()?"/>
```

utils/guidelines_xslt/odd2html/guidelines.xsl
musicEnfanthen: Or maybe go with a dual approach like in comparison.xsl for now:

```xml
<xd:doc>
        <xd:desc>
            <xd:p>Datatype references</xd:p>
        </xd:desc>
    </xd:doc>
    <xsl:template match="tei:ident[@type = ('macro','data')]" mode="guidelines">
        <xsl:variable name="text" select="string(text())" as="xs:string"/>
        <xsl:variable name="datatype.idents" select="(//tei:macroSpec/@ident, //tei:dataSpec/@ident)" as="attribute(ident)*"/>
        <xsl:choose>
            <xsl:when test="$text = $datatype.idents">
                <a class="{tools:getLinkClasses($text)}" href="#{$text}"><xsl:value-of select="$text"/></a>
            </xsl:when>
```

musicEnfanthen:
Thanks a lot @riedde , that's great. Works nicely. Some comment about making this "future-proof" above.

utils/guidelines_xslt/odd2html/guidelines.xsl
rettinghaus: @musicEnfanthen This was done in #991.

rettinghaus: This line alone would be easy to ackle, but the rest gives you the creeps. Although it is not feasable to support both, because after the switch, the old one will be out of use. 
I've opened #1782 to straighten this up.

PR #1768, opened 2026-03-27
add macro, fixing issue 919

### LINKED ISSUE #919: add macro `<exhibHist>`, `<treatHist>` and `<treatSched>`after `<provenance>`, et al.
As theses are now modelled all the same way we should confine the definition of the content model to a macro definition and just reference that from the individual element definitions.

riedde:
Adds a macro (content model) for elements on historical account, cause thir content model is identical. 

fixes #919

### DIFF
--- a/source/modules/MEI.header.xml
+++ b/source/modules/MEI.header.xml
@@ -420,12 +420,7 @@
       <memberOf key="att.lang"/>
     </classes>
     <content>
-      <rng:oneOrMore>
-        <rng:choice>
-          <rng:ref name="macro.histAccount"/>
-          <rng:text/>
-        </rng:choice>
-      </rng:oneOrMore>
+      <rng:ref name="macro.histAccount"/>
     </content>
     <remarks xml:lang="en">
       <p>The model of this element is based on the <ref target="https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-acquisition.html">acquisition</ref> element of the Text Encoding Initiative (TEI).</p>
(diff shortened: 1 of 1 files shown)


pe-ro:
Looks good to me.

pe-ro (dismissed):
Looks good.

source/modules/MEI.header.xml
```diff
@@ -396,25 +420,12 @@
...
-            <rng:ref name="model.textPhraseLike.limited"/>
-          </rng:choice>
-        </rng:zeroOrMore>
-      </rng:choice>
+      <rng:oneOrMore>
+        <rng:choice>
+          <rng:ref name="macro.histAccount"/>
+          <rng:text/>
+        </rng:choice>
+      </rng:oneOrMore>
```
musicEnfanthen: Is it intended that this is now an (obligatory) choice between the macro and text?

riedde: No, thanks for this hint. That's a copy-paste error. The content should only be the macro, because the macro should substitute the encoding of the content, but not add something.

riedde: it was redundant. The text is already part of the macro and shouldn't appear here a second time. Already fixed in 7d379dfc9d2ae7f00eb7b3fd578228678f29c598

bwbohl (approved):

bwbohl:
@riedde thanks for tackling this, @pe-ro could you please "add your review" – as I understand your comments, you do approve, right?

pe-ro (approved):
This is fine; however, the elements that utilize the macro.histAccount content model are part of the changes I've proposed.  If implemented, the new model would be --
```
<choice>
  <zeroOrMore>
    <choice>
      <ref name="mei_eventList"/>
      <ref name="mei_macro.para.basic"/>
    </choice>
  </zeroOrMore>
  <zeroOrMore>
    <choice>
      <text/>
      <ref name="mei_model.textPhraseLike.basic"/>
    </choice>
  </zeroOrMore>
</choice>
```
That is, `<head>` would be removed, and the model changed to enforce a choice between structured content (using eventList and p with basic textual content) or basic textual content.

In any case, the PR is okay for the present time.

bwbohl (approved):

PR #1770, opened 2026-04-01
schema: revise model.titleLike

pe-ro:
Revise model.titleLike; replace 'model.titleLike' with 'title' in monogr and series


riedde (commented):
As I see, these changes don't change the schema, because the `model.titleLike` contains `<title>` only. I'm not sure about the latest status of the discussion, but I remember that we talked about using more model and attribute classes instead of calling elements/attributes directly. Is there a new guideline @bwbohl ? If yes, I'll approve it, but at the moment I don't see the benefit for removing the model class here and keeping it in many other places.

pe-ro:
@riedde, in this PR, `<titlePart>` is made a member of model.titleLike PR so that both `<title>` and `<titlePart>` will be included when model.titleLike is invoked, for exmaple in textPhraseLike.basic.  Conversely, only `<title>` is desired in `<monogr>` and `<series>`, so it is invoked directly, without referencing model.titleLike.

riedde:
My GitHub preview was acting a bit strange, but I have an overview now. It’s perfectly fine to make `titlePart` an member of `model.titleLike`. Is there an issue or an ongoing discussion that this PR refers to? If we link to that, it provides more context—which is helpful for review—about the problem and the reasons why a change is necessary and should be made.

pe-ro:
@riedde, there's no issue or discussion currently. Since the change is relatively lightweight, I didn't see the need to create an issue. The change does two things: 

- makes `<title>` and `<titlePart>` co-equals in the model.titleLike class, meaning they can occur independently of each other in the same context; e.g., model.textPhraseLike.basic or `<titlePage>`;
- restricts `<monogr>` and `<series>` to use only `<title>`.

This is related to #1734.

bwbohl (commented):
While at it:
couldn’t we also simplify the content model of `title`:

```xml
<content>
<zeroOrMore>
<choice>
<rng:text/>
<ref name="macro.titlePart"/>
<ref name="model.editLike"/>
<ref name="model.transcriptionLike"/>
<ref name="titlePart"/>
</choice>
</zeroOrMore>
</content>
```

`<ref name="titlePart"/>` should be covered by `<ref name="macro.titlePart"/>`, right?

pe-ro:
Nope. 

`<ref name="macro.titlePart"/>` brings in the children of `<titlePart>`, while `<ref name="titlePart"/>` includes the titlePart element itself.  In other words, `<title>` may contain `<titlePart>` and the children of `<titlePart>`.

Similarly, `<titlePart>` may contain `<titlePart>` and the children of `<titlePart>`.

pe-ro:
`<titlePart>` could be added as a member of `macro.titlePart`, but I think that complicates the model of `<titlePart>` instead of simplifying it.  As it stands, the inclusion of both `<titlePart>` and its children is clear. Adding `<titlePart>` as a member of `macro.titlePart` obscures instead of clarifies.

bwbohl (approved):

riedde (approved):
Now I got it.

PR #1771, opened 2026-04-01
schema: develop model.locusLike

pe-ro:
This PR creates a new locusLike model of which locus and locusGrp are members. It also makes model.locusLike a member of model.textPhraseLike.

The commits related to the material element are superfluous.  These modifications will be proposed later.


source/modules/MEI.shared.xml
```diff
@@ -6984,7 +7050,6 @@
       <memberOf key="att.responsibility"/>
       <memberOf key="model.eventPart"/>
       <memberOf key="model.nameLike"/>
-      <memberOf key="model.textPhraseLike.basic"/>
```
riedde, line 6987: This fix should ot part of a PR for developing `model.locusLike`.

source/modules/MEI.shared.xml
```diff
@@ -4505,7 +4513,7 @@
   <classSpec ident="model.nameLike" module="MEI.shared" type="model">
     <desc xml:lang="en">Groups elements that contain names.</desc>
     <classes>
-      <memberOf key="model.textPhraseLike.limited"/>
```
riedde, line 4508: This fix should ot part of a PR for developing `model.locusLike`.

riedde (changes requested):
I like the idea of organizing `locus`and `locusGrp` as  `model.locusLike`. But in this PR I miss the `classSpec` that defines `model.locusLike`. Further I don't see where the new model is applyed to. Before it was part of `model.msInline`, but will `locus`and `locusGrp` allowed anywhere after applying the changes proposed here?
Moving these elements from the module `MEI.msDesc` to `MEI.shared` is fine!

source/modules/MEI.shared.xml
```diff
@@ -4480,6 +4480,14 @@
       <memberOf key="model.textPhraseLike.bare"/>
     </classes>
   </classSpec>
+  <classSpec ident="model.locusLike" module="MEI.shared" type="model">
+    <desc xml:lang="en">Groups elements that provide locational information for materials within a
+      resource's carrier; e.g., value may represent band (LP), page/page range (printed material),
+      frame number/number range (microform), start/start+end (audio/video), etc.</desc>
+    <classes>
+      <memberOf key="model.textPhraseLike.basic"/>
```
pe-ro, line 4488: Thie PR makes model.locusLIke a member of textPhraseLike.basic, which occurs in may places throughout the schema.  In other words, `<locus>` and `<locusGrp>` may now appear within phrase-level text; that is, nearly everywhere text is allowed.

source/modules/MEI.shared.xml
pe-ro: Is the intended comment -- "This fix should[ not be] part of a PR for developing model.locusLike."?

Perhaps you're right, but making nameLike a member of textPhraseLike.basic was overlooked in an earlier (already approved) PR, so I added it here.  Feel free to create a new pull request.

source/modules/MEI.shared.xml
pe-ro: See my answer above.

source/modules/MEI.shared.xml
bwbohl: Should we remedy that?

source/modules/MEI.shared.xml
riedde: I missed that. Now it's clear to me!

source/modules/MEI.shared.xml
riedde: @pe-ro Sorry for this typo. yes I meant "not be"
@bwbohl Since it is a seperate topic, a new PR might be more clean and easier to follow the development.

PR #1772, opened 2026-04-02
schema: deprecate `changeDesc` and `figDesc` elements in favor of `desc`

pe-ro:
The `<desc>` element is intended to provide metadata and/or comment on elements within the markup.  It shouldn't be used to capture the document text itself.  Therefore, `<desc>` can have a simpler model than is currently employed; i.e., it can be use the textPhraseLike.basic model and the editLike and transcriptionLike models can be dropped.

`<desc>` can also take the place of `<changeDesc>` and `<figDesc>`.


source/docs/12-facsimilesperformances.xml
```diff
@@ -148,7 +148,7 @@
...
                   </p>
-                  <p>A <gi scheme="MEI">zone</gi> element may contain <gi scheme="MEI">figDesc</gi> or <gi scheme="MEI">graphic</gi> elements that provide detailed descriptive information about the zone and additional images, e.g., at a different/higher resolution, of the rectangle defined by the zone. The data objects contained within the zone may also be specified through the use of the <att>data</att> attribute, which contains ID references to one more elements in the content tree of the MEI file, such as a <gi scheme="MEI">note</gi>, <gi scheme="MEI">measure</gi>, etc.</p>
+                  <p>A <gi scheme="MEI">zone</gi> element may contain <gi scheme="MEI">desc</gi> or <gi scheme="MEI">graphic</gi> elements that provide detailed descriptive information about the zone and additional images, e.g., at a different/higher resolution, of the rectangle defined by the zone. The data objects contained within the zone may also be specified through the use of the <att>data</att> attribute, which contains ID references to one more elements in the content tree of the MEI file, such as a <gi scheme="MEI">note</gi>, <gi scheme="MEI">measure</gi>, etc.</p>
```
riedde, line 151: So with this change it will be possible do add a description to a zone but it won't be possible to describe a figure in the zone anymore?

source/modules/MEI.figtable.xml
```diff
@@ -78,44 +78,19 @@
...
-          </rng:choice>
-        </rng:zeroOrMore>
-      </rng:choice>
-    </content>
     <remarks xml:lang="en">
       <p>Best practice suggests the use of controlled vocabulary for figure descriptions. Don't
-        confuse this entity with a figure caption. A caption is text primarily intended for display
-        with an illustration. It may or may not function as a description of the illustration.</p>
-      <p>The model of this element is based on the <ref target="https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-figDesc.html">figDesc</ref> element of the Text Encoding Initiative (TEI).</p>
+        confuse a figure caption with a figure description. A caption is text primarily intended
```
riedde, line 91: Is `desc` here a caption to the figure, or the place to descripte the illustration? The description is not clear to me.

source/modules/MEI.shared.xml
```diff
@@ -5662,9 +5662,10 @@
   <elementSpec ident="desc" module="MEI.shared">
     <gloss xml:lang="en">description</gloss>
     <desc xml:lang="en">Container for text that briefly describes the feature to which it is
-      attached, including its intended usage, purpose, or application as appropriate.</desc>
```
riedde, line 5668: When deleting this, has somebody checked all side-effects to the guidelines and element definitions, `desc` appears?

riedde (commented):
I understand the intention, but there are to many open questions at the moment. So I can't approve this PR for now. I'm sure the discussion will resolve these questions.

source/docs/12-facsimilesperformances.xml
pe-ro: `<fig>` was never allowed within `<zone>`, only its `<figDesc>` child. `<figDesc>` pre-dates the existence of `<desc>` and was a shortcut to allow description of the `<zone>`. A subordinate figure within a `<zone>` should be marked with `<zone>`, not `<fig>`.  The substitution of `<desc>` for `<figDesc>` is intended to prevent the error you just made. :-)

source/modules/MEI.figtable.xml
pe-ro: `<desc>` is a place to describe the illustration.  A figure's caption is held by the `<caption>` element.

source/modules/MEI.shared.xml
pe-ro: Since this change makes the text describing `<desc>` more inclusive, I doubt there are negative side effects in the Guidelines or in other element definitions.  If any text in these places doesn't match the modified text, it is likely to be adding some kind of restriction, and would still be compliant with the new text.

source/docs/03-metadata.xml
```diff
@@ -733,7 +733,7 @@
...
+                  <p>The log consists of a list of <gi scheme="MEI">change</gi> elements, each of which contains a detailed description of the changes made. If a number is to be associated with one or more changes (for example, a revision number), the <att>n</att> attribute may be used to indicate it. The person responsible for the change and the date of the change may be indicated by the <gi scheme="MEI">respStmt</gi> and <gi scheme="MEI">date</gi> elements. The description of the change itself is contained within the <gi scheme="MEI">desc</gi> element, which can hold one or more paragraphs.</p>
```
pe-ro:
```suggestion
                  <p>The log consists of a list of <gi scheme="MEI">change</gi> elements, each of which contains a detailed description of the changes made. If a number is to be associated with one or more changes (for example, a revision number), the <att>n</att> attribute may be used to indicate it. The person responsible for the change and the date of the change may be indicated by the <gi scheme="MEI">respStmt</gi> and <gi scheme="MEI">date</gi> elements. The description of the change itself is contained within the <gi scheme="MEI">desc</gi> element.</p>
```

source/docs/12-facsimilesperformances.xml
pe-ro: Line 736 in source/docs/03-metadata.xml needs adjustment. See the suggestion above.

source/modules/MEI.figtable.xml
pe-ro: `<desc>` = "figure description"

source/docs/03-metadata.xml
bwbohl: Nice catch, that must date back quite some time, if it was ever true :-o

source/modules/MEI.facsimile.xml
```diff
@@ -88,7 +88,7 @@
     </classes>
     <content>
       <rng:zeroOrMore>
-        <rng:ref name="model.figDescLike"/>
+        <rng:ref name="desc"/>
```
bwbohl:
```suggestion
        <rng:ref name="model.descLike"/>
```

source/modules/MEI.facsimile.xml
```diff
@@ -114,7 +114,7 @@
...
```
bwbohl:
```suggestion
        <rng:ref name="model.descLike"/>
```

source/modules/MEI.figtable.xml
```diff
@@ -78,44 +78,19 @@
     <content>
       <rng:zeroOrMore>
         <rng:choice>
-          <rng:ref name="figDesc"/>
+          <rng:ref name="desc"/>
```
bwbohl:
```suggestion
          <rng:ref name="model.descLike"/>
```

source/modules/MEI.header.xml
```diff
@@ -737,7 +737,7 @@
       <rng:optional>
         <rng:ref name="respStmt"/>
       </rng:optional>
-      <rng:ref name="changeDesc"/>
+      <rng:ref name="desc"/>
```
bwbohl:
```suggestion
      <rng:ref name="model.descLike"/>
```

source/modules/MEI.shared.xml
bwbohl: I think that it was as inclusive even before removing this due to the „as appropriate“ (at least that was my interpretation ;- ). I like the removal because it does not pre-bias or create grounds for misunderstanding anymore; with the removed passage, one might wonder whether the named possibilities are the only ones to choose from or merely examples.

bwbohl (dismissed):

source/modules/MEI.facsimile.xml
pe-ro: There is no model.descLike class.

source/modules/MEI.facsimile.xml
pe-ro: There is no model.descLike class.

source/modules/MEI.figtable.xml
pe-ro: There is no model.descLike class.

source/modules/MEI.header.xml
pe-ro: There is no model.descLike class.

source/modules/MEI.facsimile.xml
bwbohl: Oh, sorry, I was referring to MEI v5.1: https://music-encoding.org/guidelines/v5/model-classes/model.descLike.html

Do you have any idea when and why it got lost?

bwbohl: Found it here:

https://github.com/music-encoding/music-encoding/blob/05b4056bace32d10d3f4e73e7adc7d2176f8312f/source/modules/MEI.shared.xml#L4405-L4407

but somehow it didn’t make it to:
https://music-encoding.org/guidelines/dev/mei-all/model-classes.html

pe-ro:
I'm not sure if the removal of the descLike class was intentional or not.  Either way, there's not much need for it at present. It can always be added later if necessary.

bwbohl:
it’s still there in the code but it gets snapped on the way to RNG by the TEI Stylesheets if it’s not used anwhere

pe-ro:
Ah, I forgot about that. In that case, using descLike in the places you've suggested is fine.

bwbohl (approved):

source/modules/MEI.shared.xml
riedde: Okay, I see that. Thank you for clarification!

bwbohl (dismissed):

riedde (changes requested):
Thank you for the comments and clarification. There are still two aspects I'm struggling with.

1. **content model of `<desc>`.** With this PR `<desc>` will be suitable for short descriptions only. This could become problematic if more detailed documentation is necessary or required, e. g., regarding the `<change>`s. I'm arguing to allow `<p>` in `<desc>` to not force brief descriptions! Instead a second element for descriptions is needed, that allows more detailed content.
2. **description of a figure** I guess that `<desc>` should be used as a brief description of a figure in the future. As I see there isn't a place in the MEI schema where a detailed figure description is possible, is it? If not, why don't we make `<figDesc>` a header element such as `<layoutDesc>` as a dedicated place for thes descriptions and a place pointers can point to? E.g., 
```
<figDesc>
   <fig n="1 xml:id="fig-001">
      <desc>
         <p>First paragraph of the description</p>
         <p>Next paragraph of the description</p>
      </desc>
</figDesc>
```

pe-ro:
@riedde, 
>content model of `<desc>`. With this PR `<desc>` will be suitable for short descriptions only. This could become problematic if more detailed documentation is necessary or required, e. g., regarding the `<change>`s. I'm arguing to allow `<p>` in `<desc>` to not force brief descriptions! Instead a second element for descriptions is needed, that allows more detailed content.

>description of a figure I guess that `<desc>` should be used as a brief description of a figure in the future. As I see there isn't a place in the MEI schema where a detailed figure description is possible, is it? If not, why don't we make `<figDesc>` a header element such as `<layoutDesc>` as a dedicated place for thes[e] descriptions and a place pointers can point to? 

Technically, the description isn't limited in length, it just can't have structure.  Your suggestion to repurpose and move `<figDesc>` is fine with me.  I still wouldn't allow `figDesc/desc/p` there, however.  In that case, `<figDesc>` can have a model similar to what I've proposed for `<bindingDesc>` or `<watermarkDesc>`:
```
<choice>
  <group>
    <zeroOrMore>
      <ref name="mei_macro.para.basic"/>
    </zeroOrMore>
    <zeroOrMore>
      <choice>
        <ref name="mei_model.biblLike"/>
        <ref name="mei_model.dimLike"/>
        <ref name="mei_model.identifierLike"/>
        <ref name="mei_model.locrefLike"/>
        <ref name="mei_model.locusLike"/>
        <ref name="mei_model.relationLike"/>
        <ref name="mei_classification"/>
        <ref name="mei_creation"/>
        <ref name="mei_creator"/>
        <ref name="mei_dimensions"/>
        <ref name="mei_repository"/>
        <ref name="mei_watermarkDesc"/>
      </choice>
      <zeroOrMore>
        <ref name="mei_model.annotLike"/>
      </zeroOrMore>
    </zeroOrMore>
  </group>
  <zeroOrMore>
    <choice>
      <text/>
      <ref name="mei_model.textPhraseLike.basic"/>
    </choice>
  </zeroOrMore>
</choice>
```
that is, either structured or unstructured text. Keep in mind that the paragraphs defined by `mei_macro.para.basic` are also limited to paragraph components (list, table, etc.) and basic phrase level text.

pe-ro:
In spite of my comments above, I'm still not completely convinced of the need for structured text in a figure description.  Here's an AI-generated response to the query "vocabulary for book illustration description" --

To elaborate on the vocabulary for describing book illustrations, we can categorize the terminology based on **format, composition, function, and visual analysis**.

### 1. Types of Illustrations by Format and Placement

These terms describe the physical size, layout, and location of an illustration within a book.

*   **Spot Illustration**: A small, focused image without a background, highlighting a single character, object, or moment. It draws attention to a specific detail without the context of a full scene.
*   **Vignette**: Similar to a spot illustration but with a soft, fading edge. It provides a glimpse into a scene with some background context, often used to show transitions or a series of events.
*   **Full-Page Illustration**: An image that fills an entire page, typically extending to the edge (full-bleed). It can stand alone or be paired with text on the facing page.
*   **Double-Page Spread**: A single illustration that spans two facing pages, creating a large, immersive visual. Often used for climactic scenes or to establish a setting.
*   **Frontispiece**: A specific illustration placed on the page facing the title page.
*   **Endpapers**: The illustrated pages at the very front and back of a hardcover book, often featuring patterns, maps, or related artwork.
*   **Cover Illustration**: The image on the book's cover, designed to attract readers and encapsulate the story's essence.
*   **Gatefold/Fold-out**: A page that folds out beyond the book's edge, used for large maps, diagrams, or panoramic scenes.

### 2. Compositional and Technical Elements

These terms describe the visual structure and creation of the image itself.

*   **Composition**: The deliberate arrangement of visual elements (characters, objects, space) within the frame to guide the viewer's eye and create balance.
*   **Bleed**: When an image extends beyond the trim edge of the page, ensuring no white border appears after cutting.
*   **Boxed In/Booked In**: An illustration that is contained within a defined border or frame.
*   **Call-out**: A label or piece of text connected by a line to a specific part of an illustration to provide information.
*   **Panel**: A bordered frame containing a single image, especially in sequential art like comics.
*   **Implied Line**: A line not physically drawn but suggested by the arrangement of elements (e.g., a character's gaze directing the viewer's eye).
*   **Leading Lines**: Lines within the composition (like a path or a river) that guide the viewer's gaze to a focal point.
*   **Perspective**: The technique used to create the illusion of depth and three-dimensional space on a flat surface.

### 3. Visual Analysis and Descriptive Language

These terms are used to analyze the mood, style, and meaning conveyed by the illustration.

*   **Color Palette**: The specific range of colors used (e.g., warm, cool, monochromatic, high-contrast). Color is often symbolic, evoking emotions like calmness (blue) or danger (bright red/yellow).
*   **Line Quality**: The character of the lines (e.g., bold, delicate, chaotic, smooth), which can suggest energy, tension, or softness.
*   **Texture**: The visual or tactile quality of surfaces within the illustration, which can be created through the medium (e.g., watercolor wash, pencil hatching, digital brush).
*   **Focal Point**: The area of the illustration that draws the viewer's primary attention, often achieved through contrast, placement, or detail.
*   **Mood/Atmosphere**: The overall feeling evoked by the image, such as serene, tense, joyful, or mysterious, created by the combination of color, composition, and subject matter.
*   **Medium**: The physical or digital material used to create the art (e.g., watercolor, acrylic, ink, digital painting, collage).

In my opinion, each of these aspects can be addressed by a separate `<desc>` element with an appropriate value for `@type`.  Alternatively, `<desc>` can be allowed to nest; e.g.,
```
<figDesc>
   <fig n="1 xml:id="fig-001">
      <desc>
         <desc>First point of the description</desc>
         <desc>Next point of the description</desc>
      </desc>
</figDesc>
```
This way, a description can be "chunked", while still keeping the content of `<desc>` simple.

pe-ro:
Sorry, only one more comment --

The suggested constrained model for `<desc>` makes it closer to TEI's model (defined by https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-macro.limitedContent.html).

pe-ro:
The best/simplest course of action now is for `<desc>` to allow a choice between multiple paragraphs or basic text:
```
<choice>
  <group>
    <ref name="mei_macro.para.basic"/>
    <oneOrMore>
      <ref name="mei_macro.para.basic"/>
    </oneOrMore>
  </group>
  <zeroOrMore>
    <choice>
      <text/>
      <ref name="mei_model.textPhraseLike.basic"/>
    </choice>
  </zeroOrMore>
</choice>
```
Notice that with this model at least 2 paragraphs are required.

Once it's made available in another PR, `mei_macro.para.basic` can replace the reference to `model.pLike`.

bwbohl:
I think more concrete examples of the usage of the existing *Desc elements would be helpful to decide whether that is intendend use or should be modelled differently

pe-ro:
@bwbohl, normally I'd agree that it would be good to survey existing practice to determine next steps.  However, that goes against the intention of this change.  The whole point is to constrain existing practice.  The intention is move `<desc>` toward the TEI model of its `<desc>` element.  `<desc>` is not intended to hold a lengthy essay, only a short text phrase or a few paragraphs. I doubt that anyone has (ab-)used `<desc>` in the way I'm looking to disallow, but it's better to "nip it in the bud" as they say.

As I hinted above, the real "meat and potatoes" of the change won't be completely realized until the following macro is implemented:
```
<define name="mei_macro.para.basic">
  <a:documentation xmlns:a="http://relaxng.org/ns/compatibility/annotations/1.0">Redefines the
    paragraph element for those situations where the complete paragraph content model isn't
    desirable; i.e., in metadata.</a:documentation>
  <element name="p">
    <zeroOrMore>
      <choice>
        <text/>
        <ref name="mei_model.paraContentPart.basic"/>
      </choice>
    </zeroOrMore>
    <ref name="mei_att.facsimile.attributes"/>
    <ref name="mei_att.dataPointing.attributes"/>
    <empty/>
  </element>
</define>
```
This macro constrains the `<p>` element to contain only the following classes and elements:

- model.lgLike
- model.listLike
- model.quoteLike
- model.tableLike
- model.addressLike
- model.annotLike
- model.biblLike
- model.dateLike
- model.editorialLike
- model.figureLike
- model.identifierLike
- model.locusLike
- model.measurementLike
- model.nameLike
- model.qLike
- model.textPhraseLike.bare
- model.titleLike
- mei_dimensions
- mei_extent
- mei_geogFeat
- mei_material
- mei_term

by employing the `model.textPhraseLike.basic` class made available with the merging of #1734.

If, at some point in the future, this model is viewed as too limiting, then the macro can employ the `model.textPhraseLike.limited` class or the macro can be replaced by the `model.pLike` class.  The bare, basic, and limited classes provide multiple levels between the "text phrase only" and the "full-blown text with transcriptional and editorial markup" extremes.

I say, full steam ahead.

riedde:
> Sorry, only one more comment --
> 
> The suggested constrained model for `<desc>` makes it closer to TEI's model (defined by https://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-macro.limitedContent.html).

I see that. Following the TEI definition the `macro.limitedContent (paragraph content) defines the content of prose elements that are not used for transcription of extant materials. ` My question—for me still open—is where can we make a more details description, when it's not allowed here anymore?

riedde:
> <ref name="mei_macro.para.basic"/>
> […]
> Once it's made available in another PR, mei_macro.para.basic can replace the reference to model.pLike.

@pe-ro you mentioned in https://github.com/music-encoding/music-encoding/pull/1772#issuecomment-4905409310 that there is another PR containing the referred `mei_macro.para.basic`. Which PR is it? I was not able to find it.

pe-ro:
@riedde, the idea is to allow editorial elements (`choice` and `subst`) and transcriptional elements (`add`, `corr` ,`damage`, `del`, `gap`, `handShift`, `orig`, `reg`, `restore`, `sic`, `supplied`, and `unclear`) wherever "normal" paragraph content is permitted.  

If editorial or transcriptional elements are necessary, it makes sense to treat the description as part of the text of the resource, not as metadata.  In my experience, descriptions of this kind usually appear in the front or back matter.  In those very rare cases where this is not the case, the encoder may easily customize the content model of `<desc>`.

On the other hand, if the description is metadata, I think the proposed content model of `desc` is sufficient.

You didn't find the PR mentioned in my comment in https://github.com/music-encoding/music-encoding/pull/1772#issuecomment-4905409310 because it doesn't exist yet. Neither Rome nor MEI was/can be built in a day. :-)

riedde:
[This is just a fragment](https://www.oeaw.ac.at/acdh/research/musicology/resources/corpora/medieval-music-manuscripts-oenb), but I chose it because there are a lot of small sources (few leaves, unbinded) without a real front or back part. However, I see the need of having the possibility to describe or even encode the figure (e.g., marking the position) in the body and or as metadata (e.g., detailed description or transcription-like something). Concerning the watermarks we moved it to the metadata section but developed it further to make possible more detailed descriptions.
That's why I do not like restricting the figure description too much.

Here a spontainious idea:

```
<!-- header 1 (possible at the moment)-->
<fileDesc>
    <titleStmt/>
    <pubStmt/>
    <notesStmt>
       <annot xml:id="figure-description">Extended description of the figure</annot>
    </notesStmt>
</fileDesc>

<!-- header 2 (has to be modelled)-->
<physDesc>
   <figDesc>
      <figList>
         <fig/>
         <fig/>
      </figList>
   </figDesc>
</physDesc>

<!-- body -->
<fig decls="#figure-description">
    <desc>Here is the figure placed.</desc>
</fig>
```
Therefore `<fig>` needs to become a member of `att.metadataPointing`. The example `header 1` is not good, placing `notesStmt` in the fileDesc. Otherwise it has to be placed in `manifestation` or `item`, but then FRBR would be required for describing figures. The example `header 2` is an adaption of the watermark-handling.

By the way, the places where `<fig>` is allowed are impressing ..., see https://music-encoding.org/guidelines/dev/mei-all/elements/fig.html

pe-ro:
@riedde, I understand that
```
<physDesc>
   <figDesc>
      <figList>
         <fig/>
         <fig/>
      </figList>
   </figDesc>
</physDesc>
```
gathers figure descriptions in a central location; however, I don't see what it accomplishes beyond this. In fact, it encourages metadata duplication -- once in the body of the text and once in the header.  I think locating the metadata alongside the figure itself is best, especially when `<desc>` is allowed to contain `<p>`; e.g.,
```
<fig>
  <desc>
    <p>
      <fig>
        <desc>
          <p>Some text 
            <table>
              <tr/>
            </table>
            <list>
              <li/>
            </list> plus more text.</p>
        </desc>
      </fig>
    </p>
  </desc>
</fig>
```

riedde:
@pe-ro I'm fine with your last example where `<desc>` can contain `<p>`. I missed this detail in the discussion above.

PR #1781, opened 2026-05-07
schema: mark attributes in descriptions

pe-ro:


krHERO (dismissed):
looks good :)

rettinghaus:
Basically good, but it doesn't work within `valItem`:

Bildschirmfoto 2026-05-08 um 10 21 54

Please, either remove it there or adjust the Guidelines XSLT.

musicEnfanthen:
Use of `<att>` inside descriptions seems indeed not be used very widely and not rendering well.

pe-ro:
Removed `<att>` markup in valItem.  I'll leave it to someone else to modify the Guidelines XSLT.

bwbohl:
both `<att>` and `<val>` are very common inside `<remarks>` in the TEI definitions; although less in `<desc>` I think they still can be considered quite common. I recommend updating our XSLTs to support it – this would probably nicely align with #1626

PR #1782, opened 2026-05-09
schema: separate dataSpec from macroSpec

rettinghaus:
This updates `macroSpec@type="dt"` to `dataSpec` and drops the outdated `@type="pe"` on remaining `macroSpec` elements.

Stylesheets were adjusted accordingly.


utils/compare_versions/comparison.xsl
```diff
@@ -1052,7 +1052,7 @@
                     or count($added.values) gt 0
                     or count($removed.values) gt 0">
                     <tr class="c" id="{$current.macro}">
-                        <td class="macroSpec datatype ident">
+                        <td class="dataSpec datatype ident">
```
musicEnfanthen, line 1055: Not sure if this is correct. This section is about the actual macros (former pe), isnt it?

utils/compare_versions/comparison.xsl
```diff
@@ -1132,7 +1132,7 @@
 
             </xsl:for-each>
         </xsl:variable>
-        <xsl:variable name="unchanged.models" select="$kept.models[not(. = $modified.models//td[@class='macroSpec datatype ident']/text())]" as="xs:string*"/>
+        <xsl:variable name="unchanged.models" select="$kept.models[not(. = $modified.models//td[@class='dataSpec datatype ident']/text())]" as="xs:string*"/>
```
musicEnfanthen, line 1135: Same here (belongs to macro pe section; datatype section starts below)

utils/compare_versions/comparison.xsl
rettinghaus: Rechecked and it is indeed correct. (Even if it were wrong, this class is not addressed in the CSS …)

utils/compare_versions/comparison.xsl
rettinghaus: Also correct.

PR #1791, opened 2026-06-08
Update Saxon to version 13

rettinghaus:


rettinghaus:
@musicEnfanthen Any idea why it might fail in the CI? It works like a charm locally.

musicEnfanthen:
@rettinghaus My first guess would be that the CI is using the docker image where saxon is not updated yet so its still looking for version 12.9

rettinghaus:
No, it checks out this repo and just calls `ant build-customizations`.

musicEnfanthen:
I see. The error message thrown says:  `Source document file:/home/runner/work/music-encoding/music-encoding/build/mei-source_canonicalized_v6.0-dev.xml is not readable` (https://github.com/music-encoding/music-encoding/actions/runs/27125585744/job/80053664359#step:3:120)

```
[xslt] Error: odd2odd.xsl: Source document file:/home/runner/work/music-encoding/music-encoding/build/mei-source_canonicalized_v6.0-dev.xml is not readable; from file:/home/runner/work/music-encoding/music-encoding/customizations/mei-all.xml (which has an outermost element of TEI), with loc=/home/runner/work/music-encoding/music-encoding/build/mei-source_canonicalized_v6.0-dev.xml
     [xslt] /home/runner/work/music-encoding/music-encoding/lib/stylesheets/odds/odd2odd.xsl:308:34: Fatal Error! Processing terminated by xsl:message at line 308 in odd2odd.xsl
     [xslt] Failed to process /home/runner/work/music-encoding/music-encoding/customizations/mei-all.xml

```

And then the TEI stylesheets throw at line 308 in odd2odd.xsl, which is the "die"-template:

```xml
<xsl:template name="die">
    <xsl:param name="message"/>
    <xsl:message terminate="yes">
      <xsl:text>Error: odd2odd.xsl: </xsl:text> 
      <xsl:value-of select="$message"/>
    </xsl:message>
  </xsl:template>
```

triggered by the tei:workOutSource function that has this branch

```xml
<xsl:choose>
      <xsl:when test="not(doc-available($source))">
        <xsl:call-template name="die">
          <xsl:with-param name="message" as="xs:string"
                          select="concat(
                                  'Source document ',
                                  $source,
                                  ' is not readable; from ',
                                  base-uri($top),
                                  ' (which has an outermost element of ',
                                  name($top/*),
                                  '), with loc=',
                                  $loc
                                  )"/>
        </xsl:call-template>
      </xsl:when>
```

I don't know what breaking changes happened in Saxon 13 (yet), but it seems that somehow the source file is not recognized anymore for the build steps (what seems strange since it runs fine in the validation).

To narrow down the issue further: Can you try running `ant reset` locally first and see whether the build still works afterward?

rettinghaus:
In Saxon 13, the default configuration for resolving relative URIs has been completely revised. When the stylesheet odd2odd.xsl uses doc-available('file:/home/runner/work/...') to check whether the MEI source file exists, Saxon 13 blocks file access. Saxon incorrectly evaluates the path within the Ant-JAXP context as unsafe or invalid and returns `false`, causing the TEI framework to abort the transformation with an error message.

So for now we have to stay with Saxon 12.

bwbohl:
> In Saxon 13, the default configuration for resolving relative URIs has been completely revised. When the stylesheet odd2odd.xsl uses doc-available('file:/home/runner/work/...') to check whether the MEI source file exists, Saxon 13 blocks file access. Saxon incorrectly evaluates the path within the Ant-JAXP context as unsafe or invalid and returns `false`, causing the TEI framework to abort the transformation with an error message.
> 
> So for now we have to stay with Saxon 12.

Thanks for digging into this!

PR #1796, opened 2026-07-02
schema: update and clean up of `MEI.stringtab` attributes with regard to `<chordTable>`

janjusolja:
This PR updates values used in the `<chordTable>` and cleans up the deprecated attributes related to it. The following changes were made:

- The guidelines and definitions in `MEI.harmony` were updated to the current version of `MEI.stringtab` (`@fret` = `@tab.fret`, `@pos` = `@tab.pos`, `@string` = `@tab.course`)
- Deprecated attributes removed from `MEI.harmony`: `@tab.strings` and `@tab.courses` (in `<chordDef>`)
- Deprecated attributes removed from `MEI.gestural`: `@tab.strings` and `@tab.courses` (in `<staffDef>`)
- Deprecated attributes `@tab.strings` and `@tab.courses` removed from `MEI.stringtab`
- `"omit"` added as a value of `@tab.fret` in the case of an omitted course 
- Added `@tab.fret` to `<barre>`
- Deprecated `@fret` in `<barre>`
- Deprecated `@tab.fing` in favour of newly proposed attribute `@finger` (separate PR)
- Deprecated `@tab.string` in favour of existing `@tab.course`
- Removed constraints regarding `@tab.strings` and `@lines` (in `<staffDef>`) from `MEI.shared`

This PR cleans up and updates the schema and guidelines with regard to the issues [#1686](https://github.com/music-encoding/music-encoding/issues/1686) and [#1756](https://github.com/music-encoding/music-encoding/issues/1756), and outlines the necessary changes and preparations regarding the issue [#1795](https://github.com/music-encoding/music-encoding/issues/1795).

A validation test is available directly in the [mei-friend](https://mei-friend.mdw.ac.at/?file=&scale=74&breaks=encoded&select=e5rdrbo2&page=1&speed=false&notationOrientation=bottom&notationProportion=0.69). Alternatively, you can use [this minimal example](https://gist.githubusercontent.com/janjusolja/98e25b42108f3264b5bda91604a0cc2b/raw/80632b84fcc9170775d8732f30564ac60d3f4ea9/tab-chordMember-example.mei) and test it against [this RNG](https://gist.githubusercontent.com/janjusolja/1808b8eb57b8c63de8563501720a7423/raw/aaa87c67d5bcc4fe02628a5368661344264b39d9/mei-all.rng).

Special thanks to @reinierdevalk , @musicog , @DILewis  and @pe-ro  for their support!

Issue #1686, opened 2025-06-17
Enhance chordTable element and its children


pe-ro:
The `<chordTable>` element and its `<chordDef>` children support the encoding of chord tablature grids like those found in pop sheet music.  For example,

![Image](https://github.com/user-attachments/assets/e9c2f939-89f4-4421-80d7-f16f1539fafa)

can be encoded as 
```
<chordDef xml:id="chordDef_d1e300" label="C6/E" tab.pos="5">
  <chordMember xml:id="chordMember_d190e7" tab.string="1" tab.fing="x"/>
  <chordMember xml:id="chordMember_d190e9" tab.string="2" tab.fret="3" tab.fing="3"/>
  <chordMember xml:id="chordMember_d190e11" tab.string="3" tab.fret="1" tab.fing="1"/>
  <chordMember xml:id="chordMember_d190e13" tab.string="4" tab.fret="1" tab.fing="1"/>
  <chordMember xml:id="chordMember_d190e15" tab.string="5" tab.fret="1" tab.fing="1"/>
  <chordMember xml:id="chordMember_d190e17" tab.string="6" tab.fret="1" tab.fing="1"/>
  <barre tab.fing="1" startid="chordMember_d190e11" endid="chordMember_d190e17"/>
</chordDef>
```
or rather, could be encoded this way with a modification to `<barre>`, i.e., replacing `barre/@tab.fret` with `@tab.finger`, leaving the position of the chord diagram (the fret at which it starts) to be captured by `chordMember/@tab.pos`.  I own this error and so I should be the one to fix it.  Or at least the one to call attention to the problem.  :-)

Recent changes to the tablature module may have some bearing on the way the attributes on `<chordMember>` and `<barre>` are named.  I believe some of these attributes and their datatypes are deprecated by those changes. If the deprecation is carried out, then these attributes can be declared locally (in the definition of `<chordMember>`) and their names simplified, e.g., from `@tab.fret` to `@fret`.

This issue is the result of my attempt to convert into MEI a popular but now-defunct online guitar chord dictionary I created in the previous century (the data for which was stored in a plain text file, one chord per line).

![Image](https://github.com/user-attachments/assets/1b3727ca-b79a-4530-abcd-442595a1d7bc)

With this data in MEI, users may include these chords in their files where, with some tweaking, presumably Verovio can render the tab grids.

craigsapp:
[...]

PR #1798, opened 2026-07-03
schema: extension of fingering-specific attributes `@finger` and `@handRole` used on elements `<barre>`, `<chordMember>` and `<fing>`

### LINKED ISSUE #1795: Overhaul and extension of `<fing>` element
To create a generic approach to specifying fingering information with the `<fing>` element, the tablature IG suggests using the `@finger` attribute to indicate which finger to use. Its values are the first letters of the English names of the fingers: thumb (`t`), index (`i`), middle (`m`), ring (`r`) and little finger (`l`). The usage of `p`, `i`, `m`, and `a` is declined because it is limited to plucked string instruments. The usage of numeral indications is declined due to the potential for confusion as a result of different meanings between different instruments (e.g. keyboard or plucked string instruments). To specify which hand is meant, the additional `@handrole` attribute can be used, taking values `sounding|stopping` for plucked string instruments. This can be extended, if required, with values such as `left|right` for other instruments.

`@finger` and `@handrole` add a layer of abstraction that enables capturing all possible variations of indicating finger and hand combinations unambiguously, and it takes away the necessity to specify handedness. For example, the RH index finger (if you are a right-handed player) can be _i_ in classical guitar notation, a dot in French and Italian lute tablature, a modified rhythm flag in German lute tablature, or _2_ in keyboard notation. 

Actual visualisation of fingering can then be achieved in two ways: 

1. manually: by providing the symbol to be shown (e.g., _i_ or _2_) as the `<fing>`’s textual content
2. automatically: being triggered by the use of the attributes `@finger` and `@handrole` in specific cases, which solves the problems discussed in [this issue](https://github.com/rism-digital/verovio/issues/4303).

@DILewis @musicog @reinierdevalk

janjusolja:
This PR introduces two new attributes, `@finger` and `@handRole`, which are used to specify the finger (of both hands) in the `<fing>` element. This proposal is directly connected to issue [#1795](https://github.com/music-encoding/music-encoding/issues/1795). The following changes were made:
- This branch was based on changes made in PR [#1796](https://github.com/music-encoding/music-encoding/pull/1796)
- The `@finger` and `@handRole` attributes were added to the `MEI.fingering` as a new individual attribute class, `att.finger`, due to further referencing
- The attribute class `att.finger` was added to the `<chordMember>` of the `MEI.harmony` module and to the `<barre>` of the `MEI.stringtab` module, enabling a fingering-specific definition on tablature chord grids (see also issue [#1686](https://github.com/music-encoding/music-encoding/issues/1686))
- The guidelines and definitions in `MEI.harmony` were updated
- If the values of `@tab.fret` are set to either `"0"` (open string) or `"omit"` (omitted course), the use of `@finger` (never needed in these cases) is forbidden by means of a Schematron rule

A validation test is available directly [in the mei-friend](https://mei-friend.mdw.ac.at/?file=&scale=74&breaks=encoded&select=n2rrs0b&page=1&speed=false&notationOrientation=bottom&notationProportion=0.69). However, the Schematron rule must be tested in a different editor. Alternatively, you can download [this minimal example](https://gist.githubusercontent.com/janjusolja/7c5a24ae7e04d634c4be09ec22957127/raw/aebd16fdaa064b4daee37a183969a91bc8c9f92d/tab-finger-example.mei) and test it against [this RNG](https://gist.githubusercontent.com/janjusolja/cfb73d08913a9ea4dc3b9a4afec22426/raw/99003854dc11120544d18ecef8eb9a7cba003a60/mei-all.rng).

Co-authored by @musicog , @DILewis and @reinierdevalk. Special thanks to @pe-ro and @th-we for their valuable contributions to the discussion!

### DIFF
--- a/source/modules/MEI.fingering.xml
+++ b/source/modules/MEI.fingering.xml
@@ -49,7 +49,7 @@
       <memberOf key="att.timestamp2.log"/>
     </classes>
   </classSpec>
-  <classSpec ident="att.finger" module="MEI.fingering" type="atts">
+  <classSpec ident="att.fing.finger" module="MEI.fingering" type="atts">
     <desc xml:lang="en">Attributes used for a description of the finger (in physical terms).</desc>
     <attList>
       <attDef ident="finger" usage="opt">
@@ -71,6 +71,11 @@
           </valItem>
         </valList>
       </attDef>
+    </attList>
+  </classSpec>
+  <classSpec ident="att.fing.hand" module="MEI.fingering" type="atts">
+    <desc xml:lang="en">Attributes used for a description of the hand role (in physical terms).</desc>
+    <attList>
       <attDef ident="handRole" usage="opt">
         <desc xml:lang="en">Records the role of the hand in a fingering indication.</desc>
         <datatype>
@@ -120,7 +125,8 @@
       <memberOf key="att.fing.ges"/>
       <memberOf key="att.fing.log"/>
       <memberOf key="att.fing.vis"/>
-      <memberOf key="att.finger"/>
+      <memberOf key="att.fing.finger" />
+      <memberOf key="att.fing.hand" />
       <memberOf key="model.fingeringLike"/>
     </classes>
     <content>
--- a/source/modules/MEI.fingering.xml
+++ b/source/modules/MEI.fingering.xml
@@ -84,6 +84,14 @@
         <datatype>
           <rng:ref name="data.HANDROLE"/>
         </datatype>
+        <constraintSpec ident="When_sounding_hand" scheme="schematron">
+          <constraint>
+            <sch:rule context="mei:*[@handRole eq 'sounding']">
+              <sch:assert test="not(@finger eq 'none')">A sounding hand requires a finger. The value "none"
+                is not permitted for @finger when @handRole is "sounding".</sch:assert>
+            </sch:rule>
+          </constraint>
+        </constraintSpec>
       </attDef>
     </attList>
   </classSpec>
--- a/source/modules/MEI.fingering.xml
+++ b/source/modules/MEI.fingering.xml
@@ -51,19 +51,19 @@
     <attList>
       <attDef ident="finger" usage="opt">
         <valList type="closed">
-          <valItem ident="0">
+          <valItem ident="t">
             <desc xml:lang="en">Thumb</desc>
           </valItem>
-          <valItem ident="1">
+          <valItem ident="i">
             <desc xml:lang="en">Index finger</desc>
           </valItem>
-          <valItem ident="2">
+          <valItem ident="m">
             <desc xml:lang="en">Middle finger</desc>
           </valItem>
-          <valItem ident="3">
+          <valItem ident="r">
             <desc xml:lang="en">Ring finger</desc>
           </valItem>
-          <valItem ident="4">
+          <valItem ident="l">
             <desc xml:lang="en">Little finger</desc>
           </valItem>
         </valList>
--- a/source/modules/MEI.fingering.xml
+++ b/source/modules/MEI.fingering.xml
@@ -69,6 +69,9 @@
           <valItem ident="l">
             <desc xml:lang="en">Little finger</desc>
           </valItem>
+          <valItem ident="none">
+            <desc xml:lang="en">No fingers in use</desc>
+          </valItem>
         </valList>
       </attDef>
     </attList>
--- a/source/modules/MEI.fingering.xml
+++ b/source/modules/MEI.fingering.xml
@@ -48,6 +48,33 @@
       <memberOf key="att.startEndId"/>
       <memberOf key="att.timestamp2.log"/>
     </classes>
+    <attList>
+      <attDef ident="finger" usage="opt">
+        <valList type="closed">
+          <valItem ident="0">
+            <desc xml:lang="en">Thumb</desc>
+          </valItem>
+          <valItem ident="1">
+            <desc xml:lang="en">Index finger</desc>
+          </valItem>
+          <valItem ident="2">
+            <desc xml:lang="en">Middle finger</desc>
+          </valItem>
+          <valItem ident="3">
+            <desc xml:lang="en">Ring finger</desc>
+          </valItem>
+          <valItem ident="4">
+            <desc xml:lang="en">Little finger</desc>
+          </valItem>
+        </valList>
+      </attDef>
+      <attDef ident="handRole" usage="opt">
+        <desc xml:lang="en">Records the role of the hand in a fingering indication.</desc>
+        <datatype>
+          <rng:ref name="data.HANDROLE"/>
+        </datatype>
+      </attDef>
+    </attList>
   </classSpec>
   <classSpec ident="att.fingGrp.log" module="MEI.fingering" type="atts">
     <desc xml:lang="en">Logical domain attributes.</desc>
--- a/source/modules/MEI.fingering.xml
+++ b/source/modules/MEI.fingering.xml
@@ -51,6 +51,19 @@
   </classSpec>
   <classSpec ident="att.fing.finger" module="MEI.fingering" type="atts">
     <desc xml:lang="en">Attributes used for a description of the finger (in physical terms).</desc>
+    <constraintSpec ident="finger_none_not_allowed" scheme="schematron">
+      <constraint>
+        <sch:rule context="mei:chordMember[@finger]">
+          <sch:assert test="not(@finger eq 'none')">The value "none" is not permitted for
+            the @finger attribute on chordMember element.</sch:assert>
+        </sch:rule>
+        <sch:rule context="mei:barre[@finger]">
+          <sch:assert test="not(@finger eq 'none')">The value "none" is not permitted for
+            the @finger attribute on barre element.
+          </sch:assert>
+        </sch:rule>
+      </constraint>
+    </constraintSpec>
     <attList>
       <attDef ident="finger" usage="opt">
         <valList type="closed">
--- a/source/modules/MEI.fingering.xml
+++ b/source/modules/MEI.fingering.xml
@@ -25,6 +25,21 @@
   <moduleSpec ident="MEI.fingering">
     <desc xml:lang="en">Fingering component declarations.</desc>
   </moduleSpec>
+  <macroSpec ident="data.HANDROLE" module="MEI.fingering" type="dt">
+    <desc xml:lang="en">Role of hand in fingering indication relevant to plucked string instruments.
+      One of 'sounding' and 'stopping'.
+    </desc>
+    <content>
+      <valList type="closed">
+        <valItem ident="sounding">
+          <desc xml:lang="en">The plucking or strumming hand.</desc>
+        </valItem>
+        <valItem ident="stopping">
+          <desc xml:lang="en">The stopping or fretting hand.</desc>
+        </valItem>
+      </valList>
+    </content>
+  </macroSpec>
   <classSpec ident="att.fing.log" module="MEI.fingering" type="atts">
     <desc xml:lang="en">Logical domain attributes.</desc>
     <classes>
--- a/source/modules/MEI.fingering.xml
+++ b/source/modules/MEI.fingering.xml
@@ -48,6 +48,9 @@
       <memberOf key="att.startEndId"/>
       <memberOf key="att.timestamp2.log"/>
     </classes>
+  </classSpec>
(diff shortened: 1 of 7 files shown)


janjusolja:
Dear @bwbohl and @martha-thomae , please note that I had the same validation problem here as in PR [#1796](https://github.com/music-encoding/music-encoding/pull/1796) because the deprecated attribute class `att.stringtab.tuning` is mentioned in release notes in the validation source file.

th-we:
I asked about the open string in https://github.com/music-encoding/music-encoding/issues/1795#issuecomment-4804005222. I understand that you want to encode the open string with `@tab.fret="0"`, but I see two issues:

* It requires the tablature module to be active. For projects that do not use tablature and want to deactivate the module, an open string can not be expressed, as far as I can see.
* In https://github.com/music-encoding/music-encoding/issues/1795 it was discussed that a mechanism should be possible for automatically mapping `@finger` to a rendering. So all rendered fingerings can be encoded using `@finger`, except for the open string, which requires an entirely different encoding. And I don't think it's sensible to overload `@tab.fret` with rendering intentions for fingerings as it already has rendering implications other than fingerings, I guess.

But in https://github.com/music-encoding/music-encoding/issues/1795#issuecomment-4844427780 I read that you already suggested a solution to the problem:

> We would propose adding 'none' as a value for @finger attribute for this case

Could you include this in this pull request?

janjusolja:
Yes, thank you for pointing this out and breaking it down again, @th-we! I must admit that I lost track of this issue in the process.

Here is an overview of how the `@finger` and `@handRole` attributes are used:
1. In `<fing>`, both attributes are used without a direct connection to `@tab.fret`.

The encoding of fingerings for both the right and left hands for an open string in tablature would look like this:
```
<tabGrp dur="4">
    <note xml:id="note1" tab.course="1" tab.fret="0"/>
</tabGrp>

…

<fing startid="#note1" finger="none" handRole="stopping"/>
<fing startid="#note1" finger="i" handRole="sounding"/>
```
The encoding of fingerings for both the right and left hands for an open string in classical guitar notation (tablature module inactive) would look like this:

```
<note xml:id="note1" dur="4" pname="e" oct="5"/>

…

<fing startid="#note1" finger="none" handRole="stopping">0</fing>
<fing startid="#note1" finger="i" handRole="sounding">i</fing>
```
(NB: The textual content of the `<fing>` is optional, but it can be used to indicate fingerings here. In other words, there is no urgent need to map `@finger` automatically to the rendering in this case, as this will be challenging to define outside of the tablature module. However, this is a topic for another issue.)

2. The `@finger` attribute in the `<chordMember>` is used in combination with the `@tab.course` and `@tab.fret` attributes.
 
Currently, the usage of the `@finger` attribute depends on the value of the `@tab.fret` attribute: in the case of an open or omitted string, the `@finger` attribute is forbidden. This means that, by default, `@finger="none"` should either never be used, or only in combination with `@tab.fret="0"` or `@tab.fret="omit"` in this case. Personally, I would not recommend the latter, as tablature chord grids do not _necessarily_ contain fingering indications and such rules could be ambiguous.

   Furthermore, the `@handRole` attribute is unnecessary in this case. While adding `@handRole="stopping"` would not be incorrect, it would provide excess information. However, adding `@handRole="sounding"` would indeed be incorrect.

3. In the case of `<barre>`, `@finger` can be indirectly connected with `@tab.fret="omit"` as in this case:
D-Minor

```
<chordDef xml:id="chordD_minor" label="d" tab.pos="5">
    <chordMember xml:id="chordD_minor_e" tab.course="6" tab.fret="omit"/>
    <chordMember xml:id="chordD_minor_d1" tab.course="5" tab.fret="1" finger="i"/>
    <chordMember xml:id="chordD_minor_a1" tab.course="4" tab.fret="3" finger="r"/>
    <chordMember xml:id="chordD_minor_d2" tab.course="3" tab.fret="3" finger="l"/>
    <chordMember xml:id="chordD_minor_f2" tab.course="2" tab.fret="2" finger="m"/>
    <chordMember xml:id="chordD_minor_a2" tab.course="1" tab.fret="omit"/>
    <barre startid="#chordD_minor_d1" endid="#chordD_minor_a2" tab.fret="1" finger="i"/>
</chordDef>
```

There is no need to use a `@handRole="sounding"` here either, and this can also be excluded by means of the Schematron rule as well.

**To summarize, here is the list of changes that would add to the PR:**
- Add value `"none"` for `@finger`
- Add following Schematron rules:
  - The combination of `@finger="none"` and `@handRole="sounding"` should generally not be allowed
  - `@finger="none"` should never be used in `<chordMember>` or `<barre>`
  - `@handRole` should not be used in `<chordMember>` or `<barre>`

Does everyone agree with my conclusions?
Thank you so much for your support!

th-we:
@janjusolja, that's a really thorough and well comprehensible explanation of how `@finger` is intended to work! All of this now makes total sense to me. The only thing that I find odd is to use Schematron for disabling `@handRole` on `<chordMember>`. My preference would be: Define `@finger` outside of the `att.finger` `<classSpec>` and reference it there instead. Then it can also be referenced separately in `<chordMember>`, without drawing in `@handRole`.

janjusolja:
Separating `@finger` and `@handRole` into two different classes was also my initial idea, but I decided to keep them in the same class, as they belong together logically or conceptually. However, I agree that it is better practice to define such simple restrictions in the schema itself and use Schematron for more complicated rules.

Thank you so much for your help, @th-we! I have now updated all the changes, and have tested [the new RNG](https://gist.githubusercontent.com/janjusolja/f3c0ed79b33b31361b15bf267713a78b/raw/dbcc2d163d87bd12cc3ada8bec2bfc945061f211/mei-all.rng), which you will also find in the randomly encoded ["nonsense example" in the mei-friend](https://mei-friend.mdw.ac.at/?file=&scale=74&breaks=encoded&select=n2rrs0b&page=1&speed=false&notationOrientation=bottom&notationProportion=0.69) to test these restrictions.

Following changes were made:
- Attributes `@finger` and `@handRole` are now defined in two separate attribute classes (`att.fing.finger` and `att.fing.hand`)
- Value `"none"` was added to the `@finger` attribute
- Following Schematron rules were added:
  - The combination of `@finger="none"` and `@handRole="sounding"` should generally not be allowed
  - `@finger="none"` should never be used in `<chordMember>` or `<barre>`

source/modules/MEI.fingering.xml
```diff
@@ -34,6 +49,65 @@
       <memberOf key="att.timestamp2.log"/>
     </classes>
   </classSpec>
+  <classSpec ident="att.fing.finger" module="MEI.fingering" type="atts">
+    <desc xml:lang="en">Attributes used for a description of the finger (in physical terms).</desc>
+    <constraintSpec ident="finger_none_not_allowed" scheme="schematron">
+      <constraint>
+        <sch:rule context="mei:chordMember[@finger]">
```
th-we: This rule seems to assume that `chordMember/@finger` is only used for encoding chord diagrams, in conjunction with `@tab.fret`/`@tab.course`, where `@finger="none"` would be redundant. There can be other applications beyond chord diagrams, maybe like encoding the chord voicings on a [Kontra](https://en.wikipedia.org/wiki/Kontra) or [microtonal chords on a fretless guitar](https://www.musicandpractice.org/volume-4/fretless-architecture-towards-the-development-of-original-techniques-and-musical-notation-specific-to-the-fretless-electric-guitar#_ftnref35). When there are no frets, using `@finger` consistently for all `<chordMembers>` is more natural.

Is this too far-fetched? I still think the rule is over-regulating. If it should still be kept, it could be relaxed to this:
```xml
<sch:rule context="mei:chordMember[@finger][@tab.fret][@finger='none' or @tab.fret='0']">
  <sch:assert test="@finger='none' and @tab.fret='0'">
    @finger="none" and @tab.fret="0" imply each other.
  </sch:assert>
</sch:rule>
```

source/modules/MEI.harmony.xml
```diff
@@ -139,8 +139,16 @@
     <content>
       <empty/>
     </content>
+    <constraintSpec ident="chordMember_finger_requires_stopped_string" scheme="schematron">
+      <constraint>
+        <sch:rule context="mei:chordMember[@tab.fret = ('0', 'omit')]">
+          <sch:assert test="not(@finger)">When the @tab.fret attribute indicates an open or omitted string
+            (has the value "0" or "omit"), the @finger attribute must not be used.</sch:assert>
+        </sch:rule>
```
th-we, line 147: This rule is more reasonable to me than forbidding `@finger="none"` entirely in `att.fing.finger`.

pe-ro (dismissed):
I agree with @th-we's suggestions.  With those additions/modifications, I approve this PR.

source/modules/MEI.fingering.xml
janjusolja: If I understand correctly, the constraint you suggested is already included in the more general rule in `MEI.harmony` that refers to the combination of `@finger` and `@tab.fret`, which makes more sense, as you also mentioned [here](https://github.com/music-encoding/music-encoding/pull/1798/changes/BASE..5263c2782498f31517f5a25394593bc960b1f08e#r3596967261). In that case, the suggested rule would be superfluous here.

The problem you are describing would require the use of `@finger="none"` where there is no indication of `@tab.fret`, if I understand correctly. It seems to me that the best solution is just to remove the constraint in `MEI.fingering` from the `<chordMember>`. Am I right?

Shall we keep this constraint for `<barre>`, assuming that it will always be stopped by a finger, or should we also remove it for cases that might involve a capo perhaps?

th-we: I realize that the specs are somewhat contradictory on what `<chordDef>` is intended for. Maybe this has lead to some confusion in our discussion. [In the schema](https://music-encoding.org/guidelines/v5/elements/chordDef.html) it is described as "Chord tablature definition", but the documentation mentions [chord tablature grids](https://music-encoding.org/guidelines/v5/content/analysisharm.html#harmonyTablatureGrids) only briefly and concentrates on [more general uses](https://music-encoding.org/guidelines/v5/content/analysisharm.html#harmonyMetadataInscoreDef).

I'm taking the position that `<chordDef>` is not restricted to chord tablature grids. In that light:

> The problem you are describing would require the use of `@finger="none"` where there is no indication of `@tab.fret`, if I understand correctly.

No, omitting either or both is fine to me (because my understanding is that `<chordDef>` does not necessarily describe a chord grid).

> It seems to me that the best solution is just to remove the constraint in `MEI.fingering` from the `<chordMember>`. 

That's perfectly fine to me.

> Shall we keep this constraint for `<barre>`, assuming that it will always be stopped by a finger, or should we also remove it for cases that might involve a capo perhaps?

I'd keep this rule, but it's good you mention capo. I didn't think of capo situations and whether or not `<barre>` should be used to encode them. My intuition would be no, but as far as I can see, we currently don't have an alternative. As capo is a "global" thing, I think introducing a numeric `@capo` attribute on `<chordTable>` would make sense. As the capo can be shifted during the piece, multiple `<chordTable>`s with different `@capo` values would have to be allowed, though. But that's beyond the scope of this pull request.

janjusolja: Great, thank you! I now understand the problem better. This confirms to me that omitting the constraint on `<chordMember>` element is the best approach. 

I also agree with you that the capo should either have its own encoding system or remain a textual element rather than relating to the barre. That said, I will leave the second rule regarding the `<barre>` element as it is.

Thank you for the review!

th-we (approved):

pe-ro (dismissed):

bwbohl (dismissed):
thanks to all of you for your great work!

bwbohl:
This looks all good now except that:
1. The PR branch needs  to be rebased to the latest upstream/develop, which I cannot do from here since the e-laute fork does not allow upstream owners to push to the PR branch
2. The mei-source schematron validation will fail because the introduction contains an `ident` referencing the deprecated `att.stringtab.tuning` – which is part of the "About version 5.1" section and probably should be handled in a more generous way instead of ultimately failing the whole validation because it’s only relevant for the guidelines generation and the stylesheets will handle it generously by not creating the link for the unknwon target.

bwbohl:
blocked by #1815

musicog:
@bwbohl: both this PR and #1796 have now been rebased. Since this branch builds on top of #1796 (and contains its commits), please **merge #1796 before this one**. Thanks!

Side note; upstream-owners simply cannot be allowed to push changes into a PR-branch owned by an organization, it turns out; see https://github.com/orgs/community/discussions/5634

bwbohl (approved):

martha-thomae (approved):

PR #1809, opened 2026-07-20
feat: simplify font size pattern validation

bwbohl:
Replace complex exception-based pattern with single regex that enforces non-zero values. The new pattern directly matches valid font sizes (pt or vu units) with numeric values greater than zero, eliminating the need for multiple exclusion rules.

### DIFF
--- a/source/modules/MEI.xml
+++ b/source/modules/MEI.xml
@@ -1425,24 +1425,7 @@
     <desc xml:lang="en">Font size expressed as numbers; <abbr>i.e.</abbr>, points or virtual units.</desc>
     <content>
       <rng:data type="token">
-        <rng:param name="pattern">\d*(\.\d+)?(pt|vu)</rng:param>
-        <rng:except>
-          <!-- disallow no-value or all-zero patterns -->
-          <rng:choice>
-            <rng:data type="token">
-              <rng:param name="pattern">(pt|vu)</rng:param>
-            </rng:data>
-            <rng:data type="token">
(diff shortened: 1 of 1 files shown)


source/modules/MEI.xml
```diff
@@ -1425,24 +1425,7 @@
...
-            </rng:data>
-            <rng:data type="token">
-              <rng:param name="pattern">0+(\.0+)?(pt|vu)</rng:param>
-            </rng:data>
-            <rng:data type="token">
-              <rng:param name="pattern">\.0+(pt|vu)</rng:param>
-            </rng:data>
-          </rng:choice>
-        </rng:except>
+          <rng:param name="pattern">([1-9]\d*(\.\d+)?|\d*\.[0-9]*[1-9][0-9]*)(pt|vu)</rng:param>
```
rettinghaus, line 1428: Isn't there a pipe missing?
```suggestion
          <rng:param name="pattern">([1-9]\d*(\.\d+)?|\d*\.[0-9]*|[1-9][0-9]*)(pt|vu)</rng:param>
```

pe-ro (approved):
Thanks for the simplification. 

@rettinghaus, I think the suggested regex is correct; a pipe is not missing.  Adding a pipe where you suggest would allow zero values; e.g. `fontsize="0"` or `fontsize=".0"` as these would satisfy the `\d*\.[0-9]*` portion of the modified regex.

rettinghaus (approved):

PR #1810, opened 2026-07-20
schema: simplify anyXML macro definition

bwbohl:
Replace verbose RNG element definition with compact pureODD anyElement syntax. This reduces boilerplate while maintaining the same validation behaviour of permitting any XML elements except those from MEI and SVG namespaces.

### DIFF
--- a/source/mei-source.xml
+++ b/source/mei-source.xml
@@ -130,7 +130,8 @@
          <div>
             <schemaSpec ident="mei"
                         ns="http://www.music-encoding.org/ns/mei"
-                        start="mei">
+                        start="mei"
+                        defaultExceptions="http://www.music-encoding.org/ns/mei">
                <xi:include href="modules/MEI.xml"/>
                <xi:include href="modules/MEI.analytical.xml"/>
                <xi:include href="modules/MEI.cmn.xml"/>
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -28,25 +28,7 @@
   <macroSpec ident="macro.anyXML" module="MEI.shared" type="pe">
     <desc xml:lang="en">Permits any XML elements except those from the MEI or SVG namespace.</desc>
     <content>
-      <rng:element>
-        <rng:anyName>
-          <rng:except>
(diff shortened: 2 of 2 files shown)


pe-ro:
As I understand it, 
```
<anyElement except="http://www.music-encoding.org/ns/mei http://www.w3.org/2000/svg"/>
```
allows a single element.  A recursive model is necessary to address the element's descendants.

bwbohl:
The resulting RNG get’s a macro created:

```xml
<define name="anyElement_macro.anyXML_1">
      <element>
         <anyName>
            <except>
               <nsName ns="http://www.tei-c.org/ns/1.0"/>
               <name ns="http://www.tei-c.org/ns/Examples">egXML</name>
            </except>
         </anyName>
         <zeroOrMore>
            <attribute>
               <anyName/>
            </attribute>
         </zeroOrMore>
         <zeroOrMore>
            <choice>
               <text/>
               <ref name="anyElement_macro.anyXML_1"/>
            </choice>
         </zeroOrMore>
      </element>
   </define>
   ```
   
   Unfortunately RNG validation reports an error on the `<element>` But I cannot grasp the real reason behind it:
   `conflicting ID-types for attribute "id" from namespace "http://www.w3.org/XML/1998/namespace" of element "notesStmt" from namespace "http://www.music-encoding.org/ns/mei"`

**Comment by pe-ro** [comment]
@bwbohl, I think it's because elements from the TEI namespace are disallowed, but elements from the MEI namespace ARE allowed.  Because all (?) the MEI elements have an `@id` attribute and the namespace "http://www.w3.org/XML/1998/namespace", brought in by  
```
<attribute>
  <anyName/>
</attribute>
```
also defines an id attribute, there's a problem with duplication.

Although I don't know what incantation is necessary in ODD to make it happen, I think this problem will be fixed by excluding the MEI instead of the TEI namespace.  

For a more complete explanation, have a look at the book:
Van der Vlist, Eric. RELAX NG. Sebastopol, Calif: O'Reilly, 2003. pp. 176-178, "Traps to Avoid".

musicEnfanthen (approved):

PR #1816, opened 2026-07-23
schema: move `@betype` constraint to shared module

bwbohl:
Move the betype_required_when_begin_or_end constraint from performance-specific elements to the att.mediaBounds attribute class in the shared module. This fixes the duplicate ID error and improves maintainability.

### DIFF
--- a/source/modules/MEI.shared.xml
+++ b/source/modules/MEI.shared.xml
@@ -2200,6 +2200,14 @@
   </classSpec>
   <classSpec ident="att.mediaBounds" module="MEI.shared" type="atts">
     <desc xml:lang="en">Attributes that establish the boundaries of a media object.</desc>
+    <constraintSpec ident="betype_required_when_begin_or_end" scheme="schematron">
+      <constraint>
+        <sch:rule context="mei:clip[@begin or @end]">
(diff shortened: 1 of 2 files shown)


rettinghaus (approved):

source/modules/MEI.shared.xml
```diff
@@ -2200,6 +2200,14 @@
   </classSpec>
   <classSpec ident="att.mediaBounds" module="MEI.shared" type="atts">
     <desc xml:lang="en">Attributes that establish the boundaries of a media object.</desc>
+    <constraintSpec ident="betype_required_when_begin_or_end" scheme="schematron">
+      <constraint>
+        <sch:rule context="mei:clip[@begin or @end]">
+          <sch:assert role="warning" test="@betype or ancestor::mei:*[@betype]">When @begin or @end
+            is used, @betype should appear on clip or one of its ancestors.</sch:assert>
+        </sch:rule>
+      </constraint>
```
musicEnfanthen, line 2209: I read the assertion now as: when `clip` has begin or end, then clip itself or any ancestor (which is `recording` OR `avFile`) must have a `@betype` attr (first deleted rule above on clip).

Is this still triggering for the case when `recording` has begin or end (second deleted rule above on recording)?

PR #1819, opened 2026-07-23
fix: update constraint scheme from isoschematron to schematron

bwbohl:
Update two constraintSpec elements in the repeatMark definition to use the standard "schematron" scheme instead of the deprecated "isoschematron" scheme.


rettinghaus (approved):

pe-ro (approved):
Looks good.

Issue #1815, opened 2026-07-22
Deal with guidelines mentioning deprecated classes


bwbohl:
Currently, on #1792, source validation fails, and the build is thus aborted because the guidelines introduction mentions a removed attribute class in the section about 5.1. We need a strategy to handle this.

N.B. The guidelines-to-HTML-XSLTs handle this more gracefully by reporting the problem, not generating a link and outputting a suffix about the unavailability, e.g.: `The existing att.stringtab.tuning [not available in this MEI customization]`

There are several things we should do:
1. Relax the Schematron rule to reflect that the XSLTs will handle the case.
2. Decide whether/how long version-specific sections of the introduction will be kept and maintained

bwbohl:
The problematic aspect of the gracious HTML fallback, with the mention that the target is not available in this customisation, is that there’s no real reason given why it’s not available. As the template handles all such (and also for customisations) there can be different reasons:
1. typos (which it is supposed to detect)
2. removed targets (our case)
3. just not available in a drilled-down customisation
4. not available in source anywhere

Nevertheless, there can be meaningful mentions of targets that no longer exist, such as in the case of advertising changes to the schema.

We could come up with a strategy for:
* how to mark such references as deprecated 
* link those references to preceding versions of MEI

bwbohl:
@music-encoding/technical-team any ideas or thoughts?

riedde:
> link those references to preceding versions of MEI

I'm not into it at the moment, but keeping a hint-link to the former version (and information) would be nice.

bwbohl:
Moved it back to the featureFreeze milesonte because  we would like to create a tag with all build assets

ahankinson:
I thought the guidelines were version-specific, so any mention of deprecated attributes for v6 should be removed, right? It doesn’t change the older versions of the guidelines that are still available.

riedde:
@bwbohl is outdated documentation the problem or non-working documentation for specific customizations?

musicEnfanthen:
@riedde The "problem" in that specific case is that the introduction for v5.1 (that is still included in develop) mentions an attribute (and marks it up as attribute) which was recently deleted from the specs. So the validation process does not find any reference for the attribute any more. The resulting issue is that any later PR will "fail" because of the failing validation.

> Decide whether/how long version-specific sections of the introduction will be kept and maintained

[...]

PR #1829, opened 2026-08-04
Quickfix: Change assert role from error to warning for ident references

bwbohl:
Two make guidelines build pass


bwbohl:
provides a quickfix for #1815 should be implemented properly (e.g. for other elements etc) later

PR #1832, opened 2026-08-12
schema: clarifiy difference between `@head.shape` and `@glyph.*`

th-we:
**Edit:** This pull request now clarifies the difference between `@head.shape` and `@glyph.*`. Misunderstanding them lead to the original version of this PR.

~~I noticed that we have both `@head.shape`/`@head.auth` and `@glyph.name`/`@glyph.num`/`@glyph.auth` for encoding smufl (or other auth based) head shapes. I suggest only using the more generally available `@glyph.*` attributes, which also means that `@head.shape' has better validation because it currently can be set to any token.~~

~~Or am I misunderstanding the idea behind the `@glyph.*` attributes in this context and they are intended to represent the note in its "entirety"?  ("Entirety" being a vague concept for notes as it's not clear what parts of the note this would include or exclude.)~~  In that case, I'd open another pull request, documenting the difference between `@head.*` and `@glyph.*`.

### DIFF
--- a/source/modules/MEI.visual.xml
+++ b/source/modules/MEI.visual.xml
@@ -1348,7 +1348,8 @@
     </classes>
   </classSpec>
   <classSpec ident="att.note.vis" module="MEI.visual" type="atts">
-    <desc xml:lang="en">Visual domain attributes.</desc>
+    <desc xml:lang="en">Visual domain attributes. <att>glyph.name</att> and <att>glyph.num</att>
+      describe the entirety of the note, while <att>head.shape</att> encodes the head only.</desc>
     <classes>
       <memberOf key="att.altSym"/>
       <memberOf key="att.color"/>


pe-ro:
Hi @th-we,

Indeed, the intention of the `@head.* and @glyph.*` attributes is to address the head shape and the (entire) note shape separately. Others, @lpugin perhaps?, should address whether this is entirely (pun intended) necessary. 🤣

th-we:
> Indeed, the intention of the `@head.* and @glyph.*` attributes is to address the head shape and the (entire) note shape separately.

My intuition was that `@head.shape` should act in the logical domain, while `@glyph.*` attributes act in the visual domain. Currently, `@head.shape` serves as logical domain attribute if one of the [`data.HEADSHAPE.list`](https://music-encoding.org/guidelines/v5/data-types/data.HEADSHAPE.list.html) values is used, otherwise in the visual domain.

Take for example an x-shaped whole note head. When encoding, the main point I want to bring across is that it's x-shaped. To facilitate a more accurate visual rendering of the head, I might want to give some additional information about the shape variant - e.g. one of the possible SMuFL shapes:

grafik

I thought the encoding for this should be:
```xml
<note head.shape="x" glyph.auth="smufl" glyph.name="noteheadVoidWithX"/>
```

If `@glyph.*` is reserved for representing the entire note (I see there can be use cases for that), I have to decide if I want logical clarity or visual precision for head shapes, but I can't have both.

Two suggestions how to solve this:

1. Introduce `@head.name`/`@head.num`, which work like `@glyph.*` attributes, in conjunction with `@head.auth`.
    ```xml
    <note head.shape="x" head.auth="smufl" head.name="noteheadVoidWithX"/>
    ```
2. Add a `<head>` element.
    ```xml
    <note>
      <head  shape="x" glyph.auth="smufl" glyph.name="noteheadVoidWithX"/>
    </note>
    ```

ahankinson:
I’m struggling to see how `head.shape` is a “logical” attribute, given that duration and other qualities of the note are usually specified in other attributes.

SMuFL has both [full note shapes](https://smufl.formats.music/latest/tables/individual-notes.html) as well as [various head shapes](https://smufl.formats.music/latest/tables/noteheads.html) so it would make sense to allow both in different attributes. (Possibly also expanding to target stem / flag shapes as well, but that’s a separate issue).

[The docs say](https://music-encoding.org/guidelines/v5/attribute-classes/att.noteHeads.html)

“ When @head.auth matches 'smufl', @head.shape must contain a numeric glyph reference in hexadecimal notation, like "#xE000" or "U+E000".”

So setting `head.auth` to “smufl” makes your example:

```
<note head.auth="smufl" head.shape="U+E0B7”/>
```

th-we:
> I’m struggling to see how `head.shape` is a “logical” attribute, given that duration and other qualities of the note are usually specified in other attributes.

You're right, the label "logical" is not truly fitting. It's more that `@head.shape="x"` says something about the general category, while `@head.shape="U+E0B7"` (thanks for pointing out the restriction to SMuFL codepoints) can be thought of as going into graphical details.

While SMuFL codepoints always denote a very specific shape, `data.HEADSHAPE.list` values partly require showing different glyphs depending on `@dur` (except for the regular `whole`, `half`, `quarter` values). At least that's how Verovio does it, for example for `diamond` or `rectangle` :

image

image

I think this makes sense, although [the spec](https://music-encoding.org/guidelines/v5/data-types/data.HEADSHAPE.list.html) strictly speaking only specifies unfilled heads. I think this is a mistake which also needs fixing.

All in all, I think both uses of `@head.shape` (auth based and `data.HEADSHAPE.list` based) are different enough that they should be encodable independently.

ahankinson:
> While SMuFL codepoints always denote a very specific shape, data.HEADSHAPE.list values partly require showing different glyphs depending on `@dur` (except for the regular whole, half, quarter values). At least that's how Verovio does it, for example for diamond or rectangle ...

That behaviour is probably desirable, though, just like we wouldn't want to have to specify the exact glyph to be used for the "regular" noteheads. I think the expectation here is that you would vary the properties only when it differs from expected norms, e.g., an unfilled eighth note head. 

Is there a particular use case you are working with, or something you are finding it difficult to encode? Examples might help ground it in practical ideas. But I'm not particularly in favour of the changes in this PR without more discussion.

th-we:
> That behaviour is probably desirable, though, just like we wouldn't want to have to specify the exact glyph to be used for the "regular" noteheads. I think the expectation here is that you would vary the properties only when it differs from expected norms, e.g., an unfilled eighth note head.

O.K., so it seems like we can establish consent here, if you agree, @pe-ro? That could lead to another PR removing the "unfilled" part from the `data.HEADSHAPE.list` descriptions.

> Is there a particular use case you are working with, or something you are finding it difficult to encode? Examples might help ground it in practical ideas.

I'm trying to export head shapes from Sibelius and therefore I'm trying to find a good mapping from all the built-in Sibelius note styles. It doesn't have to be perfect, I guess, especially because I'll allow projects to define their own encoding for built-in and custom note styles. The note style I'm most interested in right now is slash notation like this:

image

The only relevant issue here however isn't head shape related (it doesn't matter in this case if the half note is rendered as a large diamond, like in the image, or a parallelogram, like Verovio would do it). The slashes have no ledger lines (they're not defining a pitch, so ledger lines have no meaning), but MEI has no way of encoding this AFAIK. So, that's an entirely different matter altogether.

> But I'm not particularly in favour of the changes in this PR without more discussion.

Right. Given that `@glyph.*` is intended for the entire note, not just the head, I'll rework the PR entirely.

Thanks for your thoughts and helpful input!

ahankinson:
I'll just add that there's also `@head.fill`, which is possibly why `@head.shape` specifies "unfilled" -- the expectation is that if you want filled, you use `@head.fill` to do so. But I don't think that works especially well if we want a "dynamic" fill behaviour -- like varying head fill by duration. So I'm a cautious "+1" on removing the "Unfilled" from `data.HEADSHAPE.list`, unless someone can think of a reason why that's not a good idea.

th-we:
I dropped my old changes and only added some wording to describe the different purposes of `@glyph.*` and `@head.*`.

source/modules/MEI.visual.xml
```diff
@@ -1348,7 +1348,8 @@
     </classes>
   </classSpec>
   <classSpec ident="att.note.vis" module="MEI.visual" type="atts">
-    <desc xml:lang="en">Visual domain attributes.</desc>
```
ahankinson, line 1351: This note isn't really visible to most users -- it's only really shown if they happen to go to the `att.note.vis` class page: https://music-encoding.org/guidelines/dev/mei-all/attribute-classes/att.note.vis.html

th-we: Any suggestion what could be the most suitable place?

PR #1836, opened 2026-09-18
Allow multiple timestamps in `@tstamp.real`

th-we:
When there are repeats, an element like a note may have to be played multiple times.


pe-ro (approved):
Looks good to me.

ahankinson:
1. Should `@tstamp.ges` also take multiple times?
2. What about `att.timestamp2.ges` for repeated spanning elements like slurs or ties?

th-we:
> 1. Should `@tstamp.ges` also take multiple times?

My understanding of `@tstamp.ges` is that it is relative to the parent measure. One use case might be to encode a swing feel when straight eighth notes are encoded, like in 4/4 time:

```xml
<note dur="8" tstamp.ges="1"/>
<note dur="8" tstamp.ges="1.6"/> <!-- logical @tstamp would be 1.5 -->
<note dur="8" tstamp.ges="2"/>
<note dur="8" tstamp.ges="2.6"/> <!-- logical @tstamp would be 2.5 -->
```

However, one interesting thing is that it's allowed on `<measure>` itself, and I don't know what that would mean.

>     2. What about `att.timestamp2.ges` for repeated spanning elements like slurs or ties?

I think that would be similar to `@tstamp.ges`.

